PR #8, opened 2017-07-23
Add resource for Okta auth backend

### LINKED ISSUE #7: Unable to run `terraform destroy` when configuring `okta` auth backend
Not all Vault endpoints support delete which means that if you try and set up a Vault instance using `vault_generic_secret`, such as the example below, you won't be able to successfully run `terraform destroy`. 
```hcl
resource "vault_auth_backend" "example" {
  type = "okta"
}

resource "vault_generic_secret" "example" {
  path = "auth/${vault_auth_backend.example.path}/config"
  data_json = <<EOT
{
}
EOT
}
```

The Vault team are reluctant to go through all of the endpoints to add delete (https://github.com/hashicorp/vault/pull/2889) but the way the Vault provider currently works prevents them from being properly.

One solution I thought of would be to move away from generic `vault_auth_backend` to ones specific to each auth backend or mount. The advantage that this approach has is that it avoids having to look up API endpoints to know how to configure an auth backend and helps stop `vault_generic_secret` from being used as a hammer for all of the problems.

Opinion?

wjam:
This adds a resource specific to the Okta auth backend to start to allow a more 'terraform-native' approach to maintaining the configuration of a Vault cluster, rather than having to deal with using `vault_generic_secret` with random paths. This fixes #7.


wjam:
@radeksimko @apparentlymart Any thoughts on this PR or the approach of having resources more specific to Vault rather than just having the `vault_generic_secret` resource?

wjam:
@grubernaut @apparentlymart @paddycarver @catsby Is there any appetite for merging this?

wjam:
Any update on this being reviewed?

vault/resource_okta_auth_backend_test.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	if _, ok := authMounts["okta/"]; ok {
+		return fmt.Errorf("Auth mount not destroyed")
+	}
+
+	return nil
+}
+
+func assertEquals(expected, actual interface{}) error {
```
paddycarver: I'm not super crazy about the assert* helpers; first because the Terraform style tends to just repeat the error checking and formatting, second because the way it's implemented right now makes it a bit tricky to understand which field didn't match if there's not a match, and finally because it obscures the implementation detail (that it's doing a `!=` instead of (e.g., for `time.Time`) using the built-in equality method.

vault/resource_okta_auth_backend_user.go
```diff
@@ -0,0 +1,148 @@
...
+	if err := updateOktaUser(client, path, user); err != nil {
+		return fmt.Errorf("Unable to update user %s in Vault: %s", username, err)
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s", path, username))
+
+	return nil
+}
+
+func oktaAuthBackendUserRead(d *schema.ResourceData, meta interface{}) error {
```
paddycarver: Again, we should set the data from the API response here, not just check for existence.

vault/resource_okta_auth_backend_test.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+var initialOktaAuthConfig = `
+resource "vault_okta_auth_backend" "test" {
+    description = "Testing the Terraform okta auth backend"
+    organization = "example"
+    token = "this must be kept secret"
+    ttl = "1h"
+    group {
+        group_name = "dummy"
```
paddycarver: We like making any test data that requires uniqueness randomized, so multiple tests can run at once and so if a test doesn't clean up properly after itself, it doesn't break integration tests until someone goes in and cleans it up. You can find some helpers for randomising these names in the acctest package: https://godoc.org/github.com/hashicorp/terraform/helper/acctest

Generally how this is done is by making the config a function, passing in randomised or environment-determined variables, injecting them with `fmt.Sprintf`, and initializing all these variables in the `TestAcc*` functions. You can see a few examples in the tests for the AWS auth backend.

website/docs/r/okta_auth_backend.html.md
```diff
@@ -0,0 +1,75 @@
+---
```
paddycarver: These website docs should be added to the sidebar, in `website/vault.erb`.

wjam:
Fixed issues and somehow managed to rebase the PR, although this did fix the merge conflict

wjam:
@paddycarver @grubernaut ???

grubernaut:
Hey @wjam,

Sorry for the delay in someone looking at this, unfortunately I no longer work for HashiCorp. However, I do see that there is now a merge conflict, but the majority of the issues mentioned by @paddycarver were addressed. I believe fixing the merge conflict in `website/vault.erb` should be all that's needed for this contribution. Thank you again for your patience here, and the code contribution. 💯

paddycarver:
Hey @wjam, apologies for the delay on this. I have notes here on the last few pieces of feedback, which somehow I had convinced myself I had submitted to you, and was waiting on your response. :( Sincere apologies for losing track like that.

Because you've been so patient and worked hard on this, and the pieces of feedback I have are just minor bugs and nitpicks at this point (adding docs to the sidebar, a bug with IDs not being unique enough, and a couple things I'd like to see persisted in state that aren't at the moment) I'm going to just fix them and merge this so it's in the next release of the provider, leaving your commits intact so you get credit for doing ~all the work on this.

Thank you so much for your patience and this awesome contribution to the Vault provider. Really excited about it getting merged.

paddycarver:
Really sorry for the delay on this. I've opened #72 with some really minor fixes and the merge conflicts sorted, and I'll get a coworker to look over my changes commit in the morning so we can get this merged and closed out.

paddycarver:
This has now been merged and will be released with the next release of the provider. Thanks for contributing it to Terraform!

PR #11, opened 2017-08-15
Adding resource token - creates new tokens

elliottsam:
This adds the ability to create Vault tokens that can then be used in bootstrapping instances


irconan:
This feature would be very useful to me too.
If it helps my fork has this patch rebased over current master:
https://github.com/irconan/terraform-provider-vault/tree/token_resource

elliottsam:
Hi @paddycarver thanks for the update, glad to see it will be going in soon.

PR #15, opened 2017-08-26
Make policies importable.

paddycarver:
This makes policies importable. Minor changes were needed for this:

* Read now sets name, in addition to policy.
* We added the Importer functionality and a test for it.
* Our test config needed to be updated to allow the policy name to be
  injected, to help ensure uniqueness across runs.


paddycarver:
Test failure is an acctest vendoring failure. Should be resolved by #12.

vault/import_policy_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccPolicy_importBasic(t *testing.T) {
+	name := "test-" + acctest.RandString(10)
```
catsby: There's also [RandomWithPrefix](https://github.com/hashicorp/terraform/blob/master/helper/acctest/random.go#L30), if integers are valid in the name.

vault/resource_policy.go
```diff
@@ -77,6 +80,7 @@ func policyRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.Set("policy", policy)
+	d.Set("name", name)
```
catsby: Should we set this in the `CREATE` method, `policyWrite`? Right now, `policyWrite` does not call `policyRead`, so the `name` attribute won't be set until a `refresh` operation happens. 

Normally we call `create` and it returns `read`. If that's not necessary here, I recommend we set `name` in `policyWrite` then.

vault/import_policy_test.go
paddycarver: TIL! Thanks :)

vault/resource_policy.go
paddycarver: Good call, I missed that wasn't happening here. Updated to now call `policyRead` from `policyWrite`.

PR #16, opened 2017-08-26
Make mounts importable

paddycarver:
Update mount's read function to set all the data in state, instead of just what's available in the mount config.

Update the basic test to let us inject the path in, for more resilience to dangling resources.

Add the import logic and a test to ensure that it works as expected.


paddycarver:
Test failure is vendoring acctest. Should be fixed by #12.

vault/resource_mount.go
```diff
@@ -137,13 +142,23 @@ func mountRead(d *schema.ResourceData, meta interface{}) error {
...
 
-	mount, err := client.Sys().MountConfig(path)
+	mounts, err := client.Sys().ListMounts()
 	if err != nil {
 		return fmt.Errorf("error reading from Vault: %s", err)
 	}
 
-	d.Set("default_lease_ttl_seconds", mount.DefaultLeaseTTL)
-	d.Set("max_lease_ttl_seconds", mount.MaxLeaseTTL)
+	mount, ok := mounts[strings.Trim(path, "/")+"/"]
```
catsby: So, sometimes `path` has a trailing `/`, but sometimes it doesn't? So here we ensure it has a single `/`? Is that what's happening? Please add some comments here for future readers 😄

paddycarver: Updated with comment :D

PR #23, opened 2017-10-06
Support for wrapped secrets

### LINKED ISSUE #4: Terraform should be able to generate wrapped vault secret-ids for servers
_This issue was originally opened by @gtmtech as hashicorp/terraform#12687. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Terraform all versions. Hashicorp Vault 0.6+

Terraform<->Vault is missing a fundamental feature which would allow vault AppRole authentication to work.

AppRole authentication in Vault is provided so that a server, or apps running on a server are able to authenticate itself to Vault and procure secrets.

In order to work, the authentication login step requires two bits of information - a role_id and a secret_id. Vault best practice describes how you generate a role_id, and might commit this into configuration management (e.g. Packer+Ansible => AMI) , whereas the secret_id should be supplied to the final app if possible.

It also suggests that for maximum security, the secret_id when requested should be response-wrapped using the Vault cubbyhole backend, which basically means that only one entity is able to decrypt it. If the target/server decrypts it then all is well. If a bad actor decrypts it first, then the target/server cannot decrypt it, and can raise an alert about doing so. 

All this presents a lovely case for an integration step between terraform and vaults AppRole, so that prior to spinning up an aws_instance resource, terraform requests a wrapped secret-id from Vault's AppRole API, passes the wrapped secret-id into user-data, and then the resultant ec2 instance spinning up can unwrap the secret, and get whatever secrets it needs. 

However this functionality is missing. Could we have it please?

All that would be needed is a new data resource called something like vault_approle_wrapped_secret:

```
data "vault_approle_secret_id" "jenkins" {
    role_id="1234-123456-1234-1234"
    wrapped="true"
    ttl="60s"
}
```

Then in a corresponding userdata cloud-init template (or equivalent), this can be referred to easily:

```
SECRET_ID_WRAPPED_TOKEN=${data.vault_approle_secret_id.jenkins}
```

et Voila - the instance spins up, gets the role_id from its AMI (ansible), get the wrapped token_id from user-data and does the following (probably in a bash startup script)

```
1. authenticates to vault with unwraponly:password to allow it to unwrap wrapped secrets.
2. unwraps the SECRET_ID_WRAPPED_TOKEN to yield a SECRET_ID
3. re-authenticates to vault using the role_id and the secret_id
4. pulls whichever secrets it needs out of vault.
```

jsoriano:
Adds a `wrap_ttl option` to `vault_generic_secret` so wrapped secrets can be requested in a generic way.
It also adds a `command` option so write operations can also be performed, this allows to obtain secrets that are requested with write operations as secret IDs in AppRole auth (this would fix #4) or keys in PKI module.

### DIFF
--- a/vault/data_source_generic_secret.go
+++ b/vault/data_source_generic_secret.go
@@ -4,7 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strconv"
 	"time"
+
+	"github.com/hashicorp/go-uuid"
 
 	"github.com/hashicorp/terraform/helper/schema"
 
@@ -20,6 +23,27 @@ func genericSecretDataSource() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				Description: "Full path from which a secret will be read.",
+			},
+
+			"command": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "read",
+				ValidateFunc: validateVaultCommand,
+				Description:  "Vault command to use to get the secret.",
+			},
+
+			"wrap_ttl": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Setting this value will wrap response with specified TTL.",
+			},
+
+			"write_data_json": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Data to send on write command.",
+				ValidateFunc: ValidateDataJSON,
 			},
 
 			"data_json": {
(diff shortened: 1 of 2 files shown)


melbit-michaelw:
👍 for this issue. @paddycarver is there any update on this pull request ? It looks like it's successfully built.. what's the next step ?

pielu:
Would be great to get this in. A lot of Vault's functionality would become available for Terraform. Anything we could do to get this moved forward?

jsoriano:
Branch updated with master

jsoriano:
@paddycarver thanks for your comments, I agree that not having a way to define a wrapping TTL per request is a problem. I wonder if we could instantiate a client per request as a workaround meanwhile.

Regarding the `generic_data` source, I also mostly agree with you, but I think that it provides over the `http` provider something similar to what generic `vault` cli subcommands provide over `curl`. I think it is interesting to have these generic semantics to cover use cases that don't have specific data sources or providers.

paddycarver:
> I wonder if we could instantiate a client per request as a workaround meanwhile.

You _could_, but you'd lose all connection pooling behavior, and that's a lot of overhead.

> Regarding the generic_data source, I also mostly agree with you, but I think that it provides over the http provider something similar to what generic vault cli subcommands provide over curl. I think it is interesting to have these generic semantics to cover use cases that don't have specific data sources or providers.

I agree with you in theory, but in practice I think it's generating more issues than it's replacing. Especially as the KV store iterates (0.10 just launched new improvements to it) I think deprecating the generic resources and data sources and switching to resources and data sources tooled to the API resources being managed is probably a good goal to have. I think it would make the user experience better and less confusing, personally.

jsoriano:
I see. Ok, then I'm fine with closing this by now.

paddycarver:
(I do encourage, however, opening issues for wrapping specific types of secrets! I'm definitely not against wrapping, and I could be persuaded on this issue--I just have already implemented wrapping once for #90, and the thought of doing it with `generic_secret` evokes an immediate "oh please no" reaction in me. 😄 It's also possible we could implement it using some helpers that would make it possible to add to all our secret types easily, and I'd be for that, too.)

PR #25, opened 2017-10-16
Add role tag data source.

paddycarver:
Create a data source for reading role tags from the Vault API.


vault/data_source_aws_auth_backend_role_tag.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if v, ok := d.GetOk("allow_instance_migration"); ok {
+		data["allow_instance_migration"] = v
+	}
+	if v, ok := d.GetOk("disallow_reauthentication"); ok {
+		data["disallow_reauthentication"] = v
+	}
+
+	log.Printf("[DEBUG] Reading tag data %q from Vault", path)
+	secret, err := client.Logical().Write(path, data)
```
catsby: `client.Logical().Write()` seems odd here, shouldn't it be `client.Logical().Read()`?  That's what the other two data sources have... ?

paddycarver: Because role tags [require a `POST`](https://www.vaultproject.io/api/auth/aws/index.html#create-role-tags) request, `Write()` is needed there. `Read()` would use a `GET` request.

paddycarver:
Test output:

```
TF_ACC=1 go test ./vault -v -run=TestAccDataSourceAWSAuthBackendRoleTag -timeout 120m
=== RUN   TestAccDataSourceAWSAuthBackendRoleTag_basic
--- PASS: TestAccDataSourceAWSAuthBackendRoleTag_basic (0.90s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.901s
```

PR #26, opened 2017-10-16
Add identity whitelist configuration resource.

paddycarver:
Allows users to configure the identity whitelist settings of an AWS
auth backend.


website/docs/r/aws_auth_backend_identity_whitelist.html.md
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `backend` - (Optional) The path of the AWS backend being configured.
+
+* `safety_buffer` - (Optional) The amount of extra time, in minutes, that must
```
catsby: Is `0` a valid value here, or is there a default?

vault/resource_aws_auth_backend_identity_whitelist.go
```diff
@@ -0,0 +1,153 @@
...
+func awsAuthBackendIdentityWhitelistWrite(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string)
+	data := map[string]interface{}{}
+
+	if v, ok := d.GetOk("safety_buffer"); ok {
+		data["safety_buffer"] = v
+	}
+	if v, ok := d.GetOk("disable_periodic_tidy"); ok {
```
catsby: `disable_periodic_tidy` needs to have a default value, or use `d.GetOkExists`. Without one of those, a user cannot set this to `true` and then remove it, or set it to `false`.

website/docs/r/aws_auth_backend_identity_whitelist.html.md
paddycarver: There's a default (72h: https://www.vaultproject.io/api/auth/aws/index.html#safety_buffer) that comes from Vault), but this can also be set to 0. Updating to GetOkExists so 0 can be used, thanks for pointing that out.

vault/resource_aws_auth_backend_identity_whitelist.go
paddycarver: Great catch, thanks. Updated.

PR #28, opened 2017-10-23
Add the ability to generate tokens from AWS auth backend.

paddycarver:
Add the vault_aws_auth_backend_login resource, which manages a "login",
basically a token session. Ideally, this would be a data source that
would generate a token on every read, and that's how the implementation
started. However, because the EC2 instance auth method only allows a
single authentication per host for a set amount of time, the refresh
cycle made this useless as a data source--an API error would be thrown
on every subsequent plan until the instance's identity was removed from
the whitelist. Which isn't ideal.

So instead this became a resource, so it could manage the requisite
state (namely, the nonce) that is needed to reissue tokens to an
instance that is already present in the whitelist. I opted to use the
Read method to do the actual token creation, rather than in the Create
function, to make this work essentially like a data source, issuing a
new token on every call. The plus side of this is that the resource
"just works". This could also be managed more like a typical resource,
with the Create function creating the token, but due to limitations in
the API, we can't read all the information about a token back, so the
usefulness of that approach is more limited.

Testing of this is weird, largely because Vault requires an EC2 instance
be actually running to issue tokens on its auth info. And access to
EC2's metadata server is needed, as well. Rather than trying to stand up
EC2 instances and SSH into them as part of the tests, the tests that
require the metadata server are skipped if the metadata server is
unavailable.

### DIFF
--- a/website/docs/r/aws_auth_backend_login.md
+++ b/website/docs/r/aws_auth_backend_login.md
@@ -1,0 +1,112 @@
+---
+layout: "vault"
+page_title: "Vault: vault_aws_auth_backend_login resource"
+sidebar_current: "docs-vault-aws-auth-backend-login"
+description: |-
+  Manages Vault tokens acquired using the AWS auth backend.
+---
+
+# vault\_aws\_auth\_backend\_login
+
+Logs into a Vault server using an AWS auth backend. Login can be
+accomplished using a signed identity request from IAM or using ec2
+instance metadata. For more informtion, see the [Vault
+documentation](https://www.vaultproject.io/docs/auth/aws.html).
+
+## Example Usage
+
+```hcl
+resource "vault_auth_backend" "aws" {
+  type = "aws"
+}
+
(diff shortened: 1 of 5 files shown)


website/docs/r/aws_auth_backend_login.md
```diff
@@ -0,0 +1,112 @@
...
+  removed. Can be retrieved from the EC2 metadata server.
+
+* `pkcs7` - (Optional) The PKCS#7 signature of the identity document to
+  authenticate with, with all newline characters removed. Can be retrieved from
+  the EC2 metadata server.
+
+* `nonce` - (Optional) The unique nonce to be used for login requests. Can be
+  set to a user-specified value, or will contain the server-generated value
+  once a token is issued. EC2 instances can only acquire a single token until
+  the whitelist is tidied again unless they keep track of this nonce.e
```
catsby: Extra `e` here I think, `nonce.e`

PR #30, opened 2017-10-30
Implement AppRole roles.

paddycarver:
Add a new vault_approle_auth_backend_role resource that enables users to
provision and manage AppRole roles.

This also updates to use Terraform 0.10.1 as the vendor, as we need it
for GetOkExists support. But updating to 0.10.8 requires a lot of new
dependencies and was getting us into dependency headaches land, so this
PR is the smallest possible step to achieve what we want.

A future AppRole Secret ID resource is planned, so no support for it is
in this PR. Also, an AppRole RoleID data source is forthcoming, as well,
but it was included here because users can specify their own RoleIDs and
we want to be able to support that using Terraform.


vault/resource_approle_auth_backend_role.go
```diff
@@ -0,0 +1,384 @@
...
+				Computed:    true,
+				Description: "The RoleID of the role. Autogenerated if not set.",
+			},
+			"bind_secret_id": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether or not to require secret_id to be present when logging in using this AppRole.",
+			},
+			"bound_cidr_list": {
```
catsby: Is this list guaranteed to be sorted, or does it need to be sorted? `TypeList` is great for things that should maintain an order

vault/resource_approle_auth_backend_role.go
```diff
@@ -0,0 +1,384 @@
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "List of CIDR blocks that can log in using the AppRole.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"policies": {
```
catsby: Same sorting question

vault/resource_approle_auth_backend_role.go
```diff
@@ -0,0 +1,384 @@
...
+	period, err := resp.Data["period"].(json.Number).Int64()
+	if err != nil {
+		return fmt.Errorf("Expected period %q to be a number, isn't", resp.Data["period"])
+	}
+
+	d.Set("backend", backend)
+	d.Set("role_name", role)
+	d.Set("period", period)
+	d.Set("policies", policies)
+	d.Set("bound_cidr_list", cidrs)
```
catsby: Nit: Should check err return of `d.Set` for non-scalar values, e.g. sets, lists, maps

vault/resource_approle_auth_backend_role.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	if v, ok := d.GetOk("token_max_ttl"); ok {
+		data["token_max_ttl"] = v.(int)
+	}
+
+	_, err := client.Logical().Write(path, data)
+
+	d.SetId(path)
+
+	if err != nil {
```
catsby: nit, not a required change: if we check `err` before we call `d.SetId`, and it's not `nil`, we can simply return the error and not worry about unsetting the ID, just set it after the error check block.

vault/resource_approle_auth_backend_role.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func approleAuthBackendRoleDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	path := d.Id()
+
+	log.Printf("[DEBUG] Deleting AppRole auth backend role %q", path)
+	_, err := client.Logical().Delete(path)
+	if err != nil {
+		return fmt.Errorf("Error deleting AppRole auth backend role %q", path)
```
catsby: Q: Is it possible that `Delete` here returns some kind of error akin to a 404? If so, we should catch that error and simply unset the ID

vault/resource_approle_auth_backend_role.go
paddycarver: As far as I know, it's not guaranteed to be sorted, but I can add sort to it before setting it in state if we're worried about it.

vault/resource_approle_auth_backend_role.go
paddycarver: Same sorting answer 😆

vault/resource_approle_auth_backend_role.go
paddycarver: Not 100% sure what I was thinking here. I'll update it.

vault/resource_approle_auth_backend_role.go
paddycarver: 👍

vault/resource_approle_auth_backend_role.go
paddycarver: I could've _sworn_ that Vault just returned nil on 404, instead of treating it as an error, but upon looking to prove that, discovered it _will_ treat a 404 as an error, but in a way that's incredibly frustrating to catch. :(

vault/resource_approle_auth_backend_role.go
paddycarver: Oops, no, that's a bad idea. I'll just make it a Set :)

paddycarver:
Think I've addressed all the comments here. Happy to get a second round of feedback from @catsby if he has time, but otherwise I think it's also ok for @terraform-providers/ecosystem to take a look.

PR #32, opened 2017-10-30
Add RoleID data source for AppRole.

paddycarver:
Add a data source that will read in the RoleID for an AppRole given the
backend it resides in and hte role's name.

Depends on #30.


paddycarver:
Merging this now. This should have no user-visible change; before this, the `id` attribute as well as the actual `id` are getting set to the same value, and from what I can tell, it's the value set using `d.SetId`. So I'm not considering this a BC, as all user configs should continue working just as they were, and all these fields are computed, so it's not like users could be trying to set them in configs, anyways.

vault/data_source_approle_auth_backend_role_id.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return fmt.Errorf("Error reading AppRole auth backend role %q RoleID: %s", path, err)
+	}
+	log.Printf("[DEBUG] Read AppRole auth backend role %q RoleID", path)
+
+	if resp != nil {
+		d.SetId(path + "/role-id")
+		d.Set("role_id", resp.Data["role_id"])
+	}
+
```
paultyng: should it `SetID("")` if `resp` is `nil`?

paddycarver: Probably! I'll do that.

PR #31, opened 2017-10-30
Add support for AppRole SecretIDs

paddycarver:
Allow users to generate and manage AppRole SecretIDs.

Depends on #30.

PR #34, opened 2017-11-06
 Add AppRole login

paddycarver:
Add a resource that manages AppRole auth backend logins and generating
Vault tokens.

Note: this PR depends on #31.


vault/resource_approle_auth_backend_login.go
```diff
@@ -0,0 +1,209 @@
...
+	if leaseExpiringSoon(d, client) {
+		log.Printf("[DEBUG] Lease for %q expiring soon, renewing", d.Id())
+		renewed, err := client.Auth().Token().Renew(d.Get("client_token").(string), d.Get("lease_duration").(int))
+		if err != nil {
+			log.Printf("[DEBUG] Error renewing token %q, bailing", d.Id())
+		} else {
+			resp = renewed
+			d.Set("lease_started", time.Now().Format(time.RFC3339))
+			d.Set("client_token", resp.Auth.ClientToken)
+			d.SetId(resp.Auth.Accessor)
```
catsby: Does this `resp.Auth.Accessor` value change? It's unusual to see `d.SetId` outside of a `create` method

paddycarver: It should be a permanent, stable ID, but I think I was just setting it just to be safe. I'll remove it.

paddycarver: I spoke too soon, sorry. Yes, this value changes, because we're essentially replacing the resource. I'm trading in one token (that has expired) for a new token (that has not). I dunno if it's cool to play around with resource IDs like that, but this resource is of minimal utility if I can't swap out the tokens. :/

catsby: It's unusual to swap them, yes, but the circumstances seem to warrant it

PR #38, opened 2017-11-13
Add database secret backend role resource.

paddycarver:
Add a resource that will allow the creation and management of database
secret backend roles. These roles are what are ultimately used to
generate temporary credentials for the database.

This relies on #37, which should be merged before this.


vault/resource_database_secret_backend_connection.go
```diff
@@ -0,0 +1,613 @@
...
+						"connect_timeout": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     5,
+							Description: "The number of seconds to use as a connection timeout.",
+						},
+					},
+				},
+				MaxItems:      1,
+				ConflictsWith: []string{"mongodb", "hana", "mssql", "mysql", "postgresql", "oracle"},
```
paultyng: I feel like this list repetition is pretty error prone, perhaps you could write a little func that just takes the removal in?

vault/resource_database_secret_backend_connection.go
```diff
@@ -0,0 +1,613 @@
...
+		}
+		data["allowed_roles"] = strings.Join(roles, ",")
+	}
+
+	log.Printf("[DEBUG] Writing connection config to %q", path)
+	_, err = client.Logical().Write(path, data)
+
+	d.SetId(path)
+
+	if err != nil {
```
paultyng: This check should be moved underneath the assignment if possible.

PR #63, opened 2018-01-29
Allow secret backend role paths to contain forward-slashes

jasonarewhy:
Currently vault_aws_secret_backend_role performs a validation on the role path. That validation splits the string by `/`, making the assumption that there are no forward slashes outside of those surrounding "roles". However, Vault has no such restriction on forward slashes. We use slashes in our generate Vault paths for namespacing purposes. An example:

```
/users/aws/<account>/
/services/aws/<account>/
```

vault_aws_secret_backend_role would split such a path incorrectly. This commit instead splits the string by the more complete `/roles/`, allowing for any number of slashes before and after.

The only possible hiccup with this solution (that I'm aware of) would be if you had '/roles/' in your path elsewhere, but I imagine Vault itself would have trouble with that (and it'd be a confusing choice to make).

We are successfully using this commit in production to manage vault roles with backends that include multiple forward slashes.

### DIFF
--- a/vault/resource_aws_secret_backend_role.go
+++ b/vault/resource_aws_secret_backend_role.go
@@ -84,8 +84,8 @@ func awsSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
 	path := d.Id()
-	pathPieces := strings.Split(path, "/")
-	if len(pathPieces) != 3 || pathPieces[1] != "roles" {
+	pathPieces := strings.Split(path, "/roles/")
+	if len(pathPieces) != 2 {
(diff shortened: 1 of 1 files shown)


paddycarver:
Hi @jasonarewhy! Thanks for the PR. Do you mind just adding a test case showing the bug you're observing?

martinssipenko:
This is related to #79 and currently the documentation mentions:

> The path the AWS secret backend is mounted at, with no leading or trailing /s.

However, I do think that having support leading/trailing slashes does make sense, but perhaps they could just be removed in code for backwards compatibility?

tyrannosaurus-becks:
I agree that support for this makes sense. Closing, though, because of a lack of test cases and because of the age of the PR. Happy to reconsider with tests! Thank you for the contribution @jasonarewhy !

PR #68, opened 2018-02-07
new resource: vault_pki_backend_role

roboll:
Adds a new resource, `vault_pki_backend_role`. Arguments are taken from https://www.vaultproject.io/api/secret/pki/index.html#create-update-role.

### DIFF
--- a/vault/resource_pki_backend_role.go
+++ b/vault/resource_pki_backend_role.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/hashicorp/vault/api"
 )
 
@@ -52,12 +53,12 @@ func pkiBackendRoleResource() *schema.Resource {
 				Description: "Specifies if clients can request certificates for localhost as one of the requested common names.",
 			},
 			"allowed_domains": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
-				Description: "Policies to be set on tokens issued using this role.",
+				Description: "Specifies the domains of the role.",
 			},
 			"allow_bare_domains": {
 				Type:        schema.TypeBool,
@@ -120,10 +121,11 @@ func pkiBackendRoleResource() *schema.Resource {
 				Description: "Specifies if certificates are flagged for email protection use.",
 			},
 			"key_type": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Default:     "rsa",
-				Description: "Specifies the type of key to generate for generated private keys. Currently, rsa and ec are supported.",
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "rsa",
+				ValidateFunc: validation.StringInSlice([]string{"rsa", "ec"}, false),
+				Description:  "Specifies the type of key to generate for generated private keys. Currently, rsa and ec are supported.",
 			},
 			"key_bits": {
 				Type:        schema.TypeInt,
@@ -178,10 +180,9 @@ func pkiBackendRoleResource() *schema.Resource {
 			},
 			"backend": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Required:    true,
 				Description: "Unique name of the PKI backend to configure.",
 				ForceNew:    true,
-				Default:     "pki",
 				// standardise on no beginning or trailing slashes
 				StateFunc: func(v interface{}) string {
 					return strings.Trim(v.(string), "/")
@@ -198,8 +199,6 @@ func pkiBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
 	role := d.Get("role").(string)
 
 	path := pkiBackendRolePath(backend, role)
-
-	log.Printf("[DEBUG] Writing PKI backend role %q", path)
 
 	data := map[string]interface{}{}
 
@@ -244,14 +243,14 @@ func pkiBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
 	data["generate_lease"] = d.Get("generate_lease").(bool)
 	data["no_store"] = d.Get("no_store").(bool)
 
+	log.Printf("[DEBUG] Writing PKI backend role %q", path)
 	_, err := client.Logical().Write(path, data)
 
-	d.SetId(path)
-
 	if err != nil {
-		d.SetId("")
 		return fmt.Errorf("Error writing PKI backend role %q: %s", path, err)
 	}
+
+	d.SetId(path)
 	log.Printf("[DEBUG] Wrote PKI backend role %q", path)
 
 	return pkiBackendRoleRead(d, meta)
@@ -300,7 +299,9 @@ func pkiBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("ttl", resp.Data["ttl"])
 	d.Set("max_ttl", resp.Data["max_ttl"])
 	d.Set("allow_localhost", resp.Data["allow_localhost"])
-	d.Set("allowed_domains", domains)
+	if err := d.Set("allowed_domains", domains); err != nil {
+		return err
+	}
 	d.Set("allow_bare_domains", resp.Data["allow_bare_domains"])
 	d.Set("allow_subdomains", resp.Data["allow_subdomains"])
 	d.Set("allow_glob_domains", resp.Data["allow_glob_domains"])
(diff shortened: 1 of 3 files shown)


roboll:
Hey @vancluever I see you labeled this - anything I need to do to request a review?

vancluever:
Hey @roboll, we have been working on trying to keep all new issues and pull requests tagged as soon as possible when they come in, which is you saw the labels applied quickly. You don't need to do anything else to request a review - we will try to have a look as soon as we can!

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+				Default:     true,
+				Description: "Specifies if clients can request certificates for localhost as one of the requested common names.",
+			},
+			"allowed_domains": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "Policies to be set on tokens issued using this role.",
```
paddycarver: Is this the right description? The API docs seem to suggest otherwise.

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+				Description: "The maximum TTL of certificates issued using this role, provided as a duration.",
+			},
+			"allow_localhost": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Specifies if clients can request certificates for localhost as one of the requested common names.",
+			},
+			"allowed_domains": {
+				Type:     schema.TypeList,
```
paddycarver: If the order doesn't matter here, I might suggest a `schema.TypeSet` instead.

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Specifies if certificates are flagged for email protection use.",
+			},
+			"key_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "rsa",
+				Description: "Specifies the type of key to generate for generated private keys. Currently, rsa and ec are supported.",
```
paddycarver: Could we do a validation of this here? Here's an example of how that can be done: https://github.com/terraform-providers/terraform-provider-google/blob/ca6c39031f459828f32c69ef872ec8178e3996a5/google/data_source_google_compute_zones.go#L30

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+			},
+			"key_bits": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     2048,
+				Description: "Specifies the number of bits to use for the generated keys.",
+			},
+			"key_usage": {
+				Type:     schema.TypeList,
```
paddycarver: Is order relevant to the values here? Are the values going to be useful as interpolated values for other resources?

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+				Optional:    true,
+				Default:     false,
+				Description: "If set, certificates issued/signed against this role will not be stored in the storage backend.",
+			},
+			"backend": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Unique name of the PKI backend to configure.",
+				ForceNew:    true,
+				Default:     "pki",
```
paddycarver: I'm _fairly_ certain you saw someone (almost certainly me!) using this in other resources in the provider. I think we're trying to move away from setting a default backend, in favor of making it explicit, to avoid accidents. So I think this would be better as `Required: true` with no Default. Sorry for setting a bad example there.

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+
+func pkiBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string)
+	role := d.Get("role").(string)
+
+	path := pkiBackendRolePath(backend, role)
+
+	log.Printf("[DEBUG] Writing PKI backend role %q", path)
```
paddycarver: Could we move this down so it's right over the actual API call? Just in case future log lines get introduced, or in case of panic, it'd be more helpful to claim we're writing it when we're _actually_ going to be writing it, I think.

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+	data["use_csr_common_name"] = d.Get("use_csr_common_name").(bool)
+	data["use_csr_sans"] = d.Get("use_csr_sans").(bool)
+	data["ou"] = d.Get("ou").(string)
+	data["organization"] = d.Get("organization").(string)
+	data["generate_lease"] = d.Get("generate_lease").(bool)
+	data["no_store"] = d.Get("no_store").(bool)
+
+	_, err := client.Logical().Write(path, data)
+
+	d.SetId(path)
```
paddycarver: You probably picked this up looking at the pre-existing code, but I think we're trying to move to using `d.SetId` after checking the error, which means we don't need to call `SetId` in the error handling code.

vault/resource_pki_backend_role.go
```diff
@@ -0,0 +1,380 @@
...
+	for _, iUsage := range iUsages {
+		usages = append(usages, iUsage.(string))
+	}
+
+	d.Set("backend", backend)
+	d.Set("role", role)
+	d.Set("ttl", resp.Data["ttl"])
+	d.Set("max_ttl", resp.Data["max_ttl"])
+	d.Set("allow_localhost", resp.Data["allow_localhost"])
+	d.Set("allowed_domains", domains)
```
paddycarver: Again, sorry for setting a poor example here, but `d.Set` can actually return an error. And for primitives (strings, ints, bools, etc.) we don't bother checking it, because code gets unwieldy and there's pretty much never an error there. For complex types, though (Sets, Lists, Maps) we're moving towards checking errors. So if you could check and return the error for this and `key_usage`, that would be awesome.

vault/resource_pki_backend_role_test.go
```diff
@@ -0,0 +1,236 @@
...
+		resp, err := client.Logical().Read(endpoint)
+		if err != nil {
+			return fmt.Errorf("%q doesn't exist", endpoint)
+		}
+
+		attrs := map[string]string{
+			"ttl":     "ttl",
+			"max_ttl": "max_ttl",
+		}
+		for stateAttr, apiAttr := range attrs {
```
paddycarver: I'm fairly confident you're copying a bad idea of mine on this. I _do_ regret doing tests this way, and would rather not do them like this in the future. It _is_ a valid way to test though, so I don't want to block this PR on it, I just wanted to call out that I think we should move away from code like this in the future.

vault/resource_pki_backend_role.go
roboll: Hah, yea, copy-pasta. Fixed :+1:.

vault/resource_pki_backend_role.go
roboll: Sure, order does not matter, changed.

vault/resource_pki_backend_role.go
roboll: Yep! Thanks for the pointer.

vault/resource_pki_backend_role.go
roboll: Order isn't relevant here, no. Changed to a `schema.TypeSet`.

Yes, I think interpolation is useful for these, though I don't have a particular use case in mind.

vault/resource_pki_backend_role.go
roboll: NP. Changed.

vault/resource_pki_backend_role.go
roboll: Absolutely.

vault/resource_pki_backend_role.go
roboll: Makes sense. Done.

vault/resource_pki_backend_role.go
roboll: Sure, done.

vault/resource_pki_backend_role_test.go
roboll: 🤔 Yeah, I figured this was the "tf provider way". 

LMK if you think its important to change, otherwise I think everything else is good to go.

roboll:
@paddycarver I think I addressed everything in your review, thanks a lot for getting to it quickly. This should be ready for another look.

roboll:
@paddycarver ping - I think this is in a workable state, would love to move it forward if you have a few minutes to 👀.

irconan:
@roboll 
I rebased this branch against master to try it out, I had some trouble with role resources relating to the types around the allowed_domains field.
I have some patches on my fork which seem to fix it here:
https://github.com/irconan/terraform-provider-vault/tree/pki-role
I'm not sure if this relates to terraform/go version.

PR #76, opened 2018-02-18
new resource: resource_pki_generate

roboll:
Adds a new resource, `vault_pki_generate`, which calls one of the generate endpoints depending on the `type` argument: [root](https://www.vaultproject.io/api/secret/pki/index.html#generate-root) [intermediate](https://www.vaultproject.io/api/secret/pki/index.html#generate-intermediate).

The resource generates a key/certificate pair, and _optionally_ stores the private key / certificate material into a secret path.


roboll:
@paddycarver we spoke about this briefly the other day on slack, so you may be the right person to review this :-).

Issue #78, opened 2018-03-02
AWS secret backend nested path support


lmickh:
Hi there,

### Terraform Version
v0.11.3

### Affected Resource(s)
vault_aws_secret_backend_role

### Terraform Configuration Files
```
resource "vault_aws_secret_backend" "nested-aws-path" {
  access_key                = "${var.access_key}"
  secret_key                = "${var.secret_key}"
  path                      = "nested/aws/path"
 <snip>
}

resource "vault_aws_secret_backend_role" "admin" {
  backend = "${vault_aws_secret_backend.nested-aws-path.path}"
  name    = "admin"

  policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
}
```

### Expected Behavior
`nested/aws/path` is a valid mount point.  I expect it to create `nested/aws/path/roles/admin`.  This can be done via the Vault CLI.

### Actual Behavior
TF fails providing the following error:
`Invalid id nested/aws/path/roles/admin; must be {backend}/roles/{name}`

### Steps to Reproduce

1. `terraform apply`

martinssipenko:
I think there is a related bug.

I was able to create a role ins nested path (i see it in my `vault list aws/nested/roles` output), but Terraform reported the same error.

Looking at code I think this happens because check is performed in `awsSecretBackendRoleRead` func, which is called only after the resource has been created (at the end of `awsSecretBackendRoleWrite`). Further, running terraform plan (which uses the `awsSecretBackendRoleRead` func) it triggers that error.

lmickh:
It is resolved.  Closing it.

PR #79, opened 2018-03-02
Fix path check on secret role read

lmickh:
This is to address #78 

The code currently forces a `backend/roles/name` format when it should be able to take a nested path as the backend such as `backend/path/roles/name`.  This PR changes the check to ensure it is at least as large as it should be and that `roles` is the second to last of the slice.

### DIFF
--- a/vault/resource_aws_secret_backend_role_test.go
+++ b/vault/resource_aws_secret_backend_role_test.go
@@ -16,7 +16,7 @@ const testAccAWSSecretBackendRolePolicyArn_basic = "arn:aws:iam::123456789123:po
 const testAccAWSSecretBackendRolePolicyArn_updated = "arn:aws:iam::123456789123:policy/bar"
 
 func TestAccAWSSecretBackendRole_basic(t *testing.T) {
-	backend := acctest.RandomWithPrefix("tf-test-aws/nested")
+	backend := acctest.RandomWithPrefix("tf-test-aws")
 	name := acctest.RandomWithPrefix("tf-test-aws")
 	accessKey, secretKey := getTestAWSCreds(t)
 	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


martinssipenko:
This one would be very good to get released.

martinssipenko:
@lmickh looking at your change, it appears to me that it's only a partial fix.

If you look further down the code lines 105 and 106, you can see that `pathPieces[2]` is used for `name`, but in case of nested path the name would be the last part of `pathPieces`, not the one with index 2. I think it should be something like this:

```
d.Set("backend", strings.Join(pathPieces[:len(pathPieces)-2], "/"))
d.Set("name", pathPieces[len(pathPieces)-1])
```

paultyng:
It would be good to add some testing around that alternative path if possible as well in this PR.

lmickh:
@martinssipenko nice catch.  Not sure how I even missed that.

Thanks for the feedback.  I'm still in the process of sorting out testing.   Haven't done any testing in go before much less one of these providers.

martinssipenko:
Just use nested path over here:

https://github.com/lmickh/terraform-provider-vault/blob/213f441d4a5ac60e219e586d2df4d7c3f48e3f3f/vault/resource_aws_secret_backend_role_test.go#L19

lmickh:
Made the change.  Let me know if anything else needs to be done or if that isn't what you had in mind.

vault/resource_aws_secret_backend_role_test.go
```diff
@@ -16,7 +16,7 @@ const testAccAWSSecretBackendRolePolicyArn_basic = "arn:aws:iam::123456789123:po
 const testAccAWSSecretBackendRolePolicyArn_updated = "arn:aws:iam::123456789123:policy/bar"
 
 func TestAccAWSSecretBackendRole_basic(t *testing.T) {
-	backend := acctest.RandomWithPrefix("tf-test-aws")
+	backend := acctest.RandomWithPrefix("tf-test-aws/nested")
```
paultyng: While this works I think you should just copy/paste it to `TestAccAWSSecretBackendRole_nested` or something so that we are also still testing the original path as well.

martinssipenko:
Guys, I've noticed that tests are for this resource provider are broken, and it looks that no one has actually tried to run them :)

I've created a Pull Request to @lmickh to fix tests, and hopefully when he merges it the fixed tests will land into this PR.

martinssipenko:
@lmickh can you update your branch so it's not out of date with the base?

lmickh:
@martinssipenko I rebased it on master.  Let me know if you need anything else.

martinssipenko:
This is great, hoping this gets a release soon.

PR #80, opened 2018-03-06
Added new vault_token_auth_backend_role resource

martinssipenko:
This PR adds new resource `vault_token_auth_backend_role`, arguments taken from [Vault docs](https://www.vaultproject.io/api/auth/token/index.html#create-token).

### DIFF
--- a/vault/resource_token_auth_backend_role.go
+++ b/vault/resource_token_auth_backend_role.go
@@ -1,0 +1,280 @@
+package vault
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+var (
+	tokenAuthBackendRoleNameFromPathRegex = regexp.MustCompile("^auth/token/roles/(.+)$")
+)
+
+func tokenAuthBackendRoleResource() *schema.Resource {
+	return &schema.Resource{
+		Create: tokenAuthBackendRoleCreate,
+		Read:   tokenAuthBackendRoleRead,
+		Update: tokenAuthBackendRoleUpdate,
+		Delete: tokenAuthBackendRoleDelete,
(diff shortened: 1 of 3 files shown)


martinssipenko:
@paultyng any updates on this?

vault/resource_token_auth_backend_role.go
```diff
@@ -0,0 +1,280 @@
...
+	data["explicit_max_ttl"] = d.Get("explicit_max_ttl").(string)
+	data["ttl"] = d.Get("ttl").(string)
+	data["max_ttl"] = d.Get("max_ttl").(string)
+	data["orphan"] = d.Get("orphan").(bool)
+	data["renewable"] = d.Get("renewable").(bool)
+	data["path_suffix"] = d.Get("path_suffix").(string)
+
+	_, err := client.Logical().Write(path, data)
+
+	d.SetId(path)
```
tyrannosaurus-becks: Would it be possible to move this up above the `client.Logical().Write` call so the err below gets checked _right_ after the call? Or move it into an if/else for the err?

martinssipenko: @tyrannosaurus-becks Done

martinssipenko:
@tyrannosaurus-becks I've moved the `SetId` call up a bit as you suggested. Also, I think it's obvious that this provider gets way much less attention than others and if needed I'd be happy to help in any way possible. I think that me and others send PRs here to add missing features and fix bugs as they get found by building infrastructure, and very slow release cycle prevents us from using Terraform, which leads to manual work. It would be great if this could be improved somehow.

tyrannosaurus-becks:
@martinssipenko yes, there was about a 6-month period where this provider wasn't getting any attention. It was being shifted from the Terraform team to the Vault team. Apologies! Both teams had a lot of competing projects. It's now resting with me and I'm working through the backlog of PR's and issues. It'll take some time to get caught up, but definitely should be getting better starting now.

PR #83, opened 2018-03-12
Handle generic secret deletion outside TF

paultyng:


vault/resource_generic_secret.go
```diff
@@ -147,6 +146,10 @@ func genericSecretResourceRead(d *schema.ResourceData, meta interface{}) error {
 		if err != nil {
 			return fmt.Errorf("error reading from Vault: %s", err)
 		}
+		if secret == nil {
+			d.SetId("")
```
catsby: Any time something is removed from state because it wasn't found in a `READ` method, we should log a warning:

```go
if secret == nil {
  log.Printf("[WARN] secret (%s) not found, removing from state", d.Id())
  d.SetId("")
  return nil
}
```

vault/resource_generic_secret_test.go
```diff
@@ -29,6 +28,31 @@ func TestResourceGenericSecret(t *testing.T) {
...
+			resource.TestStep{
+				Config: testResourceGenericSecret_initialConfig(path),
+				Check:  testResourceGenericSecret_initialCheck(path),
+			},
+			resource.TestStep{
+				PreConfig: func() {
+					client := testProvider.Meta().(*api.Client)
+					_, err := client.Logical().Delete(path)
+					if err != nil {
+						t.Fatal(err)
```
paddycarver: Some more context about why this is a fatal condition would be helpful for future debugging of failed tests.

paddycarver:
Tests pass for me:

```
make testacc TEST=./vault TESTARGS='-run=TestResourceGenericSecret'
TF_ACC=1 go test ./vault -v -run=TestResourceGenericSecret -timeout 120m
--- PASS: TestResourceGenericSecret (1.01s)
--- PASS: TestResourceGenericSecret_deleted (0.85s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   1.872s
```

PR #85, opened 2018-03-22
Update vendored vault api library to fix recent vault compatability

irconan:
Using latest version of vault reading vault_policy resources will fail with the error:
error reading from Vault: rules not found in response

Updating github.com/hashicorp/vault/api should resolve this.

Not sure if I've followed the correct methods for vendoring in this repo so please let me know.


irconan:
Updating the vault library is actually not necessary for the problem I was seeing.

PR #86, opened 2018-03-23
Changing role in aws_auth_backend_role must force new resource

synthdnb:
Currently role_name is not updated after `terraform apply` because changing role_name changes vault path but ForceNew is not specified in schema. Adding ForceNew fixes this.


paddycarver:
Thanks for the fix, and sorry about that oversight. :(

PR #89, opened 2018-04-05
Consul role

SteveKMin:


SteveKMin:
Created on accident, going to run through an internal review first.

PR #90, opened 2018-04-06
 Add token_auth_backend_token resource.

paddycarver:
This is an update to #11 to make issuing Vault tokens with Terraform
possible.

The updates I made to this:

1. Renamed the resource from `vault_token` to
`vault_token_auth_backend_token`, to better match the pattern the rest
of the provider uses, and to make it easier for us to add token roles
later.

2. Removed the Update definition on the resource. All our fields are
`ForceNew`, we don't need that.

3. Added `num_uses` support.

4. Updated the Read function to pull from the API and set the state
accordingly, unless wrapped. If wrapped, we don't really get anything
back from the server, so ignore it.

5. Update the tests to be a bit more comprehensive.

6. Use Computed instead of Default.

7. Use the integer seconds versions, instead of strings with arbitrary
denominations, because Vault's API returns the integers, and there's no
way to go from integer to string.

8. Add a statefunc to display_name that transforms it the same way the
Vault API does.

9. Got rid of orphan. That's... complicated, and while I think it'd be
great to support it, I didn't want it to block this iteration. Next
iteration!

10. Got rid of no_default_policy, and turned it on by default. If people
want the default policy, they should specify it in their configs;
otherwise, Terraform is going to set _exactly_ what they asked for.

11. Added the link to the docs to the sidebar of the website.

There may be other minor changes in there, but that should be the main
gist of it.


vault/resource_token_auth_backend_token_test.go
```diff
@@ -0,0 +1,181 @@
+package vault
+
+import (
+	"strconv"
+	"testing"
+
```
paultyng: extra whitespace

tyrannosaurus-becks:
@paddycarver this looks great! Are you still interested in working on it at all? I didn't initially review it because I was still getting the hang of the provider, so instead went for smaller PRs. Because of that, [a token role resource](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L99) was merged that duplicates some of the code in this PR. 

Would you be willing to merge in master and pull out the duplicated code? If so, I'd approve and merge it right away as long as acceptance tests pass, with no further TODO's.

tyrannosaurus-becks:
Closing this because it appears inactive. If you end up circling back around to it, feel free to reopen it! Thank you!

Issue #102, opened 2018-04-16
[Bug/regression] Read vault_approle_auth_backend_role crashes for Vault 0.10.0


syndbg:
### Terraform Version

`Terraform v0.11.7`

### Vault version

`Vault v0.10.0 ('5dd7f25f5c4b541f2da62d70075b6f82771a650d')`

### Affected Resource(s)

- `vault_approle_auth_backend_role`

### Terraform Configuration Files

```hcl
provider "vault" {
  # NOTE: For example sake, this is a Vault container.
  address = "http://192.168.99.100:30002"
  token = "test"
}

resource "vault_auth_backend" "approle" {
  type = "approle"
}

resource "vault_policy" "newrelic" {
  name = "newrelic"
  policy = <<EOT
path "secret/external/newrelic/*" {
  policy = "read"
}
EOT
}

resource "vault_policy" "transactions" {
  name = "transactions"
  policy = <<EOT
path "secret/transactions/*" {
  policy = "read"
}
EOT
}

resource "vault_approle_auth_backend_role" "transactions" {
  role_name = "transactions"
  policies = ["${vault_policy.transactions.name}", "${vault_policy.newrelic.name}"]

  depends_on = ["vault_policy.newrelic", "vault_policy.transactions"]
}
```

### Debug Output

https://gist.github.com/syndbg/ace5e589beacc809d8efb2e6b34e6be4

### Panic Output

https://gist.github.com/syndbg/d9b5e4fc287c043b93377bb609d0c42f

### Expected Behavior

It must successfully read the approle data returned by Vault and proceed with provision.

### Actual Behavior

Crashed during reading of approle data.

### Steps to Reproduce

1. `terraform apply`
2. Write input `yes`
3. Observe

### Important Factoids

This issue is not reproducible for Vault versions `0.8.x` and  `0.9.x`.

However it's 100% of the time reproducible when running the acceptance test suite against a Vault server with version `0.10.0`.

### References

~~None at the moment. I'll submit a PR soon.~~

* https://github.com/terraform-providers/terraform-provider-vault/pull/103 (PR)

syndbg:
Btw, it was and still is ready for review.

chrissam:
I'm facing the same issue. Terraform crashes while running _vault_approle_auth_backend_role_

Terraform v0.11.7
+ provider.vault v1.1.0
Vault v0.10.1 ('756fdc4587350daf1c65b93647b2cc31a6f119cd')

[...]

PR #103, opened 2018-04-16
[GH-102] Fix read approle_auth_role regression

syndbg:
Fixes https://github.com/terraform-providers/terraform-provider-vault/issues/102

@paddycarver


noeljackson:
@paddycarver When is the next expected release that will include this fix? Thanks!

elasticdog:
:+1: I'm also hitting this bug and was diving into how to fix it before I saw this PR. Would love an official release as well.

Issue #111, opened 2018-05-01
Feature Req: Return data from Vault Write


rorybrowne:
Howyas

It would be useful if vault_generic_secret returned the data that's returned from a vault write call. 

If for example I run vault write pki/root/generate/internal it will return me a data structure with the certificate, expiration, issuing_ca, and serial_number. It would be useful if the vault_generic_secret resource did the same.

Rory

jcfergus:
I created a PR (#120) that should address this issue (it does for me), if anyone wants to take a look and see if it works for them.

Issue #67, opened 2018-02-07
Unable to access CSR generated by vault_generic_secret


thevilledev:
### Terraform Version

- Terraform v0.11.3 
- `provider.vault` v1.0.0
- Vault v0.9.3

### Affected Resource(s)

Just `vault_generic_secret` (link to [docs](https://www.terraform.io/docs/providers/vault/r/generic_secret.html)).

### Terraform Configuration Files
```hcl
resource "vault_mount" "some_pki" {
    path = "some_pki"
    type = "pki"
    default_lease_ttl_seconds = "864000"
    max_lease_ttl_seconds     = "2592000" 
}

resource "vault_generic_secret" "intermediate_ca" {
    path = "${vault_mount.some_pki.path}/intermediate/generate/internal"
    disable_read = true
    data_json = <<EOF
{
    "common_name": "Some intermediate CA",
    "key_type": "ec",
    "key_bits": "384",
    "exclude_cn_from_sans": true
}
EOF
}
```

### Debug Output

https://gist.github.com/vtorhonen/97d37eb5ff9e2aba3afaeccaeafec5c6

HTTP request:

```
PUT /v1/some_pki/intermediate/generate/internal HTTP/1.1
Host: vault.service.consul:8200
User-Agent: Go-http-client/1.1
Content-Length: 100
X-Vault-Token: 3b957cec-862c-f883-1908-126b96623f89
Accept-Encoding: gzip
Connection: close

{"common_name":"Some intermediate CA","exclude_cn_from_sans":true,"key_bits":"384","key_type":"ec"}
```

HTTP response:

```
HTTP/1.1 200 OK
Cache-Control: no-store
Content-Type: application/json
Date: Wed, 07 Feb 2018 19:49:45 GMT
Content-Length: 626
Connection: close

[...]

PR #120, opened 2018-05-11
Added functionality to capture response from vault generic secret resources.

jcfergus:
This adds a "capture_response" setting to vault_generic_secret resources, which allows data that is only returned in a response to a Create/Write call to be accessed in the resultant data_json and data attributes.  (i.e. CSRs returned from pki endpoints and certificates returned from signing requests). This should provide a workable solution for #67 and #111  .  

[Note that this is my first foray into Go programming and has only been lightly tested.].

### DIFF
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -133,7 +133,6 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error 
 			return nil
 		}
 		jsonDataBytes, err := json.Marshal(secret.Data)
-		log.Printf("[DEBUG] Secret is %s", jsonDataBytes)
 		if err != nil {
 			return fmt.Errorf("Error marshaling JSON for %q: %s", path, err)
 		}
@@ -148,20 +147,20 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error 
 		for k, v := range secret.Data {
 			if vs, ok := v.(string); ok {
 				dataMap[k] = vs
-			} else {
-				// Again ignoring error because we know this value
-				// came from JSON in the first place and so must be valid.
-				vBytes, _ := json.Marshal(v)
-				dataMap[k] = string(vBytes)
+				continue
 			}
+			// Again ignoring error because we know this value
+			// came from JSON in the first place and so must be valid.
+			vBytes, _ := json.Marshal(v)
+			dataMap[k] = string(vBytes)
 		}
 		d.Set("data", dataMap)
 
 		d.Set("path", path)
 		return nil
-	} else {
-		return genericSecretResourceRead(d, meta)
 	}
+
+	return genericSecretResourceRead(d, meta)
 }
 
 func genericSecretResourceDelete(d *schema.ResourceData, meta interface{}) error {
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -44,6 +44,13 @@ func genericSecretResource() *schema.Resource {
 				ValidateFunc: ValidateDataJSON,
 			},
 
+			"data": &schema.Schema{
+				Type:		  schema.TypeMap,
+				Required:	  false,
+				Description:  "Data returned from the resource.  Should be a map containing the content from data_json.",
+				Computed:	  true,
+			},
+
 			"allow_read": &schema.Schema{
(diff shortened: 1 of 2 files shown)


vault/resource_generic_secret.go
```diff
@@ -104,14 +118,50 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error
...
-	return genericSecretResourceRead(d, meta)
+	captureResponse := d.Get("capture_response").(bool)
+	if captureResponse {
+		log.Printf("[DEBUG] Capture response is set.")
+		if secret == nil {
+			log.Printf("[WARN] write to (%s) returned empty response", path)
+			return nil
+		}
+		jsonDataBytes, err := json.Marshal(secret.Data)
+		log.Printf("[DEBUG] Secret is %s", jsonDataBytes)
```
seanschade: I don't know if you want to log this even under `DEBUG`.

vault/resource_generic_secret.go
```diff
@@ -104,14 +118,50 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error
...
+		// Since our "data" map can only contain string values, we
+		// will take strings from Data and write them in as-is,
+		// and write everything else in as a JSON serialization of
+		// whatever value we get so that complex types can be
+		// passed around and processed elsewhere if desired.
+		dataMap := map[string]string{}
+		for k, v := range secret.Data {
+			if vs, ok := v.(string); ok {
+				dataMap[k] = vs
+			} else {
```
seanschade: You can get rid of the `else` block if you put a `continue` in the `if` block.

```
if vs, ok := v.(string); ok {
	dataMap[k] = vs
        continue
}
// Again ignoring error because we know this value
// came from JSON in the first place and so must be valid.
vBytes, _ := json.Marshal(v)
dataMap[k] = string(vBytes)
```

vault/resource_generic_secret.go
```diff
@@ -104,14 +118,50 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error
...
+				// came from JSON in the first place and so must be valid.
+				vBytes, _ := json.Marshal(v)
+				dataMap[k] = string(vBytes)
+			}
+		}
+		d.Set("data", dataMap)
+
+		d.Set("path", path)
+		return nil
+	} else {
```
seanschade: You don't need this `else` either since the `if` block above returns.

vault/resource_generic_secret.go
jcfergus: Oops, meant to remove that.  :-O  Will take care of that.

jcfergus:
@seanschade Thanks!  Addressed those comments.  

@rorybrowne I haven't had a chance to look at your solution yet; still bogged down in the project that I need the functionality for.  I looked at the tests, and couldn't find any good examples that would be applicable to this particular use case.   I will take a look and revisit those in a bit.  

As far as the lease details, that makes sense - the case I was interested initially the lease information was irrelevant and I was trying to keep it simple.  I'll see about adding that or merging in what you've done.

stefan-caraiman:
Any updates regarding this one guys?

vault/resource_generic_secret.go
```diff
@@ -104,13 +118,48 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error
...
+	captureResponse := d.Get("capture_response").(bool)
+	if captureResponse {
+		log.Printf("[DEBUG] Capture response is set.")
+		if secret == nil {
+			log.Printf("[WARN] write to (%s) returned empty response", path)
+			return nil
+		}
+		jsonDataBytes, err := json.Marshal(secret.Data)
+		if err != nil {
+			return fmt.Errorf("Error marshaling JSON for %q: %s", path, err)
```
tyrannosaurus-becks: Could this error start with a lower-case character instead?

tyrannosaurus-becks:
@jcfergus hi! Seems like this PR is inactive, so I'm closing for now. If you get a chance to circle back to it, feel free to reopen it. Thank you so much!

pschulten:
---
**NOTE TO FUTURE SELF**

( when I'm here)

* do not do this
* always create certificates during runtime (with sidecars, initContainers, ...)
* make them short-lived

---

PR #123, opened 2018-05-14
New Resource: vault_cert_auth_backend_role

angrylogic:
This PR allows for managing the Vault Cert authentication backend.

### DIFF
--- a/vault/resource_cert_auth_backend_role.go
+++ b/vault/resource_cert_auth_backend_role.go
@@ -20,17 +20,17 @@ func certAuthBackendRoleResource() *schema.Resource {
 		Delete: certAuthResourceRead,
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-			"certificate": &schema.Schema{
+			"certificate": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-			"allowed_names": &schema.Schema{
+			"allowed_names": {
 				Type: schema.TypeSet,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
@@ -38,7 +38,7 @@ func certAuthBackendRoleResource() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
-			"required_extensions": &schema.Schema{
+			"required_extensions": {
 				Type: schema.TypeSet,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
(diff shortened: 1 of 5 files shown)


kmcquade:
@angrylogic maybe if you add the new-resource label then they'll pay attention to this? I know they're overloaded at this time but maybe it would make a visual statement that Hashi should prioritize this provider if we had a ton of "new resource" labels applied to the relevant open PRs.

vault/resource_cert_auth_backend_role.go
```diff
@@ -0,0 +1,264 @@
...
+	return &schema.Resource{
+		SchemaVersion: 1,
+
+		Create: certAuthResourceWrite,
+		Update: certAuthResourceUpdate,
+		Read:   certAuthResourceRead,
+		Delete: certAuthResourceRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
```
tyrannosaurus-becks: Would it be possible to strip the redundant field name declarations here and on the rest of the fields?

vault/resource_cert_auth_backend_role.go
```diff
@@ -0,0 +1,264 @@
...
+		data["policies"] = v.(*schema.Set).List()
+	}
+
+	if v, ok := d.GetOk("display_name"); ok {
+		data["display_name"] = v.(string)
+	}
+
+	log.Printf("[DEBUG] Writing %q to cert auth backend", path)
+	_, err := client.Logical().Write(path, data)
+	d.SetId(path)
```
tyrannosaurus-becks: Could this be moved up a line so it's not between the err and the err check?

vault/resource_cert_auth_backend_role.go
angrylogic: @tyrannosaurus-becks - I'm not sure I follow what you mean re: redundant field names here.

vault/resource_cert_auth_backend_role.go
angrylogic: No problem, will fix.

angrylogic:
I'll go ahead and get those tests and changes sorted out early next week - same for the other PR.  Thanks for taking a look at these @tyrannosaurus-becks !

vault/resource_cert_auth_backend_role.go
tyrannosaurus-becks: Sorry, I didn't phrase it very well. My code editor complains if, in a map like `map[string]*schema.Schema`, then `&schema.Schema` is also declared below. Here's a visual example:
screenshot from 2018-09-14 13-54-19
It's underlining and greying out those words because they can be safely omitted.

I know it hasn't always been that way, but it would be great to go that way going forward. I'm thinking of doing a quick pass through the pre-existing code updating stuff like that.

angrylogic:
@tyrannosaurus-becks -- Rebased and I think I've covered all the requested changes.  

Tests passing for me now:

```
BOS-M-2533:vault gharris$ go test -test.v -run TestCertAuthBackend_basic
=== RUN   TestCertAuthBackend_basic
--- PASS: TestCertAuthBackend_basic (0.18s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.926s
```

Thanks!

PR #127, opened 2018-05-20
aj/consul secret roles

amanjeev:
Adds `consul secret` engine.


amanjeev:
I guess I should have looked at the open PRs first. :laughing: 
I see https://github.com/terraform-providers/terraform-provider-vault/pull/91 already in the list so I am going to close this.

PR #129, opened 2018-05-22
Fix #128: vault_aws_auth_backend_role.resolve_aws_unique_ids setting not detected

### LINKED ISSUE #128: vault_aws_auth_backend_role.resolve_aws_unique_ids setting not detected
### Terraform Version

Terraform v0.11.7

### Affected Resource(s)

- vault_aws_auth_backend_role

```hcl
resource "vault_auth_backend" "backend" {
  type = "aws"
  path = "test-iam-auth"
}

resource "vault_aws_auth_backend_role" "broken-role" {
  backend = "${vault_auth_backend.backend.path}"
  role = "broken-role"
  auth_type = "iam"
  resolve_aws_unique_ids = false
  bound_iam_principal_arn = "arn:aws:iam::123456789012:role/MyLambdaRole"
}

resource "vault_generic_secret" "working-role" {
  path = "auth/test-iam-auth/role/working-role"
  data_json = <<JSON
{
  "auth_type": "iam",
  "resolve_aws_unique_ids": false,
  "bound_iam_principal_arn": "arn:aws:iam::123456789012:role/MyLambdaRole",
}
JSON
}
```

### Expected Behavior

Both the "working-role" and "broken-role" resources should successfully create roles for the AWS auth backend "test-iam-auth" which does not have any AWS creds configured.

### Actual Behavior

Got this error when applying the "broken-role" resource:

```failed updating the unique ID of ARN "arn:aws:iam::123456789012:role/MyLambdaRole": &errors.errorString{s:"unable to fetch client for account ID 123456789012 -- default client is for account 000011112222"}```

The error indicates that the `resolve_aws_unique_ids` setting is not being sent to the API. The "working-role" generic write works, as does manual invocation.

### Notes

I think I know what the solution is. I'm filing this to refer back to in the PR.

daveadams:
Per the schema.ResourceData docs[0], GetOkExists is needed for properly detecting whether optional boolean values have been set.

[0] https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceData.GetOkExists

### DIFF
--- a/vault/resource_aws_auth_backend_role.go
+++ b/vault/resource_aws_auth_backend_role.go
@@ -211,7 +211,7 @@ func awsAuthBackendRoleCreate(d *schema.ResourceData, meta interface{}) error {
 		if v, ok := d.GetOk("role_tag"); ok {
 			data["role_tag"] = v.(string)
 		}
-		if v, ok := d.GetOk("allow_instance_migration"); ok {
+		if v, ok := d.GetOkExists("allow_instance_migration"); ok {
 			data["allow_instance_migration"] = v.(bool)
 		}
 		if v, ok := d.GetOk("disallow_reauthentication"); ok {
@@ -228,7 +228,7 @@ func awsAuthBackendRoleCreate(d *schema.ResourceData, meta interface{}) error {
 		if v, ok := d.GetOk("inferred_aws_region"); ok {
 			data["inferred_aws_region"] = v.(string)
 		}
-		if v, ok := d.GetOk("resolve_aws_unique_ids"); ok {
+		if v, ok := d.GetOkExists("resolve_aws_unique_ids"); ok {
 			data["resolve_aws_unique_ids"] = v.(bool)
 		}
 	}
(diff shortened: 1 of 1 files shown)


vancluever:
@daveadams thanks for this!

The only issue with `GetOkExists` currently is that it actually does not work beyond the first apply. You can find detail [here](https://github.com/hashicorp/terraform/pull/17557) - the gist being that while the first apply will work as intended, the moment the value hits state, "exists" will always be true, regardless of what is in config at any point in time.

I apologize for the misleading docs. :confused: 

I'm noodling on whether or not this is good to add given that it gets us at least part of the way there or if it might be better to do it a different way. I also have a workaround that employs using strings for boolean values instead, but let me check with some of my colleagues to see if this is something that would be worthwhile recommending (especially since later versions of Terraform will be addressing the lack of nils completely).

Thanks!

tyrannosaurus-becks:
Hi @vancluever ! Thanks for pointing that out! Was a field ever added that would be more appropriate?

tyrannosaurus-becks:
In this case, I think it would be better to default these values to the same defaults as found in Vault. Thank you for surfacing this! This PR seems inactive so I'm going to close it, and do a separate PR defaulting them.

PR #131, opened 2018-05-24
Fix a misnamed resource and data source

ncorrare:
After testing the provider there were a couple of simple errors in the documentation.

The data source vault_approle_auth_backend_role_id was missing the _id in the title.
The resource vault_approle_auth_backend_login was quoted in the documentation as vault_approle_auth_backend_role_login, which was proven erroneous.

### DIFF
--- a/website/docs/r/approle_auth_backend_login.md
+++ b/website/docs/r/approle_auth_backend_login.md
@@ -1,12 +1,12 @@
 ---
 layout: "vault"
-page_title: "Vault: vault_approle_auth_backend_role_login resource"
-sidebar_current: "docs-vault-approle-auth-backend-role-login"
+page_title: "Vault: vault_approle_auth_backend_login resource"
+sidebar_current: "docs-vault-approle-auth-backend-login"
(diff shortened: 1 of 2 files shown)


website/docs/r/approle_auth_backend_login.md
```diff
@@ -1,12 +1,12 @@
 ---
 layout: "vault"
-page_title: "Vault: vault_approle_auth_backend_role_login resource"
-sidebar_current: "docs-vault-approle-auth-backend-role-login"
+page_title: "Vault: vault_approle_auth_backend_login resource"
+sidebar_current: "docs-vault-approle-auth-backend-login"
```
tombuildsstuff: looks like this [should be `docs-vault-resource-approle-auth-backend-login`](https://github.com/terraform-providers/terraform-provider-vault/blob/master/website/vault.erb#L39)

tyrannosaurus-becks:
@ncorrare if you fix that one nit, I'd be happy to approve and merge!

Issue #134, opened 2018-05-30
vault_aws_auth_backend_role missing bound_ec2_instance_id parameter


JayH5:
### Terraform Version
`Terraform v0.11.7`

(Vault provider v1.1.0)

### Affected Resource(s)
-`vault_aws_auth_backend_role`

### Expected Behavior
`vault_aws_auth_backend_role` should have the `bound_ec2_instance_id` parameter.

### Actual Behavior
`vault_aws_auth_backend_role` does not have the `bound_ec2_instance_id` parameter.

### References
- GH-92
This parameter was added to Vault 0.9.6 (hashicorp/vault#3816). Another change to the Vault API in 0.9.6 for this resource is described in GH-92 (the bound parameters changed from strings to lists - hashicorp/vault#3907).

Issue #92, opened 2018-04-09
EC2 constraints for `vault_aws_auth_backend_role` should be lists


gregsymons:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
0.11.4

### Affected Resource(s)
Please list the resources as a list, for example:
- `vault_aws_auth_backend_role`

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "vault_aws_auth_backend_role" "system" {
  role = "system"
  auth_type = "ec2"
  bound_ami_id = [ "${concat(data.aws_ami_ids.released_amis.ids, data.aws_ami_ids.testing_amis.ids)}" ]
  policies = "system"
}
```

### Expected Behavior
The list of AMI ids should be applied as constraints to the new role.

### Actual Behavior
Planning fails:

```
Acquiring state lock. This may take a few moments...
Releasing state lock. This may take a few moments...

Error: vault_aws_auth_backend_role.system: bound_ami_id must be a single value, not a list
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

### Important Factoids
_All_ of the EC2 auth constraints for the AWS auth background take lists, according to the [api docs](https://www.vaultproject.io/api/auth/aws/index.html#create-role). But looking at the source (and the TF docs) they're all implemented as strings for Terraform.

gregsymons:
Turns out you can workaround it by specifying a JSON array for the string, but that seems awkward.

paddycarver:
If I'm reading this right, the ability to use a list instead of a single item was added in hashicorp/vault#3907, which was authored three months after the resource was, which at least explains the discrepancy. 😄 

If I'm reading this right, the list feature would only be available in Vault 0.9.6, which is (at the time of this writing) the latest non-RC Vault release.

My personal 2¢ on the topic is that we have three options:

A) leave it as it is. It's definitely ergonomically weird, but there is a sensible workaround. Perhaps we could document it.

[...]

PR #135, opened 2018-05-30
vault_aws_auth_backend_role: Add bound_ec2_instance_id parameter

JayH5:
Fixes GH-134

I've left the parameter as a string instead of a list, so that the fix for GH-92 can come after this and handle all the bind parameters at once.


tyrannosaurus-becks:
Actually, https://github.com/terraform-providers/terraform-provider-vault/pull/153 may fix them, I'll play with it and see.....

tyrannosaurus-becks:
@JayH5 would you be willing to merge in master? Then I could re-run the tests and see where they're landing. If they're passing, happy to approve and merge.

JayH5:
@tyrannosaurus-becks thanks 🙏, I've merged master.

PR #137, opened 2018-06-08
Make child token generation optional

douglasfraser-cloudreach:
Add argument to stop Terraform from attempting to issue itself a new token that is a child of the one given by the `token` argument. This is necessary where the policy is not to allow update capability in order to create child tokens.

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -28,6 +28,12 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_TOKEN", ""),
 				Description: "Token to use to authenticate to Vault.",
 			},
+			"use_child_token": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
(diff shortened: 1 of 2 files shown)


tyrannosaurus-becks:
Hi @douglasfraser ! Thank you for this PR! Apologies it has taken so long to get to - this repo was in transition from the Terraform team to the Vault team. 

I'm very hesitant on this PR because my hope is to _increase_ the security level of the Vault provider, rather than decrease it. Can you elaborate on your use case and why its benefit would outweigh the potential security compromise?

My concern is that it would be so easy to set this flag, and then risk surface area would be increased dramatically due to long-lasting secrets.

douglasfraser-cloudreach:
I would have no strong objection if this PR was dropped now. I had to work with a vault installation where the administrators were not willing to implement policy from #84 . I am no longer on this project.

If helps to provide use case of sorts: This was for writing secrets retrieved from other terraform providers so vault dynamic secret exposure was not really our concern, we just wanted to use terraform a conduit for secrets generated in non-vault systems into vault (as a central secret storage). 

Please indicate if I seem to be misunderstanding the protection intended by child token generation.

PR #138, opened 2018-06-08
Correct "minutes" to "seconds" where docs are wrong

stephansnyt:
Manual testing shows that the value is actually seconds, not minutes


syndbg:
Duplicate of #100.

It would be nice if reviewers/maintainers can review and merge some of the stalling PRs.

tyrannosaurus-becks:
Thank you for submitting this! You're absolutely right. I ended up merging #100 so I'm going to go ahead and close this.

PR #148, opened 2018-07-05
Do not return error in the case of an expired accessor

### LINKED ISSUE #97: Unable to destroy secret_id after it is consumed
### Terraform Version
Terraform v0.11.7
+ provider.vault v1.1.0

### Vault Version
0.9.1

### Affected Resource(s)
- vault_approle_auth_backend_role_secret_id

```
resource "vault_approle_auth_backend_role" "app_role" {
  backend   = "approle"
  role_name = "app_role_test"
  policies  = ["${vault_policy.policy_approle.name}"]

  secret_id_num_uses = "1"
  secret_id_ttl      = "${var.secret_ttl}"
}

resource "vault_approle_auth_backend_role_secret_id" "app_secret_id" {
  backend   = "${vault_approle_auth_backend_role.app_role.backend}"
  role_name = "${vault_approle_auth_backend_role.app_role.role_name}"
}
```

### Expected Behavior
Terraform should not throw an error during the destroy process if the secret_id has been consumed in the vault.

### Actual Behavior
```
Error: Error refreshing state: 1 error(s) occurred:

* module.create_vault_approle.vault_approle_auth_backend_role_secret_id.app_secret_id: 1 error(s) occurred:

* module.create_vault_approle.vault_approle_auth_backend_role_secret_id.app_secret_id: vault_approle_auth_backend_role_secret_id.app_secret_id: Error checking if AppRole auth backend role SecretID "backend=approle::role=app_role_test::accessor="abcd1234" exists: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/approle/role/app_role_test/secret-id-accessor/lookup
Code: 500. Errors:

* 1 error occurred:

* failed to find accessor entry for secret_id_accessor:
```

### Steps to Reproduce
1. `terraform apply`
2. curl --request POST --data "{ \"role_id\" : \"$ROLE_ID\", \"secret_id\" : \"$SECRET_ID\" }" http://127.0.0.1:8200/v1/auth/approle/login
3. `terraform destroy`

jorgemarey:
Hi,

This PR makes changes to `vault_approle_auth_backend_role_secret_id` and `vault_approle_auth_backend_login`. It checks if the accessor has expired and in that case it doesn't return an error.

I don't know if the implementation is the best (it checks the error message), but I didn't find another solution for this. Let me know if someone finds a better solution for this. I'll change the code.

Should fix #97


vault/resource_approle_auth_backend_login.go
```diff
@@ -120,6 +120,10 @@ func approleAuthBackendLoginRead(d *schema.ResourceData, meta interface{}) error
 	log.Printf("[DEBUG] Reading token %q", d.Id())
 	resp, err := client.Auth().Token().LookupAccessor(d.Id())
 	if err != nil {
+		// If the token is not found (it has expired) we don't return an error
+		if strings.Contains(err.Error(), "invalid accessor") { // Not found
```
tyrannosaurus-becks: Since this check occurs twice with the same string, and twice with a different string, it would be great to zip it up into a function like `isExpiredTokenErr(err error) bool`, and have its internals just check for both strings.

jorgemarey: Yeah, you're right. I'll merge in master and add this change!

Issue #104, opened 2018-04-16
Feature request: please support ssh secret backend resources


rvandegrift:
It'd be nice if I could enable the ssh secret backend and manage our roles via terraform.

### Terraform Version
Terraform v0.11.5

### Affected Resource(s)
No resources exist yet.

PR #163, opened 2018-08-29
Add SSH secret backend CA resource

arnogeurts:
This pull request partly implements #104 

It can be used to provision Vault as a certificate authority:
https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates.html


vault/resource_ssh_secret_backend_ca.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func sshSecretBackendCACreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string)
+	generateSigningKey := d.Get("generate_signing_key").(bool)
+
+	data := map[string]interface{}{
+		"generate_signing_key": generateSigningKey,
```
tyrannosaurus-becks: I was looking at the code this feeds into [here](https://github.com/hashicorp/vault/blob/df18871704fe869e9be45b542a6b1eb2fe46c293/builtin/logical/ssh/path_config_ca.go#L154-L191) and it actually behaves differently based on whether the `generate_signing_key` flag was _explicitly_ set to false or simply omitted. So, I think that for this code to behave as expected, we would probably want to only include this flag _if it were set_.

That would mean that above, we'd want to do: 
```
generateSigningKey, ok := d.Get("generate_signing_key").(bool)
if ok {
    data["generate_signing_key"] = generateSigningKey
}
```

vault/resource_ssh_secret_backend_ca.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	if !generateSigningKey {
```
tyrannosaurus-becks: I think that this if shouldn't exist so the code linked above will be fed the way it wants to be.

vault/resource_ssh_secret_backend_ca.go
```diff
@@ -0,0 +1,140 @@
...
+		privateKey := d.Get("private_key").(string)
+		publicKey := d.Get("public_key").(string)
+		if privateKey == "" || publicKey == "" {
+			return fmt.Errorf("When generate_signing_key is false, both public_key and private_keys should be specified")
```
tyrannosaurus-becks: Is this error needed in light of how the code linked above actually does allow this case? 

It's useful, though, because we'd only want to populate the public and private key if they were actually provided, in this particular case.

vault/resource_ssh_secret_backend_ca.go
arnogeurts: I removed this check, so all values are provided to the Vault API if they are provided in the configuration. The Vault API / SDK can decide on whether a given configuration is correct.

vault/resource_ssh_secret_backend_ca.go
arnogeurts: See previous comment.

PR #167, opened 2018-09-03
Add TTL for vault_aws_access_credentials

alsar1980:


vault/data_source_aws_access_credentials.go
```diff
@@ -46,6 +46,14 @@ func awsAccessCredentialsDataSource() *schema.Resource {
 					return
 				},
 			},
+
+			"ttl": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "60m",
```
tyrannosaurus-becks: I like that you set this default and I can see you got it from [here](https://github.com/hashicorp/vault/blob/master/builtin/logical/aws/path_user.go#L32). However, Vault actually has [a secret TTL hierarchy](https://www.vaultproject.io/docs/concepts/tokens.html) that will be overridden if an explicit TTL is passed in by the Vault provider.

Would you be willing to strip this default, and then below, only append the TTL when it's explicitly provided by the user?

vault/data_source_aws_access_credentials_test.go
```diff
@@ -82,26 +84,28 @@ data "vault_aws_access_credentials" "test" {
...
+
 }
 
 data "vault_aws_access_credentials" "test" {
     backend = "${vault_aws_secret_backend.aws.path}"
     role = "${vault_aws_secret_backend_role.role.name}"
-    type = "sts"
-}`, mountPath, accessKey, secretKey)
+		type = "sts"
+		ttl  = "%s"
```
tyrannosaurus-becks: Would it be possible to have both a positive and a negative test for the TTL? One with it and one without it ensuring that it's simply not passed on to Vault when it's not explicitly set by the user?

tyrannosaurus-becks:
Closing this for now as it appears inactive. If you get a chance to circle back around, please feel free to reopen it. Thank you!

PR #170, opened 2018-09-13
Deprecate KV V1, update to KV V2

tyrannosaurus-becks:
`TestDataSourceGenericSecret` was failing for me and it was because of an incompatibility between KV V1, which the code was written for, and KV V2, the current version of KV.

This PR:
- Fixes the failing test
- Adds support for KV V2
- Removes support for KV V1
- Lower-cases a couple of error strings I encountered along the way

### DIFF
--- a/vault/data_source_generic_secret.go
+++ b/vault/data_source_generic_secret.go
@@ -72,14 +72,19 @@ func genericSecretDataSourceRead(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("error reading from Vault: %s", err)
 	}
 	if secret == nil {
-		return fmt.Errorf("No secret found at %q", path)
+		return fmt.Errorf("no secret found at %q", path)
 	}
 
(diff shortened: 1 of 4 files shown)


tyrannosaurus-becks:
@catsby another one for you!

ShimiTaNaka:
I agree with @catsby , ATM, even when you mount the KV store (not "generic"), you can still use it un-versioned, which is compatible (to my understanding) with the KV v1, as you can see here: https://www.vaultproject.io/docs/secrets/kv/index.html
More over, the default of the KV store is the v1, not the v2.
Since it's clearly stated that the v1 does in fact have advantages over the v2, I am not so sure it will go away soon

tyrannosaurus-becks:
I think you both make good points! I'm going to close this.

eedwards-sk:
is there no v2 support in terraform right now?

I'd like to use both, so if there's no feature request we should add one.

PR #184, opened 2018-09-20
Update `bindpass` in `vault_ldap_auth_backend`

lawliet89:
- Set the attribute to sensitive
- Do not read back from the API since the API does not return it

I am not sure if this is the right way to handle it. I tried to mimic how this is done in the `vault_aws_secret_backend` resource.


lawliet89:
I rebased, and had to fix `go fmt` for one of the unrelated files.

Added `bindpass` to the fixture resource created and added an extra test to check that it's not returned by Vault. Please see if this is OK.

PR #189, opened 2018-09-21
Backwards compatibility for AWS roles

### LINKED ISSUE #177: Version 1.1.3 aws_auth_backend_role breaking changes
### Terraform Version
```
$ terraform version
Terraform v0.11.8
+ provider.vault v1.1.3
```

### Affected Resource(s)
- aws_auth_backend_role
(There may be others)

```hcl
provider "vault" {
  version = "~> 1.0"
  address = "https://localhost:8200"
}

resource "vault_aws_auth_backend_role" "my_role" {
  role                           = "my-role"
  auth_type                      = "ec2"
  bound_account_id               = "123456789012"
  bound_region                   = "us-west-2"
  bound_vpc_id                   = "vpc-xxxxxxxx"
  bound_iam_instance_profile_arn = "arn:aws:iam::123456789012:instance-profile/my-instance-profile"
  ttl                            = 60
  max_ttl                        = 120
  policies                       = ["default", "my-policy"]
}
```

### Debug Output
The issue appears to be related to versioning, I'm not sure debug output is necessary here.

### Panic Output

There was no panic.

### Expected Behavior

Terraform plan & apply should still work
This issue appears where we have existing resources written to use the `String` type, where the new version of the plugin expects these to be a `List` type. We would expect that the old configurations still work when using the `~> 1.0` version specification for the `vault` provider.

### Actual Behavior
```
$ terraform plan
﻿
Error: vault_aws_auth_backend_role.my_role: bound_account_id: should be a list



Error: vault_aws_auth_backend_role.my_role: bound_iam_instance_profile_arn: should be a list



Error: vault_aws_auth_backend_role.my_role: bound_region: should be a list



Error: vault_aws_auth_backend_role.my_role: bound_vpc_id: should be a list

```

### Steps to Reproduce
With the terraform listed above:
1. `terraform init`
2. `terraform plan`

### Important Factoids
Nothing comes to mind.

### References
It looks like the issue appeared from this commit: https://github.com/terraform-providers/terraform-provider-vault/commit/768c03ba2fa3868fc3915f655884d5bdde68f1e5#diff-a97ccbd83928090a41b6c44716cd9f18L45

tyrannosaurus-becks:
Adds backwards compatibility for #153 .
Fixes #177 .

Rather than using a state migration, this PR adds fields of two types with different names - one plural as a list, and one singular as a string. That's because in testing, I found that calls were failing before reaching the state migration code due to a different field type being expected by the schema.


idubinskiy:
Seems reasonable to me. One question/concern: since the values are being read into both versions of the fields, will Terraform show a diff on subsequent plans if it's only set on one or the other in the config?

vault/resource_aws_auth_backend_role.go
```diff
@@ -319,16 +393,27 @@ func awsAuthBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("role", role)
 	d.Set("auth_type", resp.Data["auth_type"])
 
+	// Read Vault's response into the currently supported version of these fields.
+	d.Set("bound_account_ids", resp.Data["bound_account_id"])
```
catsby: When setting aggregate types (list, set, map), we need to check the error from `d.Set` to make sure the data was correctly saved to state:

- https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types

tyrannosaurus-becks:
@idubinskiy thanks for looking! Yes, I think in that version of the code that could have happened. I updated the code to closely match [the guide](https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-an-optional-attribute), and it should now be good to go.

tyrannosaurus-becks:
@catsby thanks for linking that guide, it was so helpful! I updated the code to match the guide's recommendations and will use the guide as a model going forward.

Issue #176, opened 2018-09-19
vault 0.11.1. vault_aws_secret_backend_role policy permanently change


sadpirit:
After the update to vault 0.11.1, I get a policy change on **vault_aws_secret_backend_role** every time.

### Terraform Version
Terraform v0.11.8
+ provider.external v1.0.0
+ provider.template v1.0.0
+ provider.vault v1.1.3

Vault version:   0.11.1
### Affected Resource(s)
vault_aws_secret_backend_role

### Terraform Configuration Files
```hcl
resource "vault_aws_secret_backend_role" "problem" {
    provider = "vault.current"
    backend = "secrets/aws/plr/atf01"
    name = "problem_inline"
    policy = <<EOT
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "*",
            "Resource": "*"
        }
    ]
}
EOT
}
```
With each planning I get empty policy everytime:
```
  ~ vault_aws_secret_backend_role.problem
policy:    ""  =>
 "{
    \"Version\": \"2012-10-17\",
    \"Statement\": [
        {
            \"Effect\": \"Allow\",
            \"Action\": \"*\",
            \"Resource\": \"*\"
        }
    ]
}"
```

lawliet89:
This is most likely due to [Vault 0.11 changes](https://www.vaultproject.io/guides/upgrading/upgrade-to-0.11.0.html#aws-secret-engine-roles).

icereed:
Experiencing this as well. Thanks for having a PR ready!

universam1:
any news on this, we need to move to newer Vault version

icereed:
@tyrannosaurus-becks Could you help us here to know the next steps?

dhohengassner:
thanks for the research - also facing this issue...

would love to see this provider usable with current vault version

any updates?

icereed:
Hello,

it seems there is nothing in progress to solve this issue.
Actually this is quite important for us, because we're stuck with Vault 0.10 otherwise.
How to proceed here?

swenson:
I believe this was fixed in https://github.com/hashicorp/terraform-provider-vault/pull/261

PR #194, opened 2018-09-26
[WIP] Update `aws_secret_backend_role` with changes from Vault 0.11

lawliet89:
This PR attempts to update the `aws_secret_backend_role` resource to conform to the changes made in  [Vault 0.11](https://www.vaultproject.io/guides/upgrading/upgrade-to-0.11.0.html#aws-secret-engine-roles), namely:

- New parameters for creating a role
- New return values

This should address #176.

This is still a WIP and I haven't updated the tests yet because I have a few questions (and because I am new to Go):

- To what extent should we support `Vault < 0.11`? How much logic should we write in to deal with reading back values from Vault in the "old way".
- Should we validate the parameters in the provider? I feel like the validation logic for Vault 0.11 is more complex than what was before and I don't know if we should replicate Vault's logic here or just let the application error out on Vault's side. Suffice to say some of the logic can be provided for "free" by Terraform in the form of conflicting attributes.
- How do we handle `invalidData`? I am guesing we should warn the user.But seems like all the Warnings are eaten up by Terraform.
- How do we handle more than 1 `credentialsTypes` being returned by Vault? This happened on my Vault instance which was upgraded from 0.10.4. I am guesing we should warn the user.But seems like all the Warnings are eaten up by Terraform.
- Should we attempt to "guess" the `credential_type` if it's not provided? We cannot really make it `Required` because it breaks backwards compatibility. But it's also not supported pre Vault 0.11 (which leads me back to the first question). 

FWIW I tested this with my upgraded Vault cluster (from  0.10.4 to 0.11) and I didn't update my Terraform files and the plans show no diff. So it "kind of works".

Some advice would be appreciated. I'm new to this.

### DIFF
--- a/vault/resource_aws_secret_backend_role.go
+++ b/vault/resource_aws_secret_backend_role.go
@@ -168,9 +168,14 @@ func awsSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 
 		if _, ok := d.GetOk("credential_type"); ok {
 			d.Set("credential_type", credentialTypes[0])
-			d.Set("policy_arns", secret.Data["policy_arns"])
-			d.Set("role_arns", secret.Data["role_arns"])
 			d.Set("policy_document", secret.Data["policy_document"])
+
+			if err := d.Set("role_arns", secret.Data["role_arns"]); err != nil {
+				return fmt.Errorf("error setting role_arns for role %q: %s", path, err)
+			}
+			if err := d.Set("policy_arns", secret.Data["policy_arns"]); err != nil {
+				return fmt.Errorf("error setting policy_arns for role %q: %s", path, err)
+			}
 
 		} else {
 			// Deprecated style
--- a/vault/resource_aws_secret_backend_role.go
+++ b/vault/resource_aws_secret_backend_role.go
@@ -33,17 +33,54 @@ func awsSecretBackendRoleResource() *schema.Resource {
 				ForceNew:    true,
 				Description: "The path of the AWS Secret Backend the role belongs to.",
 			},
+			"credential_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Specifies the type of credential to be used when retrieving credentials from the role.",
+				// DefaultFunc: ??,
+			},
+			"role_arns": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"policy", "policy_arn", "policy_arns", "policy_document"},
+				Description:   "Specifies the ARNs of the AWS roles this Vault role is allowed to assume.",
+			},
+			"policy_arns": {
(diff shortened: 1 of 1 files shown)


icereed:
A user could also use an old provider version if they decide to stick to 0.10.
Since Vault will have 1.0 soon, there shouldn't be so many API breaking changes by Vault in future?

universam1:
+1

tyrannosaurus-becks:
@lawliet89 thank you for working on this! I hadn't looked at it sooner because of the WIP note in its name and catching up on a backlog of PR's.

> To what extent should we support Vault < 0.11? How much logic should we write in to deal with reading back values from Vault in the "old way".

I think we shouldn't worry about Vault <0.11. The reason being, if people want functionality that supports lower versions, they can simply likewise use lower versions of the Terraform Vault Provider. However, if you're thinking of deprecating a field, best to follow [this guide](https://www.terraform.io/docs/extend/best-practices/deprecations.html). 

>Should we validate the parameters in the provider? I feel like the validation logic for Vault 0.11 is more complex than what was before and I don't know if we should replicate Vault's logic here or just let the application error out on Vault's side. Suffice to say some of the logic can be provided for "free" by Terraform in the form of conflicting attributes.

There's also a lot of validation in Vault that's free. It will often fail with helpful output if a field failed to validate. I think it's OK to let the responsibility of validation rest with Vault, and to just pass values on to it, rather than trying to have identical logic in here, which would be difficult to maintain.

>How do we handle invalidData? I am guesing we should warn the user.But seems like all the Warnings are eaten up by Terraform.

I'm uncertain on this one.

>How do we handle more than 1 credentialTypes being returned by Vault? This happened on my Vault instance which was upgraded from 0.10.4. I am guessing we should warn the user.But seems like all the Warnings are eaten up by Terraform.

I'm uncertain on this one as well. As long as you made a best effort to pass them on to the user, that's about all you can do.

>Should we attempt to "guess" the credential_type if it's not provided? We cannot really make it Required because it breaks backwards compatibility. But it's also not supported pre Vault 0.11 (which leads me back to the first question).

I think it's okay to make it required. This again is one of those things where people can just stay on an older version of the Terraform Vault Provider if they're using an older version of Vault.

Thank you! I hope that's helpful!

lawliet89:
I will look into this in the later half of the week. If someone wants to work on it, please go ahead.

icereed:
Hi there, any updates on this? Do you need help on any thing? What would be the next steps?

universam1:
@tyrannosaurus-becks Vault is already two feature versions progressed - is Hashicorp not supporting their own product any more?

tantra35:
@universam1 do you try this PR?

tyrannosaurus-becks:
@universam1 this PR is currently marked as WIP. Happy to begin the review cycle when it appears finished.

lawliet89:
Life gets in the way and I don't have the time to work on this. Someone else is welcomed to take over.

Sorry about that. Closing this for now. Will reopen if and when I can work on it again.

tyrannosaurus-becks:
@lawliet89 thank you for your efforts though!

Issue #99, opened 2018-04-13
LDAP Auth Backend Resource(s)


errygg:
Hi!

Would be great to have an auth backend resource for adding an LDAP connection similar to the Okta resources.

### Terraform Version
Terraform v0.11.7
Terraform Vault Provider v1.1.0

### Affected Resource(s)
New resource(s) for LDAP Auth Backend

### Terraform Configuration Files
```hcl
provider "vault" {
  version="~> 1.1"
}
```

### Debug Output
```
Error: module.test_vault_config.vault_ldap_auth_backend.ldap: Provider doesn't support resource: vault_ldap_auth_backend
```

### Expected Behavior
New resources would allow Vault to be configured to talk to an LDAP provider (such as Active Directory) and include the available options for LDAP auth.

### References
LDAP configuration is supported in Vault, just not yet available in this provider: https://www.vaultproject.io/docs/auth/ldap.html

icereed:
Definitely need this +1

lawliet89:
It's in [`v1.1.3`](https://github.com/terraform-providers/terraform-provider-vault/releases/tag/v1.1.3).

sidewinder12s:
@tyrannosaurus-becks This can probably be closed out.

PR #198, opened 2018-10-03
Adding GCP Auth backend resource

petems:
This is a largely a copy of the #99 backend config but for the GCP backend.

It only has 1 field, credentials. Probably needs more testing, I might need someone to pair with me on that 😄

### DIFF
--- a/vault/resource_gcp_auth_backend.go
+++ b/vault/resource_gcp_auth_backend.go
@@ -1,0 +1,196 @@
+package vault
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+const gcpAuthType string = "gcp"
+
+func gcpAuthBackendResource() *schema.Resource {
+	return &schema.Resource{
+
+		Create: gcpAuthBackendWrite,
+		Update: gcpAuthBackendUpdate,
+		Read:   gcpAuthBackendRead,
+		Delete: gcpAuthBackendDelete,
+		Exists: gcpAuthBackendExists,
+
+		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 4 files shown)


vault/resource_gcp_auth_backend_test.go
```diff
@@ -0,0 +1,90 @@
...
+			return fmt.Errorf("error checking for gcp auth backend %q: %s", rs.Primary.ID, err)
+		}
+		if secret != nil {
+			return fmt.Errorf("gcp auth backend %q still exists", rs.Primary.ID)
+		}
+	}
+	return nil
+}
+
+func testGCPAuthBackendCheck_attrs(path string) resource.TestCheckFunc {
```
tyrannosaurus-becks: `path` could be stripped from being passed in since it's not used.

vault/resource_gcp_auth_backend_test.go
```diff
@@ -0,0 +1,90 @@
...
+		instanceState := resourceState.Primary
+		if instanceState == nil {
+			return fmt.Errorf("resource has no primary instance")
+		}
+
+		return nil
+	}
+}
+
+func testGCPAuthBackendConfig_basic(credentials string) string {
```
tyrannosaurus-becks: `credentials` could be stripped because they're not used either.

petems: Fixed to actually be called by the method 👍

vault/resource_gcp_auth_backend.go
```diff
@@ -0,0 +1,198 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/hashicorp/vault/api"
```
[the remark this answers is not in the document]
petems: Done! 👍

PR #200, opened 2018-10-04
Add gcp auth role extras

petems:
Adding a few extra fields for GCE role speficic config


petems:
@tyrannosaurus-becks Good point, I added the new fields to `gcpAuthResourceRead` and rebased.

I added a new test for the new hardcoded values, but I don't think I'm actually checking that those values are set... can you help me with that test if you have some time? 😄

petems:
Closing in favour of #227

PR #201, opened 2018-10-06
Support of the backends mounted to the path with "/"

StyleT:
Currently Terraform returns an error if you try to write the following:

```hcl
resource "vault_auth_backend" "this" {
  type = "aws"
  path = "team_name/aws"
}

resource "vault_aws_auth_backend_client" "this" {
  backend = "${vault_auth_backend.this.path}"
  iam_server_id_header_value = "https://vault.devops.net"
}
```

### DIFF
--- a/vault/resource_aws_auth_backend_client.go
+++ b/vault/resource_aws_auth_backend_client.go
@@ -120,7 +120,7 @@ func awsAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 	idPieces := strings.Split(d.Id(), "/")
-	d.Set("backend", idPieces[1:len(idPieces) - 2])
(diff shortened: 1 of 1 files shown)


StyleT:
@tyrannosaurus-becks Hi! Can you pls take a look at this PR?

tyrannosaurus-becks:
@StyleT thanks for working on this! Okay, so tests pass currently, and that's good. I tried altering one of the tests to put in the path to `team_name/aws`, and I see it's now `team_nameaws` being set on the backend. Does that seem right? If so, could you add a test showing that behaviour?

I'm also wondering if this would be an issue across resources, so maybe the code needs to be something more reusable?

tyrannosaurus-becks:
This PR hasn't had any action on it for a bit so I'm going to go ahead and close it for now. If you end up circling back to it, please do feel free to reopen it. Thank you!

PR #206, opened 2018-10-10
switch to go 1.11 modules

ahilsend:
Go 1.11 introduced [go modules ](https://github.com/golang/go/wiki/Modules). With modules the language has standardized way of dealing with dependencies and versions. Instead of using one of the many other tools to do this.

Currently go modules are opt-in, but in the future they will be enabled by default. Thus it makes sense to prepare for it already.


cvbarros:
Is there an issue for this? What's the motivation for this PR?
There's no description provided, could you bring some more context please @ahilsend ?

ahilsend:
@cvbarros Sorry, I missed to add a description. Updated it now. There is no issue for this, should there be?

cvbarros:
IMHO, a change as wide as the proposed one has a lot of impact. I think it deserves proper discussion in a issue before a PR is submitted to see whether or not this is relevant/wanted/timely.

tyrannosaurus-becks:
@cvbarros I just opened a ticket to provide a forum for discussion! #223, have at it!

tyrannosaurus-becks:
@ahilsend thank you for this PR! Closing this as it seems we're not quite ready to take this step yet. Much appreciated though.

PR #209, opened 2018-10-12
Run vault and mysql in docker-compose for travis acceptance tests

ahilsend:
In order to run the acceptance tests locally one has to manually startup mysql, or similar. And travis isn't executing all the tests either. Some of those could be simplified with docker-compose.

For example running vault and mysql in docker-compose means both, developers and the CI, can use the same spec to run the tests. Which also improves test coverage from the CI.


ahilsend:
@tyrannosaurus-becks Happy to contribute. I've rebased on master now.

PR #220, opened 2018-10-22
Add identity groups and aliases

simonswine:
This allow management of groups using terraform. A generic resource
fails, as it is required to have an ID to map aliases to the group.


pfremm:
Can a new release be cut with these changes?  I see v1.3.1 was released one day before this was merged.

PR #227, opened 2018-11-02
Add gcp auth role extras Adding a few extra fields for GCE role speficic config

petems:


petems:
Woo, tests are green now, thanks for the help @tyrannosaurus-becks 😄

PR #245, opened 2018-11-23
Extra config options for auth_backend

### LINKED ISSUE #203: "vault_auth_backend" should support backend configuration options

martinssipenko:
Adds ability to configure following config parameters for `auth_backend`:
- `default_lease_ttl_seconds`
- `max_lease_ttl_seconds`
- `listing_visibility`
- `local`

Test output:
```sh
$ make testacc TESTARGS='-run=TestResourceAuth'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestResourceAuth -timeout 120m
PASS
=== RUN   TestResourceAuth
--- PASS: TestResourceAuth (0.11s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.139s
```

Fixes #203

### DIFF
--- a/vault/resource_auth_backend.go
+++ b/vault/resource_auth_backend.go
@@ -55,6 +55,38 @@ func authBackendResource() *schema.Resource {
 				Description: "The description of the auth backend",
 			},
 
+			"default_lease_ttl_seconds": {
+				Type:        schema.TypeInt,
+				Required:    false,
+				Optional:    true,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "Default lease duration in seconds",
+			},
+
+			"max_lease_ttl_seconds": {
+				Type:        schema.TypeInt,
+				Required:    false,
+				Optional:    true,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "Maximum possible lease duration in seconds",
+			},
+
+			"listing_visibility": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "Speficies whether to show this mount in the UI-specific listing endpoint",
+			},
+
+			"local": {
+				Type:        schema.TypeBool,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "Specifies if the auth method is local only",
+			},
+
 			"accessor": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


cvbarros:
This is a nicely needed improvement! I'm working on a new auth_backend resource and I've noted a few points:

IMO these auth backend resources need a bigger refactor, whereas the resource mounts/unmounts both the auth backend + configures it's options.

Adding these configuration options to the "generic" resource `auth_backend` already increases this debt as put the newly created in fd5deeda20951ae330e54ea294a08658127b3acd without these options.

IMHO, if still proceeding this PR, these options should also be added to the `gcp_auth_backend` plus any other auth backend resources that have the same approach.

martinssipenko:
@cvbarros I agree, probably the best option would be a resource per each mount type (aws, gcp, github, etc).

The update (unmount/mount) part can be fixed using partial update and mount tune.

martinssipenko:
@tyrannosaurus-becks given the above, I'm happy to add a new resource specifically for aws instead of amending the generic one. What's your view on this?

kmcquade:
@martinssipenko @tyrannosaurus-becks @cvbarros  - any chance you guys could also address the auth tune request here as well? See https://github.com/terraform-providers/terraform-provider-vault/issues/234

vault/resource_auth_backend.go
```diff
@@ -95,9 +135,7 @@ func authBackendDelete(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Deleting auth %s from Vault", path)
 
-	err := client.Sys().DisableAuth(path)
-
-	if err != nil {
```
tyrannosaurus-becks: Thank you for noticing this and cleaning it up while you're at it.

cvbarros:
@kmcquade There's an approach submitted in #255, which can be leveraged to all auth mounts regarding these options. Once that PR is discussed/merge I can happily extend it to other auth mounts as well

Issue #174, opened 2018-09-18
Feature request: Add vault entity support


OperationalDev:
# Use Case

It would be cool to mange identity entities in vault with Terraform.

Right now, we're doing it with a combination of the UI/cli, but Terraform would just be better.

## Example Cli
```
vault write identity/entity name="bob" policies="base" \
     metadata=department="the unhashibles" \
     metadata=title="overlord"

vault write identity/entity-alias name="bob" \
     canonical_id=be57e6db-bf5d-a246-e931-fda108fc2639    \
     mount_accessor=auth_ldap_b44f9dc1
```

## Example Code
```
resource "vault_entity" "bob" {
  name     = "bob"
  policy   = ["base"]
  metadata = <<EOF
    {
      "department": "the unhashibles",
      "title": "overlord"
    }
  EOF
}

resource "vault_alias" "bob" {
  name         = "bob"
  canonical_id = "${resourse.vault_entity.bob.id}"
  backend      = "ldap"
}
```

cvbarros:
I think #220 is waiting review/update that would implement this

OperationalDev:
#220 adds identity groups support. I've reused most of @simonswine contribution to add entity support. See pull request #247

tyrannosaurus-becks:
I believe this is done! See [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_identity_entity.go).

PR #243, opened 2018-11-15
Fix wrong response data from gcp_auth_backend_role read request

Phylu:
As of https://github.com/hashicorp/vault-plugin-auth-gcp/commit/438b86d4b944867fc9ca0d713280088eb72df744, the `role_type` field has been changed to `type`.

This pull request should resolve #228.

PR #247, opened 2018-11-27
This allows the management of identity entities and aliases of entities

OperationalDev:
Implementation of feature request #174 

Many thanks to @simonswine for all the groundwork on #220

### DIFF
--- a/vault/resource_identity_entity.go
+++ b/vault/resource_identity_entity.go
@@ -1,0 +1,198 @@
+package vault
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/vault/api"
+	"github.com/terraform-providers/terraform-provider-vault/util"
+)
+
+const identityEntityPath = "/identity/entity"
+
+func identityEntityResource() *schema.Resource {
+	return &schema.Resource{
+		Create: identityEntityCreate,
+		Update: identityEntityUpdate,
+		Read:   identityEntityRead,
+		Delete: identityEntityDelete,
+		Exists: identityEntityExists,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the entity.",
+				ForceNew:    true,
+			},
+
+			"metadata": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Metadata to be associated with the entity.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"policies": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "Policies to be tied to the entity.",
+			},
+
+			"disabled": {
+				Type:        schema.TypeString,
+				Optional:    true,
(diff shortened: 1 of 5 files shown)


OperationalDev:
Entity related tests pass if I bump the vault version to 0.11.5

GCP Tests pass if I include #243 and bump the vault version to 1.0.0-beta2

docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: "2.2"
 
 services:
   vault:
-    image: vault:0.11.1
+    image: vault:0.11.5
```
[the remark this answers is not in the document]
OperationalDev: Thanks for taking the time to review my PR. 

The only issue with staying on 0.11.1 is that then we cannot move an alias between entities. I don't think this is a big problem.
If we remove the test for moving an alias between entities, then the build does pass (after merging master)

Works for me as I have no intention of moving aliases between entities.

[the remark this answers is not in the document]
OperationalDev: Ok, removed the test and we're looking good on v0.11.1

OperationalDev:
Silly question, can we also submit pull requests for the docs?

tyrannosaurus-becks:
@OperationalDev 
>Can we also submit pull requests for the docs?
Yes! Docs changes can either be included with the PR that has related code changes, or done separately.

vault/resource_identity_entity.go
```diff
@@ -0,0 +1,198 @@
...
+		data["policies"] = policies
+	}
+
+	if metadata, ok := d.GetOk("metadata"); ok {
+		data["metadata"] = metadata
+	}
+
+	if disabled, ok := d.GetOk("disabled"); ok {
+		data["disabled"] = disabled
+	}
```
tyrannosaurus-becks: Not a blocker at all, but if you feel like it, updating an entity's `name` is also supported.

vault/resource_identity_entity.go
```diff
@@ -0,0 +1,198 @@
...
+		return fmt.Errorf("error reading AppRole auth backend role SecretID %q: %s", id, err)
+	}
+	log.Printf("[DEBUG] Read IdentityEntity %s", id)
+	if resp == nil {
+		log.Printf("[WARN] IdentityEntity %q not found, removing from state", id)
+		d.SetId("")
+		return nil
+	}
+
+	for _, k := range []string{"name", "metadata", "disabled"} {
```
tyrannosaurus-becks: Just curious about why `policies` isn't included here.

vault/resource_identity_entity_alias_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/hashicorp/vault/api"
+)
+
+func TestAccIdentityEntityAlias(t *testing.T) {
```
tyrannosaurus-becks: The tests won't build for me currently due to a dependency issue; I think master needs to be merged in.

OperationalDev: I thought I had merged master in, will take a look.

vault/resource_identity_entity.go
OperationalDev: No good reason, will add.

PR #259, opened 2018-12-12
Update policy_arn to be more in line with vault 0.11.0

moosilauke18:
Vault was changed from using `arn` to use `policy_arns` and `role_arns` in https://github.com/hashicorp/vault/pull/4360 

Without this change terraform will not be able to read the aws_secret_backend_role properly and assumes it always needs to be updated with policy_arn. 

Have not tested the Acceptance tests. But have tested locally to existing vault > 0.11.0 where this issue came up.

### DIFF
--- a/vault/resource_aws_secret_backend_role.go
+++ b/vault/resource_aws_secret_backend_role.go
@@ -34,16 +34,28 @@ func awsSecretBackendRoleResource() *schema.Resource {
 				ForceNew:    true,
 				Description: "The path of the AWS Secret Backend the role belongs to.",
 			},
-			"policy_arn": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				ConflictsWith: []string{"policy"},
+			"policy_arns": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				ConflictsWith: []string{"policy", "role_arns"},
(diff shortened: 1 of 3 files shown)


cvbarros:
Hi @moosilauke18 , this appears to be a breaking change by renaming this attribute. In these cases, to give opportunity for users to control the upgrade and breaking changes, it's always best to try and accomodate schema changes.

This link is a very useful resource that provides some best practices when dealing with this type of change:
https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename

Thus, such renames must be weighted against the cost-benefit of a breaking change.

Regarding the acceptance tests, do you need any assistance to run them locally?

moosilauke18:
Added back the deprecated values. Acceptance testing is passing on all except for one import test for policy_document. The old attributes were also failing the import tests.  If there is any reading/code examples of how the import actually works, I'd love to read them.

tyrannosaurus-becks:
@moosilauke18 thanks for working on this! I did recently add a [CONTRIBUTING.md](https://github.com/terraform-providers/terraform-provider-vault/blob/master/.github/CONTRIBUTING.md) to the repo that may help with tests? If you're not finding what you need there (under Testing, in Extending Terraform), let me know and I can hopefully point you to something more or improve the guide.

tyrannosaurus-becks:
Closing this for now due to inactivity. Please feel free to reopen it if you get a chance to circle back around. Thanks!

PR #264, opened 2019-01-02
Added support for EGP and RGP (sentinel)

sergeytrasko:


tyrannosaurus-becks:
@sergeytrasko thanks for working on this! 

It all looks perfect at first strike. Would it be possible to get test acceptance tests on this? Then I could spin up Vault locally and just run them to verify it works, and it would also help detect any accidental breaking changes going forward.

tyrannosaurus-becks:
@sergeytrasko also FWIW, there are some great examples of acceptance tests already written for other resources in the provider, and there's a very handy guide [here](https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html).

sergeytrasko:
Thanks for review!
I will try to work on acceptance tests later this week.

cvbarros:
Out of curiosity @tyrannosaurus-becks @sergeytrasko , is there a way to run acceptance tests against these enterprise-only features?

tyrannosaurus-becks:
@sergeytrasko for the moment I'd probably just run them against a local copy of the enterprise version of Vault. So - yes, locally, and no via Travis? I'm assuming @sergeytrasko has Vault Enterprise locally - not sure if that assumption is correct.

sergeytrasko:
I've added a couple of acceptance tests.
To enable them you should specify `TF_ACC_ENTERPRISE=1` parameter (I assume not everyone will have enterprise version of Vault)

cvbarros:
The way to toggle that was exactly what I was looking for, nice one 🎉 👍 @sergeytrasko

PR #270, opened 2019-01-09
Fix issue where boolean parameters are not being set when value is false

jyoukhana2165:
While using the **vault_aws_auth_backend_role** resource we wanted to set the parameter **resolve_aws_unique_ids** to false. However we noticed this was being ignored because **GetOk** method returns ok as false when the boolean value is false.

Therefore parameter is not supplied to underlying Vault API call. The default value for **resolve_aws_unique_ids** in Vault is true.

Method **GetOkExists** seems more appropriate for boolean parameters that do not default to being false.


krish7919:
This is precisely what I hit today.

Can we get this merged, @tyrannosaurus-becks ?

tyrannosaurus-becks:
@jyoukhana2165 thanks for working on this! Would you be willing to add a test that duplicates this issue, and that's solved by this code? That would be a huge help.

jyoukhana2165:
@tyrannosaurus-becks I've added a test to show the fix working.

PR #255, opened 2018-12-11
Implement Github Auth Backend resources

cvbarros:
Resolves #254 

In addition to the feature request implementation, this PR also introduces some common idioms/helpers (heavily present in terraform-provider-aws codebase) to flatten/expand structures from the config to the upstream API. Based the implementation on the recently-released `vault_gcp_auth_backend`, where:

* There's no distinction between the _Auth Mount_ and it's configuration. They are handled as a single resource in `resource_github_auth_backend.go`.

* Introduces a shared schema for tuning auth methods, as the configuration block `tune`. This can be reused/refactored to fix #234 and open up path to standardize other auth backends (as it is done in Vault's implementation)

* Handles `TTL` by having a human-friendly duration interface and takes care of Vault's API translation. This is an attempt to resolve some inconsistencies and "permanent diffs" when applying configurations that have `max_ttl`, or `tll` set, for example.

* Raises the bar of acceptance tests, by adding destroy checks and existence checks for resources. In theory, the optimal tests would validate resource attributes by fetching them from the API. But as long as the create/read, update/read cycle is respected, these are minimum problems.

* Tried to apply [guidelines as described here](https://github.com/MattHodge/terraform-provider-octopusdeploy/issues/25), specially related to `structures.go` and `validators.go`.

PR #273, opened 2019-01-11
base ldap_auth_backend_config on vault_auth_backend

dhohengassner:
**Changes behaviour how a ldap auth backend is defined - vault_ldap_auth renamed to vault_ldap_auth_backend_config**

**Instead of creating a auth mount itself we require now a vault_auth_backend as base**

right now the ldap auth backend does not support to set a lease-ttl
this was already added to the vault_auth_backend

change the behaviour and rename resource to ldap_auth_backend_config
the resource requires not a vault_auth_backend as base

behaviour should now be similar to kubernetes_auth_backend_config

resolves https://github.com/terraform-providers/terraform-provider-vault/issues/269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ BUG FIXES:
 * Fixes issue where the `role_name` on `token_auth_backend_role` would not be updated ([#279](https://github.com/terraform-providers/terraform-provider-vault/pull/279))
 * Fixes wrong response data from `gcp_auth_backend_role` ([#243](https://github.com/terraform-providers/terraform-provider-vault/pull/243))
 
+* **New Resource**: `ldap_auth_backend_config` ([#273](https://github.com/terraform-providers/terraform-provider-vault/pull/273))
+
+BACKWARDS INCOMPATIBILITIES / NOTES:
+* `ldap_auth_backend` is deprecated; use `ldap_auth_backend_config` instead.
+
 ## 1.4.1 (December 14, 2018)
 
 BUG FIXES:
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -134,6 +134,7 @@ func Provider() terraform.ResourceProvider {
 			"vault_okta_auth_backend":                            oktaAuthBackendResource(),
 			"vault_okta_auth_backend_user":                       oktaAuthBackendUserResource(),
 			"vault_okta_auth_backend_group":                      oktaAuthBackendGroupResource(),
+			"vault_ldap_auth_backend":                            ldapAuthBackendResource(),
 			"vault_ldap_auth_backend_config":                     ldapAuthBackendConfigResource(),
 			"vault_ldap_auth_backend_user":                       ldapAuthBackendUserResource(),
 			"vault_ldap_auth_backend_group":                      ldapAuthBackendGroupResource(),
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -134,7 +134,7 @@ func Provider() terraform.ResourceProvider {
 			"vault_okta_auth_backend":                            oktaAuthBackendResource(),
 			"vault_okta_auth_backend_user":                       oktaAuthBackendUserResource(),
 			"vault_okta_auth_backend_group":                      oktaAuthBackendGroupResource(),
-			"vault_ldap_auth_backend":                            ldapAuthBackendResource(),
+			"vault_ldap_auth_backend_config":                     ldapAuthBackendConfigResource(),
 			"vault_ldap_auth_backend_user":                       ldapAuthBackendUserResource(),
 			"vault_ldap_auth_backend_group":                      ldapAuthBackendGroupResource(),
 			"vault_policy":                                       policyResource(),
--- a/vault/resource_ldap_auth_backend.go
+++ b/vault/resource_ldap_auth_backend.go
@@ -3,26 +3,26 @@ package vault
 import (
 	"fmt"
 	"log"
-	"regexp"
 	"strings"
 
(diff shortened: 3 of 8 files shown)


dhohengassner:
tests were successful

```
» make testacc TESTARGS='-run=TestLDAPAuthBackendConfig_basic'                                                                                                                                                                                                dh@Davids-Mac-Pro
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestLDAPAuthBackendConfig_basic -timeout 120m
PASS
=== RUN   TestLDAPAuthBackendConfig_basic
--- PASS: TestLDAPAuthBackendConfig_basic (0.11s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   (cached)
```

dhohengassner:
provisioned my vault server with this fix - is already working

```
resource "vault_auth_backend" "ldap" {
  type                      = "ldap"
  path                      = "ldap"
  default_lease_ttl_seconds = 1800
  max_lease_ttl_seconds     = 3600
}

resource "vault_ldap_auth_backend_config" "my-ldap" {
  backend         = "${vault_auth_backend.ldap.path}"
.
.
.
```

image

dhohengassner:
@paddycarver @tyrannosaurus-becks need some input how to throw deprecation warnings - looks like we use an old terraform vendor where it is not supported

I added the possibility to use it - like it is in current terraform repo
is working for me now - but this is the reason the test failed I think

image

maybe you planned an update of vendor state anyway?

Any ideas? conventions?

Thanks!

vendor/github.com/hashicorp/terraform/helper/schema/data_source_resource_shim.go
```diff
@@ -32,7 +32,7 @@ func DataSourceResourceShim(name string, dataSource *Resource) *Resource {
 
 	// FIXME: Link to some further docs either on the website or in the
 	// changelog, once such a thing exists.
-	dataSource.deprecationMessage = fmt.Sprintf(
+	dataSource.DeprecationMessage = fmt.Sprintf(
```
tyrannosaurus-becks: It's odd that there are updates to the vendor files without a changed commit hash in `vendor/vendor.json`. Were these made directly? If so they'll get wiped out when `govendor` is next used to update dependencies. Would it be possible to get these changes from different versions of these libraries instead? Here's the [Govendor Cheat Sheet](https://github.com/kardianos/govendor/wiki/Govendor-CheatSheet) if it helps.

dhohengassner: @tyrannosaurus-becks thanks for the helpful reply

I was not sure which dependency solution you are using - so I adjusted manually what I needed from the terraform HEAD

Do you think it is ok to update the whole terraform dependencies to a newer version?
That make probably other adjustments necessary?

I can try that if I have some time...

tyrannosaurus-becks: Thanks for offering to do that! I'd love to leave Terraform's dependencies as unruffled as possible and shoot for the minimum change on each PR. Ah yes, it's not in the docs, but we have been using [GoVendor](https://github.com/kardianos/govendor/wiki/Govendor-CheatSheet) for this repo. Is that `DeprecationMessage` exported on later versions of Terraform? Do you happen to know what the first release was where it was exported?

dhohengassner: @tyrannosaurus-becks sorry for the delay with this PR

I tried today to come to a more current version of vendored terraform - created a new PR for it https://github.com/terraform-providers/terraform-provider-vault/pull/341

Some input how to proceed is very much appreciated!
Thanks

dhohengassner:
@tyrannosaurus-becks pipeline fail was expected - would need an tf dependency update before it is able to succeed - https://github.com/terraform-providers/terraform-provider-vault/pull/341

cvbarros:
If I may drop my opinion, I think the way this is implemented drifts a bit from the latest `auth_backend` implementations. 
Auth backends should comprise their own resource and not rely on `vault_auth_backend` at all - that one in the end is that should be deprecated. Adding an extra "config" resource is not valuable as it will always be required in order for the mount to be functional. This adds verbosity and obfuscates the HCL code.

For this, auth backends need a shared schema, to tune them. Check #255 how a `tune` block was created to allow reuse across other auth backend resources. Reuse can also be achieved via shared functions to flatten/expand these tuning settings to Vault API, making it easier to roll out to all auth backends used by the provider.

And also, you get the TTL treatment as a duration string, which IMO is a much needed enhancement to the codebase to prevent TTL serialization issues.

tyrannosaurus-becks:
Hi @dhohengassner , I'm afraid I can only consider PR's with passing tests because it's important to me that tests are passing on master. I do appreciate this PR. Contribution guidelines are available [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/.github/CONTRIBUTING.md) if it's any help. If you're able to get those tests working, I will take a closer look at the code.

tyrannosaurus-becks:
Closing, as tests aren't passing and this PR hasn't been active lately. Thanks again for your efforts. If this is still needed and you're able to get it into [this](https://github.com/terraform-providers/terraform-provider-vault/blob/master/.github/CONTRIBUTING.md) state, feel free to reopen it.

dhohengassner:
@cvbarros thanks for the great remark - good work on your backend

Maybe I find the time once to implement it for ldap auth backend - for now I will try to go with this workaround https://github.com/terraform-providers/terraform-provider-vault/issues/269#issuecomment-479707244

PR #262, opened 2018-12-21
Support for namespaces

sergeytrasko:
PR for #175

PR #274, opened 2019-01-14
add namespace to vault client config

rosteroster:
### Description

The vault client in the vault terraform provider does not support namespaces yet.
It's a relatively trivial change to include that configuration option.


chrishoffman:
Duplicate of #262

PR #289, opened 2019-01-30
Set namespace before attempting child token creation

rosteroster:
Problem with https://github.com/terraform-providers/terraform-provider-vault/pull/288 is that as it currently stands, if the terraform user has an admin token for a namespace, the child token creation will fail, as it attempts to create the child token before setting the namespace. So, the child token creation is attempted in the root namespace, not in the namespace where the user has the admin token.

This moves the namespace setter to be above the first actual Vault operations.

If we're having trouble with bugs, we should really set the client namespace to the be the namespace of the customer's token, but you can't do a token lookup in the root namespace if you only have a child namespace admin token.

An option may be to add additional configuration options ie add another "parent-token-namespace" config value so that the child can be created in the parent's namespace but still allow modification of resources in a target namespace.

Another option is to do away with the child token & instead add validation + a rule saying that the admin token cannot have a TTL > X hours

Curious for your opinion @sergeytrasko on all of the above

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -220,6 +220,11 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	if token == "" {
 		return nil, errors.New("no vault token found")
 	}
+	
+	namespace := d.Get("namespace").(string)
+	if namespace != "" {
+		client.SetNamespace(namespace)
+	}
 
 	// In order to enforce our relatively-short lease TTL, we derive a
 	// temporary child token that inherits all of the policies of the
@@ -250,11 +255,6 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	policies := childTokenLease.Auth.Policies
 
 	log.Printf("[INFO] Using Vault token with the following policies: %s", strings.Join(policies, ", "))
-
-	namespace := d.Get("namespace").(string)
-	if namespace != "" {
-		client.SetNamespace(namespace)
-	}
 
 	client.SetToken(childToken)
 
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -85,6 +85,11 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_NAMESPACE", ""),
 				Description: "The namespace to use. Available only for Vault Enterprise",
 			},
+			"token_namespace": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The namespace where the provided vault token was created, if different from the value in 'namespace'",
+			},
 		},
 
 		ConfigureFunc: providerConfigure,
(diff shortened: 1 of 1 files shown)


rosteroster:
^ Rebased off v1.5.0 / master

rosteroster:
With this change we can support admin tokens that live in any namespace ^

rosteroster:
@sergeytrasko this should be enough to avoid the bugs you were finding when prepping https://github.com/terraform-providers/terraform-provider-vault/pull/288

joestump:
This new attribute should added to [the docs](https://github.com/terraform-providers/terraform-provider-vault/blob/master/website/docs/index.html.markdown) for the provider.

vault/provider.go
```diff
@@ -221,6 +226,15 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, errors.New("no vault token found")
 	}
 
+	// if 'token_namespace' provided, set client namespace to use it for child token creation, else use 'namespace'
+	tokenNamespace := d.Get("token_namespace").(string)
+	namespace := d.Get("namespace").(string)
+	if tokenNamespace != "" {
+		client.SetNamespace(tokenNamespace)
+	} else {
+		client.SetNamespace(namespace)
```
tyrannosaurus-becks: Should this only be called if `namespace != ""`?

vault/provider.go
```diff
@@ -251,12 +265,8 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	log.Printf("[INFO] Using Vault token with the following policies: %s", strings.Join(policies, ", "))
 
-	namespace := d.Get("namespace").(string)
-	if namespace != "" {
-		client.SetNamespace(namespace)
-	}
-
 	client.SetToken(childToken)
+	client.SetNamespace(namespace)
```
tyrannosaurus-becks: Ditto here - only call if namespace is set?

kumar-shantanu:
The admin token of a particular namespace works after this change however root token stops working. Is it intentional ?

scott2449:
I have made this same update and deployed a custom version. The provider is unusable in most vault setups as is! This will not break root token because the namespace code only takes effect if namespace is defined which it should not be for root.

thomasbibb:
After quite a lot of coffee and testing, I've also found this is our issue.  Please merge 👍

tyrannosaurus-becks:
Hi! Thanks for working on this! It still needs tests, unfortunately, so I'm not confident enough to merge it right now.

jtcressy:
I'd like to chime in and see if I can move this forward. I represent an enterprise customer of hashicorp and we _really_ want to get this functionality implemented, tested and merged asap. I'm willing to step in and get it done.

@tyrannosaurus-becks are there currently any tests for the child token creation process in the provider? 
I'm imagining a test case with these steps:
- Instantiate provider in root namespace with root token, and create temp child token
- Create test namespace
- Create orphaned periodic token within test namespace 
- Instantiate second provider in test namespace with the orphaned token created in the previous step
- Test if the second provider can create a child token in the namespace

At the moment, i'm not clear on how to actually code the test case for this. The "Extending Terraform" docs help me understand how to test resource creation, but it's not very clear on how to test the actual provider implementation.

tyrannosaurus-becks:
@jtcressy yes, I'll just close this PR since it hasn't been visited by the other in a while. I will answer your further questions on the issue I believe this is regarding, and at you on it so you're pulled in.

PR #291, opened 2019-01-31
Fix documentation

martinssipenko:


martinssipenko:
Sorry about so much commit, simply was doing this via GitHub UI.

PR #296, opened 2019-02-06
jwt_auth_backend_role: add groups_claim_delimiter_pattern support

aromeyer:
Add support for groups_claim_delimiter_pattern parameters within vault_jwt_auth_backend_role resource.
Sync version of vendor/github.com/hashicorp/vault-plugin-auth-jwt with https://github.com/hashicorp/vault-plugin-auth-jwt.

### DIFF
--- a/vendor/github.com/hashicorp/vault-plugin-auth-jwt/README.md
+++ b/vendor/github.com/hashicorp/vault-plugin-auth-jwt/README.md
@@ -5,10 +5,6 @@ This plugin allows for JWTs (including OIDC tokens) to authenticate with Vault.
 
 **Please note**: We take Vault's security and our users' trust very seriously. If you believe you have found a security issue in Vault, _please responsibly disclose_ by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
 
-## IMPORTANT
-
-This plugin is in pre-release state. It is not well tested (in fact, not tested at all) and there is no documentation currently available.
-
 ## Quick Links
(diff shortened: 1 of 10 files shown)


vendor/github.com/hashicorp/vault-plugin-auth-jwt/README.md
```diff
@@ -5,10 +5,6 @@ This plugin allows for JWTs (including OIDC tokens) to authenticate with Vault.
 
 **Please note**: We take Vault's security and our users' trust very seriously. If you believe you have found a security issue in Vault, _please responsibly disclose_ by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
 
-## IMPORTANT
```
tyrannosaurus-becks: I see changes to the vendored files, but I don't see any changes in `vendor.json`. Would it be possible revert the changes in `vendor/`, and then to update them (if needed) [via Govendor instead](https://github.com/kardianos/govendor/wiki/Govendor-CheatSheet)? I'll add a note about that to our new contributing guide.

Otherwise this looks fantastic! Thank you!

aromeyer: Hello,
as far I understand, now the "plugin" is going to use JSON pointer instead of using this additional parameter (https://github.com/hashicorp/vault-plugin-auth-jwt/pull/24). I wait a bit to see what happen but I think that this PR will be soon useless.
Thanks for your comments.

PR #301, opened 2019-02-13
Fix some issues with `vault_identity_group` 

lawliet89:
- `policies` were not updated properly from reading the resource from Vault
- Update fields `policies`, `member_group_ids`, and `member_entity_ids` to be of `schema.TypeSet` instead. These lists should contain unique items and the order should not matter when calculating the diff
- Handle `member_entity_ids` for external groups (otherwise we get errors like `error updating IdentityGroup "a6c92cfa-d0b0-8d61-7122-a59cf1123a45": cannot set 'member_entity_ids' on external groups`)

Other fixes:

-  Fix typo in `vault_identity_group_alias`

Can't really write tests for the diff suppression without "logging in" with a real auth backend. It works in my local tests, though.

Before:

```
  ~ vault_identity_group.admins
      member_entity_ids.#: "1" => "0"
      member_entity_ids.0: "2b83bd73-954c-7b4d-b102-bec997008baf" => ""

  ~ vault_identity_group.users
      member_entity_ids.#: "1" => "0"
      member_entity_ids.0: "2b83bd73-954c-7b4d-b102-bec997008baf" => ""

```
After:

```
No changes. Infrastructure is up-to-date.

```

### DIFF
--- a/vault/resource_identity_group.go
+++ b/vault/resource_identity_group.go
@@ -45,7 +45,7 @@ func identityGroupResource() *schema.Resource {
 			},
 
 			"policies": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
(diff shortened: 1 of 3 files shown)


lawliet89:
The entire resource is undocumented. Is it ok for that to be done in a separate PR?

lawliet89:
That's done in #311 .

PR #303, opened 2019-02-13
ssh role add missing parameters

nand0p:
- adds `default_extensions` option


nand0p:
@tyrannosaurus-becks @martinssipenko thoughts on this? should i add the rest of the missing ssh role options here as well?

PR #310, opened 2019-02-19
Problem: Nil ca chain in response causes segfault

spheromak:
Solution: guard for a nil response for ca chain.


spheromak:
thanks @tyrannosaurus-becks  for the quick turn around. Any idea when we might see a patch release. This issue effects us cause we use our own CA in vault, and we currently can't use the tf-vault cert provisioning because of this issue.

PR #314, opened 2019-02-21
Fix TTL on PKI certificate create to use string type and not int.

### LINKED ISSUE #313: Error when generating PKI cert with TTL set to a string like 720h
When generating a PKI certificate the TTL is expected to be a `int`; whereas the [docs](https://www.vaultproject.io/api/secret/pki/index.html#generate-certificate) and CLI behaviour suggest this is incorrect and that TTLs can be passed like `720h`.

### Terraform Version
```
Terraform v0.11.11
+ provider.vault v1.5.0
```

### Affected Resource(s)
- `vault_pki_secret_backend_cert`

```hcl
resource "vault_pki_secret_backend_cert" "test_cert" {
  backend     = "${vault_mount.test_intermediate_ca.path}"
  name        = "${vault_pki_secret_backend_role.test.name}"
  common_name = "test"
  ttl         = "720h"
  format      = "pem_bundle"

  lifecycle {
    ignore_changes = ["*"]
  }
}
```

### Error output

```
Error: vault_pki_secret_backend_cert.test: ttl: cannot parse '' as int: strconv.ParseInt: parsing "720h": invalid syntax
```

jrasell:
The certificate creation call can uses a string such as 720h for
setting the TTL and shouldn't use an int.

Testing this locally the certificate creates successfully with the
following validity:
```
Not Before: Feb 21 15:25:01 2019 GMT
Not After : Mar 23 15:25:31 2019 GMT
```

Closes #313

### DIFF
--- a/vault/resource_pki_secret_backend_cert.go
+++ b/vault/resource_pki_secret_backend_cert.go
@@ -64,7 +64,7 @@ func pkiSecretBackendCertResource() *schema.Resource {
 				},
 			},
 			"ttl": {
-				Type:        schema.TypeInt,
+				Type:        schema.TypeString,
 				Optional:    true,
 				ForceNew:    false,
 				Description: "Time to leave.",
@@ -156,7 +156,7 @@ func pkiSecretBackendCertCreate(d *schema.ResourceData, meta interface{}) error 
 
 	data := map[string]interface{}{
 		"common_name":          d.Get("common_name").(string),
-		"ttl":                  d.Get("ttl").(int),
+		"ttl":                  d.Get("ttl").(string),
 		"format":               d.Get("format").(string),
 		"private_key_format":   d.Get("private_key_format").(string),
 		"exclude_cn_from_sans": d.Get("exclude_cn_from_sans").(bool),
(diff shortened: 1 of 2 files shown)


tyrannosaurus-becks:
Hey @jrasell! Thanks for submitting this! Would you be willing to add test coverage to those fields? I recently added info on how to do that by adding [a CONTRIBUTING.md file](https://github.com/terraform-providers/terraform-provider-vault/blob/master/.github/CONTRIBUTING.md).

cvbarros:
Hello @jrasell , like this one, there has been several bugs regarding the handling of TTL inside the provider. Due to lack of availability, I haven't taken a stab across equalizing all TTL treaments, but that's something I've been planning to do.

If you see #255 changes where [this](https://github.com/terraform-providers/terraform-provider-vault/blob/3298d0d7e87690c51ea8b9d7e686e7c17b663435/vault/structures.go#L102) was introduced, there is proper validation for a `duration` string. This treatment was derived from Vault API code itself, so it matches what upstream expects.

Also, I'd like to call out that a change in `schema.TypeInt` -> `schema.TypeString` . would be a breaking change, thus state migration is advised.

PR #316, opened 2019-02-22
Update `identity_entity` to use Sets for Policies

lawliet89:


nazarewk:
+1 for this, looks like extremely good idea

vault/resource_identity_entity_test.go
```diff
@@ -138,10 +140,18 @@ func testAccIdentityEntityCheckAttrs(entity string) resource.TestCheckFunc {
...
+
 					for i := 0; i < count; i++ {
-						stateData := instanceState.Attributes[stateAttr+"."+strconv.Itoa(i)]
-						if stateData != apiData[i] {
-							return fmt.Errorf("expected item %d of %s (%s in state) of %q to be %q, got %q", i, apiAttr, stateAttr, path, stateData, apiData[i])
+						found := false
+						for stateKey, stateValue := range instanceState.Attributes {
+							if strings.HasPrefix(stateKey, stateAttr) {
+								if apiData[i] == stateValue {
+									found = true
```
tyrannosaurus-becks: Should we break out of the loop here?

lawliet89: Good point.

lawliet89: That's done @tyrannosaurus-becks

tyrannosaurus-becks: :rainbow:

PR #317, opened 2019-02-22
Update token_auth_backend_role 

lawliet89:
- Add fields
- Make list of policies be sets
- Remove non-existent fields

`ttl`, and `max_ttl` does not exist. See the [API docs](https://www.vaultproject.io/api/auth/token/index.html#create-update-token-role)

- Fix fields updating due to https://github.com/hashicorp/vault/issues/6296


lawliet89:
There is some very weird bugs (?) with how Vault updates the token roles. I'll look into it and reopen this later.

lawliet89:
I have a [commit](https://github.com/terraform-providers/terraform-provider-vault/commit/9afd187c715fb39ffe15baa92641270020ccb58b) to deal with the issue that has been fixed in Vault 1.1 but the Docker image for 1.1 has not been pushed yet.

tyrannosaurus-becks:
Hey @lawliet89 , this looks very good. Would you be willing to add that commit to this branch since the fix landed?

lawliet89:
Seems like updating to 1.1 broke some of the JWT Auth engines stuff.

I will fix this on Monday.

lawliet89:
@tyrannosaurus-becks I have fixed the test failures and updated the JWT Auth Backend Role resource to fix removal of fields since Vault 1.1.

PR #321, opened 2019-02-25
Identity improvements

### LINKED ISSUE #320: Identity secret backend improvements
This is a feature request/discussion issue for [Identity backend](https://www.vaultproject.io/docs/secrets/identity/index.html).

My observations so far are:
1. there could be separate resource for `vault_identity_group` membership (separate for sub-groups and entities) management more akin to [`google_*_iam_member`](https://www.terraform.io/docs/providers/google/r/google_billing_account_iam_member.html),
1. `vault_identity_*` resources should implement/enable importers,
1. policies could be managed membership-style to resolve chicken and egg problem between groups and [templated policies](https://www.vaultproject.io/docs/concepts/policies.html#templated-policies)
1. `vault_identity_entity` and `vault_identity_group` should support renaming, maybe `ForceNew: false` suffices

### LINKED ISSUE #343: Error with 0.12-beta dev build
### Output

```
4 errors occurred:
	* resource vault_identity_group_alias: id is a reserved field name
	* resource vault_identity_entity: id is a reserved field name
	* resource vault_identity_group: id is a reserved field name
	* resource vault_identity_entity_alias: id is a reserved field name
```

### Terraform Version

```
Terraform v0.12.0-dev
+ provider.aws v1.60.0-dev20190216h00-dev
+ provider.azuread v0.2.0-dev20190216h01-dev
+ provider.vault v1.6.0-dev20190216h01-dev
```

### Affected Resource(s)
None -- fails on precheck of `terraform plan`

I've tried to find a branch/source for the dev build, but I can't seem to find one to determine if this has already been fixed or not. Builds off of master are failing with Terraform 0.12-beta1 because of the API version mismatch (4 vs 5).

nazarewk:
This PR implements #320 ideas, except for group membership management.
Fixes #343 
Fixes #320 

# Major change
Adds `vault_identity_group_policies` resource and `vault_identity_group.external_policies` attribute.
It solves chicken & egg problem of using `vault_identity_group.id` inside templated `vault_policy`.

# Bug fixes
1. adds missing error checks in Go `d.Set(...)`
1. as a result fixes `vault_identity_entity.disabled` not working at all (used `TypeString` instead of expected `TypeBool`)
1. fixes error messages being about `AppRole` instead of specific `Identity*` in `vault_identity_*` resources

# Minor changes
1. enables passthrough importers on `vault_identity_*` resources 
1. enables `name` autogeneration (by default Vault assigns `group-<UUID>` or `entity-<UUID>` names) by `Required: true` -> `Optional: true`
1. enables renaming `vault_identity_entity` and `vault_identity_group` resources (`ForceNew: false` was enough),


nazarewk:
I guess the test was a random failure, closing and reopening

vault/resource_identity_entity.go
```diff
@@ -18,13 +18,15 @@ func identityEntityResource() *schema.Resource {
...
+			State: schema.ImportStatePassthrough,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				Description: "Name of the entity.",
-				ForceNew:    true,
```
lawliet89: This might be optional, but a change would still require a `ForceNew`. Also, I think this should also have `Computed: true`

vault/resource_identity_group.go
```diff
@@ -53,6 +55,13 @@ func identityGroupResource() *schema.Resource {
 				Description: "Policies to be tied to the group.",
 			},
 
+			"external_policies": {
```
lawliet89: This field seems a bit counter-intuitive if you compare it with the `google_xxx_iam_*` or other similar resources.

May I suggest that:

- if `policies` is not provided, do a [`DiffSuppressFunc`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc) to suppress a diff on `policies`
- Likewise, do not set `policies` while writing the entity.

Then, users can use the new resources you have added to manage policies.

lawliet89: Did you miss out a `resource_identity_policies.go`?

vault/resource_identity_group_policies.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/vault/api"
+	"github.com/terraform-providers/terraform-provider-vault/util"
+)
+
+func identityGroupPoliciesResource() *schema.Resource {
```
lawliet89: Should this resource simply "ensure" that the set of policies it has is present and does not take exclusive control of all the policies for the group? (Something like [`google_project_iam_member`](https://www.terraform.io/docs/providers/google/r/google_project_iam.html#google_project_iam_member) v.s. `google_project_iam_policy`) Otherwise it does not seem to be any more useful than setting the policies directly in `vault_identity_group`.

vault/resource_identity_group.go
```diff
@@ -18,13 +18,15 @@ func identityGroupResource() *schema.Resource {
...
 		Exists: identityGroupExists,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
```
lawliet89: Ditto for `identity_entity` above.

vault/resource_identity_entity.go
nazarewk: It definitely does NOT require `ForceNew`, names of entities can be changed any time and only constant is object's ID.

I am not sure what are the interactions for `Computed` fields that are passed as parameters. Can you describe them @lawliet89 ?

vault/resource_identity_group_policies.go
nazarewk: This separates `group_id` from group creation allowing usage of `group_id` in templated policies and then assigning them to groups. Without this resource we cannot use ids in templates.

vault/resource_identity_group.go
nazarewk: Vault should auto-generate names so it might be optional as well.

vault/resource_identity_entity.go
lawliet89: `Computed is often used to represent values that are not user configurable or can not be known at time of terraform plan or apply, such as date of creation or a service specific UUID. `

Anyway, good point on the `ForceNew`. I didn't realise identity is special compared with other resources.

vault/resource_identity_group_policies.go
lawliet89: Sure.

It would still be useful if this resource does not take exclusive control on the list of policies.

lawliet89: The use case is that I would like to have multiple `vault_identity_group_policies` in more than one place to collectively define the list of policies for the entire group.

nazarewk: I am pretty sure it will lead to race conditions and unstable state, but we can try implementing single-policy management.

vault/resource_identity_group.go
nazarewk: I don't think it is possible to tell whether the value was set by user or read from the Vault.

vault/resource_identity_group_policies.go
nazarewk: I'm trying to implement single-policy management, but I am pretty sure it will lead to many hard to debug race conditions and unstable state, since policies are not external Vault object, but just a field on the Group object.

lawliet89: You're right that's unfortunate.

nazarewk: I've rolled back `vault_identity_group_policy` and defaulting to exclusive policy management on `vault_identity_group_policies`

lawliet89: Thanks for trying this out.

The GCP API has a similar issue but they get around it by using some [mutex](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/iam_project.go#L68).

Probably requires more work in the future for this.

nazarewk:
@lawliet89 i guess this is ready to be merged?

cvbarros:
@nazarewk since you got this underway, would you mind breaking a few more eggs and addressing the problem described in #343?
Apparently the reserved `id` field in the resource schema for these resources is causing problems with Terraform 0.12. AFAIK, I've never implmented any resourced that had an `id`  field set in the schema, just always used `SetId()` to make the attribute available for output.

nazarewk:
@cvbarros i'm not sure what is the interaction of `id` attribute, tests are passing after simply removing the field, so i will stick with it.

cvbarros:
@nazarewk with previous Terraform versions, one could get away of having `id` attribute in the resource schema with just a warning. 

What happens is that it is a common mistake to assume that in order to have a resource's `id` attribute exported, one would have to add it to the schema. This is not the case, since Terraform exposes it by default. It's safe (and recommended, since TF 0.12 is being more strict) to remove it.

nazarewk:
Is there some maintainer participating in conversation? It is ready for merging and I would like to proceed.

cvbarros:
@nazarewk I think repo is undergoing a codefreeze at the moment for #356. It's a pretty big change, and probably all PRs will have to be rebased once that biggie is done.

tyrannosaurus-becks:
Thanks @cvbarros and @nazarewk ! The fix for #343 would indeed be nice to have but can also be done separately, so I will go ahead and take this one in as-is. Much appreciated!

AndresPineros:
Hello, I just downloaded Terraform 0.12 and I'm getting this issue:
`Error: Failed to instantiate provider "vault" to obtain schema: Incompatible API version with plugin. Plugin version: 4, Client versions: [5]`

I'm don't really know the src of Terraform of the provider, can anyone explain why this happen? Is there a solution?

Thanks :+1:

PR #330, opened 2019-02-28
Generic sercret data

mbamber:
This makes the `generic_secret` resource export the data as a map (much in the same way as the data source of the same name.

This is useful when we want to obtain information from the resource metadata for use as an input to another resource. Data sources are not suitable in this case, as they have to be available at plan time.

Output of acceptance tests:
```bash
$ make testacc TESTARGS='-run=TestResourceGenericSecret'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestResourceGenericSecret -timeout 120m
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.021s [no tests to run]
--- PASS: TestResourceGenericSecret (6.99s)
--- PASS: TestResourceGenericSecret_deleted (5.26s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	12.264s
```

### DIFF
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -60,6 +60,12 @@ func genericSecretResource() *schema.Resource {
 				Default:     false,
 				Description: "Don't attempt to read the token from Vault if true; drift won't be detected.",
 			},
+
+			"data": {
+				Type:        schema.TypeMap,
+				Computed:    true,
+				Description: "Map of strings read from Vault.",
+			},
 		},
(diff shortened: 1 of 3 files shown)


mbamber:
Is there anything I can do to help get this merged in? We are currently building and testing against our version of this provider, but would rather use the version maintained by yourself

Thanks

tyrannosaurus-becks:
Looks good, thank you!

PR #376, opened 2019-03-31
Upgrade SDK to v0.11

appilon:
As discussed this PR will upgrade the sdk to v0.11 with govendor. I had to update a few of the transitive dependencies. Upgrading from sdk v0.10 to v0.11 required remove `id` from schema definitions, there was 1 instance of that.


vault/resource_namespace.go
```diff
@@ -19,12 +19,6 @@ func namespaceResource() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"id": {
```
tyrannosaurus-becks: Hey! I'm curious why this is stripped. [There's a general process](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal) that Terraform usually performs when removing an attribute that we may want to follow here depending on the situation.

I did run the `resource_namespace_test.go` file against the current version of Vault enterprise and it did pass, so that piece is looking good.

appilon: Ah yes! `id` is a special attribute that the SDK actually automatically provides. It does not need to be declared in the schema, when upgrading from sdk v0.10 to v0.11 it is no longer allowed to declare it in the schema. So this is not a breaking change all configs can access the `id`!

tyrannosaurus-becks: :+1:

PR #377, opened 2019-04-01
Fix approle_secret-id exists return

jorgemarey:
Fix the return value of the exists function

### DIFF
--- a/vault/resource_approle_auth_backend_role_secret_id.go
+++ b/vault/resource_approle_auth_backend_role_secret_id.go
@@ -238,7 +238,7 @@ func approleAuthBackendRoleSecretIDExists(d *schema.ResourceData, meta interface
 	if err != nil {
 		// We need to check if the secret_id has expired
 		if util.IsExpiredTokenErr(err) {
-			return true, nil
+			return false, nil
 		}
 		return true, fmt.Errorf("error checking if AppRole auth backend role SecretID %q exists: %s", id, err)
 	}


vault/resource_approle_auth_backend_role_secret_id.go
```diff
@@ -238,7 +238,7 @@ func approleAuthBackendRoleSecretIDExists(d *schema.ResourceData, meta interface
 	if err != nil {
 		// We need to check if the secret_id has expired
 		if util.IsExpiredTokenErr(err) {
-			return true, nil
+			return false, nil
 		}
 		return true, fmt.Errorf("error checking if AppRole auth backend role SecretID %q exists: %s", id, err)
```
tyrannosaurus-becks: Looks like a secret ID exists when the Approle API returns a successful response with a body. A successful response with a nil body means there's no secret, and an error from the API means something went wrong and we don't know whether there's a secret ID.

Should we return false and an error for both line 241 _and_ line 243?

jorgemarey: I don't know. I guess terraform won't check the bool if there's an error. I left this to true because I saw that behaviour on other resources like:

https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_approle_auth_backend_login.go#L182

https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_approle_auth_backend_role.go#L370

But I'll made the change if you think that's more correct.

tyrannosaurus-becks: I think it would be cool to return an error on both 241 and 243 to flag that we don't know whether the secret ID exists. As for whether to also return true or false, yes I think you're right, it would be good to return true for both because of precedent, and because it doesn't matter too much anyways.

jorgemarey: In the 241 (the one with the change), I think we shouldn't return an error if the funciont IsExpiredTokenErr returns true. If it has expired and we return an error terraform won't try to create another one and will fail, so we should just return false and let terraform create a new token. Does that makes sense to you?

The same behaviour I'm explaining is shown here:

https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_approle_auth_backend_login.go#L180

Currently, with the code as it is, when the secret_id has expired any execution of terraform will fail and instead it should just create a new secret_id.

tyrannosaurus-becks: OK! I think that's reasonable!

PR #380, opened 2019-04-04
ensure compatibility with Vault 1.0.X where role_type is not yet defined (appeared in 1.1.0)

aromeyer:
Ensure compatibility with Vault 1.0.X where role_type is not yet defined (appeared in 1.1.0)

### DIFF
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -218,7 +218,10 @@ func jwtAuthBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.Set("num_uses", tokenNumUses)
 
-	d.Set("role_type", resp.Data["role_type"].(string))
+	if resp.Data["role_type"] != nil {
+		d.Set("role_type", resp.Data["role_type"].(string))
+	}
+
 	d.Set("bound_subject", resp.Data["bound_subject"].(string))
 
 	if resp.Data["bound_cidrs"] != nil {


taldanzig:
👍 just ran into this and verified that this PR fixes the issue

tyrannosaurus-becks:
Hi @aromeyer , thanks for working on this!

So, to verify this fix, I ran the present tests for `resource_jwt_auth_backend_role` against Vault 1.0.3 and I was able to reproduce the panic. Just to make sure it wasn't happening presently, I also ran against version 1.1.2 of Vault, and found the same tests passed.

Then I pulled your branch here and ran the same tests against 1.0.3. I found that while the tests didn't panic, they did fail with this output:
```
--- FAIL: TestAccJWTAuthBackendRole_import (0.05s)
    testing.go:538: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=9) "role_type": (string) (len=3) "jwt"
        }
```
However, using this code against Vault 1.1.2 still had the tests pass. 

Just curious, do the tests pass for you if you use the version of this provider that was available at the same time as the version of Vault you're using? If so, does it make sense to stick with that?

tyrannosaurus-becks:
Closing because we'd prefer not to attempt to support past versions of Vault with the present Terraform Vault Provider. If possible, please use the version of the provider that was available at that time. Thank you!

PR #386, opened 2019-04-05
implement renewal of pki certs with min_seconds_remaining argument

joemiller:
This implements renewal for certificates in the `vault_pki_secret_backend_cert` and `vault_pki_secret_backend_sign` resources.  Related issue: https://github.com/terraform-providers/terraform-provider-vault/issues/353

Some notes on the implementation:

* Adds a new optional argument `min_seconds_remaining`. The  default is 7 days (604800s). If the certificate's `expiration` is within this many seconds a new certificate will be fetched from Vault. The reason I chose to go with seconds instead of days was to cover cases where short-lived (perhaps less than a day) certificates are desired. I suppose "hours" might be a reasonable choice here as well. Days would be too long.

* Adds a new computed attribute `expiration` to these resources. This value was already being returned by the Vault API and now it will be stored in TF state.

* Based loosely off of the way in which the acme TF provider handles cert refresh, except they use `min_days_remaining`. This seems fine for certs issued by public CAs but some internal CAs may desire shorter expiration - https://github.com/terraform-providers/terraform-provider-acme/blob/0077b51dadc8b9fc4485e5b2a2844dedb063d9ae/acme/resource_acme_certificate.go#L83-L109

* I bumped the `mitchellh/copystructure` and `mitchellh/reflectwalk` vendor'd deps to `v1.0.0`. I ran into a bug `panic: interface conversion: interface {} is schema.schemaMap, not *schema.schemaMap` when implementing the custom Diff funcs and this was the resolution.

### DIFF
--- a/vault/resource_pki_secret_backend_cert.go
+++ b/vault/resource_pki_secret_backend_cert.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
@@ -12,10 +13,11 @@ import (
 
 func pkiSecretBackendCertResource() *schema.Resource {
 	return &schema.Resource{
-		Create: pkiSecretBackendCertCreate,
-		Read:   pkiSecretBackendCertRead,
-		Update: pkiSecretBackendCertUpdate,
-		Delete: pkiSecretBackendCertDelete,
+		Create:        pkiSecretBackendCertCreate,
+		Read:          pkiSecretBackendCertRead,
+		Update:        pkiSecretBackendCertUpdate,
+		Delete:        pkiSecretBackendCertDelete,
+		CustomizeDiff: pkiSecretBackendCertDiff,
 
 		Schema: map[string]*schema.Schema{
 			"backend": {
@@ -91,6 +93,18 @@ func pkiSecretBackendCertResource() *schema.Resource {
 				Description: "Flag to exclude CN from SANs.",
 				ForceNew:    true,
 			},
+			"auto_renew": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "If enabled, a new certificate will be generated if the expiration is within min_seconds_remaining",
+			},
+			"min_seconds_remaining": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     604800,
+				Description: "Generate a new certificate when the expiration is within this number of seconds",
+			},
 			"certificate": {
 				Type:        schema.TypeString,
 				Computed:    true,
@@ -121,6 +135,11 @@ func pkiSecretBackendCertResource() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 				Description: "The serial number.",
+			},
+			"expiration": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The certificate expiration.",
 			},
 		},
 	}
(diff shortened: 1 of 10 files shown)


joemiller:
tests are passing now. The issue was that the test case's pki role has a 1 hour max ttl, but min_seconds_remaining defaults to 7days. I added `min_seconds_remaining = 60` to the test so that it would not immediately be marked for renewal

joemiller:
Any feedback on this PR? thanks

vault/resource_pki_secret_backend_cert.go
```diff
@@ -192,16 +205,47 @@ func pkiSecretBackendCertCreate(d *schema.ResourceData, meta interface{}) error
...
+	expireTime := time.Unix(int64(expiration), 0)
+
+	minSeconds := 0
+	if v, ok := d.GetOk("min_seconds_remaining"); ok {
+		minSeconds = v.(int)
+	}
+
+	renewTime := expireTime.Add(-time.Duration(minSeconds) * time.Second)
+	if time.Now().After(renewTime) {
+		log.Printf("[DEBUG] certificate %q is due for renewal, expires %s, renewal time %s", d.Id(), expireTime, renewTime)
```
tyrannosaurus-becks: It would be great to get some test coverage of this code path. I see that the rest of the code in this method is currently hit, but not in here.

vault/resource_pki_secret_backend_cert.go
```diff
@@ -192,16 +205,47 @@ func pkiSecretBackendCertCreate(d *schema.ResourceData, meta interface{}) error
...
+}
+
 func pkiSecretBackendCertRead(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
 func pkiSecretBackendCertUpdate(d *schema.ResourceData, m interface{}) error {
+	// If the certificate or private_key have been marked as changed by our custom diff function we need
+	// to create a new certificate and key
+	if d.HasChange("certificate") || d.HasChange("private_key") {
```
tyrannosaurus-becks: This code path doesn't appear to have test coverage yet. Would be great to get some here too.

vault/resource_pki_secret_backend_sign.go
```diff
@@ -194,16 +206,46 @@ func pkiSecretBackendSignCreate(d *schema.ResourceData, meta interface{}) error
...
+
+	renewTime := expireTime.Add(-time.Duration(minSeconds) * time.Second)
+	if time.Now().After(renewTime) {
+		log.Printf("[DEBUG] certificate %q is due for renewal, expires %s, renewal time %s", d.Id(), expireTime, renewTime)
+		d.SetNewComputed("certificate")
+		return nil
+	}
+
+	log.Printf("[DEBUG] certificate %q is not due for renewal, expires %s, renewal time %s", d.Id(), expireTime, renewTime)
+	return nil
```
tyrannosaurus-becks: It would be great to get test coverage on all the code changes in this file too - the new fields that have been added, and the certificate that's generated.

vault/resource_pki_secret_backend_cert.go
joemiller: ok, i'll work on that

joemiller:
@tyrannosaurus-becks I am happy to make this opt-in for backwards compatibility. i'll work on the requested changes

joemiller:
@tyrannosaurus-becks I've added an `auto_renew = [true|false] (default: false)` attribute to both the `vault_pki_secret_backend_sign` and `vault_pki_secret_backend_cert` resources  to aid with backward compat, as requested. Also added acceptance tests covering renewal. Though I notice a switch to circleci is in progress and unlike the previous travis tests it's not running the acceptance tests currently.

I also include a commit with some doc fixes. 1) cleaned up some trailing whitespace and 2) "ttl - time to leave" I am assuming should be "time to live" per the related vault docs

joemiller:
updated to remove vendor/vendor.json which i noticed i had accidentally re-added in my most recent rebase from master

PR #387, opened 2019-04-06
Allow vault_ldap_auth resources to be imported

lucymhdavies:
Fixes https://github.com/terraform-providers/terraform-provider-vault/issues/383

Having pushed this, I now see that the `ldap_auth_backend` part of this may become redundant, given https://github.com/terraform-providers/terraform-provider-vault/pull/273

### DIFF
--- a/vault/resource_ldap_auth_backend.go
+++ b/vault/resource_ldap_auth_backend.go
@@ -21,6 +21,9 @@ func ldapAuthBackendResource() *schema.Resource {
 		Read:   ldapAuthBackendRead,
 		Delete: ldapAuthBackendDelete,
 		Exists: ldapAuthBackendExists,
+		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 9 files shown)


lucymhdavies:
Okay, that's _weird_

I see it's failing in Travis because it needs a `make fmt`

```
0.35s$ make test
gofmt needs running on the following files:
./vault/resource_pki_secret_backend_intermediate_cert_request.go
./vault/resource_pki_secret_backend_root_cert.go
You can use the command: `make fmt` to reformat code.
make: *** [fmtcheck] Error 1
The command "make test" exited with 2.
```

Which is the exact thing I added 6118e49fd2e4d773b4bd5e30d5847deb79bed17e to fix. Though looking more closely at what `make fmt` did on my machine to create 6118e49, it's the opposite way round to what I'd expect.

I'll remove 6118e49 from my branch, and this should work.

PR #395, opened 2019-04-12
Handle "all policies deleted" corner case

marcinwyszynski:
When the user intends to delete all policies, they don't delete any.
This also leads to an inconsistent state which will show changes on every `plan`.

### DIFF
--- a/vault/resource_aws_auth_backend_role.go
+++ b/vault/resource_aws_auth_backend_role.go
@@ -535,9 +535,8 @@ func awsAuthBackendRoleUpdate(d *schema.ResourceData, meta interface{}) error {
 	if v, ok := d.GetOk("period"); ok {
 		data["period"] = v.(int)
 	}
-	if len(policies) > 0 {
-		data["policies"] = policies
-	}
+
+	data["policies"] = policies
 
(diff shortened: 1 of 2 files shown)


tyrannosaurus-becks:
Hi @marcinwyszynski ! Thanks for working on this.

Question. Correct me if I'm mistaken, but aren't policies required for a role to be valid? Why would a person want to delete policies without deleting the entire role? 

Just trying to understand the use case. Thanks!

marcinwyszynski:
Hey @tyrannosaurus-becks, I believe that if don't set policies explicitly, the role will be assigned a `default` policy, well... by default :) That's what's causing the problem, really - you're trying to set an empty list from Terraform, but Vault set the `default` policy, which gets reported back when you do the `read`, and makes Terraform want to change the resource again. Hope I'm making some sense :)

marcinwyszynski:
Referring to your question explicitly:

> Why would a person want to delete policies without deleting the entire role?

To make the role only use the implicit `"default"` role.

tyrannosaurus-becks:
Thanks @marcinwyszynski ! That makes sense!

PR #396, opened 2019-04-16
Don't overwrite vault_namespace path with response value.

williams-brian:
Overwriting the path with the one returned by Vault causes an invalid namespace to be set when using nested namespaces.
ex: `namespace-b` which is inside `namespace-a` will have the path overwritten with `namespace-a/namespace-b/` when the correct value is `namespace-b/`

### DIFF
--- a/vault/resource_namespace.go
+++ b/vault/resource_namespace.go
@@ -75,7 +75,6 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	// remove trailing slash
 	namespacePath := filepath.Clean(resp.Data["path"].(string))
 
-	d.Set("path", namespacePath)
 	d.SetId(resp.Data["id"].(string))
 
 	return nil
--- a/vault/resource_namespace.go
+++ b/vault/resource_namespace.go
@@ -3,7 +3,6 @@ package vault
 import (
 	"fmt"
 	"log"
-	"path/filepath"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/vault/api"
@@ -71,9 +70,6 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("error reading from Vault: %s", err)
 	}
-
-	// remove trailing slash
-	namespacePath := filepath.Clean(resp.Data["path"].(string))
 
 	d.SetId(resp.Data["id"].(string))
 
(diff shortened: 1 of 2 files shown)


tyrannosaurus-becks:
Hi @williams-brian , thanks for working on this! 

First question, would you be willing to add a test that duplicates the issue, and then shows that this code solves it? 

Also, @petems , would you want to take a look at this one too?

williams-brian:
@tyrannosaurus-becks I have attempted to add a test case which demonstrates the issue, but I am not sure how to run the tests with Vault enterprise, so I haven't been able to verify that it works correctly. 

I hope my explanation of the issue is clear. After running `terraform apply` the namespace path that appears in the Terraform state is different from the one which is set on the terraform resource. Subsequent runs of `terraform apply` then fail. My test case attempts to check that the path inside the terraform state is the same as what is on the "vault_namespace" resource.

williams-brian:
I figured out how to run Vault Enterprise locally and use that for acceptance tests. The new test case should properly illustrate the issue.

Running the test before this change:
```
$ go test -v github.com/terraform-providers/terraform-provider-vault/vault -run TestNamespace_basic
=== RUN   TestNamespace_basic
--- FAIL: TestNamespace_basic (0.06s)
    testing.go:538: Step 2 error: Check failed: expected path to be child-namespace-5645055735786586850, got test-namespace-7756255336919402260/child-namespace-5645055735786586850
    testing.go:599: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: Error refreshing: 1 error(s) occurred:
        
        * vault_namespace.test_child: 1 error(s) occurred:
        
        * vault_namespace.test_child: vault_namespace.test_child: error reading from Vault: Error making API request.
        
        URL: GET http://localhost:8200/v1/sys/namespaces/test-namespace-7756255336919402260/child-namespace-5645055735786586850
        Code: 400. Errors:
        
        * "/" is not allowed in namespace names
        
        State: <nil>
FAIL
FAIL	github.com/terraform-providers/terraform-provider-vault/vault	0.082s
```

Running the test after this change:
```
$ go test -v github.com/terraform-providers/terraform-provider-vault/vault -run TestNamespace_basic
=== RUN   TestNamespace_basic
--- PASS: TestNamespace_basic (0.08s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	0.098s
```

petems:
Good catch, I forgot to remove the logic around removing slashes after we added the validation function, which would break nested namespaces.

This solution makes sense to me! 😄

PR #397, opened 2019-04-17
allowed_organization_units maps incorrectly to vault api.

damomurf:
The allowed_organization_units field on a cert_auth_backend_role resource in Terraform is labelled incorrectly when it should be allowed_organizational_units. This change adapts to the difference and ensures the value is actually set and read properly from vault.

The intent behind the existing tests is unclear to me and certainly doesn't complain before or after the change in this PR.

### DIFF
--- a/vault/resource_cert_auth_backend_role.go
+++ b/vault/resource_cert_auth_backend_role.go
@@ -167,7 +167,7 @@ func certAuthResourceWrite(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if v, ok := d.GetOk("allowed_organization_units"); ok {
-		data["allowed_organization_units"] = v.(*schema.Set).List()
+		data["allowed_organizational_units"] = v.(*schema.Set).List()
 	}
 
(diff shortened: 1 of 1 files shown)


tyrannosaurus-becks:
Hi @damomurf , good catch!

It looks like the code was attempting to check correctness [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_cert_auth_backend_role_test.go#L150), but was skipping the check [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_cert_auth_backend_role_test.go#L161-L162). 

Would you be willing to at least update the field names in the tests? Like check the project for any further usages of `allowed_organization_units` and replace it?

Assuming the tests pass at that point, should be good to go, though it looks like the tests on this neck of code could use some additional :heart: .

tyrannosaurus-becks:
Closing due to inactivity.

damomurf:
@tyrannosaurus-becks So this has been closed and not fixed? The issue still remains in the provider.

PR #399, opened 2019-04-23
Add pki config urls

### LINKED ISSUE #363: No ability to set PKI secret engines CRL
Hi There,

Before generating a root certificate, one should set the configuration URLs for the issuing cert and the distribution endpoint: see https://www.vaultproject.io/docs/secrets/pki/index.html, step 4.

There are no options in the `vault_pki_secret_backend_root_cert` resource to do this. I tried using the `vault_generic_secret` and it worked, however `pki/config/urls`  does not have a DELETE operation, instead you have to set the parameters to empty. As a result, I couldn't get terraform to roll back itself automatically since it kept trying to initiate a delete command. 

Finally I had to simply remove that particular resource from the state file via the command: `terraform state rm vault_generic_secret.pki_config` to get things working again.

We need a means to initiate a set on that particular path and in the case of rollback a setting of an empty string instead.

### Terraform Version
Terraform v0.11.13
Vault Provider 1.6.0
Vault v1.0.3

### Affected Resource(s)
- vault provider
- vault_pki_secret_backend_root_cert
- vault_generic_secret

```hcl
resource "vault_generic_secret" "pki_config" {
  path         = "pki/config/urls"
  disable_read = true

  data_json = <<EOT
  {
    "issuing_certificates" : "https://172.0.0.1:8200/v1/pki/ca",
    "crl_distribution_points" : "https://172.0.0.1:8200/v1/pki/crl"
  }
EOT
}
```

### Expected Behavior
Correctly update/delete the resource and parameters in Vault

### Actual Behavior
```
* vault_generic_secret.pki_config: error deleting "pki/config/urls" from Vault: "Error making API request.\n\nURL: DELETE https://172.40.1.96:8200/v1/pki/config/urls\nCode: 405. Errors:\n\n* 1 error occurred:\n\t* unsupported operation\n\n"
```

### Steps to Reproduce
1. enable pki engine in Vault
2. `terraform apply` to get resource created.
3. Add count = 0 to resource
4. `terraform apply` resource will attempt to be deleted.

edmeister:
This PR adds functionality to set the issuing certificate endpoints, CRL distribution points, and OCSP server endpoints that will be encoded into issued certificates.

Fixes #363

### DIFF
--- a/vault/resource_pki_secret_backend_config_urls.go
+++ b/vault/resource_pki_secret_backend_config_urls.go
@@ -96,7 +96,6 @@ func pkiSecretBackendConfigUrlsRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("crl_distribution_points", config.Data["crl_distribution_points"])
 	d.Set("ocsp_servers", config.Data["ocsp_servers"])
 
-
 	return nil
 }
 
--- a/vault/resource_pki_secret_backend_config_urls.go
+++ b/vault/resource_pki_secret_backend_config_urls.go
@@ -1,0 +1,137 @@
+package vault
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+func pkiSecretBackendConfigUrlsResource() *schema.Resource {
+	return &schema.Resource{
+		Create: pkiSecretBackendConfigUrlsCreate,
+		Read:   pkiSecretBackendConfigUrlsRead,
+		Update: pkiSecretBackendConfigUrlsUpdate,
+		Delete: pkiSecretBackendConfigUrlsDelete,
+
+		Schema: map[string]*schema.Schema{
+			"backend": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Optional:    false,
+				ForceNew:    false,
+				Description: "The path of the PKI secret backend the resource belongs to.",
+			},
+			"issuing_certificates": {
+				Type:        schema.TypeString,
+				Required:    false,
+				Optional:    true,
+				Description: "Specifies the URL values for the Issuing Certificate field. This can be a comma-separated string list.",
+				ForceNew:    false,
+			},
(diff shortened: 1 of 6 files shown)


vault/resource_pki_secret_backend_config_urls.go
```diff
@@ -0,0 +1,136 @@
...
+		return fmt.Errorf("error updating URL config for PKI secret backend %q: %s", backend, err)
+	}
+	log.Printf("[DEBUG] Updated URL config on PKI secret backend %q", backend)
+
+	return pkiSecretBackendConfigUrlsRead(d, meta)
+
+}
+
+func pkiSecretBackendConfigUrlsDelete(d *schema.ResourceData, meta interface{}) error {
+	return nil
```
tyrannosaurus-becks: Will the linked issue be solved if this isn't implemented?

edmeister: There is no delete method in the Vault API.

The linked issue uses a generic secret as a workaround, for which the Vault provider expects to have a delete method. Since Vault does not allow `DELETE /pki/config/urls`, a terraform destroy fails, leaving an orphaned mount behind. The only option to proceed is to manually disable the mounted secret. 

A workaround for the lack of delete method would be to update with empty strings. However, it might make more sense to just stick with the Vault API.

tyrannosaurus-becks: Ah thanks for the explanation! Does that mean, then, that we would get the same behavior from the Terraform Vault Provider if we stripped having this PKI Delete method? Like if we just struck [this](https://github.com/terraform-providers/terraform-provider-vault/pull/399/files#diff-7a80efbc38ff34d692c9de568dc5080aR17) whole line?

edmeister: According to [Terraform's guide](https://www.terraform.io/docs/extend/writing-custom-providers.html#defining-resources), the Delete method is mandatory for the resource to function.

tyrannosaurus-becks: Gotcha! Thanks!

PR #407, opened 2019-05-03
aws_secret_backend_role: support role_arns argument

oxlay:
This PR adds support in the provider for the [`role_arns` parameter](https://www.vaultproject.io/api/secret/aws/index.html#parameters-3) of AWS secret backend roles, [added in Vault 0.11](https://github.com/hashicorp/vault/pull/4360).

### DIFF
--- a/vault/resource_aws_secret_backend_role.go
+++ b/vault/resource_aws_secret_backend_role.go
@@ -37,7 +37,7 @@ func awsSecretBackendRoleResource() *schema.Resource {
 			"policy_arns": {
 				Type:          schema.TypeList,
 				Optional:      true,
-				ConflictsWith: []string{"policy", "policy_arn"},
+				ConflictsWith: []string{"policy", "policy_arn", "role_arns"},
 				Description:   "ARN for an existing IAM policy the role should use.",
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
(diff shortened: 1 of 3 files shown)


vault/resource_aws_secret_backend_role.go
```diff
@@ -70,6 +70,16 @@ func awsSecretBackendRoleResource() *schema.Resource {
 				Required:    true,
 				Description: "Role credential type.",
 			},
+			"role_arns": {
```
tyrannosaurus-becks: Do `role_arns` conflict with other fields in practice? [The documentation](https://www.vaultproject.io/api/secret/aws/index.html#role_arns) doesn't say it does. Should we remove it from the conflicts with clauses?

vault/resource_aws_secret_backend_role.go
```diff
@@ -97,8 +107,14 @@ func awsSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
...
 
-	if policy == "" && len(policyARNs) == 0 {
-		return fmt.Errorf("either policy or policy_arn must be set.")
+	var roleARNs []string
+	roleARNsIfc := d.Get("role_arns")
+	for _, roleIfc := range roleARNsIfc.([]interface{}) {
+		roleARNs = append(roleARNs, roleIfc.(string))
+	}
+
+	if policy == "" && len(policyARNs) == 0 && len(roleARNs) == 0 {
```
tyrannosaurus-becks: Not sure this check is needed.

vault/resource_aws_secret_backend_role.go
oxlay: * `policy_document` apparently does not conflict with `role_arns` (see [test using both](https://github.com/joelthompson/vault/blob/f88c1b37b751a9add6ed3c0487fa9a0bf2bb0362/builtin/logical/aws/backend_test.go#L657-L658)), so I've removed the ConflictsWith clause for this one
* `policy_arns` and `policy_arn` are only valid when `credential_type` is `iam_user`, while `role_arns` is only valid when `credential_type` is `assumed_role` and prohibited otherwise (per [the doc](https://www.vaultproject.io/api/secret/aws/index.html#parameters-3)), so in practice, they are mutually exclusive
* for `policy`, the doc says "cannot be mixed with the parameters listed above", hence the `ConflictsWith` clause

vault/resource_aws_secret_backend_role.go
oxlay: Are you talking about the whole `if { }` or just the `&& len(roleARNs) == 0` part?

hExPY:
+1

asksmruti:
Could you please make release out of this change? Desperately need this feature

mgerlach:
+1

zee-ahmed:
+1

oxlay:
@tyrannosaurus-becks any chance you can take a second look soon? 😄

PR #420, opened 2019-05-17
Implement "accessor" attribute on vault_okta_auth_backend

jtcressy:
The okta backend needed a smidge of love. Both ldap and github auth backends implemented the "accessor" attribute, and okta should have one too.

Story:
As a consumer of Vault Identities backends, I like to set aliases on my groups to bind them to groups from an auth provider. These aliases require that you know the `accessor` of an auth mount to properly bind the identity. With the github auth mount, you can simply use the interpolation syntax: 

`${vault_github_auth_backend.mybackend.accessor}`

However since okta hasn't exposed this attribute, I've had to use more hacky ways of programmatically fetching the accessor (like reading `sys/mfa/method/okta/myorg.okta.com` and getting the `mount_accessor` from the returned data). 

If this gets merged, it can be done just like `github` and `ldap` and I can reference the accessor directly from the auth mount resource. `${vault_okta_auth_backend.myokta.accessor}`

Output from unit tests:
```
=== RUN   TestOktaAuthBackendGroup
--- PASS: TestOktaAuthBackendGroup (0.16s)
=== RUN   TestOktaAuthBackend
--- PASS: TestOktaAuthBackend (0.18s)
=== RUN   TestOktaAuthBackendUser
--- PASS: TestOktaAuthBackendUser (0.14s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	9.954s
```

Documentation added for new attribute as well.

### DIFF
--- a/vault/resource_okta_auth_backend.go
+++ b/vault/resource_okta_auth_backend.go
@@ -192,6 +192,12 @@ func oktaAuthBackendResource() *schema.Resource {
 				},
 				Set: resourceOktaUserHash,
 			},
+
+			"accessor": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The mount accessor related to the auth mount.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


jtcressy:
Question I have for everyone:
Should I also implement the `tune` parameter? I should probably just bring this inline with the github and ldap auth methods but I didn't want to blow out scope too much.

PR #415, opened 2019-05-13
Implement tests alongside an improved fix for #368

jtcressy:
This is both a fix and a feature add around #368 and an improvement upon #405 

Adds optional `token_namespace` parameter to provider config.

Hopefully this allows you to auth to /ns1 and perform resource creation in a sub-namespace like /ns1/ns2
You can also auth to / but configure /ns1 by setting `token_namespace` to "/" and `namespace` to "/ns1"
The default behavior is to auth to `namespace` and configure resources in `namespace`.

Todo:
- [x]  Docs for new parameters to provider config

I would like to get some constructive feedback on this too. I'm not an expert on TF's unit testing framework so I've likely done it incorrectly despite the tests passing. (I watched the logs of the vault server when It did tests and they appear to be doing what I want)

Output from acceptance testing:
```
=== RUN   TestAccNamespaceProviderConfigure
--- PASS: TestAccNamespaceProviderConfigure (0.30s)
PASS
ok  	terraform-provider-vault/vault	6.641s
```
Output of `vault status` for the test acceptance server I was running:
```
> $ vault status
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         1.1.0+pro
Cluster Name    vault-cluster-996fb35d
Cluster ID      f9c2dd34-62b2-02a5-3a93-aaf74dca82ef
HA Enabled      false
```

PR #433, opened 2019-06-04
Read Namespace Information from Token

chrishoffman:
This is an alternative approach to #415 where instead of providing the token namespace explicitly, we just read it from the token.  I  believe this is a better approach and will result in a more natural behavior from the provider.


jtcressy:
This is super neat! Definitely better than my solution.

mack1040:
@chrishoffman / @tyrannosaurus-becks has this been in released in version 1.9 or 2.0.0?

jtcressy:
@mack1040 This was released in 1.9, and rolled into 2.0

PR #414, opened 2019-05-12
Add support for assumed_role and STS

ShimiTaNaka:
Add support to get assumed_role credentials and add option to set default STS TTL and max STS TTL

PR #444, opened 2019-06-13
Add STS TTLs to `vault_aws_secret_backend_role`

lawliet89:
Also make lists become sets


ShimiTaNaka:
Also covered in #414

tyrannosaurus-becks:
Happy to take a look at this once the tests are passing!

lawliet89:
The tests are fixed. It was a flaky test in the Vault Policy data source.

vault/resource_aws_secret_backend_role.go
```diff
@@ -80,6 +81,18 @@ func awsSecretBackendRoleResource() *schema.Resource {
 				ConflictsWith: []string{"policy", "policy_arn", "policy_arns"},
 				Description:   "ARNs of AWS roles allowed to be assumed. Only valid when credential_type is 'assumed_role'",
 			},
+			"default_sts_ttl": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "The default TTL in seconds for STS credentials. When a TTL is not specified when STS credentials are requested, and a default TTL is specified on the role, then this default TTL will be used. Valid only when credential_type is one of assumed_role or federation_token.",
+			},
+			"max_sts_ttl": {
```
tyrannosaurus-becks: In looking at the Vault code [here](https://github.com/hashicorp/vault/blob/master/builtin/logical/aws/path_roles.go#L88), I'm seeing that these are both fields of TypeDurationSecond. Those are a little funny in Vault because they can also be represented as strings. We did that to increase usability, so users could give times like "10s" to represent 10 seconds. Anyways, [this PR](https://github.com/terraform-providers/terraform-provider-vault/pull/314/files) gives a really good example of handling TTL fields.

lawliet89: I originally set them to strings, but Vault always returns them as integer number of seconds (for this endpoint at least), IIRC. So if someone set this to `"1h"`, Vault will return `3600`  and I didn't want to implement logic to "unify" this with the `"1h"` string that the user might have provided.

Let me double check this and get back.

lawliet89: (I should also probably add `ValidateFunc` to these two fields so that they are not silently ignored when the credential type is wrong.)

lawliet89:
```
--- FAIL: TestAccAWSSecretBackendRole_basic (0.17s)
    testing.go:568: Step 1 error: Check failed: Check 2/15 error: vault_aws_secret_backend_role.test_policy_inline: Attribute 'default_sts_ttl' expected "1h", got "3600"
--- PASS: TestAccAWSSecretBackendRole_import (0.17s)
--- FAIL: TestAccAWSSecretBackendRole_nested (0.20s)
    testing.go:568: Step 1 error: Check failed: Check 3/15 error: vault_aws_secret_backend_role.test_policy_inline: Attribute 'default_sts_ttl' expected "1h", got "3600"
FAIL
FAIL	github.com/terraform-providers/terraform-provider-vault/vault	0.549s
```

Vault returns `3600` when the input is set to `1h`.

tyrannosaurus-becks: Alright! I think that will be fine.

PR #446, opened 2019-06-13
Upgrade to 0.12

paddycarver:
This monster PR upgrades the provider to use the Terraform 0.12 SDK. Because a breaking change was required for that, it also removes deprecated fields.

I also ran all the acceptance tests, and quite a few were broken, so I fixed tests as I went. This now passes all acceptance tests, including tests that rely on external databases or APIs.

A manual examination of whether the `github_auth_backend.tune`, `okta_auth_backend.user`, `okta_auth_backend.group`, and `jwt_auth_backend.tune` fields fall prey to the [optional computed block issue](https://www.terraform.io/docs/extend/terraform-0.12-compatibility.html#computed-resource-attributes). I'll update in this PR when that's done, but I figured it best to share what I had when all the tests pass, to get a start on review.

It may be helpful to browse changes by commit.


vault/resource_auth_backend.go
```diff
@@ -146,6 +147,7 @@ func authBackendRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	for path, auth := range auths {
+		path = strings.TrimSuffix(path, "/")
```
lawliet89: This should fix https://github.com/terraform-providers/terraform-provider-vault/issues/182

tyrannosaurus-becks:
Wow, @paddycarver , thank you!

paddycarver:
@tyrannosaurus-becks I had to push c3e9342 (commit message has more details) to fix one last set of issues, this should be code complete and good to go now. Does this still look good to you? If so, I'll push an upgrade guide PR in the morning and we can roll this out.

grahamc:
Note this PR did not update the documentation for how to use `vault_rabbitmq_secret_backend_role`, however last May someone opened a PR correcting the documentation. Perhaps that -1/+6 PR could get reviewed: https://github.com/hashicorp/terraform-provider-vault/pull/787

PR #462, opened 2019-07-09
Exposing local flag on vault_mount resource

hashitop:
We've encountered an issue where the `-local` flag is supported by Vault API as well as CLI according to this document below

https://learn.hashicorp.com/vault/operations/mount-filter#cli-command-3

Current version of terraform-provider-vault does not support this attribute

https://www.terraform.io/docs/providers/vault/r/mount.html

> ## Argument Reference
> 
> The following arguments are supported:
> 
> * `path` - (Required) Where the secret backend will be mounted
> 
> * `type` - (Required) Type of the backend, such as "aws"
> 
> * `description` - (Optional) Human-friendly description of the mount
> 
> * `default_lease_ttl_seconds` - (Optional) Default lease duration for tokens and secrets in seconds
> 
> * `max_lease_ttl_seconds` - (Optional) Maximum possible lease duration for tokens and secrets in seconds
> 
> * `options` - (Optional) Specifies mount type specific options that are passed to the backend

I have examined the source code in Vault API and found that the data structure `MountInput` has already accommodated this flag hence API is working, only the provider is not currently exposing the flag at the `vault_mount` resource. 

The file `resource_mount.go` is modified to allow flag `local` as boolean type as optional for cluster environment, in case local mount is required, to pass through to API via `MountInput` data structure. Please review and share comment if any concern.


hashitop:
Hi @tyrannosaurus-becks 
another test case added, trying to keep the same pattern, and it's been unit test locally as well but I think circleci failed because of some other test.

tyrannosaurus-becks:
Thanks for adding the test! I double-checked the test by merging in master and pushing it up, and I found that the build does succeed with master merged in, so we should be safe.

PR #466, opened 2019-07-10
Add tool for analyzing Vault API coverage

tyrannosaurus-becks:
This PR adds a tool for analyzing the level of Vault API coverage we have, as well as what is and isn't supported. 

The report it produces is like this:
```
/auth/kubernetes/config is in the Terraform Vault Provider multiple times
/sys/policy/{name} is in the Terraform Vault Provider multiple times
/auth/kubernetes/role/{name} is in the Terraform Vault Provider multiple times
 
16 percent coverage
70 of 429 vault paths are supported
359 of 429 vault paths are unsupported
 
SUPPORTED
    /auth/approle/login
    /auth/approle/role/{role_name}
    /auth/approle/role/{role_name}/custom-secret-id
    ...more...
 
UNSUPPORTED
    /ad/config
    /ad/creds/{name}
    /ad/roles
    ...more...
```

It will also flag resources or datasources that haven't gotten inventoried by causing tests to fail with useful output.

This report will now be run as part of the CI tests, so it'll be possible to view it by simply viewing the CI tests run with the latest commit. For an example, see the "Details" of the last successful build of this PR.


cmd/coverage/main.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// TODO if an endpoint's only method is LIST, omit it from the comparison
+// TODO actually, make this an api endpoint in the plugin? or put a list in the readme or linked there?
+// TODO ensure the paths in the inventory match something in openapi but exclude enterprise endpoints
+// TODO also make sure they're only covered once and have a match?
+// TODO make an exception for GenericPath and UnknownPath
+var pathToOpenAPIDoc = flag.String("openapi-doc", "", "path/to/openapi.json")
+
+// This tool is used for generating a coverage reports regarding
```
malnick: report, singular?

cmd/coverage/main_test.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"flag"
+	"testing"
+)
+
+func Test_Main(t *testing.T) {
+	pathToDoc := "/home/tbex/go/src/github.com/terraform-providers/terraform-provider-vault/testdata/openapi.json"
```
malnick: May want to consider using a path in `/tmp` or something that isn't unique to your local machine.

cmd/coverage/main_test.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"flag"
+	"testing"
+)
+
+func Test_Main(t *testing.T) {
```
malnick: This should be MixedCaps: https://golang.org/doc/effective_go.html#mixed-caps

If you do use `TestMain()` note that it does have special intentions: https://golang.org/pkg/testing/#hdr-Main

Mainly (no pun intended), `TestMain()` is used for setup and tear-down of tests, and other test cases won't be called directly.

vault/provider.go
```diff
@@ -93,88 +113,342 @@ func Provider() terraform.ResourceProvider {
...
+	// from Vault's home directory, running "$ ./scripts/gen_openapi.sh", and then by
+	// drilling into the paths with "$ cat openapi.json | jq ".paths" | jq 'keys[]'".
+	// Here's a short example of how paths and their path variables should be represented:
+	//		"/transit/keys/{name}/config"
+	//		"/transit/random"
+	//		"/transit/random/{urlbytes}"
+	//		"/transit/sign/{name}/{urlalgorithm}"
+	PathInventory []string
+
+	// Defaults to false, but should be marked true if a resource is enterprise only.
```
malnick: I'm not sure how deep we go on lint checks, but you might consider the form:
```
// EnterpriseOnly defaults to...
```

Reference https://golang.org/doc/effective_go.html#commentary

vault/provider.go
```diff
@@ -93,88 +113,342 @@ func Provider() terraform.ResourceProvider {
...
+}
 
-		DataSourcesMap: map[string]*schema.Resource{
-			"vault_approle_auth_backend_role_id":   approleAuthBackendRoleIDDataSource(),
-			"vault_kubernetes_auth_backend_config": kubernetesAuthBackendConfigDataSource(),
-			"vault_kubernetes_auth_backend_role":   kubernetesAuthBackendRoleDataSource(),
-			"vault_aws_access_credentials":         awsAccessCredentialsDataSource(),
-			"vault_generic_secret":                 genericSecretDataSource(),
-			"vault_policy_document":                policyDocumentDataSource(),
+// TODO add a tool that lists the path inventory, or even performs a check against Vault so it's super easy
```
malnick: I think this is outdated as of this PR?

cmd/coverage/main_test.go
tyrannosaurus-becks: Weird that showed up for you, it was that way during my first commit but not the second. Should be currently resolved on the branch to be a relative directory.

cmd/coverage/main_test.go
tyrannosaurus-becks: Good point! I think I will remove the test, since `Main` has no immediately testable data returned. 

Instead, for now, I think I will have the tool run as a CI step so it's easy to view there as more of a one-off thing. If it's helpful to make it _more_ evident and usable down the line, I could also later add it to a local file and reference it in the README.

vault/provider.go
tyrannosaurus-becks: :+1: Will do.

vault/provider.go
tyrannosaurus-becks: Odd, I think you must have been looking at the first commit here rather than the PR as a whole. This is no longer present when viewing the code at the second commit.

cmd/coverage/main.go
tyrannosaurus-becks: Fixed!

tyrannosaurus-becks:
All feedback is addressed. Merging so other contributors can work off this code.

PR #468, opened 2019-07-11
Add token_type to vault_auth_backend resource

ryansheppard:
This PR adds `token_type` to the generic `vault_auth_backend`. It will use the backend default unless specified in the `resource` block. I do not see any issues referencing this problem in particular, but we have a use case requiring batch tokens for an AWS auth backend.

### DIFF
--- a/vault/resource_auth_backend.go
+++ b/vault/resource_auth_backend.go
@@ -73,6 +73,15 @@ func authBackendResource() *schema.Resource {
 				Description: "Specifies whether to show this mount in the UI-specific listing endpoint",
 			},
 
+			"token_type": {
+				Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


tyrannosaurus-becks:
Hi @ryansheppard ! Thanks for working on this, it's much appreciated.

I'd previously thought of this resource as being mainly for the `/sys/auth/{path}` endpoint [here](https://www.vaultproject.io/api/system/auth.html#enable-auth-method), which doesn't appear to support a token type field. Are you using it for a different endpoint? If so, can you link me to it?

tyrannosaurus-becks:
Closing due to lack of activity. Thank you, though, and feel free to reopen this later if you still want it and end up circling back.

PR #476, opened 2019-07-26
Minor enhancements to `vault_gcp_secret_roleset`

### LINKED ISSUE #469: vault_gcp_secret_roleset doesn't reflect service _account_email change
I'm using the vault 2.1.0 provider and having an issue with the resource not updating properly.

I'm using the following resource configuration to set up service account access:

resource "vault_gcp_secret_roleset" "role" {
  backend      = "gcp"
  roleset      = "role"
  secret_type  = "access_token"
  token_scopes = ["https://www.googleapis.com/auth/cloud-platform"]
  project      = data.google_project.terraform.project_id

  binding {
    resource = data.google_folder.details.name

    roles = [
      "roles/compute.osAdminLogin",
      "roles/storage.objectViewer",
      "roles/iam.serviceAccountUser"
    ]
  }
}

### Actual Behavior
If I add a new role to the roles list (example "roles/compute.viewer") , the resource correctly updates and adds the role, but in the process it recreates the service account email.  The terraform plan doesn't indicate that the email will change and other resources that depend on the output value of the service_account_email don't update.

### Expected Behavior
Either the role should be updated without changing the email, or the plan should note that the email will change and dependent resources should update.

lawliet89:
- Document Import Support
- Added some more test cases
- Fix #469  ~but this seems to cause a permanent diff in the plan. See https://github.com/hashicorp/terraform/issues/17411#issuecomment-515314101~ (Worked around the Terraform bug)

### DIFF
--- a/vault/resource_gcp_secret_roleset.go
+++ b/vault/resource_gcp_secret_roleset.go
@@ -121,7 +121,7 @@ func gcpSecretRolesetCreate(d *schema.ResourceData, meta interface{}) error {
 		d.SetId("")
 		return fmt.Errorf("Error writing GCP Secrets backend roleset %q: %s", path, err)
 	}
-	log.Printf("[DEBUG] Wrote GCP Secrets backend rolese %q: %s", path, err)
+	log.Printf("[DEBUG] Wrote GCP Secrets backend roleset %q", path)
 
 	return gcpSecretRolesetRead(d, meta)
 }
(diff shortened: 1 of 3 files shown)


lawliet89:
```
make testacc TESTARGS='-run=TestGCPSecretRoleset' 
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestGCPSecretRoleset -timeout 120m
PASS
=== RUN   TestGCPSecretRoleset
--- PASS: TestGCPSecretRoleset (37.54s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	37.558s
```

PR #477, opened 2019-07-26
Add Resources for Transit Secrets Backend

jtcressy:
Hi all, 

I wanted a structured way to create transit backends and encryption keys via Terraform, so I went ahead and implemented it for us all.

This PR implements two new resources:
- `vault_transit_secret_backend`
- `vault_transit_secret_backend_key`

The transit secret backend is fairly trivial and primitive, since it's simply wrapping `vault_mount` by setting `type = "transit"`. However I wanted to control the keys within the mount so it seemed natural for there to be resource for the backend.

There is more that could be done on top of this, as I have an idea to implement some new data sources around transit:
- Data sources for encrypting/decrypting/signing/verifying some input value for use elsewhere in terraform (similar to how data.template_file is used)
- Data sources for exporting existing keys, both public and private if applicable.

Checklist:
- [x] Tests cases created & pass
- [x] Documentation created/updated

Acceptance test output:
```
=== RUN   TestTransitSecretBackendKey_basic
--- PASS: TestTransitSecretBackendKey_basic (0.56s)
=== RUN   TestTransitSecretBackendKey_rsa4096
--- PASS: TestTransitSecretBackendKey_rsa4096 (2.19s)
=== RUN   TestTransitSecretBackendKey_import
--- PASS: TestTransitSecretBackendKey_import (0.33s)
=== RUN   TestTransitSecretBackend_basic
--- PASS: TestTransitSecretBackend_basic (0.39s)
=== RUN   TestTransitSecretBackend_import
--- PASS: TestTransitSecretBackend_import (0.32s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	3.986s

Process finished with exit code 0

```


jtcressy:
Howdy, @tyrannosaurus-becks! I'm back with more commits. Let me know if there's any problems with this PR so I can fix them.

vault/provider.go
```diff
@@ -446,6 +446,14 @@ var (
 			Resource:      pkiSecretBackendSignResource(),
 			PathInventory: []string{"/pki/sign/{role}"},
 		},
+		"vault_transit_secret_backend": {
+			Resource:      transitSecretBackendResource(),
+			PathInventory: []string{UnknownPath},
```
tyrannosaurus-becks: In looking at the code, it looks like this is for mounting the Transit backend. I believe this is already covered with actually a couple more fields by the [mount resource](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L362-L365). Is there more to this than I think? Or should we take this out since the mount resource already covers this?

jtcressy: Well, I've been taking many hints from the pki resources, and it seemed like it would fit the pattern to have a transit backend resource to accompany the transit key resource. 

technically `vault_mount` covers what `vault_pki_secret_backend` already does.

jtcressy: edit: See #L409 and note that `vault_pki_secret_backend` is also registered with a null path.

tyrannosaurus-becks: Ah! That actually explains a lot. I had wondered what endpoint it was hitting and had been moving quite quickly through everything trying to list endpoints. When it wasn't immediately evident I marked it as an `UnknownPath`. I would actually much prefer to deprecate that endpoint rather than making it a pattern, because it would be quite a lot to maintain if more and more were introduced. I actually will go and add a deprecation message to that resource; my apologies for the confusion.

tyrannosaurus-becks: FYI #490

jtcressy: Alright, I removed `vault_transit_secret_backend` and updated docs and tests to use `vault_mount` with `type` set to `transit`.

PR #478, opened 2019-07-29
Update various Authentication Resources for Vault 1.2

lawliet89:
Vault 1.2 has the change:

> auth/token: Token store roles use new, common token fields for the values that overlap with other auth backends. period, explicit_max_ttl, and bound_cidrs will continue to work, with priority being given to the token_ prefixed versions of those parameters. They will also be returned when doing a read on the role if they were used to provide values initially; however, in Vault 1.4 if period or explicit_max_ttl is zero they will no longer be returned. (explicit_max_ttl was already not returned if empty.)

This causes the provider to panic when used with Vault 1.2.

This PR attempts to fix these panics by deprecating some fields and adding new ones.

TODOs:
- [x] Update `resource_cert_auth_backend_role`
- [ ] Update Documentation
- [x] Some more tests

Questions for @tyrannosaurus-becks :
- Some resources requires setting all fields or omitted fields get reset to defaults. This makes it hard for the provider to provide backward compatibility because we now have to set all new `token_` fields when updating or they get reset. I think this can be done by querying Vault for its version but it adds complexity.
- There's quite a bit of complexity for backwards support and it's hard to test.  Do you think it's worth doing a major version bump and breaking compatibility?


cmd/coverage/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"os"
 	"sort"
 
-	"github.com/hashicorp/vault/logical/framework"
+	"github.com/hashicorp/vault/sdk/framework"
```
tyrannosaurus-becks: Nice catch! Thank you!

lawliet89:
I will submit a new pr to update the documentation.

StephenWithPH:
@lawliet89 and @tyrannosaurus-becks and @paddycarver  : this is great work; we've been waiting for this.

Purely an FYI: we were waiting on this so much that we enabled [GitHub release notifications](https://help.github.com/en/articles/watching-and-unwatching-releases-for-a-repository), but we didn't get a notification for this release. After digging, [I learned that](https://github.community/t5/How-to-use-Git-and-GitHub/Not-getting-release-notification-despite-quot-Watching-releases/m-p/19725#M5890):
>Without adding a message to the release, it only counts as a "tag" and not as a full release.

I suspect this also explains why the releases page doesn't emphasize release `2.2.0`... it thinks this was just a tag.

Screenshot from 2019-08-09 11-39-32

42wim:
This breaks nomad's vault integration, it still expects `Period`.

https://github.com/hashicorp/nomad/blob/83a75b310d60b6e9f3c6ca6d43c0a5ee104ef3fd/nomad/vault.go#L902-L904

kalafut:
Note to future readers: This change first appeared in version 2.2.0 of the provider.

PR #479, opened 2019-07-31
document auto_renew attribute

joemiller:
Follow-up to #386. Documents the `auto_renew` attribute.

Addresses https://github.com/terraform-providers/terraform-provider-vault/issues/353#issuecomment-516623427


joemiller:
This is a docs only PR which I don't think caused the CI failure. There's a failed build a few runs earlier with a similar failure. Possibly a transient failure? I am not able re-run the build though

PR #481, opened 2019-08-02
Add Azure Secret Backend resource

petems:
First of the two! (See https://github.com/terraform-providers/terraform-provider-vault/issues/460)


vault/provider.go
```diff
@@ -240,6 +240,10 @@ var (
 			Resource:      awsSecretBackendRoleResource(),
 			PathInventory: []string{"/aws/roles/{name}"},
 		},
+		"vault_azure_secret_backend": {
+			Resource:      azureSecretBackendResource(),
+			PathInventory: []string{"/azure/config/"},
```
tyrannosaurus-becks:
```suggestion
			PathInventory: []string{"/azure/config"},
```

vault/resource_azure_secret_backend.go
```diff
@@ -0,0 +1,183 @@
...
+					if strings.HasSuffix(value, "/") {
+						errs = append(errs, fmt.Errorf("path cannot end in '/'"))
+					}
+					return
+				},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return old+"/" == new || new+"/" == old
+				},
+			},
+			"description": {
```
tyrannosaurus-becks: Question! I note that this isn't a field that's present on this API endpoint in Vault. Just curious about why it's added, and also why `ForceNew` is true. Doesn't that mean that if the description were to change, the object would be deleted and recreated remotely?

vault/resource_azure_secret_backend.go
```diff
@@ -0,0 +1,183 @@
...
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Human-friendly description of the mount for the backend.",
+			},
+			"subscription_id": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				Description: "The subscription id for the Azure Active Directory.",
+			},
```
tyrannosaurus-becks: Should this be marked sensitive too?

vault/resource_azure_secret_backend.go
petems: All mounts can have a description added (https://www.vaultproject.io/api/system/mounts.html#description) but yes, probably shouldn't be ForceNew, let me fix that

petems: Fixed! 😄

vault/resource_azure_secret_backend.go
petems: Yep, fixed

PR #483, opened 2019-08-02
add new JWKS options for the JWT auth mount

jcarrothers-sap:
Vault v1.2.0 added:
- jwks_url
- jwks_ca_pem

This change provides support for the new options.

Also did some refactoring on resource_jwt_auth_backend.go and its
documentation in order to make them more consistent.

### DIFF
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -130,7 +130,7 @@ func jwtCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
 	_, jwksUrlExists = d.GetOk("jwks_url")
 	_, jwtPubKeysExists = d.GetOk("jwt_validation_pubkeys")
 
-	if (discUrlExists && (jwksUrlExists || jwtPubKeysExists)) || (jwksUrlExists && (discUrlExists || jwtPubKeysExists)) || (jwtPubKeysExists && (discUrlExists || jwksUrlExists)) {
+	if !(discUrlExists || jwksUrlExists || jwtPubKeysExists) {
 		return errors.New("exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided")
 	}
 
@@ -138,6 +138,7 @@ func jwtCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
 }
 
 var (
+	// TODO: build this from the Resource Schema?
 	matchingJwtMountConfigOptions = []string{
 		"oidc_discovery_url",
 		"oidc_discovery_ca_pem",
@@ -235,7 +236,10 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	configuration := map[string]interface{}{}
 	for _, configOption := range matchingJwtMountConfigOptions {
-		configuration[configOption] = d.Get(configOption)
+		// Set the configuration if the user has specified it, or the attribute is in the Diff
+		if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
+			configuration[configOption] = d.Get(configOption)
+		}
 	}
 
 	_, err := client.Logical().Write(jwtConfigEndpoint(path), configuration)
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -17,6 +17,8 @@ func jwtAuthBackendResource() *schema.Resource {
 		Read:   jwtAuthBackendRead,
 		Update: jwtAuthBackendUpdate,
 
+		CustomizeDiff: jwtCustomizeDiff,
+
 		Schema: map[string]*schema.Schema{
 
 			"path": {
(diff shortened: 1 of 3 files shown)


vault/resource_jwt_auth_backend.go
```diff
@@ -106,6 +123,35 @@ func jwtAuthBackendResource() *schema.Resource {
...
 }
 
+func jwtCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
+	var _ interface{}
+	var discUrlExists, jwksUrlExists, jwtPubKeysExists bool
+	_, discUrlExists = d.GetOk("oidc_discovery_url")
+	_, jwksUrlExists = d.GetOk("jwks_url")
+	_, jwtPubKeysExists = d.GetOk("jwt_validation_pubkeys")
+
+	if (discUrlExists && (jwksUrlExists || jwtPubKeysExists)) || (jwksUrlExists && (discUrlExists || jwtPubKeysExists)) || (jwtPubKeysExists && (discUrlExists || jwksUrlExists)) {
```
lawliet89: Isn't this check handled by `ConflictsWith` on the fields?

vault/resource_jwt_auth_backend.go
```diff
@@ -192,37 +233,9 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
...
-		configuration["oidc_client_id"] = v.(string)
-	}
-	if v, ok := d.GetOkExists("oidc_client_secret"); ok {
-		configuration["oidc_client_secret"] = v.(string)
-	}
-	if v, ok := d.GetOkExists("default_role"); ok {
-		configuration["default_role"] = v.(string)
+	configuration := map[string]interface{}{}
+	for _, configOption := range matchingJwtMountConfigOptions {
+		configuration[configOption] = d.Get(configOption)
```
tyrannosaurus-becks: This logic appears subtly different from the logic above. Above, most of the parameters are only set _if they exist_ whereas here they're just set. I'm concerned that this might cause the update to behave differently than before.

vault/resource_jwt_auth_backend_test.go
```diff
@@ -101,17 +108,40 @@ resource "vault_jwt_auth_backend" "jwt" {
 `, path)
 }
 
-func testAccJWTAuthBackendConfigFull(path string, oidcDiscoveryUrl string, validationPublicKeys string, boundIssuer string, supportedAlgs string) string {
+func testAccJWTAuthBackendConfigODIC(path string, oidcDiscoveryUrl string, boundIssuer string, supportedAlgs string) string {
```
kalafut:
```suggestion
func testAccJWTAuthBackendConfigOIDC(path string, oidcDiscoveryUrl string, boundIssuer string, supportedAlgs string) string {
```

vault/resource_jwt_auth_backend.go
jcarrothers-sap: You're right, as written this check is redundant.  What I had intended to do was replace the check that was originally part of the update method to ensure that at least one of these properties is set.

vault/resource_jwt_auth_backend.go
jcarrothers-sap: Good point, I'll adjust the logic here.

jcarrothers-sap:
@tyrannosaurus-becks I've left my review history in my branch as I'm expecting you to squash this PR.  Let me know if you'd like me to do otherwise.

jcarrothers-sap:
@tyrannosaurus-becks Unfortunately this change seems to have missed v2.2.0.  When do you think the next release will be and do you expect this will be in it?

Thanks

vault/resource_jwt_auth_backend.go
```diff
@@ -106,6 +123,36 @@ func jwtAuthBackendResource() *schema.Resource {
 	}
 }
 
+func jwtCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
```
kalafut: This is checking that existence is >= 1. Consider something like https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_config.go#L175 to check that exactly one method is present.

vault/resource_jwt_auth_backend_test.go
```diff
@@ -101,17 +108,40 @@ resource "vault_jwt_auth_backend" "jwt" {
 `, path)
 }
 
-func testAccJWTAuthBackendConfigFull(path string, oidcDiscoveryUrl string, validationPublicKeys string, boundIssuer string, supportedAlgs string) string {
+func testAccJWTAuthBackendConfigFullODIC(path string, oidcDiscoveryUrl string, boundIssuer string, supportedAlgs string) string {
```
kalafut:
```suggestion
func testAccJWTAuthBackendConfigFullOIDC(path string, oidcDiscoveryUrl string, boundIssuer string, supportedAlgs string) string {
```

kalafut:
@jcarrothers-sap Sorry for the delay. I've had a chance to look through this and try it out, and it all looks good aside from a tiny typo. Thanks!

PR #488, opened 2019-08-05
Add resources for Vault Identity Tokens

lawliet89:
- Add `vault_identity_oidc_key`
- Add `vault_identity_oidc_role` resource
- Add `vault_identity_oidc` resource
- Add `vault_identity_oidc_key_allowed_client_id`

The name for the last resource might be a bit too long... :sweat_smile:


vault/provider.go
```diff
@@ -400,6 +404,22 @@ var (
...
+			Resource:      identityOidc(),
+			PathInventory: []string{"/identity/oidc/config"},
+		},
+		"vault_identity_oidc_key": {
+			Resource:      identityOidcKey(),
+			PathInventory: []string{"/identity/oidc/key/{name}"},
+		},
+		"vault_identity_oidc_key_allowed_client_id": {
+			Resource:      identityOidcKeyAllowedClientId(),
+			PathInventory: []string{"/identity/oidc/key/{key_name}"},
```
tyrannosaurus-becks: Can you point me towards the API docs for this one?

vault/resource_identity_oidc_role.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+
+			"ttl": {
+				Type:        schema.TypeInt,
+				Description: "TTL of the tokens generated against the role in number of seconds.",
+				Optional:    true,
+				Default:     86400,
+			},
+
+			"client_id": {
```
tyrannosaurus-becks: I'm curious about why this is added, I don't see it listed [here](https://www.vaultproject.io/api/secret/identity/tokens.html#create-or-update-a-role). I also don't see it listed below in lines 62:64 so I'm thinking it might have slipped in.

lawliet89:
Hi @tyrannosaurus-becks. I will answer several of your questions in a comment since they are mostly related.

## Client IDs

The steps to use an Identity Token Role involve:

1. Creating a Named key (`vault_identity_oidc_key`). When creating the key, you can either configure the `allowed_client_ids` to be `["*"]` and skip step 3 later, or leave it empty for now.
1. Create a Role (`vault_identity_oidc_role`). Upon creation, Vault will generate a `client_id` which is a Computed only argument in the resource.
1. You will need to go back to the named key and add the generated `client_id` into the list of `allowed_client_ids`.

I hope you can see that there is a circular dependency here. You cannot know the `client_id` from a Role since it's computed and generated by Vault. But the role requires a Named Key and the Named Key needs a list of `allowed_client_ids`. Thus, I "invented" the `vault_identity_oidc_key_allowed_client_id` to allow us to decouple the adding of the `client_id` of the role after it's been created.

## Mutex

The issue now is that we cannot have both `vault_identity_oidc_key` and `vault_identity_oidc_key_allowed_client_id` resources write to the same key at the same time. Thus, I have the following code in functions that write to the key:

```go
vaultMutexKV.Lock(path)
defer vaultMutexKV.Unlock(path)
```

where the `path` is the path to the named key in Vault.

This pattern is first used in the AWS Provider for resources like [`aws_security_group`](https://www.terraform.io/docs/providers/aws/r/security_group.html) and [`aws_security_group_rule`](https://www.terraform.io/docs/providers/aws/r/security_group_rule.html) resources. The AWS Security groups also have the same issue in that the authors wanted to allow decoupled declaration of rules but AWS only has one single API resource for rules and security group. 

The AWS Provider also [declares](https://github.com/terraform-providers/terraform-provider-aws/blob/947c5b4d1d8b9c3b83824b407ec0dc354fe1bcbc/aws/provider.go#L1077) a global provider KV Mutex which is what I did here.

Hope this answers your questions. I will add tests for the slice utility functions.

vault/provider.go
```diff
@@ -25,6 +26,9 @@ const (
 	UnknownPath = "unknown"
 )
 
+// This is a global MutexKV for use within this plugin.
+var vaultMutexKV = mutexkv.NewMutexKV()
```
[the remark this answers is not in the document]
lawliet89: Please see my explanation for the [mutex](https://github.com/terraform-providers/terraform-provider-vault/pull/488#issuecomment-518463146) below. I will add an explanation to the mutex.

vault/provider.go
lawliet89: Please see [comment](https://github.com/terraform-providers/terraform-provider-vault/pull/488#issuecomment-518463146).

vault/resource_identity_oidc_role.go
lawliet89: Please see [comment](https://github.com/terraform-providers/terraform-provider-vault/pull/488#issuecomment-518463146).

util/util.go
```diff
@@ -141,3 +141,31 @@ func ShortDur(d time.Duration) string {
 	}
 	return s
 }
+
+func SliceHasElement(list []interface{}, search interface{}) (bool, int) {
+	for i, ele := range list {
+		if ele == search {
```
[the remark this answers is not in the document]
lawliet89: - Modified to use `DeepEqual` for comparison
- Added bunch of tests.

util/util.go
```diff
@@ -141,3 +141,31 @@ func ShortDur(d time.Duration) string {
...
+
+	return list
+}
+
+func SliceRemoveIfPresent(list []interface{}, search interface{}) []interface{} {
+	if found, index := SliceHasElement(list, search); found {
+		// Set the index we found to be the last item
+		list[index] = list[len(list)-1]
+		// Return slice sans last item
+		return list[:len(list)-1]
```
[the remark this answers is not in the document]
lawliet89: Added a bunch of tests.

kalafut:
Hi @lawliet89. I'm reviewing your PR and everything is looking good so far. One not-obvious thing about the relationship between roles and keys is that you can reference a non-existent named key when you make a role. Of course it will fail if _used_, but it won't fail at config time. So you should be able to:

1. Make a role with the desired (but not yet existing) named key
2. Save the client_id from the new role
3. Make the named key, referencing the client_id

That might simplify some logic.

lawliet89:
Hi @kalafut I am afraid what you suggested is not really possible because the resource model of Terraform combined with the API of Vault do not make that possible.

> Make a role with the desired (but not yet existing) named key

Vault will error if the key does not exist. Terraform does not have a concept of "deferring" writing of resources unless with implicit or explicit dependencies. But as I have explained, the resources have circular dependencies.

> Save the client_id from the new role

To continue from the point above, this does not exist until after Vault has created the role.

lawliet89:
@kalafut I see what you mean. But I think there is, I think, there is a use case for the `vault_identity_oidc_key_allowed_client_id` resource to allow more flexible adding of Client IDs to a named key.

I will add another example and explanation to the documentation to make the two workflows clearer.

vault/resource_identity_oidc_key_test.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func testAccCheckIdentityOidcKeyDestroy(s *terraform.State) error {
+	client := testProvider.Meta().(*api.Client)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "vault_identity_oidc_key" {
+			continue
+		}
+		secret, err := client.Logical().Read(identityEntityIDPath(rs.Primary.ID))
```
kalafut: Do you instead want something like: `resp, err := identityOidcKeyApiRead(rs.Primary.Attributes["name"], client)` ?

Not that when looking at this I discovered that Vault is incorrectly returning a 400 for key not found instead of 404. I've just fixed that, but in the interim you may need the subsequent check to be something like:

`if err != nil && !strings.Contains(err.Error(), "no named key found") {`

kalafut:
@lawliet89 Yeah I can see that use, and actually I'd misinterpreted things the first time around and thought the `vault_identity_oidc_key_allowed_client_id` was required. So I'm OK with this.

kalafut:
@lawliet89 Key deletion will be blocked if there are roles referencing the key. For this provider I see a couple of issues:

1. If you're trying to remove a key and there are referencing roles, the returned error isn't very helpful "Error: error IdentityOidcKey "mykey".  The API response is: `unable to delete key "mykey" because it is currently referenced by these roles: myrole`
2. If you're trying to remove both a key and associated roles, this will error the first time during the key delete step (there are still roles), but then it deletes the roles and an second `terraform apply` will work. The deletes should be ordered so roles are always removed before keys.

kalafut:
@lawliet89 By the way, I think this is good as-is if we hit our release/merge window, and the latest comments can be follow-on improvements if need be.  Thanks!

vault/resource_identity_oidc_key_test.go
lawliet89: I've fixed this and added the check in `identityOidcKeyApiRead`(commit 34517b60662fbe3e3e560cfa3382f00f19da7442).

lawliet89:
> If you're trying to remove a key and there are referencing roles, the returned error isn't very helpful "Error: error IdentityOidcKey "mykey". The API response is: unable to delete key "mykey" because it is currently referenced by these roles: myrole

Fixed the error formatting for that. 

> If you're trying to remove both a key and associated roles, this will error the first time during the key delete step (there are still roles), but then it deletes the roles and an second terraform apply will work. The deletes should be ordered so roles are always removed before keys.

I don't think this can be done from the provider. Terraform will delete things in reverse dependency order. So the user will have to implicitly or explicitly have the role depend on the key for the destroy to work in one single operation.

tyrannosaurus-becks:
I'm going to go ahead and merge this because I see that the linked PR was taken off this branch, so its diff will be smaller when this is merged. Going to merge this then go take a look.

PR #492, opened 2019-08-07
Use Mutex to lock Identity Group Updates

lawliet89:
Use the same mutex mechanism used in #488 to lock Identity groups updates.


tyrannosaurus-becks:
Thank you!

PR #493, opened 2019-08-07
Add vault azure secret backend role

petems:
Now #481 is merged, now lets create the role for azure secrets! 😄 

This one's a bit trickier to test, as it requires real subscription information...

So for the test you'll require the following environment variables:

ARM_TENANT_ID
ARM_CLIENT_ID
ARM_CLIENT_SECRET
ARM_RESOURCE_GROUP

The resource group you can create with the CLI, or use an existing one if needed:

`az group create --name 'azure-vault-tf-testing' && export ARM_RESOURCE_GROUP="azure-vault-tf-testing"`

So, you'll need to create a service principal, then enable the permissions to create new accounts from that account, which is only possible with Powershell 😭  (so you have to do it from the Azure Cloud Shell)

Doc to do that: https://www.terraform.io/docs/providers/azuread/auth/service_principal_configuration.html


petems:
From my personal testing with my setup:

```hcl
	subscription_id = "<REDACTED>"
	tenant_id = "<REDACTED>"
	client_id = "<REDACTED>"
	client_secret = "<REDACTED>"
}

resource "vault_azure_secret_backend_role" "test" {
  backend                     = "${vault_azure_secret_backend.azure.path}"
  role                        = "test"
  ttl                         = 300
	max_ttl                   = 600
	description				  = "Test for Vault Provider"

	azure_roles {
    role_name = "Reader"
    scope =  "/subscriptions/<REDACTED>/resourceGroups/psouter-vault-secret-azure-backend"
  }
}
```

```
$ 2019-08-07T22:23:09.490+0100 [INFO]  core: successful mount: namespace= path=azure/ type=azure
```

```
$ vault list azure/roles
Keys
----
test
```


```
$ vault read azure/creds/test
Key                Value
---                -----
lease_id           azure/creds/test/[...]
[...]
```

petems:
To remind myself, the full CLI command I ran was:

`envchain azure ENV ARM_RESOURCE_GROUP=psouter-vault-secret-azure-backend TF_ACC=1 go test -timeout 30s github.com/terraform-providers/terraform-provider-vault/vault -run '^(TestAzureSecretBackendRole)$' -v`

vault/resource_azure_secret_backend_role.go
```diff
@@ -0,0 +1,210 @@
...
+			},
+			"application_object_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Application Object ID for an existing service principal that will be used instead of creating dynamic service principals.",
+			},
+			"ttl": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Human-friendly description of the mount for the backend.",
```
tyrannosaurus-becks: Needs updating, not accurate for ttl.

vault/resource_azure_secret_backend_role.go
```diff
@@ -0,0 +1,210 @@
...
+			},
+			"max_ttl": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Human-friendly description of the mount for the backend.",
```
tyrannosaurus-becks: Here too.

vault/resource_azure_secret_backend_role.go
```diff
@@ -0,0 +1,210 @@
...
+func azureSecretBackendRoleUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	path := d.Id()
+
+	data := map[string]interface{}{}
+	azureSecretBackendRoleUpdateFields(d, data)
+
+	log.Printf("[DEBUG] Updating role %q in Azure Secret backend", path)
+	_, err := client.Logical().Write(path, data)
+	if err != nil {
```
tyrannosaurus-becks: This is unimportant and the code looks great as-is. I wanted to share a tip here - sometimes with stuff like this I instead do:
```
if _, err := client.Logical().Write(path, data); err != nil {
    ...
}
```
The benefit is that the err then is only available inside the if statement. It's a slightly more defensive way to code because it doesn't leave the err available for somebody else to accidentally misuse lower down - like for them to accidentally use it in a different goroutine (which would cause a race), or to accidentally interpret an err from here as an err from there. 

Not at all a necessity, just thought I would share in case you felt like adopting this somewhere down the line. This code still looks very good and this is more cake frosting stuff.

vault/resource_azure_secret_backend_role.go
petems: Damn, too much copypasta, will fix!

vault/resource_azure_secret_backend_role.go
petems: Yep, makes a lot of sense, happy to fix! 👍

PR #495, opened 2019-08-08
Fix Non-Exclusive Behaviour of `identity_group_policies`

lawliet89:
This PR uses the same utility functions added in https://github.com/terraform-providers/terraform-provider-vault/pull/488 and should not result in any merge conflicts.

### DIFF
--- a/vault/resource_identity_group.go
+++ b/vault/resource_identity_group.go
@@ -255,7 +255,7 @@ func readIdentityGroupPolicies(client *api.Client, groupID string) ([]interface{
 		return nil, err
 	}
 	if resp == nil {
-		return nil, fmt.Errorf("error reading IdentityGroup does not exist: %s", groupID)
+		return nil, fmt.Errorf("error IdentityGroup %s does not exist", groupID)
 	}
 
 	if v, ok := resp.Data["policies"]; ok && v != nil {
@@ -264,6 +264,7 @@ func readIdentityGroupPolicies(client *api.Client, groupID string) ([]interface{
 	return make([]interface{}, 0), nil
 }
 
+// This function may return `nil` for the IdentityGroup if it does not exist
 func readIdentityGroup(client *api.Client, groupID string) (*api.Secret, error) {
 	path := identityGroupIDPath(groupID)
 	log.Printf("[DEBUG] Reading IdentityGroup %s from %q", groupID, path)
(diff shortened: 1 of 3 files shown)


tyrannosaurus-becks:
Hi @lawliet89 ! Weirdly, GH is still showing those utility functions in the diff even though I merged them to master. Would you be willing to pull master and push it up to fix the diff?

lawliet89:
@tyrannosaurus-becks I have done a rebase with master.

vault/resource_identity_group.go
```diff
@@ -249,23 +249,28 @@ func identityGroupIDPath(id string) string {
...
-	if resp, err := readIdentityGroup(client, groupId); err != nil {
-		return nil, fmt.Errorf("error reading IdentityGroup policies %q: %s", groupId, err)
-	} else {
-		presentPolicies = resp.Data["policies"].([]interface{})
+func readIdentityGroupPolicies(client *api.Client, groupID string) ([]interface{}, error) {
+	resp, err := readIdentityGroup(client, groupID)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil {
```
tyrannosaurus-becks: Nice catch here that the response can be nil.

vault/resource_identity_group.go
```diff
@@ -249,23 +249,28 @@ func identityGroupIDPath(id string) string {
...
+	if v, ok := resp.Data["policies"]; ok && v != nil {
+		return v.([]interface{}), nil
+	}
+	return make([]interface{}, 0), nil
 }
 
-func readIdentityGroup(client *api.Client, groupId string) (*api.Secret, error) {
-	path := identityGroupIDPath(groupId)
-	log.Printf("[DEBUG] Reading IdentityGroup %s from %q", groupId, path)
+func readIdentityGroup(client *api.Client, groupID string) (*api.Secret, error) {
```
tyrannosaurus-becks: Might be nice to add a comment just above this function letting folks know that the resp can be nil.

vault/resource_identity_group_policies.go
```diff
@@ -128,11 +91,7 @@ func identityGroupPoliciesRead(d *schema.ResourceData, meta interface{}) error {
 
 	resp, err := readIdentityGroup(client, id)
 	if err != nil {
-		// We need to check if the secret_id has expired
-		if util.IsExpiredTokenErr(err) {
```
tyrannosaurus-becks: Should we strip this out?

lawliet89: It doesn't really make sense to be here. I think it was just a copy/paste from another resource by the original author. If the user's Vault Token has expired, the whole operation should fail anyway.

vault/resource_identity_group.go
lawliet89: Done.

PR #500, opened 2019-08-13
Fix missing sidebar links and broken links

lawliet89:


greut:
:+1: also `vault_transit_secret_backend_key`.

https://github.com/terraform-providers/terraform-provider-vault/pull/477

PR #506, opened 2019-08-15
Feature/pki crl config

### LINKED ISSUE #472: Unable to Configure CRL Expiry for PKI engine
It currently does not seem possible to configure the PKI CRL expiry through any of the associated resources.

### Terraform Version
Terraform v0.12.2

billyaustin84:
This resolves #472.
This is my first attempt at Go, so all feedback is gratefully received.

### DIFF
--- a/vault/resource_pki_secret_backend_crl_config.go
+++ b/vault/resource_pki_secret_backend_crl_config.go
@@ -42,7 +42,7 @@ func pkiSecretBackendCrlConfigCreate(d *schema.ResourceData, meta interface{}) e
 
 	backend := d.Get("backend").(string)
 
-	path := pkiSecretBackendConfigUrlsPath(backend)
+	path := pkiSecretBackendCrlConfigPath(backend)
 
 	expiry := d.Get("expiry")
 	disable := d.Get("disable")
@@ -59,7 +59,7 @@ func pkiSecretBackendCrlConfigCreate(d *schema.ResourceData, meta interface{}) e
 	}
 	log.Printf("[DEBUG] Created CRL config on PKI secret backend %q", backend)
 
-	d.SetId(fmt.Sprintf("%s/config/crl", backend))
+	d.SetId(path)
 	return pkiSecretBackendCrlConfigRead(d, meta)
 }
 
@@ -87,9 +87,8 @@ func pkiSecretBackendCrlConfigRead(d *schema.ResourceData, meta interface{}) err
 func pkiSecretBackendCrlConfigUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
-	backend := d.Id()
-
-	path := pkiSecretBackendCrlConfigPath(backend)
+	path := d.Id()
+	backend := pkiSecretBackendCrlConfigPath(path)
 
 	expiry := d.Get("expiry")
 	disable := d.Get("disable")
--- a/vault/resource_pki_secret_backend_crl_config.go
+++ b/vault/resource_pki_secret_backend_crl_config.go
@@ -1,0 +1,119 @@
+package vault
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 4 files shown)


vault/resource_pki_secret_backend_crl_config.go
```diff
@@ -0,0 +1,118 @@
...
+		Read:   pkiSecretBackendCrlConfigRead,
+		Update: pkiSecretBackendCrlConfigUpdate,
+		Delete: pkiSecretBackendCrlConfigDelete,
+
+		Schema: map[string]*schema.Schema{
+			"backend": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The path of the PKI secret backend the resource belongs to.",
+			},
```
tyrannosaurus-becks: A lot of endpoints leave this as optional and default it to the engine name. If we went that way, this is probably what it'd look like:
```
"backend": {
	Optional:    true,
	Description: "The path of the PKI secret backend the resource belongs to.",
	ForceNew:    true,
	Default:     "pki",
	// standardise on no beginning or trailing slashes
	StateFunc: func(v interface{}) string {
		return strings.Trim(v.(string), "/")
	},
},
```
Do you think we should make it more like that, or are there reasons it's better as-is?

vault/resource_pki_secret_backend_crl_config.go
```diff
@@ -0,0 +1,118 @@
...
+
+func pkiSecretBackendCrlConfigCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string)
+
+	path := pkiSecretBackendCrlConfigPath(backend)
+
+	expiry := d.Get("expiry")
+	disable := d.Get("disable")
```
tyrannosaurus-becks: For these two, since they're optional, I think we might want to do something more like this instead:
```
data := make(map[string]interface{})
if expiry, ok := d.GetOK("expiry"); ok {
    data["expiry"] = expiry
}
if disable, ok := d.GetOK("disable"); ok {
    data["disable"] = disable
}
```
Then they're only sent to Vault if they were actually specified by the user. This will allow Vault to see that they're unset and use its normal defaults.

vault/resource_pki_secret_backend_crl_config.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	log.Printf("[DEBUG] Updated CRL config on PKI secret backend %q", backend)
+
+	return pkiSecretBackendCrlConfigRead(d, meta)
+
+}
+
+func pkiSecretBackendCrlConfigDelete(d *schema.ResourceData, meta interface{}) error {
+	return nil
+}
```
tyrannosaurus-becks: We should probably go ahead and add this. I think [the AWS auth resource](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_aws_auth_backend_role.go) is a really good one to look at for examples because it's probably very highly used and if it were buggy, we would know! :-)

billyaustin84: I had a look at incorporating this, however Vault currently doesn't appear to support a DELETE on this endpoint.

vault/resource_pki_secret_backend_crl_config.go
billyaustin84: I have taken on board you suggestion and incorporated this in both the Create and Update functions.

vault/resource_pki_secret_backend_crl_config.go
billyaustin84: I have added half of your suggestion here, but "Optional" wouldn't work in conjunction with "Default"

billyaustin84:
@tyrannosaurus-becks Many thanks for your positive and constructive review.  I have addressed where possible.
My circleci build failed but for apparently unrelated reasons:
`fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out`
I don't have permissions to start it again, would you mind restarting it please?

PR #507, opened 2019-08-21
Fix name for all references to organization_units.

damomurf:
Retry of #397 as it was closed:

The allowed_organization_units field on a cert_auth_backend_role resource in Terraform is labelled incorrectly when it should be allowed_organizational_units. This change adapts to the difference and ensures the value is actually set and read properly from vault.

The intent behind the existing tests is unclear to me and certainly doesn't complain before or after the change in this PR.

All references to the incorrectly named field have bee changed.

### DIFF
--- a/vault/resource_cert_auth_backend_role.go
+++ b/vault/resource_cert_auth_backend_role.go
@@ -62,7 +62,7 @@ func certAuthBackendRoleResource() *schema.Resource {
 			Optional: true,
 			Computed: true,
 		},
-		"allowed_organization_units": {
+		"allowed_organizational_units": {
 			Type: schema.TypeSet,
 			Elem: &schema.Schema{
 				Type: schema.TypeString,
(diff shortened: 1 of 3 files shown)


lawliet89:
This might be a breaking change and you might need to deprecate the old attribute first.

>The intent behind the existing tests is unclear to me and certainly doesn't complain before or after the change in this PR.

The test is checking whether an API returned attribute, if it exists, is set correctly in the state or not. Since the previous attribute being tested for did not exist, the test for that attribute is simply skipped.

tyrannosaurus-becks:
@damomurf thanks for your work on this. Yes, @lawliet89 is correct and I should have pointed that out on my previous review that you linked. For this, we'll need to follow [best practices](https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-an-optional-attribute) for renaming an optional attribute, rather than immediately updating it. That will help folks with upgrading to the corrected spelling without experiencing their current configs breaking.

tyrannosaurus-becks:
Closing due to lack of activity. Please feel free to reopen this if you get a chance to circle back around. Thanks!

PR #530, opened 2019-09-05
Add root_rotation_statements to vault_database_secret_backend_connection

### LINKED ISSUE #527: Support root_rotation_statements attribute
The `root_rotation_statements` argument should be supported as a top level attribute on `vault_database_secret_backend_connection` I think.
See the Vault API docs here: https://www.vaultproject.io/api/secret/databases/index.html#configure-connection

You can sorta workaround this by putting it in the `data` map but its not a sensitive attribute and would be nice if it was supported properly as it's own thing!

Related to #202

hamishforbes:
Fixes #527


hamishforbes:
Sure, added!

PR #531, opened 2019-09-06
 Update resource_namespace.go

Kuleshov-Arsentii:
Namespace import fail because of wrong API path:
```
URL: GET https://VAULT_ADDR/v1/sys/namespaces
Code: 405. Errors:

* 1 error occurred:
	* unsupported operation
```
When import resource, func `namespaceRead` get empty path from `d.Get("path")`, so use `d.Id()` as path.
Set path as `Id`, received as import argument.
If call from `namespaceWrite`, we just reset path as was in resource definition.

### DIFF
--- a/vault/resource_namespace.go
+++ b/vault/resource_namespace.go
@@ -64,6 +64,9 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
 	path := d.Get("path").(string)
+	if path == "" {
+		path = d.Id()
+	}
 
 	resp, err := client.Logical().Read("sys/namespaces/" + path)
 
@@ -78,6 +81,7 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.SetId(resp.Data["id"].(string))
+	d.Set("path", path)
 
 	return nil
 }


vault/resource_namespace.go
```diff
@@ -64,6 +64,9 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
 	path := d.Get("path").(string)
+	if path == "" {
```
tyrannosaurus-becks: Since the `"path"` field is marked as required above, is it possible to hit this case?

vault/resource_namespace.go
```diff
@@ -78,6 +81,7 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.SetId(resp.Data["id"].(string))
+	d.Set("path", path)
```
tyrannosaurus-becks: Is it possible to have this method return an error? If so, should we check it?

petems:
I think this PR has the right idea, as it's not possible to import an existing namespace with the current code: The namespace ID isn't used to lookup an existing namespace, it's the path. So I've opened a new PR to move the ID to it's own attribute and use the path as ID

https://github.com/terraform-providers/terraform-provider-vault/pull/570

tyrannosaurus-becks:
@Kuleshov-Arsentii would you mind if I closed this in favor of #570 ? Let me know if that doesn't solve it for you.

Kuleshov-Arsentii:
> @Kuleshov-Arsentii would you mind if I closed this in favor of #570 ? Let me know if that doesn't solve it for you.

Yeah I do not mind.

PR #548, opened 2019-09-24
Add support for configuring the Transit cache

kalafut:


website/docs/r/transit_secret_backend_cache_config.html.md
```diff
@@ -0,0 +1,41 @@
...
+layout: "vault"
+page_title: "Vault: vault_transit_secret_backend_cache_config resource"
+sidebar_current: "docs-vault-resource-transit-secret-backend-cache-config"
+description: |-
+  Configure the cache for the Transit Secret Backend in Vault.
+---
+
+# vault\_transit\_secret\_backend\_cache\_config
+
+Configured the cache for the Transit Secret Backend in Vault.
```
malnick: `Configured` -> `Configure`

website/docs/r/transit_secret_backend_cache_config.html.md
```diff
@@ -0,0 +1,41 @@
...
+  backend = "${vault_mount.transit.path}"
+  size    = 500
+}
+
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `backend` - (Required) The path the transit secret backend is mounted at, with no leading or trailing `/`s.
```
catsby: Should this be named `path` instead? Seems it's expected to receive the `path` attribute of other resources, but we call it `backend`.

catsby: What happens if they do include a trailing or leading `/`? Is it conceivable that whatever it's interpolated would have that, and they wouldn't actually input that as the example shows? If they shouldn't have it, it may be worth adding a state and diff functions to the attribute.

vault/resource_transit_cache_config.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func transitSecretBackendCacheConfigUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	size := d.Get("size").(int)
+
+	backend := d.Get("backend").(string) + "/cache-config"
+
+	log.Printf("[DEBUG] Setting transit cache size to: %d", size)
+	d.SetId(backend)
```
catsby: We probably don't want to write the ID of the backend until it completes successfully, right? I don't see it explicitly stated in [Terraform Plugin Best Practices][1] but I believe you want to call `SetId` as late as possible, when _the thing_ has actually been created. Otherwise there's the risk that the State gets recorded as having a thing, but it doesn't actually exist. 

[1]: https://www.terraform.io/docs/extend/best-practices/index.html

vault/resource_transit_cache_config.go
```diff
@@ -0,0 +1,86 @@
...
+		return fmt.Errorf("error reloading transit plugin: %v", err)
+	}
+
+	return transitSecretBackendCacheConfigRead(d, meta)
+}
+
+func transitSecretBackendCacheConfigRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string) + "/cache-config"
```
catsby: If you do choose to add a state / diff function on `backend`, then remember to use `d.Id()` here instead of asking for the `backend` value again

vault/resource_transit_cache_config.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil
+	}
+
+	d.Set("size", secret.Data["size"])
+
+	return nil
+}
+
+func transitSecretBackendCacheConfigDelete(d *schema.ResourceData, meta interface{}) error {
+	return nil
```
catsby: Can we add a small code comment that explains why this is a no-op? If nothing else, for future readers to understand it's not simply an omission.

website/docs/r/transit_secret_backend_cache_config.html.md
kalafut: fixed

vault/resource_transit_cache_config.go
kalafut: Moved it to the end

vault/resource_transit_cache_config.go
kalafut: Added comment

website/docs/r/transit_secret_backend_cache_config.html.md
kalafut: I agree that `path` is a bit nicer, but`backend` was chosen mainly for consistency with the other transit resource: https://github.com/terraform-providers/terraform-provider-vault/blob/754a58aa908d524e0c91136b7e65d660c1f46107/vault/resource_transit_secret_backend_key.go#L31

Regarding slashes, adding a trailing slash doesn't cause any issues... just a path name diff on the first apply, but that's it. Adding a leading slash is an entirely different path, so it would need to match the path already mounted.

kalafut:
@catsby Added the StateFunc like we discussed.

Issue #428, opened 2019-05-28
[Feature Request] Authenticate Vault provider via approle instead of token


jtcressy:
I'm just shooting this out here to determine the value-add of such a feature.

For the purposes of auditing and easier federation of access, approle is nicer to use than the default token auth backend. It supports the vault identities api, which means roles can be tied to entities which can be members of groups. And, you can get very detailed with the policies attached to the identities structure.

My current issues with using periodic orphan tokens:
- Token lifecycle not strictly managed, is a manual process
- Access control is limited & opaque
- Not namespace-aware (only relevant for us vault enterprise users)
- Audit log always says "token" for the display name, which doesn't help identify the "who did it"

Approle, as we all may know, can solve each of those points:
- Token lifecycle managed by role and role ttl's
- Role Secret ID is generated once and is the only manual process, never expires unless explicitly rotated
- Roles can be tied to entities via entity-alias, thus into groups, and have per-group policies.
- Namespace-aware when group structure permeates into namespaces
- Roles appear in audit log for easier trace of actions performed by terraform, and to differentiate between other user's actions.

Proposal:
The `token` parameter could be made optional and we add an `approle` auth parameter to the provider:
```
provider "vault" {
  address = "https://localhost:8200"
  approle {
    role_id = "${var.vault_role_id}"
    secret_id = "${var.vault_secret_id}"
  }
}
```
On the backend, you'd fork the auth flow from the normal child-token creation method, and use approle to generate the limited child token.
The api client doesn't expose an explicit approle auth method, but this can be done by simply writing to the `auth/approle/login` path and extracting the token out of the response:
```
data := map[string]interface{}{
        "role_id":   roleID,
        "secret_id": secretID,
    }
    resp, err := client.Logical().Write("auth/approle/login", data)
    if err != nil {
        return err
    }
    if resp.Auth == nil {
        return fmt.Errorf("no auth info returned")
    }
```

If this feature gets implemented, it becomes monumentally easier to manage dozens of terraform workspaces without the risk of leaking orphaned tokens anywhere. The audit log will be able to tell me which workspace is accessing any particular path because I can associate specific app roles to each workspace.

Anyway, thanks for hearing me out. Thoughts anyone?

jtcressy:
approle now supported via #552

Issue #346, opened 2019-03-13
vault_approle_auth_backend_login doesn't work, unless you already have a Vault token.


darnold-harness:
### Terraform Version
Terraform v0.11.13
+ provider.google v2.2.0
+ provider.vault v1.6.0

### Affected Resource(s)
- vault_approle_auth_backend_login

### Terraform Configuration Files
```hcl
resource "vault_approle_auth_backend_login" "login" {
  backend   = "approle"
  role_id   = "${var.approle_role_id}"
  secret_id = "${var.approle_secret_id}"
}

provider "vault" {
    alias = "vault_approle"
    token = "${vault_approle_auth_backend_login.login.client_token}"
}
```

### Debug Output
Error: Error refreshing state: 1 error(s) occurred:

* provider.vault: no vault token found

cvbarros:
Hi @darnold-harness, what you're trying to achieve is not possible - terraform has to initialize the provider before it can provision any resources. So you got a chicken and egg situation here.

I suggest you wrap your terraform plan/apply cycle from the context you're trying to login using AppRole by providing the token as a input variable to your configuration.
You can do that by leveraging Vault's CLI and some wrapper script to your terraform commands.

cvbarros:
@tyrannosaurus-becks Recommendation: Close

darnold-harness:
So, what is the purpose of the approle login?

cvbarros:
To grab a token using the `approle` secrets engine. You could then use that token in many other contexts other than Vault - to provision a machine, for example and inject the token to it.
 
But how could you grab a token with this provider if in order to initialize the provider and provision a resource that would require the creation of a child token, from another token? I don't believe this is meant to be used as a "passthrough" to apply Terraform - unless you use different modules, and that would be a similar approach to wrapping plan/apply.
Either way, you'd need a token as a starting point - that is just the way the Terraform and this provider works. Potentially, the provider authentication could be revamped, but it's a totally different matter and would require large scale changes to the codebase.

MichaelSmyth0184:
Yeah I am currently having this problem too, makes the Approle sort of useless then If your having to provider a token to begin with for the Vault provider just so you can retrieve another token? Makes no sense to me tbh.

PR #552, opened 2019-09-27
Add support for alternative auth methods using a method-agnostic implementation

### LINKED ISSUE #351: private terraform enterprise with Provider Vault using Approle instead of Token
### Terraform Version
0.11.10

### Affected Resource(s)
Reading secrets 

provider "vault" "login_vault" {
  address = ""
  token = ""
}

### Expected Behavior
There is no built-in integration between Terraform Enterprise and Vault using any of Approle or other models the only way to login to Vault is using VAULT_TOKEN which has to get issued every time we want to run queue the job or apply it 

How can we use Approle id and Approle Secret id in our Terraform code so these gets saved as Env variables so everytime we want to run the code it doesn't need a temporary token ?
Something like 
provider "vault" "login_vault" {
  address = ""
  approle_id= ""
  secret_id=""
}

We saved our GCP JSON credentials in Vault and in PTFE we want to have a permanent credentials saved in env. variables so we can read the roleset and get the token from Vault for Provider Google .

Any idea or suggestion is highly appreciated 

Thanks

jtcressy:
Hi, I decided to take a crack at #438 and I'm attempting to implement a method-agnostic authentication mechanism.

Additionally, this should still close #351, #428, and #346.

Use of this new config block overrides any token fetched from the token helpers or environment, including if you provide a token via the `token` provider argument.

So far, I've only tested this with approle and userpass. In theory, all auth methods should work as long as they don't require anything complex. An acceptance test was only written for approle, so if we need to add coverage for other login methods anyone may feel free to do so.

My implementation introduces a new config block called `auth_login`:
```hcl-terraform
provider vault {
  address = "http://localhost:8200"
  auth_login {
    path = "auth/approle/login"
    namespace = "my-team/sub-namespace" # only needed if auth method is mounted within another namespace (enterprise only)
    parameters = {
      role_id = var.role_id
      secret_id = var.secret_id
    }
  }
}
```
Some auth methods use parameters as part of the url, and those can just be included in the `path` argument:
```hcl-terraform
provider vault {
  address = "http://localhost:8200"
  auth_login {
    path = "auth/userpass/login/${var.login_username}"
    parameters = {
      password = var.login_password
    }
  }
}
```

Acceptance tests have been ran against a local Vault Enterprise 1.1.5 premium server with `TF_ACC_ENTERPRISE=1` following the regex `TestAcc.*ProviderConfigure`:
```
=== RUN   TestAccAuthLoginProviderConfigure
2019/09/27 15:10:50 [INFO] Using Vault token with the following policies: root
--- PASS: TestAccAuthLoginProviderConfigure (0.73s)
=== RUN   TestAccNamespaceProviderConfigure
--- PASS: TestAccNamespaceProviderConfigure (0.93s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	14.142s
```

Documentation included.
Again, I think I could've done more test coverage but this should at least validate this PR.


vault/provider.go
```diff
@@ -557,6 +581,30 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+		authLoginParameters := authLogin["parameters"].(map[string]interface{})
+
+		secret, err := client.Logical().Write(authLoginPath, authLoginParameters)
+		if err != nil {
+			return nil, err
+		}
+		token = secret.Auth.ClientToken
+	}
+
 	if err != nil {
```
tyrannosaurus-becks: Should this err be checked immediately below `token, err := providerToken(d)` on line 583?

vault/provider.go
```diff
@@ -68,6 +68,30 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_CAPATH", ""),
 				Description: "Path to directory containing CA certificate files to validate the server's certificate.",
 			},
+			"auth_login": {
```
tyrannosaurus-becks: This question might be naive -- I'm wondering why the login support was added here rather than as a data source? I'm not saying it "must" be done that way, I'm just curious for my own understanding.

vault/provider.go
```diff
@@ -68,6 +68,30 @@ func Provider() terraform.ResourceProvider {
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Login to vault with an existing auth method using auth/<mount>/login",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"path": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"namespace": {
```
tyrannosaurus-becks: In the Terraform Vault Provider presently, you can set the namespace _once_ per Terraform run. We've been thinking about making it more like this, where you could designate the namespace on the client per-resource or per-data source. 

I'm just curious if you have a use case where you'd have a different namespace for different login endpoints on the same run of Terraform? If so, would you be willing to share? I've not directly heard any of the use cases yet, though I know they exist, so I'm just wanting to learn so I can understand it more.

vault/provider.go
jtcressy: Doesn't a data source require that the provider be initialized (i.e. configured) first before anything can be done? I'm going off of #346 however I do realize they were trying to use a `resource` instead of `data`.

side note: should I rename this to `token_provider` like y'all were considering in the previous PR? just semantics but doesn't hurt to ask.

vault/provider.go
jtcressy: You're right on that, i'll move this statement up in the file.

vault/provider.go
jtcressy: Currently, the vault provider supports auth'ing to a namespace other than set in the provider if the token's namespace differs. (i.e. a token where s.<token-id>.<namespace-id> is provided)
Our current installation of vault uses Okta auth at the root namespace, but our teams are split into multiple namespaces. I make use of identity-groups and inheritance to translate an identity at the root namespace to policies in the team's namespace. 
Of course, this is only for user access, as an automated pipeline is hardly expected to use okta for logins.
However I expect a pattern to emerge: a team has a suite of sub-namespaces to control access to prod/nonprod secrets but controls programmatic access to these secrets by having a central approle or other auth method in the parent namespace. 
It is expected at this point that a terraform run would auth to 'ns1/' and perform actions in 'ns1/subns1'. 
Therefore, if we configure the provider like so:
```
provider vault {
  address = "something:8200"
  namespace = "ns1/subns1"
  auth_login {
    path = "auth/approle/login"
    namespace = "ns1"
    parameters = {...}
  }
}
```
One would expect it to get a token from the approle mount in 'ns1' and perform actions in 'ns1/subns1' assuming the policies it got from the initial login allow it to do so. 

Even if this pattern does not appear, it at least provides flexibility.

jtcressy: By the way, I already designate namespaces on a per-resource, per-data-source basis by using aliased providers. 
e.g.
```
provider vault {
  alias = "ns1"
  namespace = "ns1"
}
provider vault {
  alias = "ns2"
  namespae = "ns2"
}
```
then:
```
resource vault_generic_secret {
  provider = "vault.ns1"
  ...
}
```
But the auth context can still be different for the sake of controlling the sources of identities.

vault/provider.go
paultyng: A provider is configured prior to any of its data sources or resources being used, but a separate data source can (in some cases) be interpolated in a provider block (ie. you could chain vault providers, but UX wise that is probably undesirable). 

There are unfortunately a lot of nuances and issues with provider schema interpolation outside of just variable usage, so the best bet if you need something in a provider block is to add the reading of it to the provider configuration itself.

vault/provider.go
tyrannosaurus-becks: That is very helpful! Thank you!

StephenWithPH:
@jtcressy ... this is great work! Thank you for contributing.

Do you know how this will interact with `token` (typically set with the `VAULT_TOKEN` environment variable)?

The use case is for human users to set `VAULT_TOKEN`, while build automation provides valid values for `role_id` and `secret_id`.

```hcl
variable role_id {
  type = string
  default = ""
}

variable secret_id {
  type = string
  default = ""
}

provider vault {
  address = <some vault>
  auth_login {
    path = "auth/approle/login"
    parameters = {
      role_id = var.role_id
      secret_id = var.secret_id
    }
  }
}
```

( cc @tyrannosaurus-becks )

jtcressy:
Currently, the order of precedence is such:
- if `auth_login` defined and results in successful login, this token is used.
(if login fails, the provider will panic instead of falling back to another method)
- else, If `token` is defined in the provider config, this is used.
- else, If `VAULT_TOKEN` defined in the environment, this is used.
- else, provider panics with "no vault token found"

note, `auth_login` even overrides the `token` parameter if both are defined in the provider config.

Suggestions:
- You could still use `approle` for human-driven tf plans
- Or, extrapolate the auth method entirely and use variables for `path` and `parameters`. Then, you can still use `approle` in CI and then something like `userpass` or `ldap` on the workstation.

StephenWithPH:
>Or, extrapolate the auth method entirely and use variables for `path` and `parameters`. Then, you can still use `approle` in CI and then something like `userpass` or `ldap` on the workstation.

That was my plan B. It has the (small) downside of breaking our existing solution around `token`. I'll noodle on some other approaches.

PR #553, opened 2019-09-27
Add support for Vault 1.2+ token fields to LDAP auth

kalafut:


vault/resource_ldap_auth_backend.go
```diff
@@ -283,6 +288,10 @@ func ldapAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
+	if err := readTokenFields(d, resp); err != nil {
```
catsby: What is the upgrade story like here? IF a user upgrades versions is there a moderate chance they'll see diffs in the next `terraform plan` due to new information?

kalafut: I just tested the upgrade and it's clean (no diffs). In this case they're all new fields so there isn't even any deprecation to worry about.

PR #558, opened 2019-10-07
provider: Migrate to standalone plugin SDK v1.1.1

radeksimko:
This is a result of `tf-sdk-migrator migrate` and makes this provider not depend on Terraform core anymore. 🎉 

I would recommend cutting a separate release (2.4.0) with bumped version of Terraform before merging and releasing this, just to make it easier to pinpoint potential issues related to either of these two. There should be none, but it's a good practice to keep these in separate releases.


tyrannosaurus-becks:
@radeksimko sounds great, and thank you for this PR. I'm aiming to cut a release in the next week or two. Immediately after that, I'll test this through against master, and as long as there are no surprises, I'll cut a second, separate release for this alone.

tyrannosaurus-becks:
@radeksimko alright! I just cut a release and I'm ready to get started on doing this one on its own. A couple of questions:

- Do you recommend a patch, minor, or major release? I'm thinking minor but it's unclear to me so I'd love to hear your feedback.
- I merged in master and pushed it to see how the tests do against current master. I am getting one new failure, [here's the link](https://github.com/terraform-providers/terraform-provider-vault/compare/sdk-migration-tests).

radeksimko:
> Do you recommend a patch, minor, or major release? I'm thinking minor but it's unclear to me so I'd love to hear your feedback.

I think the convention so far was for providers to cut minor release.

> I merged in master and pushed it to see how the tests do against current master. I am getting one new failure, here's the link.

I'm not sure if you meant to link to a branch, but I see a branch with a merge commit. I just did a fresh rebase, so my branch in this PR should be up to date. Can you try running acceptance tests from that branch and name the failing test? I can try to look into it.

tyrannosaurus-becks:
Thanks @radeksimko . On the build currently, I'm seeing:
```
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.013s
# github.com/terraform-providers/terraform-provider-vault/vault [github.com/terraform-providers/terraform-provider-vault/vault.test]
vault/provider.go:304:4: cannot use consulSecretBackendRoleResource() (type *"github.com/hashicorp/terraform/helper/schema".Resource) as type *"github.com/hashicorp/terraform-plugin-sdk/helper/schema".Resource in field value
vault/resource_consul_secret_backend_role_test.go:18:3: cannot use testProviders (type map[string]"github.com/hashicorp/terraform-plugin-sdk/terraform".ResourceProvider) as type map[string]"github.com/hashicorp/terraform/terraform".ResourceProvider in field value
FAIL	github.com/terraform-providers/terraform-provider-vault/vault [build failed]
```

I'm going to hit the re-run button, but I suspect it will give the same failure.

radeksimko:
@tyrannosaurus-becks That looks like a vendoring issue. Did you remember to `go mod vendor`? Is there still `github.com/hashicorp/terraform` in the `vendor` folder, is it still in `go.mod`? It shouldn't be.

radeksimko:
Oh, that is from my branch actually - sorry for the momentary confusion - I will look into it!

radeksimko:
@tyrannosaurus-becks I believe this is now ready - I just forgot to rewrite import paths for the new files which were brought through the rebase from master, which then brought back `hashicorp/terraform` and hence these odd errors. 🙈 

PTAL

PR #559, opened 2019-10-11
Fix panic when importing from cassandra db

jackkleeman:
protocol_version is treated as an int by Terraform, but is returned by
Vault as a string. As such, the code currently panics when importing it.
We should convert the string to an int64.

### DIFF
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tyrannosaurus-becks:
Hi @jackkleeman ! Thanks for this PR.

It's strange that it's coming back as a string for you. When I look at the Vault API docs, [it's described as an int](https://www.vaultproject.io/api/secret/databases/cassandra.html#protocol_version) and also when I further look at the code, I see it's [returned here](https://github.com/hashicorp/vault/blob/master/builtin/logical/cassandra/path_config_connection.go#L110) and that the protocol version is [an int here](https://github.com/hashicorp/vault/blob/master/builtin/logical/cassandra/backend.go#L75).

Could you give further steps to reproduce this, and include what Vault and Terraform versions you're on? 

Thank you!

jackkleeman:
Sorry for slow response! It seems to come out as a string for me. I wonder if I wrote this variable to Vault wrong somehow. I never really get how the Vault CLI knows its a string or an int

```
$ vault version
Vault v1.1.2 ('0082501623c0b704b87b1fbc84c2d725994bac54')

$ vault status
Key                    Value
---                    -----
Version                1.2.3

$ vault read database/config/cassandra-premium1 -format=json
{
  "request_id": "redacted",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "allowed_roles": [
      "*"
    ],
    "connection_details": {
      "connect_timeout": "2s",
      "consistency": "local_quorum",
      "hosts": "redacted",
      "protocol_version": "4",
      "skip_verification": "true",
      "username": "redacted"
    },
    "plugin_name": "cassandra-database-plugin",
    "root_credentials_rotate_statements": []
  },
  "warnings": null
}
```

tyrannosaurus-becks:
Interesting! Thanks for providing that! I was able to reproduce that on our current master branch. Since that is the current state, I do think this is the best way to go. Thanks again!

Issue #550, opened 2019-09-26
[Feature Request] Skip creating child token if token used is itself an orphan


mjtice:
Right now the vault provider creates a child token with a reduced TTL and uses that token to access vault.(1)  The problem with this is that it requires the end user to use a non-orphaned token in the `token` argument.  This is because by their nature, non-token based authentication methods don't have a parent token(2).

I propose that terraform-provider-vault skips the child token creation if the token used to authenticate is an orphan token.

What do you think (I'm looking at you @tyrannosaurus-becks! :) )?  

1: https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L597
2: https://groups.google.com/d/msg/vault-tool/7Mq0t53mszA/VBCJGz1uAAAJ

tstoermer:
Actually, I like the usage of a child token in the terraform run making sure everything is cleaned up and has a scoped lifetime.

In addition I would like to enable the child token with referencing information for auditing, see https://github.com/terraform-providers/terraform-provider-vault/pull/594

Maybe the creation of a child token should be controlled by a provider argument?

tyrannosaurus-becks:
@mjtice I think that @tstoermer makes some good points. Also, there's a code note [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L639-L650) regarding why that approach was taken.

What auth method are you using to obtain the token you'd like to use for running Vault? Also, can you provide more information about your use case that leads to your request for this change?

Thank you! For now I'm going to mark the associated PR as one not to be merged until we've reached a conclusion here.

mjtice:
Hi @tyrannosaurus-becks 
Sure, I understand why the child token approach was taken.   Here's my use case:

1. We use Stackstorm ([and the associated terraform pack](https://github.com/StackStorm-Exchange/stackstorm-terraform) automate the creation and destruction of resources.
2. We've configured a local vault agent on this host and the agent is authenticated via approle method.
3. Using the token from the agent, I stuff it into my workflow (and therefor into my terraform.tfvars file) where it's consumed by the vault provider.
4. Because this token is an orphan and unable to create its own child token the terraform operations fail.

I'm not sure of the points @tstoermer is making - it sounds more like a feature request than an argument against what I'm proposing?  Maybe we take a route similar to what @tstoermer suggests (making the child token an argument) - but the default is to continue with what is currently done and an the function to skip creating a child is explicitly requested?

[...]

PR #561, opened 2019-10-15
Skip creating a child token if the token used is in fact an orphan

mjtice:
This is related to:
[Feature Request] Skip creating child token if token used is itself an orphan #550 

All we're doing here is just skipping creating a child token if the token used is an orphan.

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -652,24 +652,26 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		}
 	}
 
-	renewable := false
-	childTokenLease, err := client.Auth().Token().Create(&api.TokenCreateRequest{
-		DisplayName:    "terraform",
-		TTL:            fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		ExplicitMaxTTL: fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		Renewable:      &renewable,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create limited child token: %s", err)
+	if tokenInfo.Data["orphan"] == false {
+		renewable := false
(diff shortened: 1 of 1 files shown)


vault/provider.go
```diff
@@ -652,24 +652,26 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
-	childTokenLease, err := client.Auth().Token().Create(&api.TokenCreateRequest{
-		DisplayName:    "terraform",
-		TTL:            fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		ExplicitMaxTTL: fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		Renewable:      &renewable,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create limited child token: %s", err)
-	}
+	if tokenInfo.Data["orphan"] == false {
```
tyrannosaurus-becks: In looking at Vault's present code, I see that this field is _always_ added and accurate. However, if someone were to edit Vault's code to sometimes omit the field, or if somebody changed which method we used to look up a token above so the response now omitted it, this could always evaluate to `false`. 

Should we instead change this to:
```
isOrphan, found := tokenInfo.Data["orphan"]
if !found {
    // return an error
}
if !isOrphan {
    // the rest 
}
```
It might provide some future-proofing just in case. :-)

tyrannosaurus-becks:
Just circling back on this. Am I right in thinking that the discussion on the linked ticket means that there's a workaround and we don't have a necessity to add this code? If we can avoid it, I would prefer so, because it decreases the security stance of the provider. Let me know what you think. Thank you!

mjtice:
Well, now I think I need to do some more testing on my end.  Let's just close this out and if need be I'll re-create the PR.

PR #565, opened 2019-10-16
Add information about Vault 1.2 for `token_*` fields

lawliet89:
https://github.com/terraform-providers/terraform-provider-vault/pull/487
deprecated some fields that were deprecated in Vault 1.2

Add more context to deprecation messages that the new `token_*` fields
are only for Vault >= 1.2


Techcadia:
Would love to see this merged. Would have save us some time when upgrading.

website/docs/d/kubernetes_auth_backend_role.md
```diff
@@ -72,6 +72,12 @@ These attributes are common across several Authentication Token resources since
 * `token_max_ttl` - The maximum lifetime for generated tokens in number of seconds.
   Its current value will be referenced at renewal time.
 
+* `token_period` - (Optional) If set, indicates that the
+  token generated using this role should never expire. The token should be renewed within the
+  duration specified by this value. At each renewal, the token's TTL will be set to the
+  value of this field. The maximum allowed lifetime of token issued using this
```
kalafut: I think the `The maximum allowed lifetime of token issued using this role` is a copy/paste error?  (Here and in the same block elsewhere)

lawliet89: @kalafut  I've fixed the text.

PR #566, opened 2019-10-16
Add new JWT Auth Role Fields introduced with Vault 1.2

lawliet89:
Fixes https://github.com/terraform-providers/terraform-provider-vault/issues/544

I have a couple of options for implementation:

- Currently, I set the default values for the new fields as indicated by Vault API docs. Since these fields were newly introduced with Vault 1.2, I am not sure if they will cause issues with Vault < 1.2
- If I don't set the default values, then to "unset" the fields after having set them, the user will have to change the fields to `-1` or `false` respectively. Simply "removing" the fields will cause an infinite diff in plans and Vault does not actually disable them.

Which option should I go for? The former is "cleaner" but might have compatibility issues. The latter is "yucky". @tyrannosaurus-becks @kalafut


kalafut:
@lawliet89 Thanks! The "clean" option will likely be fine but let me double check.

kalafut:
@lawliet89 Sending the values to pre-1.2 Vault is fine as they'll be ignored. For the default clock skews, wouldn't you want 0 instead of -1?  -1 is explicitly disabling any clock skew.

lawliet89:
> For the default clock skews, wouldn't you want 0 instead of -1? -1 is explicitly disabling any clock skew.

lawliet89:
> For the default clock skews, wouldn't you want 0 instead of -1? -1 is explicitly disabling any clock skew.

I was under the impression that the default is -1 because the docs weren't very clear.

I have changed the defaults to `0`.

kalafut:
> I was under the impression that the default is `-1` because the docs weren't very clear.

Yeah that was one of the ickier settings we had to add (well past initial release). There was already a reasonable default skew built in that we needed to retain if the user did nothing or had upgraded, but then others wanted a custom skew or none at all. 😕

kalafut:
Thanks!

PR #569, opened 2019-10-18
Fix build on Go 1.13 

lawliet89:
- Removed unnecessary `replace`
- Should still build on Go 1.12
- Ran `go mod tidy`


lawliet89:
Hi @tyrannosaurus-becks, thanks for your response.

I haven't actually changed the build to use 1.13. `go.mod` still says `1.12` and the CircleCI tests are still run on `1.12`. This PR just tidies up some dependencies so that it will build on both 1.12 and 1.13.

tyrannosaurus-becks:
Hi @lawliet89! Yes, I saw that you left the build on 1.12, which I thought was great because it definitely demonstrates that this is backwards compatible. :-) Thank you!

We have run into situations on Vault where some developers were on X version of Go, and others were on Y version of Go, and whenever each committed, developers on X would switch dependencies back to X version of Go, and then developers on Y version would revert those changes. They were essentially "flapping" dependencies back and forth, if you will, and so they ended up spending some time debugging why simple PRs had caused so many dependency changes, only to realize it was a Go versioning issue. If we stay in step with Vault, people are less likely to encounter this and will only need to maintain one version of Go locally to contribute to them.

That's mainly what we're trying to avoid by waiting on this one. It's quite clear it's a good PR and backwards-compatible. We appreciate you tackling it.

lawliet89:
Thanks for the explanation @tyrannosaurus-becks 

I am going to close this for now because it's going to be tedious to keep resolving conflicts. It's relatively simple to do, anyway, when Vault switches to Go 1.13.

PR #570, opened 2019-10-21
Change vault_namespace ID to path

petems:
* The ID isn’t super useful, it’s only used by Vault internally
* We should use the path as it allows imports of existing namespaces
* Move the namespace ID to its own computed attribute `namespace_id`

### DIFF
--- a/vault/resource_namespace.go
+++ b/vault/resource_namespace.go
@@ -3,6 +3,7 @@ package vault
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/vault/api"
@@ -24,6 +25,12 @@ func namespaceResource() *schema.Resource {
 				Required:     true,
 				Description:  "Path of the namespace.",
 				ValidateFunc: validateNoTrailingSlash,
+			},
+
+			"namespace_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "ID of the namepsace.",
 			},
 		},
 	}
@@ -63,7 +70,9 @@ func namespaceDelete(d *schema.ResourceData, meta interface{}) error {
 func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
-	path := d.Get("path").(string)
+	upgradeNonPathdNamespaceID(d)
+
+	path := d.Id()
 
 	resp, err := client.Logical().Read("sys/namespaces/" + path)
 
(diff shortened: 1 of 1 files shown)


petems:
After:

```
$ vault namespace create import_tester
Key     Value
---     -----
id      ghA9v
path    import_tester/
$ terraform import vault_namespace.import_tester import_tester
vault_namespace.import_tester: Importing from ID "import_tester"...
vault_namespace.import_tester: Import prepared!
  Prepared vault_namespace for import
vault_namespace.import_tester: Refreshing state... [id=import_tester]

Import successful!

$ cat terraform.tfstate | jq .'resources'
[
  {
    "mode": "managed",
    "type": "vault_namespace",
    "name": "import_tester",
    "provider": "provider.vault",
    "instances": [
      {
        "schema_version": 0,
        "attributes": {
          "id": "import_tester",
          "namespace_id": "ghA9v",
          "path": "import_tester"
        },
        "private": "eyJzY2hlbWFfdmVyc2lvbiI6IjAifQ=="
      }
    ]
  }
]
```

petems:
Before:

```
$ terraform import vault_namespace.import_tester import_tester
vault_namespace.import_tester: Importing from ID "import_tester"...
vault_namespace.import_tester: Import prepared!
  Prepared vault_namespace for import
vault_namespace.import_tester: Refreshing state... [id=import_tester]

Error: error reading from Vault: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/namespaces
Code: 405. Errors:

* 1 error occurred:
	* unsupported operation

```

vault/resource_namespace.go
```diff
@@ -77,7 +83,9 @@ func namespaceRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
-	d.SetId(resp.Data["id"].(string))
+	d.SetId(resp.Data["path"].(string))
```
tyrannosaurus-becks: Question! If people are already using this, and then they upgrade the Terraform Vault Provider, and Terraform detects that it needs to update the ID, will that result in Terraform outputting that it's detected that a namespace needs to be modified? Or I suppose in a broader sense, what will the migration path look like?

Just asking about backwards compat so I can factor that into whether this belongs in a major, minor, or patch release. Thank you!

petems:
@tyrannosaurus-becks Happy to do so, but I'll need a hand helping with that. 

Are there examples of doing this in TF providers? Have you got time for a quick pairing session on it as well? 😄

tyrannosaurus-becks:
@petems actually, I have a sample! We went through the same discussion on this PR, here's a direct link to the convo: https://github.com/terraform-providers/terraform-provider-vault/pull/581#discussion_r340846068. You should be able to check out the subsequent commits to see how it was implemented.

petems:
@tyrannosaurus-becks Perfect! I'm not sure how to add a test for this new scenario... is there a way to create a special resource to run tests on in the old state?

petems:
I can see it working in the logs with a manual test of manually editing the statefile:

```
2019-11-07T02:06:18.453Z [DEBUG] plugin.terraform-provider-vault: 2019/11/07 02:06:18 [INFO] Using Vault token with the following policies: root
2019-11-07T02:06:18.455Z [DEBUG] plugin.terraform-provider-vault: 2019/11/07 02:06:18 [DEBUG] Upgrading old ID to path - abcd123 to import_tester
2019-11-07T02:06:18.455Z [DEBUG] plugin.terraform-provider-vault: 2019/11/07 02:06:18 [DEBUG] Setting namespace_id to old ID - abcd123
2019-11-07T02:06:18.455Z [DEBUG] plugin.terraform-provider-vault: 2019/11/07 02:06:18 [DEBUG] Vault API Request Details:
```

PR #571, opened 2019-10-21
Wait longer for AWS creds to propagate

tyrannosaurus-becks:
This PR adds further delay for AWS creds to propagate. Instead of waiting for 3 total tests of the credentials for them to be considered valid, waits for 5 sequential tries + 5 seconds. 

Since sts credentials are immediately consistent, doesn't wait to test them.

### DIFF
--- a/vault/data_source_aws_access_credentials.go
+++ b/vault/data_source_aws_access_credentials.go
@@ -11,11 +11,29 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/sts"
-	cleanhttp "github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-
 	"github.com/hashicorp/vault/api"
+)
+
+const (
+	// sequentialSuccessesRequired is the number of times the test of an eventually consistent
+	// credential must succeed before we return it for use.
+	sequentialSuccessesRequired = 5
+
+	// sequentialSuccessTimeLimit is how long we'll wait for eventually consistent AWS creds
+	// to propagate before giving up. In real life, we've seen it take up to 15 seconds, so
+	// this is ample and if it's unsuccessful there's something else wrong.
+	sequentialSuccessTimeLimit = time.Minute
+
+	// retryTimeOut is how long we'll wait before timing out when we're retrying credentials.
+	// This corresponds to Vault's default 30-second request timeout.
+	retryTimeOut = 30 * time.Second
+
+	// propagationBuffer is the added buffer of time we'll wait after N sequential successes
+	// before returning credentials for use.
+	propagationBuffer = 5 * time.Second
 )
 
 func awsAccessCredentialsDataSource() *schema.Resource {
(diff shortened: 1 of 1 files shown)


vault/data_source_aws_access_credentials.go
```diff
@@ -118,13 +118,27 @@ func awsAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}
...
 
 	d.SetId(secret.LeaseID)
-	d.Set("access_key", secret.Data["access_key"])
-	d.Set("secret_key", secret.Data["secret_key"])
-	d.Set("security_token", secret.Data["security_token"])
-	d.Set("lease_id", secret.LeaseID)
-	d.Set("lease_duration", secret.LeaseDuration)
-	d.Set("lease_start_time", time.Now().Format(time.RFC3339))
-	d.Set("lease_renewable", secret.Renewable)
+	if err := d.Set("access_key", secret.Data["access_key"]); err != nil {
```
kalafut: While I normally would 👍 the error handing, I've seen error checking on `Set()` only occasionally, and even the "how to extend TF" docs omit it. I've wondered about what the correct pattern is. Thoughts on this @catsby ?

vault/data_source_aws_access_credentials.go
```diff
@@ -138,35 +152,45 @@ func awsAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}
...
-				} else if err != nil {
-					return resource.NonRetryableError(err)
-				}
-				log.Printf("[DEBUG] Checked if AWS sts token %q is valid", secret.LeaseID)
-			}
-			return nil
-		})
-		if err != nil {
+	if credType == "sts" {
+		// STS credentials are immediately consistent. Let's ensure they're working.
```
kalafut: Consider adding a link to AWS doc re: the different consistency expectations since we're now splitting them out.

vault/data_source_aws_access_credentials.go
```diff
@@ -138,35 +152,45 @@ func awsAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}
...
+			return resource.RetryableError(err)
+		} else if err != nil {
+			log.Printf("[DEBUG] Error checking if creds %q are valid: %s", secret.LeaseID, err)
+			return resource.NonRetryableError(err)
+		}
+		sequentialSuccesses++
+		log.Printf("[DEBUG] Checked if AWS creds %q are valid", secret.LeaseID)
+		return nil
+	}
+	for sequentialSuccesses < 5 {
```
kalafut: If for some reason we're getting a mix of results but never reaching 5 successes, we could be stuck here indefinitely. We probably want something like a max time condition, after which we'd return an error.

vault/data_source_aws_access_credentials.go
tyrannosaurus-becks: Ah! Thank you! I'll update that.

vault/data_source_aws_access_credentials.go
tyrannosaurus-becks: Ah, good idea. Will do.

vault/data_source_aws_access_credentials.go
```diff
@@ -138,35 +152,45 @@ func awsAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}
...
-			return nil
-		})
-		if err != nil {
+	if credType == "sts" {
+		// STS credentials are immediately consistent. Let's ensure they're working.
+		log.Printf("[DEBUG] Checking if AWS sts token %q is valid", secret.LeaseID)
+		if _, err := stsconn.GetCallerIdentity(&sts.GetCallerIdentityInput{}); err != nil {
+			return err
+		}
+		log.Printf("[DEBUG] Checked if AWS sts token %q is valid", secret.LeaseID)
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Hrm, yes, I think that's true. It was more useful when it was in the retry loop, but now that it's been moved out it's not necessary. I'll strip it.

vault/data_source_aws_access_credentials.go
tyrannosaurus-becks: Added a time limit of a minute.

vault/data_source_aws_access_credentials.go
```diff
@@ -138,35 +156,51 @@ func awsAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}
...
+			return resource.NonRetryableError(err)
+		}
+		sequentialSuccesses++
+		log.Printf("[DEBUG] Checked if AWS creds %q are valid", secret.LeaseID)
+		return nil
+	}
+
+	start := time.Now()
+	for sequentialSuccesses < sequentialSuccessesRequired {
+		if time.Now().Sub(start) > sequentialSuccessTimeLimit {
```
kalafut: Simpler:
```suggestion
		if time.Since(start) > sequentialSuccessTimeLimit {
```

PR #573, opened 2019-10-23
Fix panic when trying to write an entity alias that already exists

### LINKED ISSUE #523: Adding new vault_identity_entity_alias causes runtime panic
```
panic: runtime error: invalid memory address or nil pointer dereference
2019-08-29T15:14:42.161+0200 [DEBUG] plugin.terraform-provider-vault_v2.1.0_x4: [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x1b9db64]
2019-08-29T15:14:42.161+0200 [DEBUG] plugin.terraform-provider-vault_v2.1.0_x4:
2019-08-29T15:14:42.161+0200 [DEBUG] plugin.terraform-provider-vault_v2.1.0_x4: goroutine 45 [running]:
2019-08-29T15:14:42.161+0200 [DEBUG] plugin.terraform-provider-vault_v2.1.0_x4: github.com/terraform-providers/terraform-provider-vault/vault.identityEntityAliasCreate(0xc00015b5e0, 0x1eb1bc0, 0xc00015a700, 0x2, 0x2c5fb20)
```

### Terraform Version
I have tried all possible combinations between `Terraform v0.12.3`, `Terraform v0.12.7`, `provider.vault v2.1.0` and `provider.vault v2.2.0`

### Affected Resource(s)
- vault_identity_entity_alias

Data has been slightly anonymized to get rid of server and application names.
We manage around 45-50 applications in terraform:
<details><summary>Setup</summary>

We are using this module (`maint.tf`)

```hcl
resource "vault_identity_entity" "application_identity" {
    name      = "${var.application_name}"
    policies  = ["application"]
}

resource "vault_identity_entity_alias" "aliases" {
    count           = length(var.servers)
    name            = "${var.deployment_type}-${var.servers[count.index]}"
    mount_accessor  = "${var.vault_accessor_id}"
    canonical_id    = "${vault_identity_entity.application_identity.id}"
}

resource "vault_mount" "transit" {
    count = var.has_transit_engine ? 1 : 0
    path = "company/transit/apps/${var.application_name}"
    type = "transit"
}
```
To define identities for applications with aliases to a certificate authority which looks like this: 
```hcl
resource "vault_auth_backend" "applications_certificate_auth" {
    path = "company/cert/applications"
    type = "cert"
}
resource "vault_cert_auth_backend_role" "applications_certificate_auth_cert" {
    name          = "CAName"
    certificate   = "${file("./path/to/cert.pem")}"
    backend       = "${vault_auth_backend.applications_certificate_auth.path}"
    policies      = []
}
```
</details>

I was trying to add a new server (`server13`) to one application. No other changes. This was the diff on our config:
old: 
```hcl
module "application_s" {
    source = "./modules/application"
    application_name = "application_s"
    deployment_type = "application-staging"
    servers = ["server1", "server2", "server3", "server4", "server5", "server6", "server8", "server9", "server10", "server11", "server12"]
    vault_accessor_id = "${module.setup.applications_certificate_accessor}"
    has_transit_engine = true
}
```
new:
```hcl
module "application_s" {
    source = "./modules/application"
    application_name = "application_s"
    deployment_type = "application-staging"
    servers = ["server1", "server2", "server3", "server4", "server5", "server6", "server8", "server9", "server10", "server11", "server12", "server13"]
    vault_accessor_id = "${module.setup.applications_certificate_accessor}"
    has_transit_engine = true
}
```
### Panic Output

I also removed application names and servernames from this gist, it is also not the full debug log as that was a huge log and purging it of information was a lot of effort. If you need the entire log, or a bigger part of the log let me know and i'll take the time to generate it.
https://gist.github.com/JeroenBrinkman/0efb73725a1774af38e5bee5eafbb2ed

### Expected Behavior
I expected the planned alias to be added to vault

### Actual Behavior
Vault crashes with a runtime panic

### Steps to Reproduce
1. `terraform apply` old config to empty vault and no terraform state (dry run)
2. `terraform apply` extra server
3. planned changes are: 
```
  # module.application_s.vault_identity_entity_alias.aliases[11] will be created
  + resource "vault_identity_entity_alias" "aliases" {
      + canonical_id   = "d957a64e-1980-f502-9bdc-e5bc88b8329e"
      + id             = (known after apply)
      + mount_accessor = "auth_cert_b4e16cb9"
      + name           = "application-staging-server13"
    }
```
4. confirm with `yes`
5. crash
6. No new alias and a broken state when combined with other changes that do apply succesfully

### Important Factoids
- Vault version `1.2.2` running in HA mode
- A (assumed) fault in our config causes terraform to recreate the kv2 store every run
- The issue is 100% reproducible on this vault instance however trying the same kind of change on a different instance with about the same number of apps with different names worked perfectly fine. The other instance has less servers however, so in total will contain less aliases.

### References
Not that i know about

kalafut:
Fixes #523

### DIFF
--- a/vault/resource_identity_entity_alias.go
+++ b/vault/resource_identity_entity_alias.go
@@ -63,9 +63,22 @@ func identityEntityAliasCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
 	}
-	log.Printf("[DEBUG] Wrote IdentityEntityAlias %q", name)
 
-	d.SetId(resp.Data["id"].(string))
+	if resp != nil {
+		d.SetId(resp.Data["id"].(string))
+	} else {
+		// A nil response is possible if the alias already exists and TF doesn't know about it.
+		// But we still need and don't have the alias ID, so a search is used to find it.
+		log.Printf("[DEBUG] Entity alias ID is unknown. Reading aliases for entity %q", canonicalID)
+
+		id, err := findAliasID(client, canonicalID, name, mountAccessor)
+		if err != nil {
+			return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
+		}
+		d.SetId(id)
+	}
+
+	log.Printf("[DEBUG] Wrote IdentityEntityAlias %q", name)
 
 	return identityEntityAliasRead(d, meta)
 }
@@ -177,4 +190,23 @@ func identityEntityAliasNamePath(name string) string {
 
 func identityEntityAliasIDPath(id string) string {
 	return fmt.Sprintf("%s/id/%s", identityEntityAliasPath, id)
+}
+
+func findAliasID(client *api.Client, canonicalID, name, mountAccessor string) (string, error) {
+	path := identityEntityIDPath(canonicalID)
+
+	resp, err := client.Logical().Read(path)
+	if err != nil {
+		return "", fmt.Errorf("error reading entity aliases: %s", err)
+	}
+
+	aliases := resp.Data["aliases"].([]interface{})
+	for _, aliasRaw := range aliases {
+		alias := aliasRaw.(map[string]interface{})
+		if alias["name"] == name && alias["mount_accessor"] == mountAccessor {
+			return alias["id"].(string), nil
+		}
+	}
+
+	return "", fmt.Errorf("unable to determine alias ID. canonical ID: %q  name: %q  mountAccessor: %q", canonicalID, name, mountAccessor)
 }
--- a/vault/resource_identity_entity_alias.go
+++ b/vault/resource_identity_entity_alias.go
@@ -64,21 +64,19 @@ func identityEntityAliasCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
 	}
 
-	if resp != nil {
-		d.SetId(resp.Data["id"].(string))
-	} else {
-		// A nil response is possible if the alias already exists and TF doesn't know about it.
-		// But we still need and don't have the alias ID, so a search is used to find it.
-		log.Printf("[DEBUG] Entity alias ID is unknown. Reading aliases for entity %q", canonicalID)
+	if resp == nil {
+		aliasIDMsg := "Unable to determine alias id."
 
-		id, err := findAliasID(client, canonicalID, name, mountAccessor)
-		if err != nil {
-			return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
+		if aliasID, err := findAliasID(client, canonicalID, name, mountAccessor); err == nil {
+			aliasIDMsg = fmt.Sprintf("Alias resource ID %q may be imported.", aliasID)
 		}
-		d.SetId(id)
+
+		return fmt.Errorf("IdentityEntityAlias %q already exists. %s", name, aliasIDMsg)
 	}
 
 	log.Printf("[DEBUG] Wrote IdentityEntityAlias %q", name)
+
+	d.SetId(resp.Data["id"].(string))
 
 	return identityEntityAliasRead(d, meta)
 }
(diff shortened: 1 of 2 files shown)


vault/resource_identity_entity_alias.go
```diff
@@ -63,9 +63,22 @@ func identityEntityAliasCreate(d *schema.ResourceData, meta interface{}) error {
...
+		// A nil response is possible if the alias already exists and TF doesn't know about it.
+		// But we still need and don't have the alias ID, so a search is used to find it.
+		log.Printf("[DEBUG] Entity alias ID is unknown. Reading aliases for entity %q", canonicalID)
+
+		id, err := findAliasID(client, canonicalID, name, mountAccessor)
+		if err != nil {
+			return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
+		}
+		d.SetId(id)
+	}
```
catsby: This part is a bit magical from the end user perspective. From a Vault perspective this sounds great, it's fault tolerant. From a Terraform perspective, the user _thinks_ they just created the thing, but it actually already exists. 

Assuming the response is `nil` _because_ the alias already exists, would it be possible to check the response status code and make a decision? If it's `204` (I assume this means the alias exists) maybe we should error here, and tell the user this thing already exists. Terraform does have an `import` command, it may be more Terraform-like to error and have them import, than to automagically assume ownership of this thing. What do you think?

kalafut: That is a fair critique and it probably comes down to UX and Terraform norms, which you're more familiar with. We can definitely provide an error message instead if that would be more inline with typically behavior. I've not looked into the `import` support for this resource, though I suspect it would need to be beefed up with much of the same code code (i.e. finding an alias by traits).

catsby: Could we add the `findAliasID` call to the `READ` method?

```go
func identityEntityAliasRead(d *schema.ResourceData, meta interface{}) error {
  client := meta.(*api.Client)
  id := d.Id()
  if id == "" {
    id,err := findAliasID()
    // err handle
    if id != nil {
      d.SetId(id)
    }
  }
  // other stuff
}
```

That would cover the import functionality, so we could just use `schema.ImportStatePassthrough`:

- https://www.terraform.io/docs/extend/resources/import.html#importer-state-function

The only real difference between these approaches is that one is in `CREATE` and the other is `READ`, and the consequences of "magic" behind the scenes are different to end users. 

If you add configuration for an alias that happens to exist already, then:

`findAliasID` in create:

- Users run `plan`, Terraform refreshes state and plans to create the alias (it wont)
- Users run `apply`, Terraform claims it created the alias (it didn't)


`findAliasID` in read:

- Users run `plan`, Terraform refreshes state and `plan` says there's nothing to do (it added the ID locally to state though)
- Users run `apply`, Terraform doesn't do anything with this resource

kalafut: I'm not sure how the magic in the `import` case would actually work. Import wants an ID to fetch, but we don't have an ID. It complains if you don't provide one, and other ways to fake that check and allow the findAliasID to run get hacky.

I'm thinking we can keep this really simple and if you want to import the alias, provide the ID like normal. You can find the alias ID via the CLI or UI. Then just:

```
terraform import vault_identity_entity_alias.test_alias <alias_id>
```

That seems to work fine, has no magic, and would mean the only change would be to detect that the ID already exists (the `nil` response on Create) and return an error message there.

kalafut: I just pushed an update that retains the alias ID search (which would be a tedious step for the operator trying to locate the ID in order to import or otherwise deal with it), but it only uses the result to provide a more helpful error message.  So, retain the utility, but remove the magic.

vault/resource_identity_entity_alias.go
```diff
@@ -63,6 +63,17 @@ func identityEntityAliasCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("error writing IdentityEntityAlias to %q: %s", name, err)
 	}
+
+	if resp == nil {
+		aliasIDMsg := "Unable to determine alias id."
+
+		if aliasID, err := findAliasID(client, canonicalID, name, mountAccessor); err == nil {
+			aliasIDMsg = fmt.Sprintf("Alias resource ID %q may be imported.", aliasID)
```
tyrannosaurus-becks: Love this.

vault/resource_identity_entity_alias.go
```diff
@@ -63,6 +63,17 @@ func identityEntityAliasCreate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+
+		return fmt.Errorf("IdentityEntityAlias %q already exists. %s", name, aliasIDMsg)
```
tyrannosaurus-becks: This is interesting, I previously thought all errors in Go should be lowercase, but I see [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) that if they're full strings (like these are), it's okay sometimes to capitalize them.

vault/resource_identity_entity_alias.go
```diff
@@ -178,3 +189,24 @@ func identityEntityAliasNamePath(name string) string {
...
+
+func findAliasID(client *api.Client, canonicalID, name, mountAccessor string) (string, error) {
+	path := identityEntityIDPath(canonicalID)
+
+	resp, err := client.Logical().Read(path)
+	if err != nil {
+		return "", fmt.Errorf("error reading entity aliases: %s", err)
+	}
+
+	if resp != nil {
```
tyrannosaurus-becks: Nit: we could avoid nesting below if we just returned above if response == nil.

kalafut: Yeah, I didn't love it. Chose this route to only have the one instance of the long-ish error message.

PR #577, opened 2019-10-25
Added vault_database_secret_backend_static_role resource

martinssipenko:
### Community Note

### Description

This change provides new resource to manage Vaults database secret engine [static roles](https://www.vaultproject.io/docs/secrets/databases/index.html#static-roles).

https://www.vaultproject.io/api/secret/databases/index.html#create-static-role

```release-note
Adds a resource for the "/database/static-roles/{name}" endpoint (#577).
```

```
$ make testacc TESTARGS='-run=TestAccDatabaseSecretBackendStaticRole'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDatabaseSecretBackendStaticRole -timeout 120m
?     github.com/terraform-providers/terraform-provider-vault [no test files]
?     github.com/terraform-providers/terraform-provider-vault/cmd/coverage  [no test files]
PASS
ok    github.com/terraform-providers/terraform-provider-vault/util  (cached) [no tests to run]
=== RUN   TestAccDatabaseSecretBackendStaticRole_import
--- PASS: TestAccDatabaseSecretBackendStaticRole_import (0.22s)
=== RUN   TestAccDatabaseSecretBackendStaticRole_basic
--- PASS: TestAccDatabaseSecretBackendStaticRole_basic (0.30s)
PASS
ok    github.com/terraform-providers/terraform-provider-vault/vault 0.544s
```


vault/resource_database_secret_backend_static_role.go
```diff
@@ -0,0 +1,210 @@
...
+			},
+			"db_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Database connection to use for this role.",
+			},
+			"rotation_statements": {
+				Type:        schema.TypeList,
+				Required:    true,
```
tyrannosaurus-becks: I noticed on the docs that this isn't required on Vault's end. Was there a reason for requiring it here?

vault/resource_database_secret_backend_static_role.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Unique name for the static role.",
+			},
+			"backend": {
```
tyrannosaurus-becks: I've noticed that some folks name this "path", make it optional, and fall back to the secret engine name if it's not included (in this case, it would be "database" which is pretty generic.) I'm not dedicated to naming or handling it any one way. I was just curious, do you think we should do that here, or does that not make as much sense in this case?

vault/resource_database_secret_backend_static_role_test.go
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
+
+resource "vault_database_secret_backend_static_role" "test" {
+  backend = "${vault_mount.db.path}"
+  db_name = "${vault_database_secret_backend_connection.test.name}"
+  name = "%s"
+  username = "%s"
+  rotation_period = 1800
+  rotation_statements = ["SELECT 2;"]
```
tyrannosaurus-becks: If we end up marking rotation statements optional, we should probably also add one test where they're not included.

vault/resource_database_secret_backend_static_role.go
```diff
@@ -0,0 +1,210 @@
...
+	d.Set("name", name)
+	d.Set("username", role.Data["username"])
+	d.Set("db_name", role.Data["db_name"])
+
+	if v, ok := role.Data["rotation_period"]; ok {
+		n, err := v.(json.Number).Int64()
+		if err != nil {
+			return fmt.Errorf("unexpected value %q for rotation_period of %q", v, path)
+		}
+		d.Set("rotation_period", n)
```
tyrannosaurus-becks: For primitive values, we don't need to worry about the potential err returned from `d.Set`, but I recently learned from one of my coworkers that it would be wise to check it when setting more complex values.

tyrannosaurus-becks: Whoops! I meant this for line 155 below where it reads `d.Set("rotation_statements", rotation)`. :blush:

website/docs/r/database_secret_backend_static_role.md
```diff
@@ -0,0 +1,76 @@
...
+Creates a Database Secret Backend static role in Vault. Database secret backend
+static roles can be used to manage 1-to-1 mapping of a Vault Role to a user in a
+database for the database.
+
+~> **Important** All data provided in the resource configuration will be
+written in cleartext to state and plan files generated by Terraform, and
+will appear in the console output when Terraform runs. Protect these
+artifacts accordingly. See
+[the main provider documentation](../index.html)
+for more details.
```
tyrannosaurus-becks: It's awesome you point this out. Do you think we should mark any of the new schema fields Sensitive to protect against this further?

vault/resource_database_secret_backend_static_role.go
martinssipenko: I thought so too initially, however, while testing with bellow commands I discovered it actually is required.

Started postgres  container in tab 1.
```sh
docker run -it --rm \
    -p 5432:5432 \
    -e POSTGRES_PASSWORD=secret \
    -e POSTGRES_USER=root \
    postgres
```

Started vault in tab 2.
```sh
vault server -dev
```

Run following commands in tab 3.
```sh
vault secrets enable database

vault write database/config/my-postgresql-database \
    plugin_name=postgresql-database-plugin \
    allowed_roles="my-role" \
    connection_url="postgresql://{{username}}:{{password}}@127.0.0.1:5432/?sslmode=disable" \
    username="root" \
    password="secret"

vault write database/static-roles/my-role \
    username="test" \
    rotation_period="24h" \
    db_name="my-postgresql-database"
```

When executing last command Vault returns error:
```
Error writing data to database/static-roles/my-role: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/database/static-roles/my-role
Code: 500. Errors:

* 1 error occurred:
	* error setting credentials: empty rotation statements
```

I guess we could pass the `rotation_statements` parameter to backend, but leave it empty by default. Should we do that?

vault/resource_database_secret_backend_static_role.go
martinssipenko: I tried to stay consistent with already existing [`vault_database_secret_backend_connection`](https://www.terraform.io/docs/providers/vault/r/database_secret_backend_role.html) resource, but we could change this for this. Or we could leave it as is and change this in another pull request in both db role resources. WDYT?

website/docs/r/database_secret_backend_static_role.md
martinssipenko: I'm removing this because I after second look it appears to me that there are no secret schema fields.

martinssipenko:
After trying to change `rotation_statements` to be optional, I now get the following error when running acceptance test. @tyrannosaurus-becks perhaps you can help me out here?

```sh
make testacc TESTARGS='-run=TestAccDatabaseSecretBackendStaticRole'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDatabaseSecretBackendStaticRole -timeout 120m
PASS
=== RUN   TestAccDatabaseSecretBackendStaticRole_import
--- FAIL: TestAccDatabaseSecretBackendStaticRole_import (3.70s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: error creating static role "staticrole-2970162460807754599" for backend "tf-test-db-6356817738787049945": Error making API request.
        
        URL: PUT http://127.0.0.1:8200/v1/tf-test-db-6356817738787049945/static-roles/staticrole-2970162460807754599
        Code: 500. Errors:
        
        * 1 error occurred:
        	* error setting credentials: empty rotation statements
        
        
        
          on /var/folders/md/52wjnhf54x50bjskcyvbpjkm0000gn/T/tf-test162060127/main.tf line 17:
          (source code not available)
        
        
=== RUN   TestAccDatabaseSecretBackendStaticRole_basic
--- FAIL: TestAccDatabaseSecretBackendStaticRole_basic (3.27s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: error creating static role "staticrole-3847025391580588612" for backend "tf-test-db-795078128907647134": Error making API request.
        
        URL: PUT http://127.0.0.1:8200/v1/tf-test-db-795078128907647134/static-roles/staticrole-3847025391580588612
        Code: 500. Errors:
        
        * 1 error occurred:
        	* error setting credentials: empty rotation statements
        
        
        
          on /var/folders/md/52wjnhf54x50bjskcyvbpjkm0000gn/T/tf-test018150633/main.tf line 17:
          (source code not available)
        
        
FAIL
FAIL	github.com/terraform-providers/terraform-provider-vault/vault	7.519s
make: *** [testacc] Error 1
```

vault/resource_database_secret_backend_static_role.go
tyrannosaurus-becks: Ah, I see! Interesting. When I saw that output, I thought it actually might be a bug on Vault's end, so I poked at it a little. I found that for the postgres engine, the rotation statements are required [here](https://github.com/hashicorp/vault/blob/master/plugins/database/postgresql/postgresql.go#L103). _But_, for [the MSSQL secrets engine](https://github.com/hashicorp/vault/blob/master/plugins/database/mssql/mssql.go), "statements.Rotation" aren't used at all, and in [the MySQL secrets engine](https://github.com/hashicorp/vault/blob/master/plugins/database/mysql/mysql.go#L275), they're defaulted if they're not provided. So, the behavior is particular to each back-end, which puts us in an unusual spot.

That makes me think that maybe we should leave it marked as optional, only include it in calls if it's set, pass it through to Vault, and immediately return any error that Vault gives. Does that seem like the best way to handle it to you? Or are there potential problems I'm not thinking of?

martinssipenko: Sure, so the code remains as it currently is, but I need to fix the tests, because acceptance tests use Postgres we are required to set the rotation statements. We could change acceptance to use MySQL, or even add additional cases to test on both Postgres and MySQL, WDYT?

tyrannosaurus-becks: Hmmm, for a long time, I've been wanting to add the container for at least one database to the tests. I think MySQL would be a good one to start with. Let me see if I can do it just now, and if so, I'll PR it to master so it can be used here after it's merged.

If that holds you up, though, feel free to take a different approach, though one doesn't occur to me off-hand.

martinssipenko:
Changed acceptance tests to use MySQL, after which they pass:

```sh
make testacc TESTARGS='-run=TestAccDatabaseSecretBackendStaticRole'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDatabaseSecretBackendStaticRole -timeout 120m
PASS
=== RUN   TestAccDatabaseSecretBackendStaticRole_import
--- PASS: TestAccDatabaseSecretBackendStaticRole_import (0.21s)
=== RUN   TestAccDatabaseSecretBackendStaticRole_basic
--- PASS: TestAccDatabaseSecretBackendStaticRole_basic (0.28s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	1.013s
```

PR #581, opened 2019-10-28
Add missing attributes and allow importing on consul_secret_backend_role

hamishforbes:
Also changes the ID to follow the pattern of other vault resources where the Vault API path is the ID, rather than a comma separated backend,role format.

This breaks backward compatibility, not sure if some kind of migration is required?

### DIFF
--- a/vault/resource_consul_secret_backend_role.go
+++ b/vault/resource_consul_secret_backend_role.go
@@ -3,17 +3,23 @@ package vault
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/vault/api"
 )
 
+var (
+	consulSecretBackendRoleBackendFromPathRegex = regexp.MustCompile("^(.+)/roles/.+$")
+	consulSecretBackendRoleNameFromPathRegex    = regexp.MustCompile("^.+/roles/(.+$)")
+)
+
 func consulSecretBackendRoleResource() *schema.Resource {
 	return &schema.Resource{
-		Create: consulSecretBackendRoleCreate,
+		Create: consulSecretBackendRoleWrite,
 		Read:   consulSecretBackendRoleRead,
-		Update: consulSecretBackendRoleUpdate,
+		Update: consulSecretBackendRoleWrite,
 		Delete: consulSecretBackendRoleDelete,
 		Exists: consulSecretBackendRoleExists,
 		Importer: &schema.ResourceImporter{
@@ -27,11 +33,11 @@ func consulSecretBackendRoleResource() *schema.Resource {
 				ForceNew:    true,
 				Description: "The name of an existing role against which to create this Consul credential",
 			},
-			"path": {
+			"backend": {
 				Type:        schema.TypeString,
-				ForceNew:    true,
 				Required:    true,
-				Description: "Unique name of the Vault Consul mount to configure",
+				ForceNew:    true,
+				Description: "The path of the Consul Secret Backend the role belongs to.",
 			},
 			"policies": {
 				Type:        schema.TypeList,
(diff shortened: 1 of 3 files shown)


vault/resource_consul_secret_backend_role.go
```diff
@@ -24,11 +33,11 @@ func consulSecretBackendRoleResource() *schema.Resource {
 				ForceNew:    true,
 				Description: "The name of an existing role against which to create this Consul credential",
 			},
-			"path": {
+			"backend": {
```
tyrannosaurus-becks: If we rename this field, we'll need to go through [a renaming process](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename). I'd prefer to avoid that if possible because path/backend is inconsistent in the provider and I'm thinking it would be best to take them all through the deprecation process at once. But, I may be overlooking something. What do you think?

vault/resource_consul_secret_backend_role.go
```diff
@@ -38,53 +47,95 @@ func consulSecretBackendRoleResource() *schema.Resource {
...
+	if _, err := client.Logical().Write(path, payload); err != nil {
+		return fmt.Errorf("error writing role configuration for %q: %s", path, err)
 	}
-	d.SetPartial("name")
-	d.SetPartial("path")
-	d.SetPartial("policies")
-	d.Partial(false)
 
-	return nil
+	d.SetId(path)
```
tyrannosaurus-becks: Since this will be a different path now, how will this effect folks migrating from a previous version to this?

vault/resource_consul_secret_backend_role.go
```diff
@@ -93,78 +144,67 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+		return "", fmt.Errorf("no name found")
+	}
+	res := consulSecretBackendRoleNameFromPathRegex.FindStringSubmatch(path)
+	if len(res) != 2 {
+		return "", fmt.Errorf("unexpected number of matches (%d) for name", len(res))
+	}
+	return res[1], nil
+}
+
+func consulSecretBackendRoleBackendFromPath(path string) (string, error) {
```
tyrannosaurus-becks: Would it be possible to get a unit test on this one?

vault/resource_consul_secret_backend_role.go
```diff
@@ -93,78 +144,67 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
 	return secret != nil, nil
 }
 
-func consulSecretBackendRolePath(path, name string) string {
-	return strings.Trim(path, "/") + "/roles/" + name
+func consulSecretBackendRolePath(backend, name string) string {
+	return strings.Trim(backend, "/") + "/roles/" + name
+}
+
+func consulSecretBackendRoleNameFromPath(path string) (string, error) {
```
tyrannosaurus-becks: Would probably be good to get a unit test on this one.

vault/resource_consul_secret_backend_role.go
hamishforbes: Yeah I'm guessing this will break? Is there a way to migrate IDs in a provider too?
I kinda went for the nuclear option initially here, but am happy to make changes to improve backwards compatibility.

The current ID is completely different from other role resources, which was a bit confusing when I initially enabled importing and tried to import existing roles.

vault/resource_consul_secret_backend_role.go
tyrannosaurus-becks: Hm, I think you make a valid point. That sounds reasonable to me.

Let's do it! If you're cool with starting the renaming process, that would be great.

vault/resource_consul_secret_backend_role.go
tyrannosaurus-becks: I'm open to updating it, or not updating it. I think you make a good point. If we go that way, it would be best to make the upgrade path for folks seamless, mainly because I should only make breaking changes in major releases of the provider, and we don't have a major release planned soon. 

If we decide to go with changing it, on the question of how, it may be possible using the `StateUpgrader` on the resource. Perhaps it could be used to check for old paths and upgrade them to the new paths. Another idea might be supporting both paths in the Exists and Read funcs, and then upgrading to the new path.

vault/resource_consul_secret_backend_role.go
hamishforbes: Done

vault/resource_consul_secret_backend_role.go
hamishforbes: Done

vault/resource_consul_secret_backend_role.go
hamishforbes: I've added code to handle the old format IDs. I'm not sure how to write tests for this though.
I can't see a way to get a state with the old ID format in the tests?

It does work for me in manual testing though

vault/resource_consul_secret_backend_role.go
hamishforbes: I think I've correctly started the renaming process here now.

PR #589, opened 2019-10-30
WIP: Add a testing package providing Dockerized MySQL

tyrannosaurus-becks:
Relates to #577 

This PR aims to make it easier to test DB plugins by creating an easy way to incorporate using a MySQL Docker container locally and remotely for testing. 

We can extend this pattern to add more test containers that can be easily used. We can also go back and incorporate these test containers into preexisting tests.

This code is a close copy-paste from [here](https://github.com/hashicorp/vault/blob/master/builtin/logical/mysql/backend_test.go#L19) and [here](https://github.com/hashicorp/vault/blob/master/helper/testhelpers/docker/testhelpers.go#L11).


.circleci/config.yml
```diff
@@ -19,6 +20,12 @@ jobs:
           sleep 5
           docker port vault
           docker logs vault
+    - run:
+        name: "Start MySQL Container"
+        command: |
+          docker run --detach -e MYSQL_ROOT_PASSWORD=mysql --name mysql -p 3306:3306 mysql:latest
+          sleep 30
```
martinssipenko: Instead of sleeping for exact 30 seconds you could use [dockerize](https://github.com/jwilder/dockerize):
```sh
dockerize -wait tcp://127.0.0.1:3306 -timeout 30s
```

martinssipenko:
Just curios, what is the reason to run `machine` builds?

martinssipenko:
@tyrannosaurus-becks I got carried away and did a PoC of different way of building in #590 , took your branch as basis. Let me know what you think about it.

In future the build could be improved to run against different database types in parallel.

PR #591, opened 2019-10-31
Monthly maintenance

tyrannosaurus-becks:
This PR makes the following changes:
- Simplifies the path used for the CircleCI builds. I'd over-complicated it earlier by attempting to directly use the $GOPATH provided by Circle, which is actually multiple directories, only some of which we have write access to. I later realized we could just hard-code the one with write access and use it all the way through to improve the code's readability.
- Updates the `openapi.json` doc against present Vault master. It's a _helpful and rather accurate_ report, not a _perfectly accurate_ report, so I use bleeding edge data.
- Begins the deprecation process on a dupe resource the report uncovered.


vault/resource_identity_oidc_key_allowed_client_id.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func identityOidcKeyAllowedClientId() *schema.Resource {
 	return &schema.Resource{
+		DeprecationMessage: "This resource is a less-featured dupe and is being deprecated in favor of the `vault_identity_oidc_key` resource. Please update your configuration.",
```
lawliet89: Hey @tyrannosaurus-becks, this resource isn't really a duper per se. This is meant to allow the user to specify the `alllowed_client_ids` of an OIDC key in a decoupled manner.

We discussed this in https://github.com/terraform-providers/terraform-provider-vault/pull/488

PR #594, opened 2019-11-04
Add token name provider argument

tstoermer:
### Community Note

Introduce a provider argument to define the Vault child token name used by Terraform. This is useful to provide a reference of the Terraform run traceable in Vault audit log, e.g. commit hash or id of the CI/CD execution job.

```release-note
Add `token_name` provider argument defining the Vault child token name
```

N/A


tstoermer:
This will no longer work with https://github.com/terraform-providers/terraform-provider-vault/pull/561

tyrannosaurus-becks:
Hi @tstoermer ! Thanks for this PR!

Happy to look at it when it's ready for review. Per [CONTRIBUTING.md](https://github.com/terraform-providers/terraform-provider-vault/blob/master/.github/CONTRIBUTING.md) it still needs test coverage.

Also, the linked PR will likely move forward because it simply provides _the option_ of not using a child token. I haven't checked out this code yet, but if there's some way to ensure this code works with that PR, that would be awesome.

tstoermer:
Hi @tyrannosaurus-becks, thanks for your first feedback.

Please correct me, if I overlook anything - I was thinking about the two PRs.

The linked PR will skip the child token creation for all tokens that are orphan tokens, e.g.:
- a token created with Auth methods AppRole or AWS for CI/CD
- personal token created with OIDC

This will limit the usage of token name for the vault provider child token.

I would propose to change the other PR using a provider argument flag controlling the creation of a child token (default: create child token for backward compatibility). This would:
- provide full control for terraform user
- enable full usage of both PR features

Another thing to consider:
If one of the above orphan token examples will create a child token (that will have orphan=false) and pass it to terraform, the provider would again create a child token. 
What would be the intention for this? Maybe I'm missing a fact?

tstoermer:
Hi @tyrannosaurus-becks,
I added some acceptance tests checking the token_name usage. Please have a look, hope I did it the right way :) 

Local results:
```
make testacc TESTARGS='-run=TestAccTokenName'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccTokenName -timeout 120m
PASS
=== RUN   TestAccTokenName
--- PASS: TestAccTokenName (1.94s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	1.975s
```

Regarding the other failed tests in circleci build:
Seems they are no longer working with new vault:1.3.0

vault/provider.go
```diff
@@ -56,6 +56,12 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_TOKEN", ""),
 				Description: "Token to use to authenticate to Vault.",
 			},
+			"token_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VAULT_TOKEN_NAME", "terraform"),
```
tyrannosaurus-becks:
```suggestion
				DefaultFunc: schema.EnvDefaultFunc("VAULT_TOKEN_NAME", ""),
```

vault/provider.go
```diff
@@ -634,6 +640,11 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, errors.New("no vault token found")
 	}
 
+	tokenName := d.Get("token_name").(string)
+	if tokenName == "" {
```
tyrannosaurus-becks: This will never be an empty string if the default func on line 62 defaults to "terraform". That's why I proposed the change above. That way it'll be like:
- Is it in the env variable? If so use that.
- Is it set as token_name? If so use that.
- Otherwise, use "terraform".

vault/provider_test.go
```diff
@@ -513,3 +513,137 @@ func TestAccProviderToken(t *testing.T) {
 		})
 	}
 }
+
+func TestAccTokenName(t *testing.T) {
+
+	// unset VAULT_TOKEN_NAME
+	defer func() {
+		if err := os.Unsetenv("VAULT_TOKEN_NAME"); err != nil {
```
tyrannosaurus-becks: I learned a few months ago that we don't actually need to do this. Turns out that all os env variables set by a test get wiped out before the next test starts, and they don't apply to tests in parallel either.

vault/provider.go
tstoermer: Thanks, you are right, the "terraform" fallback/default is duplicated.

By adding the Acc tests, I discovered, that the token name was empty, if token_name is defined as empty string "". That's why I added this fallback.

I will change it to your proposal and I will enhance documentation regarding default value.

tstoermer: Done

vault/provider_test.go
tstoermer: Thanks for your explanation. I removed this.

tstoermer:
Thanks @tyrannosaurus-becks, I added your suggestions and performed a merge with current master, to make sure everything is ok.

PR #606, opened 2019-11-14
Add auth_backend data source

adongy:
### Community Note

Add support for querying `vault_auth_backend` data with a datasource.
Some resources need the auth backend accessor (e.g. for identity groups). If it's managed outside Terraform, we can query the accessor through this datasource.

```release-note
Add `vault_auth_backend` datasource
```

```
$ make testacc TESTARGS='-run=TestDataSourceAuthBackend'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestDataSourceAuthBackend -timeout 120m
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.008s [no tests to run]
=== RUN   TestDataSourceAuthBackend
--- PASS: TestDataSourceAuthBackend (0.21s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	0.217s
```

### DIFF
--- a/vault/data_source_auth_backend.go
+++ b/vault/data_source_auth_backend.go
@@ -1,0 +1,87 @@
+package vault
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+func authBackendDataSource() *schema.Resource {
+	return &schema.Resource{
+		Read: authBackendDataSourceRead,
+		Schema: map[string]*schema.Schema{
+			"path": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The auth backend mount point.",
+			},
+			"type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The name of the auth backend.",
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description of the auth backend.",
+			},
+			"default_lease_ttl_seconds": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Default lease duration in seconds",
+			},
+			"max_lease_ttl_seconds": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Maximum possible lease duration in seconds",
+			},
+			"listing_visibility": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Specifies whether to show this mount in the UI-specific listing endpoint.",
+			},
+			"local": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Specifies if the auth method is local only",
+			},
+			"accessor": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The accessor of the auth backend.",
+			},
+		},
+	}
+}
+
+func authBackendDataSourceRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	targetPath := d.Get("path").(string)
+
+	auths, err := client.Sys().ListAuth()
+	if err != nil {
+		return fmt.Errorf("error reading from Vault: %s", err)
+	}
+
+	for path, auth := range auths {
+		path = strings.TrimSuffix(path, "/")
+		if path == targetPath {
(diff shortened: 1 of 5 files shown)


website/docs/d/auth_backend.html.md
```diff
@@ -0,0 +1,41 @@
+---
```
lawliet89: You need to add a link to `website/vault.erb` or this page will not show up in the sidebar.

adongy: Updated, thanks!

adongy:
Test failures seem unrelated.

adongy:
Hey @tyrannosaurus-becks,

Thanks for the review!

My use case is for bootstrapping Vault configuration. In my case, to avoid doing a manual "first time" Terraform configuration with a root token, I run a bootstrap script that registers some basic config.

In my case, I register an admin policy, as well as a OIDC mount to allow GSuite admin users to login and get that policy. After that, I can revoke the initial root token, and work from short lived tokens.

Through that bootstrap script, the OIDC mount is created, and I'd like to reuse it to allow non-admin users to also login through that auth backend. This config is managed through Vault identities, aliases shenanigans are used so it's a bit more complex.

vault/data_source_auth_backend.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+func authBackendDataSource() *schema.Resource {
+	return &schema.Resource{
+		Read: authBackendDataSourceRead,
+		Schema: map[string]*schema.Schema{
+			"path": {
```
tyrannosaurus-becks: Question, would it be possible to drop this "path" parameter and to simply record all the auths returned by the list response? I'm thinking that the advantage might be that if you needed information on _multiple_ paths, you could get them in only one call. Also, since this is a list endpoint, I wonder if getting the full response might be more what a user would _expect_, so I'm wondering whether we'd eventually get a request for that down the line. Let me know what you think.

vault/data_source_auth_backend.go
```diff
@@ -0,0 +1,87 @@
...
+	targetPath := d.Get("path").(string)
+
+	auths, err := client.Sys().ListAuth()
+	if err != nil {
+		return fmt.Errorf("error reading from Vault: %s", err)
+	}
+
+	for path, auth := range auths {
+		path = strings.TrimSuffix(path, "/")
+		if path == targetPath {
```
tyrannosaurus-becks: If we dropped the "path" match here and recorded everything here, would we need to sort the auths alphabetically by name so that Terraform wouldn't think there was a constant diff due to Go's random map ordering?

vault/data_source_auth_backend.go
adongy: It is of course possible to add a `auth_backends` data source, as equivalents already exist in other providers:
- https://www.terraform.io/docs/providers/aws/d/availability_zones.html
- https://www.terraform.io/docs/providers/aws/d/availability_zone.html

The advantage of such a data source would be to avoid requesting multiple times the list of auth backends (knowing that our implem, does in fact list all of them).

We could return a map of `path` to auth backend, which should amount to the same data. However I feel like readability counts, and doing that incurs a non negligible cost:

```
data "vault_auth_backends" "mounts" {}
...
resource ".." ".." {
  mount_accessor = vault_auth_backend.mounts["oidc"].mount_accessor
} 
```
vs
```
data "vault_auth_backend" "oidc" {
  path = "oidc"
}
...
resource ".." ".." {
  mount_accessor = data.vault_auth_backend.oidc.mount_accessor
} 
```

For people iterating on the list of mounts, it will surely be useful so we can preemptively add it if you want. LMK.

vault/data_source_auth_backend.go
adongy: I believe Terraform is smart enough to ignore the randomness in maps and compare by key/value, TypeMap being the schema type.
Here's a test storing a `map[string]string` as a computed `schema.TypeMap`

https://github.com/hashicorp/terraform/blob/master/builtin/providers/test/resource.go

vault/data_source_auth_backend.go
tyrannosaurus-becks: I think that seems reasonable! Thanks for talking it through with me.

PR #625, opened 2019-12-02
Fix build

tyrannosaurus-becks:
When the latest version of Vault was released, tests started failing on master. It was because the expected `lease_duration` for a `ttl` of 30 seconds was 30 seconds. However, in the tests, they were now coming back as 29 seconds. 

The tests are performed by creating a token, and then in a separate, second API call reading the remaining time left on the lease. Having even a hair of a change in how long the second API call takes would result in this, as would having a hair of change in how Vault calculates the remaining time on the lease. 

Since it's only 1 second, it isn't cause for concern and I think it's sufficient to simply change test expectations.


tyrannosaurus-becks:
I'm going to go ahead and merge this so folks can start getting this merged into their PR's to help me review them.

PR #627, opened 2019-12-05
Prevent panic when Cassandra protocol_version is set

tyrannosaurus-becks:
I circled back to #559 to see if it was really working for me by running Cassandra in Docker locally. I found that if I added `protocol_version = 4` line to our tests, I received a panic on line 544 stating that `v` was a `json.Number`, not a `string`.

This PR fixes that and adds test coverage for when the field is _actively_ set. Test coverage already exists for when it defaults.


tyrannosaurus-becks:
Merging this so it won't hold up the release.

PR #641, opened 2020-01-06
Fix panic when reading unconfigured PKI mount URLs

kalafut:


vault/resource_pki_secret_backend_config_urls.go
```diff
@@ -81,6 +82,10 @@ func pkiSecretBackendConfigUrlsRead(d *schema.ResourceData, meta interface{}) er
 	log.Printf("[DEBUG] Reading URL config from PKI secret backend %q", backend)
 	config, err := client.Logical().Read(path)
 
+	if config == nil && err == nil {
```
catsby: I'm wondering, should we simply return the error if `err` is not nil from `client.Logical().Read(path)`? I think the only case we'd want to remove from state is if `config == nil` and/or `err == nil`, which represents the 404. 

The way the code reads now, and I do realize this is not code you changed, it seems any error will result in removing it from state, which I'm not sure is the behavior we want

kalafut: I think that makes sense and is an improvement we can make here.

PR #557, opened 2019-10-05
Allow update vault_auth_backend resource

leominov:
Should I add other fields from documentation?

It's about fixing #315 #290 #269
(ref: https://www.vaultproject.io/api/system/auth.html)

Let me know what you think about this changes.

Issue #269, opened 2019-01-08
Unable to set LDAP auth TTL settings


oarmstrong:
Using the `vault_ldap_auth_backend` resource it doesn't appear to be possible to set the token TTL settings `default-lease-ttl` and `max-lease-ttl`.

For example, the AWS backend requires the use of a generic `vault_auth_backend` resource which does accept the TTL settings but for the LDAP auth method there is the specific `vault_ldap_auth_backend` resource which does not accept these settings.

I can see two potential solutions.

1. Add the TTL options as parameters to the `vault_ldap_auth_backend` resource which can set them when it mounts the auth method.
1. Require that a `vault_auth_backend` resource be used before the `vault_ldap_auth_backend` and have the latter **not** mount the auth method.

The second option would be preferable to me but would be backwards incompatible. I'm open to opinions and happy to go ahead and submit a PR if there's consensus as to how this should be implemented.

dhohengassner:
Run into the same issue!  
Would be super helpful if we could set this parameter.

I also like the way it is already solved with the kubernetes auth backend - should match your 2nd suggested solution.
https://www.terraform.io/docs/providers/vault/r/kubernetes_auth_backend_config.html

oarmstrong:
Bingo, that's exactly what I meant @dhohengassner. I'll see if I can get some time to look into it this week.

(For now I've worked around it by setting my global TTLs in the Vault config which works for my use case.)

dhohengassner:
@oarmstrong created a PR for that
https://github.com/terraform-providers/terraform-provider-vault/pull/273

please feel free to comment if that meet your expectations

oarmstrong:
LGTM @dhohengassner, cheers.

jsmilani:
FYI, there is a new `vault_generic_endpoint` resource which makes some of the random api endpoints more accessible. I solved this problem with the sys/mount tune api using the following:

    resource "vault_ldap_auth_backend" "ldap" {
      path        = "ldap"
      ...
    }

    resource "vault_generic_endpoint" "ldap-mount-tune-ttl" {
      depends_on = ["vault_ldap_auth_backend.ldap"]
      path = "sys/mounts/auth/${vault_ldap_auth_backend.ldap.path}/tune"
      disable_delete = true
      ignore_absent_fields = true
      write_fields = ["default_lease_ttl", "max_lease_ttl"]
      data_json = <<EOT
    {
      "default_lease_ttl": 43200,
[...]

PR #650, opened 2020-01-15
Add support for standard auth mount tune configuration to vault_auth_backend resource

### LINKED ISSUE #315: `vault_auth_backend` should not `ForceNew` on every field
### Affected Resource(s)
- `vault_auth_backend`

`vault_auth_backend` should not `ForceNew` on every change to fields. By recreating the mount, we revoke all existing tokens issued by the existing mount. This is not desirable in most cases.

The API method [`tune`](https://www.vaultproject.io/api/system/auth.html#tune-auth-method) exists to allow us to modify some of the configuration.

ianferguson:
### Community Note

### Description

Currently, updates to `default_lease_ttl_seconds`, `max_lease_ttl_seconds` or `listing_visibility` attributes of the `vault_auth_backend` resource force creation of a new resource, which invalidates existing tokens previously issued by the backend, and can result in roles under that backend no longer exiting/not being recreated, leading to outages for users depending on that backend to login.

This PR follows up on the PR @leominov opened in #557, but implemented using the suggestion by @shwuandwing in this comment: https://github.com/terraform-providers/terraform-provider-vault/pull/557#issuecomment-539193352

It does not strictly speaking resolve #315, in that the existing parameters will still force a new resource, but it addresses the root issue but deprecating those existing parameters and introducing a `tune` configuration block consistent with what has been done in the `vault_github_auth_backend` and `vault_jwt_auth_backend`. 

The common tune schema and implementation does allow for updating those values without forcing a new resource.

Relates #269 
Relates #315
Relates #557 

```release-note
- Adds support to `vault_auth_backend` for common backend tune parameters. Also allows updating Max TTL, Default TTL and Visibility Listing tuning settings on `vault_auth_backend` without forcing a new resource.
```

```
> env TESTARGS="-run TestAccAuthBackend.*" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccAuthBackend.* -timeout 120m
PASS
=== RUN   TestAccAuthBackend_importBasic
--- PASS: TestAccAuthBackend_importBasic (0.14s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.168s

> env TESTARGS="-run TestResourceAuth.*" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestResourceAuth.* -timeout 120m
PASS
=== RUN   TestResourceAuth
--- PASS: TestResourceAuth (0.23s)
=== RUN   TestResourceAuthTune
--- PASS: TestResourceAuthTune (0.21s)
=== RUN   TestResourceAuthTuneTtlConflict
--- PASS: TestResourceAuthTuneTtlConflict (0.02s)
=== RUN   TestResourceAuthMigrateToTune
--- PASS: TestResourceAuthMigrateToTune (0.23s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.734s
...
```

### DIFF
--- a/vault/resource_auth_backend.go
+++ b/vault/resource_auth_backend.go
@@ -16,6 +16,7 @@ func authBackendResource() *schema.Resource {
 		Create: authBackendWrite,
 		Delete: authBackendDelete,
 		Read:   authBackendRead,
+		Update: authBackendUpdate,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
@@ -49,28 +50,35 @@ func authBackendResource() *schema.Resource {
 			},
 
 			"default_lease_ttl_seconds": {
-				Type:        schema.TypeInt,
-				Required:    false,
-				Optional:    true,
-				Computed:    true,
-				ForceNew:    true,
-				Description: "Default lease duration in seconds",
+				Type:          schema.TypeInt,
+				Required:      false,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"tune.0.default_lease_ttl"},
+				Deprecated:    "Use the tune configuration block to avoid forcing creation of new resource on an update",
+				Description:   "Default lease duration in seconds",
 			},
 
 			"max_lease_ttl_seconds": {
-				Type:        schema.TypeInt,
-				Required:    false,
-				Optional:    true,
-				Computed:    true,
-				ForceNew:    true,
-				Description: "Maximum possible lease duration in seconds",
+				Type:          schema.TypeInt,
+				Required:      false,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"tune.0.max_lease_ttl"},
+				Deprecated:    "Use the tune configuration block to avoid forcing creation of new resource on an update",
+				Description:   "Maximum possible lease duration in seconds",
 			},
 
 			"listing_visibility": {
-				Type:        schema.TypeString,
-				ForceNew:    true,
-				Optional:    true,
-				Description: "Specifies whether to show this mount in the UI-specific listing endpoint",
+				Type:          schema.TypeString,
+				ForceNew:      true,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"tune.0.listing_visibility"},
+				Deprecated:    "Use the tune configuration block to avoid forcing creation of new resource on an update",
+				Description:   "Specifies whether to show this mount in the UI-specific listing endpoint",
 			},
 
 			"local": {
@@ -85,6 +93,8 @@ func authBackendResource() *schema.Resource {
 				Computed:    true,
 				Description: "The accessor of the auth backend",
 			},
+
+			"tune": authMountTuneSchema(),
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


vault/resource_auth_backend.go
```diff
@@ -49,28 +50,35 @@ func authBackendResource() *schema.Resource {
...
-				Optional:    true,
-				Computed:    true,
-				ForceNew:    true,
-				Description: "Default lease duration in seconds",
+				Type:          schema.TypeInt,
+				Required:      false,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"tune.0.default_lease_ttl"},
```
tyrannosaurus-becks: I'm curious why you added this as a `ConflictsWith` field. `ConflictsWith` is described as:
```
// ConflictsWith is a set of schema keys that conflict with this schema.
```
So far I've only seen it used when deprecating a different key in the schema, as described [here](https://www.terraform.io/docs/extend/best-practices/deprecations.html). So then the string used will exactly match a key in the Schema map here. 

Can you explain more about why you added this here? This question also pertains to lines 69 and 79.

ianferguson: `default_lease_ttl_seconds`, `listing_visibility` and `max_lease_ttl_seconds` all overlap with fields in the ` authMountTuneSchema` that this PR is reusing from other mount's resources.

I had marked these 3 fields as deprecated (in favor of the more "standard" set of configuration for mount tuning), and utilized Conflicts with to avoid the `max_lease_ttl_seconds` being set in both this top level block and in the tune block, which would cause an issue with one value winning over the other and/or the configuration flapping between the two. 

My understanding is that this is what is recommended when safely renaming an optional field (https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-an-optional-attribute), which is effectively what this PR is doing, I think, but if that was a misunderstanding on my part I'm happy to readjust how that is done.

ianferguson:
@tyrannosaurus-becks thank you for the review! To answer you questions: 

> I have a general question regarding this PR - is it fully backwards compatible? 

Yes!

> So, would folks who are presently using this need to make any changes to their configs?

Current users will not need to make any changes, and users currently using the existing attributes will be able to move from the now deprecated attributes to the new/standard tune block attributes safely without generating a new resource.

I think this test case exercises a scenario where the user has just the configuration that exists today (the first step) as well as what happens when a user edits the configuration to use the same ttl attributes that are now in the tune block, in the second step of the test:

https://github.com/terraform-providers/terraform-provider-vault/pull/650/files#diff-e2b851590904030e7bfcdb6a3f03663fR302-R339

vault/resource_auth_backend.go
tyrannosaurus-becks: That makes sense! Thank you!

ianferguson:
@tyrannosaurus-becks thank you so much!

Ninir:
@tyrannosaurus-becks can we add this PR to the Changelog? 🙏  it seems to be missing

tyrannosaurus-becks:
@Ninir , whoops, oversight on my part! Just added, thanks for bringing this to my attention.

Andor:
I think you need to update the documentation as well: mark old parameters as deprecated and provide the example.

threemachines:
Agreed, this new `tune` parameter needs to be in the docs.

ianferguson:
@Andor @threemachines I've opened #697  to update the documentation to match these changes, sorry about that.

PR #651, opened 2020-01-15
Set VAULT_ADDR in process env when using token helper

### LINKED ISSUE #526: Export VAULT_ADDR env variable when calling a token helper
Hi there,

I'm not sure if this is possible, but that would be nice if terraform provider (or the vault lib it uses) would export the `VAULT_ADDR` environmental variable whet it calls the token helper.

My use case is that I use a complex terraform configuration with a lot of root modules. Each root module has its individual vault provider with an address parameter:
```hcl
# infra-1/main.tf
provider "vault" {
  address = "https://infra-1.example.com"
}
``` 
```hcl
# infra-2/main.tf
provider "vault" {
  address = "https://infra-2.example.com"
}
``` 
and so on.

Currently I have to run `vault login` every time when I switch between directories, because the provider relies on the default token helper (`~/.vault-token`)  by default.

The idea of [Token Helpers](https://www.vaultproject.io/docs/commands/token-helper.html) is very cool, but it is quite useless in the current implementation, because it doesn't define how the Vault cluster address will be passed to the helper executable.

Most of implementations of token helpers (including the example from https://www.hashicorp.com/blog/building-a-vault-token-helper) are relying on `VAULT_ADDR` env variable. 
In my case, I don't use `VAULT_ADDR`  env variable in my shell because I use multiple vaults. 
So maybe the provider can define it to the expected Vault server address when calling the helper?

What do you think about it? If it's not possible - do you have suggestions how to orginize the workflow for the case I described above?

Thanks!

P.s. Maybe, the vault provider should have a parameter for defying `VAULT_CONFIG` instead? Then people with the similar cases could write small shell-scripts which will wrap the actual helper and export the desired env var before that?

ianferguson:
### Community Note

### Description

Closes #526 

I'm opening this as a draft pull request because I'd like to get the maintainers thoughts on the approach being taken.

I've been using a token helper that stores tokens based on the configured value of VAULT_ADDR in the environment while using terraform across many Vault clusters, and having support for this would make it much easier to run changes en masse and avoid needing to manage setting and unsetting the VAULT_ADDR for different applies based on what was configured in the Vault terraform provider.

```release-note
VAULT_ADDR is always set to the provider's address attribute when a Vault token helper is used to resolve a Vault token.
```

- None at this time -- if the general approach is ok with the maintainers, I can write up some specific tests for this.

### DIFF
--- a/vault/provider_test.go
+++ b/vault/provider_test.go
@@ -616,8 +616,22 @@ func TestAccProviderVaultAddrEnv(t *testing.T) {
 	}
 
 	// Clear the config file env var and restore it after the test.
-	reset, err := tempUnsetenv(config.ConfigPathEnv)
-	defer failIfErr(t, reset)
+	resetConfigPathEnv, err := tempUnsetenv(config.ConfigPathEnv)
+	defer failIfErr(t, resetConfigPathEnv)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// clear BASH_ENV for this test so any cmd.Exec invocations do not source the BASH_ENV
+	// file which is configured with a VAULT_ADDR here:
+	// https://github.com/terraform-providers/terraform-provider-vault/blob/f42716aae3aebc8daf9702dfa20ce3f8d09d9f4d/.circleci/config.yml#L27
+	// All values set in that BASH_ENV file will still be in the process environment of this
+	// test, they just won't clobber any values this test modifies in the process environment
+	// when the ExternalTokenHelper's command is formatted into a shell invocation to exec here:
+	// https://github.com/hashicorp/vault/blob/master/command/token/helper_external.go#L117-L132
+	//
+	resetBashEnvEnv, err := tempUnsetenv("BASH_ENV")
+	defer failIfErr(t, resetBashEnvEnv)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/vault/provider_test.go
+++ b/vault/provider_test.go
@@ -116,8 +116,7 @@ func getTestRMQCreds(t *testing.T) (string, string, string) {
 }
 
 // A basic token helper script.
-const tokenHelperScript = `
-#!/usr/bin/env bash
+const tokenHelperScript = `#!/usr/bin/env bash
 echo "helper-token"
 `
 
@@ -387,17 +386,8 @@ func TestAccProviderToken(t *testing.T) {
 	}
 
 	// Clear the config file env var and restore it after the test.
-	origConfigPath, ok := os.LookupEnv(config.ConfigPathEnv)
-	if ok {
-		// A config path env var was set; ensure we restore it.
-		defer func() {
-			err := os.Setenv(config.ConfigPathEnv, origConfigPath)
-			if err != nil {
-				t.Fatal(err)
-			}
-		}()
-	}
-	err = os.Unsetenv(config.ConfigPathEnv)
+	reset, err := tempUnsetenv(config.ConfigPathEnv)
+	defer failIfErr(t, reset)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


tyrannosaurus-becks:
Thank you for your PR. I have thoughts and questions on it and to avoid splitting the conversation, I'm going to comment on the linked issue further.

ianferguson:
@tyrannosaurus-becks sounds good! I was offline all last week, so will check in on the discussion on the original issue, thanks for following up on this

vault/provider_test.go
```diff
@@ -116,8 +116,7 @@ func getTestRMQCreds(t *testing.T) (string, string, string) {
 }
 
 // A basic token helper script.
-const tokenHelperScript = `
-#!/usr/bin/env bash
+const tokenHelperScript = `#!/usr/bin/env bash
```
ianferguson: The extra newline added after the backtick caused the script to come out as 

```

#!/usr/bin/env bash
echo "helper-token"
```

With the shebang on the 2nd, rather than 1st, line. When executed as an exec by the cmd package for a user using fish shell the follow error silently happened:

```
> /usr/local/bin/fish -c $tmpScriptPath
Failed to execute process '$tmpScriptPath'. Reason:
exec: Invalid argument
The file '$tmpScriptPath' is marked as an executable but could not be run by the operating system.
```

Removing the line break here fixes that otherwise perplexing test failure that non-bash users may see

vault/provider_test.go
```diff
@@ -387,17 +386,8 @@ func TestAccProviderToken(t *testing.T) {
...
-		// A config path env var was set; ensure we restore it.
-		defer func() {
-			err := os.Setenv(config.ConfigPathEnv, origConfigPath)
-			if err != nil {
-				t.Fatal(err)
-			}
-		}()
-	}
-	err = os.Unsetenv(config.ConfigPathEnv)
+	reset, err := tempUnsetenv(config.ConfigPathEnv)
```
ianferguson: I ended up using TestAccProviderToken as a template for writing TestAccProviderVaultAddEnv, and in the process of reading them and writing another similar test consolidated some of the setup/tear down refactoring -- if this is not ok I can unwind the changes here

ianferguson:
@tyrannosaurus-becks I've updated the PR with a flag (`add_address_to_env`) for turning the new functionality on and off (and it defaults to the current non-adding behavior), added tests and updated the documentation.

The tests pass for me against a local `vault server -dev`, but fail on circleci with:
```
--- FAIL: TestAccProviderVaultAddrEnv (0.02s)
    --- FAIL: TestAccProviderVaultAddrEnv/AddAddressToEnvNotConfigured (0.01s)
        provider_test.go:683: bad token value: want "https://pretest-env-var.example.com", got "http://127.0.0.1:8200"
    --- FAIL: TestAccProviderVaultAddrEnv/AddAddressToEnvIsFalse (0.01s)
        provider_test.go:683: bad token value: want "https://pretest-env-var.example.com", got "http://127.0.0.1:8200"
    --- FAIL: TestAccProviderVaultAddrEnv/AddAddressToEnvIsTrue (0.00s)
        provider_test.go:683: bad token value: want "https://provider.example.com", got "http://127.0.0.1:8200"
```

Which I'll take a look at on Monday or Tuesday, but otherwise I think this should be about ready for review based on what we discussed on #526

vault/provider_test.go
```diff
@@ -640,3 +600,206 @@ func testTokenName_check(expectedTokenName string) resource.TestCheckFunc {
...
+
+	// clear BASH_ENV for this test so any cmd.Exec invocations do not source the BASH_ENV
+	// file which is configured with a VAULT_ADDR here:
+	// https://github.com/terraform-providers/terraform-provider-vault/blob/f42716aae3aebc8daf9702dfa20ce3f8d09d9f4d/.circleci/config.yml#L27
+	// All values set in that BASH_ENV file will still be in the process environment of this
+	// test, they just won't clobber any values this test modifies in the process environment
+	// when the ExternalTokenHelper's command is formatted into a shell invocation to exec here:
+	// https://github.com/hashicorp/vault/blob/master/command/token/helper_external.go#L117-L132
+	//
+	resetBashEnvEnv, err := tempUnsetenv("BASH_ENV")
```
ianferguson: I think its possible that a better fix here might be to move all of the vars set here: https://github.com/terraform-providers/terraform-provider-vault/blob/f42716aae3aebc8daf9702dfa20ce3f8d09d9f4d/.circleci/config.yml#L26-L32 to an `Environment` object in the circleci yaml configuration for the container these tests are run in, as documented here: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-container

That being said, I don't know the finer details of these tests or how circle manages those environment variables/containers across multiple steps well enough to be confident that that wouldn't have any unintended side effects, so I took a more surgical route to fixing these tests that rely on testing the VAULT_ADDR change when a token helper is called.

PR #685, opened 2020-02-24
Support metadata parameter when creating vault tokens

dpedu:
This PR adds support for the [meta](https://www.vaultproject.io/api-docs/auth/token/#inlinecode-meta) field when creating a vault token.

### Community Note

```release-note
Support meta field when creating vault tokens
```

Even without my changes the tests were already broken:

```
$ git log -1
commit 2505d4575f3fdecbe42c52df4954b1dad5fe11df (HEAD, upstream/master, master)
Author: Yoan Blanc <yoan@dosimple.ch>
Date:   Fri Feb 21 20:41:47 2020 +0100

    website: add consul_secret_backend_role to menu (#678)

    Signed-off-by: Yoan Blanc <yoan@dosimple.ch>


$ VAULT_ADDR=http://127.0.0.1:8200 VAULT_TOKEN=xxxx make testacc TESTARGS='-run TestResourceToken_full'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestResourceToken_full -timeout 120m
PASS
=== RUN   TestResourceToken_full
--- FAIL: TestResourceToken_full (0.33s)
    testing.go:640: Step 0 error: Check failed: Check 10/12 error: vault_token.test: Attribute 'lease_duration' expected "59", got "60"
FAIL
FAIL	github.com/terraform-providers/terraform-provider-vault/vault	0.418s
FAIL
make: *** [testacc] Error 1
```

If I comment out the `lease_duration` check in that test, it passed with out without my changes:

```
$ VAULT_ADDR=http://127.0.0.1:8200 VAULT_TOKEN=s.rWodq7XNoL0eVypoDQHjbgf0 make testacc TESTARGS='-run TestResourceToken_full'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestResourceToken_full -timeout 120m
PASS
=== RUN   TestResourceToken_full
--- PASS: TestResourceToken_full (0.37s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	0.433s
```


vault/resource_token.go
```diff
@@ -89,6 +89,12 @@ func tokenResource() *schema.Resource {
 				Default:     "token",
 				Description: "The display name of the token.",
 			},
+			"metadata": {
```
tyrannosaurus-becks: I like that you made this more readable by extending its name. I also think it could easily confuse people looking for the `"meta"` field specifically. Could we rename this to `"meta"` to exactly match the field this maps to in the Vault API?

vault/resource_token.go
```diff
@@ -89,6 +89,12 @@ func tokenResource() *schema.Resource {
...
+				Type:        schema.TypeMap,
+				Optional:    true,
+				ForceNew:    true,
```
tyrannosaurus-becks: Should we use `ForceNew` here? It seems to me like even if metadata changes, it still could essentially be the same object, we wouldn't need a brand new one.

vault/resource_token.go
```diff
@@ -89,6 +89,12 @@ func tokenResource() *schema.Resource {
...
+				Description: "A map of string to string valued metadata.",
```
tyrannosaurus-becks:
```suggestion
				Description: "A map of string to string valued metadata. This is passed through to the audit devices.",
```

vault/resource_token.go
```diff
@@ -407,3 +419,11 @@ func tokenCheckLease(d *schema.ResourceData) bool {
 
 	return false
 }
+
+func metaFromMap(input map[string]interface{}) map[string]string {
+	result := make(map[string]string, len(input))
+	for k, v := range input {
+		result[k] = v.(string)
```
tyrannosaurus-becks: If a user erroneously provides a bool or int here, the string that comes out of `v.(string)` will be `""` and the code will continue along. Would it be possible to instead fail here with explicit output saying that they must provide a string but they provided something else for X key?

website/docs/r/token.html.md
```diff
@@ -49,6 +49,8 @@ The following arguments are supported:
 
 * `policies` - (Optional) List of policies to attach to this token
 
+* `metadata` - (Optional) A map of string to string valued metadata to be set on the token
```
tyrannosaurus-becks:
```suggestion
* `metadata` - (Optional) A map of string to string valued metadata to be set on the token. This is passed through to the audit devices.
```

vault/resource_token.go
tyrannosaurus-becks: Hm, yes, good point, I do see that the `meta` currently can't be changed by any of the token endpoints after it's been created.

However, that's not a guarantee going forward. In the future there may be a valid use case for the metadata to change without the token itself changing, for example, if we wanted to enrich tokens with additional metadata for audit logs. In theory, the metadata _could_ change without require a brand new token, it's not part of the fields contract or documentation that it _must_ always be static on a token.

vault/resource_token.go
tyrannosaurus-becks: Ah yes! Thanks for the test coverage on that!

vault/resource_token.go
dpedu: I did a quick test setting `ForceNew: false` to see how Terraform would behave. I created a token with some metadata then added/removed some of the keys. On the first apply after editing the Terraform hcl code, it said it updated the token. But on every single apply after that, Terraform said it's making the same change again and again. I don't think that's desirable behavior; it doesn't line up with my understanding of Terraform. The future use case you mentioned makes sense but I think the provider should be tailored to the constraints Vault has today.

I noticed some other properties both in this resource and others in the Vault provider that exhibit this behavior - they're written in a way that lets Terraform say a change was made and nothing happened - maybe I'm missing something?

dpedu:
Is there anything we can do to move this along?

vault/resource_token.go
tyrannosaurus-becks: Hi! Regarding `ForceNew`, its functionality is documented [here](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#forcenew) and I'm afraid I'm not going to be comfortable merging this until it's not set to false, for the reasons stated earlier. 

Having a permanent diff is a separate issue and can be addressed through the `SchemaDiffSuppressFunc` function.

Thanks again!

dpedu: Okay. I'm not comfortable with submitting code with behavior that I view as a bug. Open to the idea, but I don't understand why it's not a bug. So apologies, I must decline. Thanks for taking a look!

PR #704, opened 2020-03-09
Elasticsearch support in vault_database_secret_backend_connection

tv:
Add basic support for elasticsearch database. By basic, I mean that we only have properties for url, username and password. No ca certificates, or client_certs because they would need to reside inside vaults filesystem.

Current vault's elasticsearch support is not great:
1) it's using legacy API
2) CA certs needs to be in filesystem (can't define them in requests)
3) Role definitions can't modify kibana ACLs

Example of vault-cli walkthrough: https://github.com/hashicorp/vault-plugin-database-elasticsearch#example-walkthrough

This would allow us to setup user provision through vault with terraform same way as with vault-cli. Good enough for now, but when ES support gets better in vault, this could be extended to be better as well.

### Community Note

```release-note
Adds support for elasticsearch in vault_database_secret_backend_connection
```

```
make testacc TESTARGS='-run=TestAccDatabaseSecretBackendConnection_elasticsearch'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDatabaseSecretBackendConnection_elasticsearch -timeout 120m
PASS
=== RUN   TestAccDatabaseSecretBackendConnection_elasticsearch
--- PASS: TestAccDatabaseSecretBackendConnection_elasticsearch (0.19s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   (cached)
```

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -21,3 +21,12 @@ services:
       MYSQL_DATABASE: "main"
       MYSQL_USER: "vault"
       MYSQL_PASSWORD: "vault"
+
+  elastic:
+    image: elasticsearch:7.6.0
+    ports:
+    - "9200:9200"
+    environment:
+      "discovery.type": "single-node"
+      "xpack.security.enabled": "true"
+      "ELASTIC_PASSWORD": "elastic"
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -74,17 +74,17 @@ func databaseSecretBackendConnectionResource() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"url": {
 							Type:        schema.TypeString,
-							Optional:    true,
+							Required:    true,
 							Description: "The URL for Elasticsearch's API",
 						},
 						"username": {
 							Type:        schema.TypeString,
-							Optional:    true,
+							Required:    true,
 							Description: "The username to be used in the connection URL",
 						},
 						"password": {
 							Type:        schema.TypeString,
-							Optional:    true,
+							Required:    true,
 							Description: "The password to be used in the connection URL",
 							Sensitive:   true,
 						},
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -16,7 +16,7 @@ import (
 var (
 	databaseSecretBackendConnectionBackendFromPathRegex = regexp.MustCompile("^(.+)/config/.+$")
 	databaseSecretBackendConnectionNameFromPathRegex    = regexp.MustCompile("^.+/config/(.+$)")
-	dbBackendTypes                                      = []string{"cassandra", "hana", "mongodb", "mssql", "mysql", "mysql_rds", "mysql_aurora", "mysql_legacy", "postgresql", "oracle"}
+	dbBackendTypes                                      = []string{"cassandra", "hana", "mongodb", "mssql", "mysql", "mysql_rds", "mysql_aurora", "mysql_legacy", "postgresql", "oracle", "elasticsearch"}
 )
 
 func databaseSecretBackendConnectionResource() *schema.Resource {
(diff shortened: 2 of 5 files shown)


tv:
Hi @tyrannosaurus-becks ! Here is a 
> Hi @tv, thanks for this PR! I have a couple questions before starting a review on this one.
> 
> > it's using legacy API
> 
> By that, do you mean the version 6.6 of the API noted [here](https://github.com/hashicorp/vault-plugin-database-elasticsearch/blob/master/client.go#L6)? Has that API version been deprecated by Elastic?

Yep, it has been deprecated since 7. The new _security api is almost identical, but in different path. https://www.elastic.co/guide/en/elasticsearch/reference/7.6/security-api.html . So this is not yet an issue and I think the  _xpack/security endpoints will still work in 8.x elasticsearch.

> > CA certs needs to be in filesystem (can't define them in requests)
> 
> Are you aware of a different method being available? If so can you point me in the right direction? We could open a ticket internally to add support for it.

I'd think base64 encoded certificate would suffice most cases. For example, cassandra database plugins TLS support is done that way. https://github.com/hashicorp/vault/blob/master/plugins/database/cassandra/connection_producer.go#L114 .
This will be a blocker for some use cases for sure. We are currently using let's encrypt, so dodged this issue.

> > Role definitions can't modify kibana ACLs
> 
> Can you clarify on the feature you're wishing for here? Would be great to get your use case, and a brief example of what you wish it did.

Now that I look more to the [vault-plugins](https://github.com/hashicorp/vault-plugin-database-elasticsearch/blob/master/client.go) source code, roles are passed as interface. So I'm pretty sure that I missed something in there and that point can be discarded :)

docker-compose.yaml
```diff
@@ -21,3 +21,12 @@ services:
       MYSQL_DATABASE: "main"
       MYSQL_USER: "vault"
       MYSQL_PASSWORD: "vault"
+
+  elastic:
+    image: elasticsearch:7.6.0
```
tyrannosaurus-becks: :heart: :heart: :heart: Thanks for adding this! :-)

vault/resource_database_secret_backend_connection.go
```diff
@@ -66,6 +66,34 @@ func databaseSecretBackendConnectionResource() *schema.Resource {
...
+							Description: "The URL for Elasticsearch's API",
+						},
+						"username": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "The username to be used in the connection URL",
+						},
+						"password": {
+							Type:        schema.TypeString,
+							Optional:    true,
```
tyrannosaurus-becks: Should `url`, `username`, and `password` all be required? They're required by the engine itself noted [here](https://www.vaultproject.io/api/secret/databases/elasticdb.html#parameters).

tyrannosaurus-becks:
Also! I just finished running tests locally and it does look like there's a related failure:
```
=== RUN   TestAccDatabaseSecretBackendConnection_elasticsearch
--- FAIL: TestAccDatabaseSecretBackendConnection_elasticsearch (0.11s)
    testing.go:640: Step 0 error: errors during apply:
        
        Error: error configuring database connection "tf-test-db-8864949240827763805/config/db-5604694732052959644": Error making API request.
        
        URL: PUT http://localhost:8200/v1/tf-test-db-8864949240827763805/config/db-5604694732052959644
        Code: 400. Errors:
        
        * error creating database object: plugin not found in the catalog: elasticsearch-database-plugin
        
          on /tmp/tf-test606206532/main.tf line 7:
          (source code not available)
```
The way I'm running is very simple:
```
docker-compose up
export TF_ACC=1
go test -v ./...
```

vault/resource_database_secret_backend_connection.go
tv: Sure, it just seemed like convention to have everything as optional. I though it might have something to do with multiple different types of optional blocks and so that their internals needed to be optional as well. That doesn't seem to be the case as I can run mysql acceptance tests with these fields as required. Pushing them.

tv:
Ensure that you are running recent enough vault in the local docker compose. `docker pull vault:latest`. And that you have `export ELASTIC_URL="http://elastic:9200"; export ELASTIC_USERNAME="elastic"; export ELASTIC_PASSWORD="elastic"` in env :)

PR #713, opened 2020-03-24
Add Azure data source that can validate new creds

### LINKED ISSUE #580: Timing issue when using dynamic secrets with Vault using "vault_generic_secret" for "azurerm" Terraform provider
**Versions**
Terraform 0.12.10
provider.azurerm v1.35.0
provider.vault v2.0.0

**Description**
I'm trying to use dynamic secrets, generated by Vault via the Azure Secret Engine, to log into Azure. However, it seems there is a timing issue, in a way that the newly generated secrets are not ready for all Azure services, which causes Terraform to run into either on of the following error messages.

Terraform plan/apply/destroy error messages
Actual IDs removed

a)

Error: Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/ID/providers?api-version=2016-02-01: StatusCode=400 -- Original Error: adal: Refresh request failed. Status Code = '400'. Response body: {"error":"unauthorized_client","error_description":"AADSTS700016: Application with identifier 'ID' was not found in the directory 'ID'. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You may have sent your authentication request to the wrong tenant.\r\nTrace ID: f717c659-5fe0-4eaa-b6af-2c73d9159a00\r\nCorrelation ID: ID\r\nTimestamp: 2019-10-14 09:08:13Z","error_codes":[700016],"timestamp":"2019-10-14 09:08:13Z","trace_id":"f717c659-5fe0-4eaa-b6af-2c73d9159a00","correlation_id":"ID","error_uri":"https://login.microsoftonline.com/error?code=700016"}

b)

Error: Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: resources.ProvidersClient#List: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'ID' with object id 'ID' does not have authorization to perform action 'Microsoft.Resources/subscriptions/providers/read' over scope '/subscriptions/ID' or the scope is invalid. If access was recently granted, please refresh your credentials."

The documentation for the AWS Secret Engine even mentions something that kind of matches the behaviour I'm seeing here and I tend to believe, this will not be much different with Azure:

"Unfortunately, IAM credentials are eventually consistent with respect to other Amazon services. If you are planning on using these credential in a pipeline, you may need to add a delay of 5-10 seconds (or more) after fetching credentials before they can be used successfully."

I've already successfully configured and tested the Azure Secret Engine configuration and I'm able to generate credentials (client_id and secret_id).

If I'm providing these manually generated secrets along with the corresponding subscription_id and tenant_id in my provider block, everything works fine.

But If I'm trying to use the automated way below, it is a gamble, if a Terraform plan succeeds or running into one of the two error messages mentioned above. Sometimes a back to back Terraform apply leads to different results:

➜ Azure tfa
data.vault_generic_secret.azure: Refreshing state...
...
azurerm_resource_group.main: Creating...
azurerm_resource_group.main: Creation complete after 2s [id=/subscriptions/ID/resourceGroups/pschulz-dev-temp-resources]
azurerm_virtual_network.main: Creating...
azurerm_virtual_network.main: Still creating... [10s elapsed]
azurerm_virtual_network.main: Creation complete after 12s [id=/subscriptions/ID/resourceGroups/pschulz-dev-temp-resources/providers/Microsoft.Network/virtualNetworks/pschulz-dev-network]
Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
➜ Azure tfa
data.vault_generic_secret.azure: Refreshing state...
Error: Unable to list provider registration status, ...

In addition to that I've cross checked the dynamically generated credentials and extracted them from the debug logs and they work just fine a couple of seconds later.

Unfortunately there is no way to configure a wait timer or build a workaround using a local provisioner as "depends_on" is reserved for future use.

**Terraform Configuration Files**
```
provider "vault" {
  address      = "https://vault:8200"
  token          = var.token
  ca_cert_file = "/Users/patrickschulz/Documents/GitHub/_DEMO/tls/vault/ca.crt.pem"
}

data "vault_generic_secret" "azure" {
  path = "azure/creds/azure-temp-creds"
}

provider "azurerm" {
  client_id     = "${data.vault_generic_secret.azure.data["client_id"]}"
  client_secret = "${data.vault_generic_secret.azure.data["client_secret"]}"
  tenant_id = var.tenant_id
  subscription_id = var.subscription_id
} 
```

Moved from #7717 hashicorp/vault to here, as it seems to make more sense.

tyrannosaurus-becks:
Closes #580 

This PR adds a data source for Azure that returns a client ID and secret. By default, these are just returned without testing whether they've fully propagated throughout Azure, which causes Terraform scripts to run quickly. However, by adding the information needed to validate credentials, it defaults to a retrying loop that one of our colleagues has found successful in the past. This retrying loop won't be appropriate for all environments and use cases, so every element of the loop is configurable.

```release-note
Adds Azure data source that can validate new creds
```

```
=== RUN   TestAccDataSourceAzureAccessCredentials_basic
--- PASS: TestAccDataSourceAzureAccessCredentials_basic (36.81s)
PASS
```


vault/data_source_azure_access_credentials.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func azureAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	backend := d.Get("backend").(string)
+	role := d.Get("role").(string)
+	path := backend + "/creds/" + role
+
+	data := make(map[string][]string)
```
kalafut: Is this needed? Can we use `client.Logical().Read()` instead?

vault/data_source_azure_access_credentials.go
```diff
@@ -0,0 +1,243 @@
...
+	if subscriptionID == "" || tenantID == "" {
+		return fmt.Errorf(`if 'validate_creds' is true, both 'subscription_id' and 'tenant_id' must also be provided`)
+	}
+
+	// Let's, test the credentials before returning them.
+	vnetClient := network.NewVirtualNetworksClient(subscriptionID)
+
+	// You can't directly pass in a client ID and secret.
+	// The closest thing you can do is provide them through the environment.
+	revert, err := setAndPrepareReversion(auth.TenantID, tenantID)
```
kalafut: Maybe rename `setAndPrepareReversion` to something reflecting that it's updating env vars. "Prepare reversion" doesn't tell me much.

vault/data_source_azure_access_credentials.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil
+	}
+	if subscriptionID == "" || tenantID == "" {
+		return fmt.Errorf(`if 'validate_creds' is true, both 'subscription_id' and 'tenant_id' must also be provided`)
+	}
+
+	// Let's, test the credentials before returning them.
+	vnetClient := network.NewVirtualNetworksClient(subscriptionID)
+
+	// You can't directly pass in a client ID and secret.
```
kalafut: The Azure Secrets plugin does something like this which perhaps you can copy?  https://github.com/hashicorp/vault-plugin-secrets-azure/blob/master/provider.go#L132

vault/data_source_azure_access_credentials.go
```diff
@@ -0,0 +1,243 @@
...
+	credValidationTimeoutSecs := d.Get("max_cred_validation_seconds").(int)
+	sequentialSuccessesRequired := d.Get("num_sequential_successes").(int)
+	secBetweenTests := d.Get("num_seconds_between_tests").(int)
+
+	startTime := time.Now()
+	endTime := startTime.Add(time.Duration(credValidationTimeoutSecs) * time.Second)
+
+	// This URL merely lists providers, and should return a 200 if the credentials
+	// have propagated to the Azure server receiving the call.
+	u := fmt.Sprintf("https://management.azure.com/subscriptions/%s/providers?api-version=2019-10-01", subscriptionID)
```
kalafut: We probably need to account for the different environment since they have different endpoints. (https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L34).  The environment should already be configured elsewhere for Azure Secrets to work, so presumably we can grab the endpoint from existing config.

website/docs/d/azure_access_credentials.html.md
```diff
@@ -0,0 +1,116 @@
...
+* `backend` - (Required) The path to the Azure secret backend to
+read credentials from, with no leading or trailing `/`s.
+
+* `role` - (Required) The name of the Azure secret backend role to read
+credentials from, with no leading or trailing `/`s.
+
+* `validate_creds` - (Optional) Whether generated credentials should be 
+validated before being returned. Defaults to `false`, which returns 
+credentials without checking whether they have fully propagated throughout
+Azure Active Directory. Designating `true` activates testing. Defaults
```
kalafut: "Default to `false`" has already been mentioned.

website/docs/d/azure_access_credentials.html.md
```diff
@@ -0,0 +1,116 @@
...
+
+* `max_cred_validation_seconds` - (Optional) If 'validate_creds' is true, 
+the number of seconds after which to give up validating credentials. Defaults
+to 1,200 (20 minutes).
+
+* `subscription_id` - (Required if `validate_creds`) If 'validate_creds' is true, 
+the subscription id to use to test whether the client ID and secret are valid.
+
+* `tenant_id` - (Required if `validate_creds`) If 'validate_creds' is true, 
+the tenant id  to use to test whether the client ID and secret are valid.
```
kalafut:
```suggestion
the tenant id to use to test whether the client ID and secret are valid.
```

vault/data_source_azure_access_credentials_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// retries real credentials until they're eventually consistent.
+	if testing.Short() {
+		t.SkipNow()
+	}
+	mountPath := acctest.RandomWithPrefix("tf-test-azure")
+	subscriptionID, tenantID, clientID, clientSecret, scope := getTestAzureCreds(t)
+	resource.Test(t, resource.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
```
kalafut: A failure case would be good too. Maybe set up an impossible to satisfy set of requirements (1000 successes in 5 seconds or something like that) and make sure it fails.

vault/data_source_azure_access_credentials.go
tyrannosaurus-becks: Sure, I renamed it a bunch of times because I wasn't sure what to call it. I like the idea of saying something about env vars.

vault/data_source_azure_access_credentials.go
tyrannosaurus-becks: Ah! Yes, that was copy pasta. Updating.

tyrannosaurus-becks: https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/c2525e132a31c199cc614fc101220373b1b7e5e0

vault/data_source_azure_access_credentials.go
tyrannosaurus-becks: https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/c2525e132a31c199cc614fc101220373b1b7e5e0

vault/data_source_azure_access_credentials.go
tyrannosaurus-becks: Niiiiice. Thanks! https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/754040d02fce9d92ddf0b9375c15d316a0eff339

vault/data_source_azure_access_credentials.go
tyrannosaurus-becks: @kalafut do you think I should also read the tenant ID and subscription ID from there? https://www.vaultproject.io/api/secret/azure/index.html#sample-response. I was kind of torn on it because on the one hand, it's easy and prevents drift. On the other hand, it's slightly magical?

website/docs/d/azure_access_credentials.html.md
tyrannosaurus-becks: :+1: https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/3d64cb5b713f456488b2799ed4cd1d1892ef258f

vault/data_source_azure_access_credentials_test.go
tyrannosaurus-becks: Just added this: https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/529625c5f98f49425be1782bcbd4716ff23eaf56

Test output remains as:
```
=== RUN   TestAccDataSourceAzureAccessCredentials_basic
--- PASS: TestAccDataSourceAzureAccessCredentials_basic (25.37s)
PASS
```

vault/data_source_azure_access_credentials.go
kalafut: I didn't think of that before, but I think that makes sense. If you're generating creds for (sub,tenant,env), then I can't imagine why you'd want to, or be successful with, verifying them against something other than (sub,tenant,env). I suppose we can start with the magic approach and if we get a request for explicit config we could easily add it without breaking compatibility (it would just be an override). But I really can't think of the use, so I'd say pull them all from there.

tyrannosaurus-becks: This was done in https://github.com/terraform-providers/terraform-provider-vault/pull/713/commits/d9389dc4c6b1175a483a6becd3e89e2e913a4d6f. I also found a more direct way to _only_ do the token refresh call, which I think is preferable over my previous approach because it doesn't require any permissions for some separate call.

PR #717, opened 2020-04-03
Support sts_region in AWS auth's client resource

### LINKED ISSUE #689: No sts_region in vault_aws_auth_backend_client
Hi, 

If define sts_endpoint, sts_region is also required, but we can't see this parameter in https://www.terraform.io/docs/providers/vault/r/aws_auth_backend_client.html? Anything I'm missing?

Reference:
https://www.vaultproject.io/api/auth/aws/index.html#inlinecode-sts_endpoint-1

tyrannosaurus-becks:
Closes #689 

This PR adds support for `sts_region` for those using Terraform to configure the client used in AWS auth.

### DIFF
--- a/vault/resource_aws_auth_backend_client.go
+++ b/vault/resource_aws_auth_backend_client.go
@@ -60,6 +60,12 @@ func awsAuthBackendClientResource() *schema.Resource {
 				Optional:    true,
 				Description: "URL to override the default generated endpoint for making AWS STS API calls.",
 			},
+			"sts_region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Description: `Region to override the default region for making AWS STS API calls. Should only be set 
+if sts_endpoint is set. If so, should be set to the region in which the custom sts_endpoint resides.`,
+			},
 			"iam_server_id_header_value": {
 				Type:        schema.TypeString,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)


vault/resource_aws_auth_backend_client.go
```diff
@@ -78,6 +84,7 @@ func awsAuthBackendWrite(d *schema.ResourceData, meta interface{}) error {
 	ec2Endpoint := d.Get("ec2_endpoint").(string)
 	iamEndpoint := d.Get("iam_endpoint").(string)
 	stsEndpoint := d.Get("sts_endpoint").(string)
+	stsRegion := d.Get("sts_region").(string)
```
catsby, line 87: Is there a situation where `sts_region` would be an empty string, and would that have an effect on the Vault side here https://github.com/hashicorp/vault/blob/314bc50de40afa8545c52849c55bb204b79e95fc/builtin/credential/aws/path_config_client.go#L239 ?

tyrannosaurus-becks:
Just closing this one because it was to aid in fixing a user issue that was fixed through other means.

yongzhang:
@tyrannosaurus-becks I see you closed this PR, so what's the final solution for this?

lcgkm:
Hi @tyrannosaurus-becks 
Please reopen this PR, I think the sts_region is mandatory when we want to set sts_endpoint.
```
sts_endpoint (string: "") - URL to override the default generated endpoint for making AWS STS API calls. If set, sts_region should also be set.
sts_region (string: "") - Region to override the default region for making AWS STS API calls. Should only be set if sts_endpoint is set. If so, should be set to the region in which the custom sts_endpoint resides.
```
Reference:
https://www.vaultproject.io/api-docs/auth/aws

snoby:
um how was this fixed in other ways?  Definitely please reopen.  @tyrannosaurus-becks

tyrannosaurus-becks:
Hi! I moved on from HashiCorp, but @kalafut will know who the present maintainer is here.

bradpoulton:
I am in need of this feature. Thanks!

bradpoulton:
@kalafut Can this be reopened and pulled in?

kalafut:
@bradpoulton Looks like GitHub won’t let me reopen it, but we’ll re-review and make a new PR if needed.

kalafut:
Fixed in: https://github.com/hashicorp/terraform-provider-vault/pull/931

PR #723, opened 2020-04-09
add `use_child_token` to control child token creation

### LINKED ISSUE #722: Feature Request: add provider argument `use_child_token` to control child token creation
### Terraform Version
v0.12.24

### Affected Resource(s)
- `vault_token`

### Expected Behavior
User can choose whether to use child token or use the injected token itself.

### Actual Behavior
One cannot use the injected token itself.
Since the usage of child token is default and cannot be refused, `vault_token` resources created with terraform will be revoked in 20minutes(default of `max_lease_ttl_seconds`.)
There are [periodic service tokens](https://learn.hashicorp.com/vault/security/tokens#periodic-service-tokens) which are meant to have possibility to live no max TTL.

For the developers using terraform to create vault resources, they might not be aware of the child token because the create `vault_token` resources won't show that their parent has a short TTL. I've been struggling while dealing with periodic tokens disappearing in 20 minutes without a notion.

### Steps to Create Periodic Tokens
```

resource "vault_token_auth_backend_role" "test_role" {
  role_name           = "test_role"
  token_period        = "2592000"
}

resource "vault_token" "periodic_token" {
  role_name = vault_token_auth_backend_role.test_role.role_name
}

```

### References
- https://learn.hashicorp.com/vault/security/tokens#periodic-service-tokens
- https://github.com/terraform-providers/terraform-provider-vault/issues/550#issuecomment-552360100

rokrokss:
### Community Note

Closes #722 

```release-note
add `use_child_token` to control child token creation
```

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -157,6 +157,12 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_NAMESPACE", ""),
 				Description: "The namespace to use. Available only for Vault Enterprise",
 			},
+			"use_child_token": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VAULT_USE_CHILD_TOKEN", true),
+				Description: "Set this to false only for skipping the creation of a child token.",
+			},
 		},
 		ConfigureFunc:  providerConfigure,
 		DataSourcesMap: dataSourcesMap,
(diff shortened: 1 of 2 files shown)


rokrokss:
I don't think errors from circleci are caused by this change.

tyrannosaurus-becks:
Hi! Thanks for this PR!

I believe this use case is already achievable using the [`auth_login`](https://www.terraform.io/docs/providers/vault/index.html#auth_login) parameter. Here's [one example](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider_test.go#L287) of it in use. That should be usable for creating a token as well [at these endpoints](https://www.vaultproject.io/api/auth/token#create-token). This would allow you to use a token role that's set with a TTL that's as long as you need.

Will that solve this for you, or is this a different situation?

Thanks again!

rokrokss:
@tyrannosaurus-becks the usecase for this PR is like creating token with unlimited max ttl, we use our root token to create a periodic token with some ttl value, and the periodic token will be able to renew itself unlimited times as long as the renewal is processed before that ttl.
If we use approle or some other non-token auth method, that might solve the problem. But to me it doesn't feel strateforward that we can't use our token to be the direct parent of the created tokens.
It's also hard for users to recognize that another 20min child token is used for created tokens. outputs from the tokens don't provide that info.

tyrannosaurus-becks:
Ah, thanks for explaining that!

When I run the Terraform Vault Provider on a command line where I've set the Vault token through `$ export VAULT_TOKEN=root`, then by the time I arrive [here](https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L644), the returned token is still "root". Then, if the logic were to go into the `auth_login` logic and hit the [create token](https://www.vaultproject.io/api-docs/auth/token#create-token) endpoint, the created token would be a direct child of "root". Also, since you can pass in parameters there, you can use the `period` field to obtain a periodic token.

Marking this as `do-not-merge` for now because the use case described can already be satisfied.

rokrokss:
@tyrannosaurus-becks I'm very sorry to take your time too long, I think I'm not fully understanding the answer.
I meant the usecase as creating a periodic token as the result of the `terraform apply`, with `resource "vault_token"`.
I understand your comment as using a periodic child token for creating other resources, which is different with creating a periodic token as a result. Am I missing something?

tyrannosaurus-becks:
Continuing this conversation on #722

trv-msieczko:
@tyrannosaurus-becks - if my Vault administrator doesn't allow us to create child tokens, all tokens in our namespace have a TTL of 60m, how can I connect with the provider without this change?

tyrannosaurus-becks:
@trv-msieczko please see the discussion [here](https://github.com/terraform-providers/terraform-provider-vault/issues/722#issuecomment-629495117) and do feel free to weigh in. Thanks!

benashz:
Duplicate of #775

PR #724, opened 2020-04-09
Add an identity_group_member_entity_ids resource to manage group entity member in a decoupled manner

shwuandwing:
### Community Note

```release-note

```

```
make testacc TESTARGS='-run=TestAccIdentityGroup
MemberEntityIdsExclusive'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccIdentityGroupMemberEntityIdsExclusive -timeout 120m
PASS
=== RUN   TestAccIdentityGroupMemberEntityIdsExclusive
--- PASS: TestAccIdentityGroupMemberEntityIdsExclusive (0.23s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.241s

make testacc TESTARGS='-run=TestAccIdentityGroupMemberEntityIdsNonExclusive'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccIdentityGroupMemberEntityIdsNonExclusive -timeout 12
0m
PASS
ok      github.com/terraform-providers/terraform-provider-vault/util    0.009s [no tests to run]
=== RUN   TestAccIdentityGroupMemberEntityIdsNonExclusive
--- PASS: TestAccIdentityGroupMemberEntityIdsNonExclusive (0.33s)
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   0.344s
...
```


tyrannosaurus-becks:
Hi! Thanks for this PR! Can you talk me through the use case for this resource? It's not evident to me from looking at the code alone. Thanks again!

vault/provider.go
```diff
@@ -481,6 +481,10 @@ var (
 			Resource:      identityGroupAliasResource(),
 			PathInventory: []string{"/identity/group-alias"},
 		},
+		"vault_identity_group_member_entity_ids": {
+			Resource:      identityGroupMemberEntityIdsResource(),
+			PathInventory: []string{"/identity/lookup/group"},
```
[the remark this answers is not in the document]
shwuandwing: Done

vault/resource_identity_group_member_entity_ids.go
```diff
@@ -0,0 +1,167 @@
...
+	id := d.Get("group_id").(string)
+
+	log.Printf("[DEBUG] Updating IdentityGroupMemberEntityIds %q", id)
+	path := identityGroupIDPath(id)
+
+	vaultMutexKV.Lock(path)
+	defer vaultMutexKV.Unlock(path)
+
+	data := make(map[string]interface{})
+	memberEntityIds := d.Get("member_entity_ids").(*schema.Set).List()
```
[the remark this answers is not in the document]
shwuandwing: I have added two tests where the member_entity_ids are empty; they are always populated and there were no panics.

vault/provider.go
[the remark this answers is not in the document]
shwuandwing: Done

vault/resource_identity_group_member_entity_ids.go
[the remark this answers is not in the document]
shwuandwing: I have added two tests to check on empty memberEntityIds; they did not panic so it is always populated.   The 2 tests did help catch a bug through, so good call.

shwuandwing:
@kalafut -- do you have a timeframe for reviewing this diff?

website/docs/r/identity_group.html.md
```diff
@@ -60,6 +60,8 @@ The following arguments are supported:
 
 * `external_policies` - (Optional) `false` by default. If set to `true`, this resource will ignore any policies return from Vault or specified in the resource. You can use [`vault_identity_group_policies`](identity_group_policies.html) to manage policies for this group in a decoupled manner.
 
+* `external_member_entity_ids` - (Optional) `false` by default. If set to `true`, this resource will ignore any policies return from Vault or specified in the resource. You can use [`vault_identity_group_member_entity_ids`](identity_group_member_entity_ids.html) to manage member entities for this group in a decoupled manner.
```
austingebauer, line 63: Suggest to replace: 
1. `will ignore any policies return` with `will ignore any Entity IDs returned`
2. `to manage member entities for` with `to manage Entity IDs for`

website/docs/r/identity_group.html.md
```diff
@@ -60,6 +60,8 @@ The following arguments are supported:
 
 * `external_policies` - (Optional) `false` by default. If set to `true`, this resource will ignore any policies return from Vault or specified in the resource. You can use [`vault_identity_group_policies`](identity_group_policies.html) to manage policies for this group in a decoupled manner.
```
austingebauer, line 61: I know it's not your change, but it would be kind if you could change `will ignore any policies return` with `will ignore any policies returned` here. Thanks!

vault/resource_identity_group_member_entity_ids.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if d.Get("exclusive").(bool) {
```
austingebauer, line 64: I'm wondering if we should guard populating `data["member_entity_ids"]` here with a check to ensure the identity group is not of type `external`.

Something like: 
```go
resp, err := readIdentityGroup(client, id)
if err != nil {
    return err
}
t, ok := resp.Data["type"]
if ok && t != "external" {
    // rest of logic for d.Get("exclusive").(bool) here
}
```

The reason for this would be to have consistency with the behavior of `external` type groups that don't use the decoupled `vault_identity_group_member_entity_ids` introduced in this PR. 

For example, this config results in zero `member_entity_ids` in `g1` and no errors from `terraform apply`:
```
resource "vault_identity_group" "g1" {
  name = "g1"
  type = "external"
  external_member_entity_ids = false
  member_entity_ids = ["${vault_identity_entity.user.id}"]
}
resource "vault_identity_entity" "user" {
  name = "user"
}
```

This config results in an error from `terraform apply`:
```
resource "vault_identity_group" "g1" {
  name = "g1"
  type = "external"
  external_member_entity_ids = true
}
resource "vault_identity_group_member_entity_ids" "members" {
  exclusive = true
  member_entity_ids = ["${vault_identity_entity.user.id}"]
  group_id = vault_identity_group.g1.id
}
resource "vault_identity_entity" "user" {
  name = "user"
}
```

Error:
```
Error: error updating IdentityGroupMemberEntityIds "478bbe18-8ef8-7488-101a-6e738cf58b4c": Error making API request.

URL: PUT http://127.0.0.1:8200/v1/identity/group/id/478bbe18-8ef8-7488-101a-6e738cf58b4c
Code: 400. Errors:

* member entities can't be set manually for external groups
```

shwuandwing: I rather let this exposed via Vault.

The problem with the vault_identity_group approach is from the terraform a user thinks the group has these members for identity -- when in reality it ignored the memberentityid's entirely.

Here Vault tells you there is an error.

website/docs/r/identity_group.html.md
shwuandwing: Fixed

website/docs/r/identity_group.html.md
shwuandwing: Fixed

shwuandwing:
@austingebauer -- let me know if you still want the check for member_entity_ids.

vault/resource_identity_group_member_entity_ids.go
austingebauer: I actually don't have a strong opinion about whether an error should be returned or not in this case. I think it could be fine either way.

My reason for proposing this is for sake of consistency. I'd prefer the behavior for external groups work the same whether you're using `vault_identity_group_member_entity_ids` or `member_entity_ids`. I think having two different experiences could be confusing. I'm also reluctant to change the current behavior of `member_entity_ids` to return an error because it's been around for a while.

For that reason, I'd prefer to add the check. Let me know your thoughts!

website/docs/r/identity_group.html.md
austingebauer: Thank you!

austingebauer:
LGTM! Thanks for the updates, @shwuandwing.

PR #730, opened 2020-04-15
Allow setting HTTP headers

Lucretius:
This PR adds the ability to set HTTP headers for Vault API requests made by the Terraform Vault provider.  A sample configuration:

```hcl
provider "vault" {
  address = "https://testvault.123"

  headers {
    name = "header1"
    value = "valueofheader1"
  }

  headers {
    name = "X-Special-Header"
    value = "specialheadervalue"
  }
}
```

I have run the acceptance tests locally and also ran this against a Vault cluster in the cloud that requires specific headers in order for requests to forward to the cluster properly.

### Community Note

Relates OR Closes #0000

```release-note
Added `headers` provider configuration setting to allow setting HTTP headers for all requests to the Vault server.
```

```
PASS
ok      github.com/terraform-providers/terraform-provider-vault/vault   94.623s
```

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -161,6 +161,7 @@ func Provider() terraform.ResourceProvider {
 			"headers": {
 				Type:        schema.TypeList,
 				Optional:    true,
+				Sensitive:   true,
 				Description: "The headers to send with each Vault request.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -659,7 +659,11 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	}
 	// Set headers if provided
 	headers := d.Get("headers").([]interface{})
-	parsedHeaders := make(http.Header)
+	parsedHeaders := client.Headers().Clone()
+
+	if parsedHeaders == nil {
+		parsedHeaders = make(http.Header)
+	}
 
 	for _, h := range headers {
 		header := h.(map[string]interface{})
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -164,10 +164,10 @@ func Provider() terraform.ResourceProvider {
 				Description: "The headers to send with each Vault request.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"key": {
+						"name": {
 							Type:        schema.TypeString,
 							Required:    true,
-							Description: "The header key",
+							Description: "The header name",
 						},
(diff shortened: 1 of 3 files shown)


vault/provider.go
```diff
@@ -157,6 +158,25 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_NAMESPACE", ""),
 				Description: "The namespace to use. Available only for Vault Enterprise",
 			},
+			"headers": {
```
tyrannosaurus-becks, line 161: Should this field be marked [sensitive](https://www.terraform.io/docs/cloud/workspaces/variables.html#sensitive-values)? I ask because I suspect that sometimes people will send authentication headers through.

vault/provider.go
```diff
@@ -157,6 +158,25 @@ func Provider() terraform.ResourceProvider {
...
+				Optional:    true,
+				Description: "The headers to send with each Vault request.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The header name",
+						},
+						"value": {
```
tyrannosaurus-becks, line 172: Would it be possible to get test coverage of the  `name` and `value` fields? It would be cool to see what the Terraform config looks like when they're populated.

vault/provider.go
```diff
@@ -637,6 +657,21 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to configure Vault API: %s", err)
 	}
+	// Set headers if provided
+	headers := d.Get("headers").([]interface{})
```
tyrannosaurus-becks, line 661: It would also be great to get a test where they're explicitly not provided. That probably already exists, and if so, if you could just point me to it, that'd be great.

vault/provider.go
Lucretius: Yeah, thats a good idea, I'll change that.

vault/provider.go
Lucretius: Yeah, I can find one - this would be new so the existing tests definitely aren't testing for this behavior.

vault/provider.go
Lucretius: I copied the basic token name test, and just added a custom provider configuration which included the headers.  It doesn't test any header related functionality (which would depend on individual use cases), only that the headers don't cause errors when Vault is doing things (in the case of this test, token lookup)

PR #737, opened 2020-04-23
Move provider to separate package

tyrannosaurus-becks:
In most Terraform providers, all resources and data sources live in the same package as `provider.go`. That's because the provider needs to import all the resources and data sources so it can return them, _and_ all the tests require the full provider to test. So, if they were moved into separate packages, this would create a cyclic dependency:

- `provider` package needs `resources` package to return them
- `resources` package need `provider` package to test

We would like to move towards generating more resources and data sources to cover the Vault API. Vault has a total of 544 endpoints, which translates to at least 544 resources and data sources if we were to fully cover all of Vault's endpoints. Each file should have an accompanying test, so full coverage would add up to 1,088 files + `provider.go` and `provider_test.go` (1,090) files all in one single package. It would be difficult to find a particular file or endpoint you're looking for, and quite awkward and unusual.

This PR moves the provider into its own package and allows resources and data sources to register with it. This breaks the import cycle because now, the provider _won't know_ about the specific resources and data sources registering with it. This creates a one-way path of knowledge. Resources and data sources know about the provider, can register with it, and can test with it. 

Because of breaking the import cycle, this frees us from being required to keep all potential 1,090 files in the same directory, and allows us to organize them in a way that may make more sense to someone looking for the code for a particular endpoint.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -1,0 +1,32 @@
+package provider
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func New(seed *schema.Provider) *Provider {
+	return &Provider{
+		provider: seed,
+	}
+}
+
+type Provider struct {
+	provider *schema.Provider
+}
+
+func (p *Provider) RegisterDataSource(name string, dataSource *schema.Resource) {
+	p.provider.DataSourcesMap[name] = dataSource
+}
+
+func (p *Provider) RegisterResource(name string, resource *schema.Resource) {
+	p.provider.ResourcesMap[name] = resource
+}
+
+func (p *Provider) SchemaProvider() *schema.Provider {
+	return p.provider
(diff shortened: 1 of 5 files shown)


provider/provider.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func New(seed *schema.Provider) *Provider {
+	return &Provider{
+		provider: seed,
+	}
+}
+
+type Provider struct {
```
jasonodonnell, line 14: Question: `Provider.provider` seems weird.  Are we able to do something such as: `type Provider *schema.Provider` instead?  Maybe this isn't possible :)

tyrannosaurus-becks: Thanks for chatting with me about it and letting me bounce thoughts off you. Updated in https://github.com/terraform-providers/terraform-provider-vault/pull/737/commits/5fb25abe536f80b0965eb3f251b96f27cd889ca4.

PR #738, opened 2020-04-23
Add directory structure

tyrannosaurus-becks:
This PR adds the code that will be used to create a directory structure for the generated code in the Terraform Vault Provider.

There's a 1:3 ratio of Vault endpoints to files in the Terraform Vault Provider. Each endpoint (like `"/database/roles/{name}"` requires:
- A `.go` file with its actual code
- A `_test.go` file with its associated tests
- A `.md` file with its docs

Creating a less flat directory structure makes it easier to look for the files related to one of Vault's >500 endpoints. The list of the endpoints for the `transit` engine alone is:
```
"/transit/backup/{name}"
"/transit/cache-config"
"/transit/datakey/{plaintext}/{name}"
"/transit/decrypt/{name}"
"/transit/encrypt/{name}"
"/transit/export/{type}/{name}"
"/transit/export/{type}/{name}/{version}"
"/transit/hash"
"/transit/hash/{urlalgorithm}"
"/transit/hmac/{name}"
"/transit/hmac/{name}/{urlalgorithm}"
"/transit/keys"
"/transit/keys/{name}"
"/transit/keys/{name}/config"
"/transit/keys/{name}/rotate"
"/transit/keys/{name}/trim"
"/transit/random"
"/transit/random/{urlbytes}"
"/transit/restore"
"/transit/restore/{name}"
"/transit/rewrap/{name}"
"/transit/sign/{name}"
"/transit/sign/{name}/{urlalgorithm}"
"/transit/verify/{name}"
"/transit/verify/{name}/{urlalgorithm}"
```
Having a less flat directory structure that directly mirrors the endpoint you're seeking makes it very straight-forward and intuitive to find the file you're looking for.

On alternative approaches, I considered other ideas like making all the endpoints be under "transit" with no further directories, but this makes code generation more complicated. It makes it more likely for collisions to occur with regards to how functions and variables are named in each file. It would be possible to prefix all functions with the "thing" they refer to, like for all methods in `"/transit/sign/{name}/{urlalgorithm}"` to be prefixed with `signNameUrlalgorithmRead`, but I thought that made the generated code read awkwardly.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -8,6 +8,10 @@ default: build
 build: fmtcheck
 	go install
 
+dev: fmtcheck
+	go build -o terraform-provider-vault
+	mv terraform-provider-vault ~/.terraform.d/plugins
+
 test: fmtcheck
 	go test -i $(TEST) || exit 1
 	echo $(TEST) | \
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -8,10 +8,6 @@ default: build
 build: fmtcheck
 	go install
 
-dev: fmtcheck
-	go build -o terraform-provider-vault
-	mv terraform-provider-vault ~/.terraform.d/plugins/
-
 test: fmtcheck
 	go test -i $(TEST) || exit 1
 	echo $(TEST) | \
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -10,7 +10,7 @@ build: fmtcheck
 
 dev: fmtcheck
 	go build -o terraform-provider-vault
-	mv terraform-provider-vault ~/.terraform.d/plugins
+	mv terraform-provider-vault ~/.terraform.d/plugins/
 
 test: fmtcheck
 	go test -i $(TEST) || exit 1
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -3,6 +3,7 @@ package codegen
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
@@ -43,8 +44,8 @@ we eventually cover all >500 of them and add tests.
 			│   └── name.go
 			└── transformation.go
 */
-func codeFilePath(tmplType templateType, path string) string {
-	filename := fmt.Sprintf("%s%s.go", tmplType.String(), path)
+func codeFilePath(tmplType templateType, endpoint string) string {
+	filename := fmt.Sprintf("%s%s.go", tmplType.String(), endpoint)
 	path := filepath.Join(pathToHomeDir, "generated", filename)
 	return stripCurlyBraces(path)
 }
@@ -76,8 +77,8 @@ we eventually cover all >500 of them and add tests.
 			│   └── name.md
 			└── transformation.md
 */
-func docFilePath(tmplType templateType, path string) string {
-	filename := fmt.Sprintf("%s%s.md", tmplType.String(), path)
+func docFilePath(tmplType templateType, endpoint string) string {
+	filename := fmt.Sprintf("%s%s.md", tmplType.String(), endpoint)
 	path := filepath.Join(pathToHomeDir, "website", "docs", "generated", filename)
 	return stripCurlyBraces(path)
 }
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -1,0 +1,89 @@
+package codegen
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// pathToHomeDir yields the path to the terraform-vault-provider
+// home directory on the machine on which it's running.
+// ex. /home/your-name/go/src/github.com/terraform-providers/terraform-provider-vault
+var pathToHomeDir = func() string {
+	repoName := "terraform-provider-vault"
+	wd, _ := os.Getwd()
+	pathParts := strings.Split(wd, repoName)
+	return pathParts[0] + repoName
+}()
+
+/*
+codeFilePath creates a directory structure inside the "generated" folder that's
+intended to make it easy to find the file for each endpoint in Vault, even if
+we eventually cover all >500 of them and add tests.
+	terraform-provider-vault/generated$ tree
+	.
+	├── datasources
+	│   └── transform
+	│       ├── decode
+	│       │   └── role_name.go
+	│       └── encode
+	│           └── role_name.go
+	└── resources
+		└── transform
+			├── alphabet
+			│   └── name.go
+			├── alphabet.go
+			├── role
+			│   └── name.go
+			├── role.go
+			├── template
+			│   └── name.go
+			├── template.go
+			├── transformation
+			│   └── name.go
+			└── transformation.go
+*/
+func codeFilePath(tmplType templateType, path string) string {
+	return stripCurlyBraces(fmt.Sprintf("%s/generated/%s%s.go", pathToHomeDir, tmplType.String(), path))
+}
+
+/*
+docFilePath creates a directory structure inside the "website/docs/generated" folder
+that's intended to make it easy to find the file for each endpoint in Vault, even if
+we eventually cover all >500 of them and add tests.
+	terraform-provider-vault/website/docs/generated$ tree
+	.
+	├── datasources
+	│   └── transform
+	│       ├── decode
+	│       │   └── role_name.md
+	│       └── encode
+	│           └── role_name.md
+	└── resources
+		└── transform
+			├── alphabet
+			│   └── name.md
+			├── alphabet.md
+			├── role
+			│   └── name.md
+			├── role.md
+			├── template
+			│   └── name.md
+			├── template.md
+			├── transformation
+			│   └── name.md
+			└── transformation.md
+*/
+func docFilePath(tmplType templateType, path string) string {
+	result := fmt.Sprintf("%s/website/docs/generated/%s%s.md", pathToHomeDir, tmplType.String(), path)
+	return stripCurlyBraces(result)
+}
+
+// stripCurlyBraces converts a path like
+// "generated/resources/transform-transformation-{name}.go"
+// to "generated/resources/transform-transformation-name.go".
+func stripCurlyBraces(path string) string {
+	path = strings.Replace(path, "{", "", -1)
+	path = strings.Replace(path, "}", "", -1)
+	return path
+}
--- a/codegen/generate_test.go
+++ b/codegen/generate_test.go
@@ -1,0 +1,92 @@
+package codegen
+
+import (
+	"testing"
+)
+
+// testPaths is a sampling of real paths that
+// originate in the OpenAPI doc, pulled via:
+// $ cat testdata/openapi.json | jq '.paths' | jq 'keys[]'
+var testPaths = []string{
+	"/database/roles",
+	"/database/roles/{name}",
+	"/auth/userpass/users/{username}/password",
+	"/auth/userpass/users/{username}/policies",
+	"/transit/export/{type}/{name}/{version}",
+}
+
+func TestCodeFilePath(t *testing.T) {
+	expected := []struct {
+		DataSourceFilePath string
+		ResourceFilePath   string
+	}{
+		{
+			DataSourceFilePath: "/generated/datasources/database/roles.go",
+			ResourceFilePath:   "/generated/resources/database/roles.go",
+		},
+		{
+			DataSourceFilePath: "/generated/datasources/database/roles/name.go",
+			ResourceFilePath:   "/generated/resources/database/roles/name.go",
+		},
+		{
+			DataSourceFilePath: "/generated/datasources/auth/userpass/users/username/password.go",
+			ResourceFilePath:   "/generated/resources/auth/userpass/users/username/password.go",
+		},
+		{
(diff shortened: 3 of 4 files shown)


GNUmakefile
```diff
@@ -8,6 +8,10 @@ default: build
 build: fmtcheck
 	go install
 
+dev: fmtcheck
+	go build -o terraform-provider-vault
+	mv terraform-provider-vault ~/.terraform.d/plugins
```
tyrannosaurus-becks, line 13: This is unrelated to directory generation, just a helpful method for developers.

catsby: nit: if `~/.terraform.d/plugins` directory doesn't already exist, this moves the compiled binary to that path _as `~/.terraform.d/plugins`. 

<details>

```
[terraform-provider-vault][add-directory-structure](4)$ ls -la ~/.terraform.d
total 16
drwxr-xr-x   4 clint  staff   128 Apr 27 13:59 .
drwxr-xr-x+ 73 clint  staff  2336 Apr 27 13:59 ..
-rw-r--r--   1 clint  staff   319 Apr 23 06:55 checkpoint_cache
-rw-r--r--   1 clint  staff   394 Apr 20 07:49 checkpoint_signature

[terraform-provider-vault][add-directory-structure](4)$ make dev
mv terraform-provider-vault ~/.terraform.d/plugins

[terraform-provider-vault][add-directory-structure](4)$ ls -la ~/.terraform.d
total 72944
drwxr-xr-x   5 clint  staff       160 Apr 27 13:59 .
drwxr-xr-x+ 73 clint  staff      2336 Apr 27 13:59 ..
-rw-r--r--   1 clint  staff       319 Apr 23 06:55 checkpoint_cache
-rw-r--r--   1 clint  staff       394 Apr 20 07:49 checkpoint_signature
-rwxr-xr-x   1 clint  staff  37335636 Apr 27 13:59 plugins

[terraform-provider-vault][add-directory-structure](4)$ ~/.terraform.d/plugins
This binary is a plugin. These are not meant to be executed directly.
Please execute the program that consumes these plugins, which will
load any plugins automatically
```
</details>

I believe either a trailing slash `/` on the path, or using the `-p` option for `mkdir` will resolve that.

Terraform searches the current directory for plugins as a convenience for developing, and recommends ~/.terraform.d/plugins` for third-party plugins. I'm not sure we should encourage placing dev provider builds there.

https://www.terraform.io/docs/extend/how-terraform-works.html#plugin-locations

tyrannosaurus-becks:
Hi! Thanks for the review!

>1 thing on the dev step of the Makefile. Is that a step that's consistent with other providers? It's been a while since I was fluent in this area.
>
>I also didn't see a make step for running this new generation code, how is it invoked?

Regarding that, I haven't added it _yet_ but I do have plans to. The reason I didn't add it is because I haven't PR'd the script yet that `make generate` will call. That's why I'm PR'ing all of this into a branch. That way I can keep the PRs small, and there can be a final large PR of vetted code at the end.

GNUmakefile
tyrannosaurus-becks: https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/386b63148cc3be9c6bf78c71f12a71ffddad5835

codegen/generate.go
```diff
@@ -0,0 +1,89 @@
...
+			├── role.go
+			├── template
+			│   └── name.go
+			├── template.go
+			├── transformation
+			│   └── name.go
+			└── transformation.go
+*/
+func codeFilePath(tmplType templateType, path string) string {
+	return stripCurlyBraces(fmt.Sprintf("%s/generated/%s%s.go", pathToHomeDir, tmplType.String(), path))
```
pcman312, line 47:
```suggestion
	filename := fmt.Sprintf("%s%s.go", tmplType.String(), path)
	path := filepath.Join(pathToHomeDir, "generated", filename)
	return stripCurlyBraces(path)
```

codegen/generate.go
```diff
@@ -0,0 +1,89 @@
...
+			├── template
+			│   └── name.md
+			├── template.md
+			├── transformation
+			│   └── name.md
+			└── transformation.md
+*/
+func docFilePath(tmplType templateType, path string) string {
+	result := fmt.Sprintf("%s/website/docs/generated/%s%s.md", pathToHomeDir, tmplType.String(), path)
+	return stripCurlyBraces(result)
```
pcman312, line 79:
```suggestion
	filename := fmt.Sprintf("%s%s.md", tmplType.String(), path)
	path := filepath.Join(pathToHomeDir, "website", "docs", "generated", filename)
	return stripCurlyBraces(path)
```

codegen/generate.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// stripCurlyBraces converts a path like
+// "generated/resources/transform-transformation-{name}.go"
+// to "generated/resources/transform-transformation-name.go".
+func stripCurlyBraces(path string) string {
+	path = strings.Replace(path, "{", "", -1)
+	path = strings.Replace(path, "}", "", -1)
```
pcman312, line 87: optional nit:
```suggestion
	path = strings.ReplaceAll(path, "{", "")
	path = strings.ReplaceAll(path, "}", "")
```

codegen/templates.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func (t templateType) String() string {
+	switch t {
+	case templateTypeDataSource:
+		return "datasources"
+	case templateTypeResource:
+		return "resources"
+	}
+	return "docs"
```
pcman312, line 18: Since the `templateType` is ultimately an integer that could be set to any value, I recommend adding a case for `templateTypeDoc` that returns `"docs"` and having a default value - something like `"unrecognizedtype"` or something like that. That way we can easily distinguish files that were created from bad `templateType`s.

codegen/generate_test.go
```diff
@@ -0,0 +1,92 @@
...
+var testPaths = []string{
+	"/database/roles",
+	"/database/roles/{name}",
+	"/auth/userpass/users/{username}/password",
+	"/auth/userpass/users/{username}/policies",
+	"/transit/export/{type}/{name}/{version}",
+}
+
+func TestCodeFilePath(t *testing.T) {
+	expected := []struct {
```
pcman312, line 19: This table test feels very flaky. 1) It's relying upon the ordering of the `testPaths` which are defined outside of this test. 2) You have to keep the number and order of test cases in line with the `testPaths` slice.

Instead, I recommend embedding the input path into the test case struct:
```go
[]struct {
  input string
  expectedDataSourcePath string
  expectedResourcePath string
}{
  {
    input: "/database/roles",
    expectedSourcePath: "/generated/datasources/database/roles.go",
    expectedResourcePath: "/generated/resources/database/roles.go",
  },
  ...
}
```
This helps ensure that adding/removing/changing test cases is self-contained and not reliant on external values that could be manipulated by other tests. It also makes the tests easier to understand because all of the values are in the test case struct definition and not in multiple locations.

GNUmakefile
```diff
@@ -8,6 +8,10 @@ default: build
 build: fmtcheck
 	go install
 
+dev: fmtcheck
+	go build -o terraform-provider-vault
+	mv terraform-provider-vault ~/.terraform.d/plugins/
```
catsby, line 13: With 386b631 this will error if `~/.terraform.d/plugins/` doesn't already exist. Is a `dev` target consistent with other Terraform providers?  I think for now we should either remove this target or leave it as `go build -o terraform-provider-vault` and leave it in the current directory to be picked up by Terraform, since `.` is a valid search path as a convenience during development 

<details>

```
[terraform-provider-vault][add-directory-structure](3)$ make dev                                                                                             
mv: rename terraform-provider-vault to /Users/clint/.terraform.d/plugins/: No such file or directory
make: *** [dev] Error 1
```

</details>

codegen/generate_test.go
```diff
@@ -0,0 +1,92 @@
...
+			DataSourceFilePath: "/generated/datasources/database/roles.go",
+			ResourceFilePath:   "/generated/resources/database/roles.go",
+		},
+		{
+			DataSourceFilePath: "/generated/datasources/database/roles/name.go",
+			ResourceFilePath:   "/generated/resources/database/roles/name.go",
+		},
+		{
+			DataSourceFilePath: "/generated/datasources/auth/userpass/users/username/password.go",
+			ResourceFilePath:   "/generated/resources/auth/userpass/users/username/password.go",
```
catsby, line 33: This seems very verbose in naming. `/generated/resources/auth/userpass` seems very reasonable, but curious of the benefit of going deeper into `user/username/`. Looking at the [API docs][userpass] I more likely expect a `.go` file for each endpoint under `username`, e.g.  `/generated/resources/auth/userpass/user.go` and `/generated/resources/auth/userpass/login.go`. What do you think about reducing the nesting here?


[userpass]: https://www.vaultproject.io/api-docs/auth/userpass

codegen/templates.go
tyrannosaurus-becks: Sure, I can do that.

tyrannosaurus-becks: https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/0817194e75a179dfda560319faf19fe0ed2499c3

I made it the first option, corresponding to a value of 0, because of it being the default/unset value for an integer.

codegen/generate.go
tyrannosaurus-becks: I had to slightly tweak this to keep tests passing because path is a variable name already passed in. https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/52e1328c533e05f2e864bc3ee23d859e2065264d

codegen/generate.go
tyrannosaurus-becks: Had to tweak this too: https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/52e1328c533e05f2e864bc3ee23d859e2065264d

codegen/generate_test.go
tyrannosaurus-becks: Sure, I could do that.

GNUmakefile
tyrannosaurus-becks: Hey, regarding this, did you see my question earlier?
>I haven't seen it in other providers, I simply did it to speed my own development workflow. Is there a different approach you're aware of that people tend to do?

codegen/generate_test.go
tyrannosaurus-becks: Maybe, I'm not entirely certain the way you're thinking of. 

Here's an example of all the userpass endpoints, with the .go file that currently gets generated for it:
```
/auth/userpass/duo/access =>                /generated/resources/auth/userpass/duo/access.go
/auth/userpass/duo/config =>                /generated/resources/auth/userpass/duo/config.go
/auth/userpass/login/{username} =>          /generated/resources/auth/userpass/login/username.go
/auth/userpass/mfa_config =>                /generated/resources/auth/userpass/mfa_config.go
/auth/userpass/users =>                     /generated/resources/auth/userpass/users.go
/auth/userpass/users/{username} =>          /generated/resources/auth/userpass/users/username.go
/auth/userpass/users/{username}/password => /generated/resources/auth/userpass/users/username/password.go
/auth/userpass/users/{username}/policies => /generated/resources/auth/userpass/users/username/policies.go
```
Can you copy pasta that and put what the result would be the way you're thinking of?

codegen/generate_test.go
tyrannosaurus-becks: This is done in https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/79a3b6c1e0ecf4dce6ccc947657b595c8b116965

GNUmakefile
catsby: From what I saw in https://github.com/terraform-providers/ other providers just use the default make step of `build`, which just does `go install`. For development purposes I think we should leave it as is, unless `make` alone wasn't working for you? I have heard some changes may come here in the search paths for provider binaries but from what I did gather I believe if we wanted to change beyond `go install` then the right thing would be to build it in the current directory.

codegen/generate_test.go
catsby: I was thinking something like this would be simpler:

```
/auth/userpass/duo/access =>                /generated/resources/auth/userpass/duo.go
/auth/userpass/duo/config =>                /generated/resources/auth/userpass/duo.go
/auth/userpass/login/{username} =>          /generated/resources/auth/userpass/login.go
/auth/userpass/mfa_config =>                /generated/resources/auth/userpass/mfa_config.go
/auth/userpass/users =>                     /generated/resources/auth/userpass/users.go
/auth/userpass/users/{username} =>          /generated/resources/auth/userpass/users.go
/auth/userpass/users/{username}/password => /generated/resources/auth/userpass/users.go
/auth/userpass/users/{username}/policies => /generated/resources/auth/userpass/users.go
```

GNUmakefile
tyrannosaurus-becks: This is done in https://github.com/terraform-providers/terraform-provider-vault/pull/738/commits/68eba094b3ff8a9af375c50cb0a119ab6215215b. I hadn't realized I can do essentially the same thing with just `make`.

PR #743, opened 2020-05-05
aws_secret_backend_role: fix attribute removal

### LINKED ISSUE #742: aws_secret_backend_role: cannot remove attribute values
### Terraform Version
```
Terraform v0.12.24
+ provider.vault v2.10.0
```

### Affected Resource(s)
- vault_aws_secret_backend_role

```hcl
resource "vault_aws_secret_backend_role" "foo" {
  backend         = vault_aws_secret_backend.aws.path
  name            = "foo"
  credential_type = "assumed_role"

  role_arns = [
    "arn:aws:iam::${local.account_id}:role/foo",
  ]
  policy_arns = [
    "arn:aws:iam::${local.account_id}:policy/AmazonEC2ReadOnlyAccess",
  ]
}
```

Given the configuration file on top, applying, then removing `policy_arns` and applying will give the following plan:

```
  # vault_aws_secret_backend_role.foo will be updated in-place
  ~ resource "vault_aws_secret_backend_role" "foo" {
        backend         = "aws"
        credential_type = "assumed_role"
        default_sts_ttl = 0
        id              = "aws/roles/foo"
        max_sts_ttl     = 0
        name            = "foo"
      ~ policy_arns     = [
          - "arn:aws:iam::aws:policy/AmazonEC2ReadOnlyAccess",
        ]
        role_arns       = [
            "arn:aws:iam::...:role/foo",
        ]
    }
```

Everything looks okay, however applying does not perform the change. This results in a permanent config drift.

### Expected Behavior
`policy_arns` gets removed.

### Actual Behavior
`policy_arns` stays, state stored in the `.tfstate` correctly reflects that, next plans will still try to apply the change.

### Steps to Reproduce
1. `terraform apply`
2. remove `policy_arns`
3. `terraform apply`
4. plan again.

Ultimately, this comes down to the fact that we don't send empty attributes
https://github.com/terraform-providers/terraform-provider-vault/blob/4a089b73916b4c36a8d89543eaded08ed967ed1b/vault/resource_aws_secret_backend_role.go#L136-L138

On the Vault side, updates are performed by reading the stored role, then changing the data sent. Thus, as `policy_arns` is not in the data payload, it does not get removed.
https://github.com/hashicorp/vault/blob/3d2e84128bb0d9d73e5008bc77d6504354189b82/builtin/logical/aws/path_roles.go#L197-L312

adongy:
### Community Note

Closes #742 

There's a change in util (that only affects the `aws_secret_backend_role` resource). The pathological case treats newlines/spaces as non meaningful, for cases where a template could expand to whitespace only:

```
resource "" "" {
policy_document = <<EOF
EOF
}
```

```release-note
aws_secret_backend_role: fix attribute removal (role_arns, policy_arns, policy_document, sts_ttl settings)
```

```
$ make test
go test -i $(go list ./... |grep -v 'vendor') || exit 1
echo $(go list ./... |grep -v 'vendor') | \
	xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-vault github.com/terraform-providers/terraform-provider-vault/cmd/coverage github.com/terraform-providers/terraform-provider-vault/util github.com/terraform-providers/terraform-provider-vault/vault
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.017s
ok  	github.com/terraform-providers/terraform-provider-vault/vault	0.477s

$ make testacc TESTARGS='-run=TestAccAWSSecretBackend'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSSecretBackend -timeout 120m
PASS
--- PASS: TestAccAWSSecretBackendRole_basic (0.24s)
--- PASS: TestAccAWSSecretBackendRole_import (0.28s)
--- PASS: TestAccAWSSecretBackendRole_nested (0.28s)
=== RUN   TestAccAWSSecretBackend_basic
--- PASS: TestAccAWSSecretBackend_basic (0.30s)
=== RUN   TestAccAWSSecretBackend_import
--- PASS: TestAccAWSSecretBackend_import (0.11s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	1.218s
```

Thanks!


vault/resource_aws_secret_backend_role.go
```diff
@@ -130,25 +130,15 @@ func awsSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
...
 	maxStsTTL, maxStsTTLOk := d.GetOk("max_sts_ttl")
 	if credentialType == "assumed_role" || credentialType == "federation_token" {
-		if defaultStsTTLOk {
-			data["default_sts_ttl"] = strconv.Itoa(defaultStsTTL.(int))
-		}
-		if maxStsTTLOk {
-			data["max_sts_ttl"] = strconv.Itoa(maxStsTTL.(int))
-		}
+		data["default_sts_ttl"] = strconv.Itoa(defaultStsTTL.(int))
+		data["max_sts_ttl"] = strconv.Itoa(maxStsTTL.(int))
```
tyrannosaurus-becks, line 141: For these (line 140 and 141), should we only set them if they've been sent by the caller? 

What I'm thinking about is, what if they send nothing, and the code here resolves to `data["max_sts_ttl"] = "0"`. Then, Vault gets the call and sees the `max_sts_ttl` key _included_ and views it as an intentional 0, rather than as _not included_. I'm not sure what Vault's behavior would be in that situation, and I'm not sure it will always behave the same way in the future as it does today.

vault/resource_aws_secret_backend_role.go
```diff
@@ -130,25 +130,15 @@ func awsSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
...
-	}
-	if len(policyARNs) != 0 {
-		data["policy_arns"] = policyARNs
-	}
-	if len(roleARNs) != 0 {
-		data["role_arns"] = roleARNs
-	}
+	data["policy_document"] = strings.TrimSpace(policy.(string))
+	data["policy_arns"] = policyARNs
+	data["role_arns"] = roleARNs
```
tyrannosaurus-becks, line 135: Are these being added even if they're zero so that people will be able to set them to empty again? If so, what would you think about instead putting them behind a `GetOk` where, if OK, they get set?

adongy: I feel like knowing that to remove these attributes if they were previously set, you need to set them to an empty list (or empty string, for `policy_document`) is counter intuitive. Two seemingly similar configurations have different results depending on the current state:

```tf
resource "" "" {
  policy_arns = []
}

# vs

resource "" "" {
  # nothing
}
```

If you apply these two from a fresh state, you get the same result. I'd expect if I applied this on another cluster that already has configuration would result in the same. If we use `GetOk` the non-explicit version will keep the old value.

vault/resource_aws_secret_backend_role.go
adongy: Unfortunately, I did not find an endpoint that allows us to overwrite the whole role instead of only updating it, so it's a hard call: either we follow the Vault behavior (keep previous values), or we favor the Terraform principle where given a configuration we need a stable output.

For integer values, the "default" value being 0 stays consistent with go's default value, which is why I believe this won't change anytime soon. I don't see another way for us to tell Vault to use the default value for these parameters on this specific API call, other than using the actual default value.

adongy:
Another way to implement this, although a bit more convoluted, would be to use the current state to determine which values to send, although it is subject to the same issues: if the default value changes, we won't catch it.

vault/resource_aws_secret_backend_role.go
tyrannosaurus-becks: I like the first way of unsetting a field.

It's great for backwards compatibility. Since these 5 fields are all optional, some users will have omitted those fields from their configurations because they're using the present intentional or default settings. This could overwrite their previous settings.

It's great for forwards compatibility too. In Vault, [0 indicates the some TTLs should never expire](https://www.vaultproject.io/docs/concepts/tokens#token-time-to-live-periodic-tokens-and-explicit-max-ttls). If we ever changed STS TTLs to behave that way too, overwriting with 0 here would dramatically extend an STS token's expiration.

AndreasSko:
@adongy thank you for this PR! We are facing the same problem you described in #742: We are migrating from using `policy_document` to `policy_arn` for existing resources and are now in a weird state where every apply claims to remove the `policy_document` but in the end fails to. Your work looks like to fix this issue. 
I'm just wondering what the state of this PR is, as there haven't been any updates for over a year. Were you able to work around it in a different way? Are there other things that are blocking this fix and where help could be beneficial, so the PR can keep moving? 🙂

adongy:
Hello @AndreasSko,

I have changed companies and thus do not require the PR anymore. If other people are interested in taking/rebasing/finishing it, please feel free to do so.

PR #749, opened 2020-05-08
Remaining code generation and endpoints

tyrannosaurus-becks:
This PR adds the following:

- A template for generating data sources
- A template for generating docs
- A README on how to use code generation

It also adds support for the following endpoints:

- `"/transform/alphabet/{name}"`
- `"/transform/decode/{role_name}"`
- `"/transform/encode/{role_name}"`
- `"/transform/template/{name}"`
- `"/transform/transformation/{name}"`

<details>
<summary>Output from acceptance tests</summary>

```
tbex@pop-os:~/go/src/github.com/terraform-providers/terraform-provider-vault/generated$ go test -v ./...
?   	github.com/terraform-providers/terraform-provider-vault/generated	[no test files]
=== RUN   TestDecodeBasic
--- PASS: TestDecodeBasic (0.20s)
=== RUN   TestDecodeBatch
--- PASS: TestDecodeBatch (0.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/decode	(cached)
=== RUN   TestEncodeBasic
--- PASS: TestEncodeBasic (0.20s)
=== RUN   TestEncodeBatch
--- PASS: TestEncodeBatch (0.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/encode	(cached)
=== RUN   TestAlphabetName
--- PASS: TestAlphabetName (0.37s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/alphabet	(cached)
=== RUN   TestRoleName
--- PASS: TestRoleName (0.38s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/role	(cached)
=== RUN   TestTemplateName
--- PASS: TestTemplateName (0.42s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/template	(cached)
=== RUN   TestTransformationName
--- PASS: TestTransformationName (0.38s)
PASS
```

</details>


codegen/templates.go
```diff
@@ -69,61 +69,65 @@ type templateHandler struct {
 }
 
 // Write takes one endpoint and uses a template to generate text
-// for it. This template is written to the given writer.
-func (h *templateHandler) Write(wr io.Writer, tmplType templateType, endpoint string, endpointInfo *framework.OASPathItem) error {
+// for it. This template is written to the given writer. It's exported
+// because it's the only method intended to be called by external callers.
+func (h *templateHandler) Write(wr io.Writer, endpoint string, endpointInfo *framework.OASPathItem, addedInfo *additionalInfo) error {
```
pcman312, line 74: Same question here re: pointer receiver.

codegen/templates.go
```diff
@@ -69,61 +69,65 @@ type templateHandler struct {
...
-	for i := 1; i < len(parameters); i++ {
-		if parameters[j] == parameters[i] {
-			continue
+	if len(parameters) > 2 {
+		// De-duplicate the parameters in place, because often parameters
+		// are at both the top-level and in the post body. This in-place
+		// approach is directly recommended here:
+		// https://github.com/golang/go/wiki/SliceTricks#in-place-deduplicate-comparable
+		j := 0
+		for i := 1; i < len(parameters); i++ {
```
pcman312, line 108: Can de-duplication be put into the `collectParameters` function? That seems like a common requirement for that function.

pcman312, line 108: Do we need to handle out-of-order de-duplication or are we guaranteed the parameters will always be ordered properly?

codegen/templates_test.go
```diff
@@ -25,7 +25,7 @@ func TestClean(t *testing.T) {
 		},
 		{
 			input:    "{role_name}",
-			expected: "rolename",
+			expected: "roleName",
```
pcman312, line 28: Can you add some additional tests around camel casing? I'm thinking less happy-path tests like:
`{role_name_}`
`{_role_name}`
`{role__name}`
`{role_name_here}`
`{rOlE_nAmE}`
`{ROLE_NAME}`

generated/datasources/transform/decode/role_name_test.go
```diff
@@ -0,0 +1,124 @@
...
+	p.RegisterResource("vault_transform_role_name", role.NameResource())
+	p.RegisterDataSource("vault_transform_decode_role_name", RoleNameDataSource())
+	return p
+}()
+
+func TestDecodeBasic(t *testing.T) {
+	isEnterprise := os.Getenv("TF_ACC_ENTERPRISE")
+	if isEnterprise == "" {
+		t.Skip("TF_ACC_ENTERPRISE is not set, test is applicable only for Enterprise version of Vault")
+	}
```
pcman312, line 31: This could be extracted to a function:
```go
func SkipIfNotEnterprise(t *testing.T) {
	t.Helper()
	isEnterprise := os.Getenv("TF_ACC_ENTERPRISE")
	if isEnterprise == "" {
		t.Skip("TF_ACC_ENTERPRISE is not set, test is applicable only for Enterprise version of Vault")
	}
}
```

website/docs/generated/datasources/transform/decode/role_name.md
```diff
@@ -0,0 +1,52 @@
+---
```
pcman312, line 1: I feel like these should be moved out of the `generated/` folder. The URL will have this path and the docs are currently generate-once with manual manipulation after that. What do you think?

codegen/README.md
```diff
@@ -0,0 +1,33 @@
...
+- If it doesn't:
+  - Make a new Vault dev binary from the `vault-enterprise` repo.
+  - Export a Vault license that includes the `transform` secrets engine: `export VAULT_LICENSE=foo`.
+  - In the Vault or Vault Enterprise repo, run `bash scripts/gen_openapi.sh`.
+  - Move the resulting file to `testdata/openapi.json`.
+- Add the 1 endpoint you wish to generate to `codegen/endpoint_registry.go`.
+- From the home directory of `terraform-provider-vault`, run:
+```
+make generate
+make fmt
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Oh, yes, I suppose so. :-)

https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/8b8d0b00f4c4154635f80621ac59cc8a7e655723

codegen/generate.go
```diff
@@ -33,18 +33,22 @@ func Run(logger hclog.Logger, paths map[string]*framework.OASPathItem) error {
 		templateHandler: h,
 	}
 	createdCount := 0
-	for endpoint, templateType := range endpointRegistry {
-		logger.Info(fmt.Sprintf("generating %s for %s\n", templateType.String(), endpoint))
-		if err := fCreator.GenerateCode(endpoint, paths[endpoint], templateType); err != nil {
+	for endpoint, addedInfo := range endpointRegistry {
+		logger.Info(fmt.Sprintf("generating %s for %s\n", addedInfo.TemplateType.String(), endpoint))
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Good eye!

https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/689b6cbb9ef1ab46c2bb77f946af6cb5ce971d32

codegen/generate.go
```diff
@@ -58,23 +62,39 @@ type fileCreator struct {
...
-// TODO in a separate PR - add a GenerateDoc method.
+// GenerateDoc is exported to indicate it's intended to be directly used.
+func (c *fileCreator) GenerateDoc(endpoint string, endpointInfo *framework.OASPathItem, addedInfo *additionalInfo) error {
+	pathToFile, err := docFilePath(addedInfo.TemplateType, endpoint)
+	if err != nil {
+		return err
+	}
+	// If the doc already exists, no need to generate a new one, especially
+	// since these get hand-edited after being first created.
+	if _, err := os.Stat(pathToFile); err == nil {
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Yes, good point, I was torn about that too. I thought about it and ended up returning a bool to indicate whether something was created, and I updated the script to give better output around what happened.

https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/689b6cbb9ef1ab46c2bb77f946af6cb5ce971d32

codegen/generate.go
```diff
@@ -58,23 +62,39 @@ type fileCreator struct {
...
+		// The file already exists, nothing further to do here.
+		return nil
+	}
+	// From here on, addedInfo will be used to select the template to
+	// use. Since we want it to be for docs, we need to update that now.
+	addedInfo.TemplateType = templateTypeDoc
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Hm, I think this is a good point. It made me re-think my object model. Essentially, the type of file you want to generate is _separate_ from the type of template to be used. Terraform has two main file types being generated, resources and data sources. But there are three main types of templates - resources, data sources, and docs.

To improve it, I decided to add a new `tfType` field for people to describe the kind of file they want to generated, and that is never overwritten. Then I also kept the template types, and passed it in as you suggested. 

Hopefully this makes more sense but let me know what you think. https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/e331011829da696d7e80e1f58cbac17ef5fa3936

website/docs/generated/datasources/transform/decode/role_name.md
tyrannosaurus-becks: Oh - the docs aren't generate-once. There are two main reasons it's like this:

1. To flag to folks that these are generated files and shouldn't be manually edited, because there are a lot of community contributions to the Terraform Vault Provider.
2. Each time a new OpenAPI doc is added and `$ make generate` is run, if an endpoint has changed, the new file will overwrite the previous one. Then there'll be a visible diff - a parameter appeared or disappeared! Then we can both add immediate support for the current state, and it flags us to begin [this process](https://www.terraform.io/docs/extend/best-practices/deprecations.html). 

On handling deprecations, renames, removals, etc., I think where the file should live during that process is currently TBD but can be determined as needed moving forward. One option would be it's left in place, and manually edited, but future additional changes are easy to see from the endpoint. Another option is to move it into the `vault` folder for manual upkeep, which might be nicer, but would have the downside of not highlighting changes.

codegen/endpoint_registry.go
```diff
@@ -1,10 +1,104 @@
...
+				},
+				Computed: true,
+			},
+		},
+	},
+}
+
+type additionalInfo struct {
+	TemplateType         templateType
+	AdditionalParameters []*templatableParam
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Hm, this would be a departure from the coding approach that's been taken by the team so far. I'm hesitant to do it because I don't want folks looking at it later and wondering what I was thinking. Rather than changing this here, can we instead take it into a team meeting for discussion before changing it?

codegen/generate.go
```diff
@@ -58,23 +62,39 @@ type fileCreator struct {
 // GenerateCode is exported because it's the only method intended to be used by
 // other objects. Unexported methods may be available to other code in this package,
 // but they're not intended to be used by anything but the fileCreator.
-func (c *fileCreator) GenerateCode(endpoint string, endpointInfo *framework.OASPathItem, tmplType templateType) error {
-	pathToFile, err := codeFilePath(tmplType, endpoint)
+func (c *fileCreator) GenerateCode(endpoint string, endpointInfo *framework.OASPathItem, addedInfo *additionalInfo) error {
```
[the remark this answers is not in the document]
tyrannosaurus-becks: Same comment as above - this would be a big departure from all the code I've seen so far at Vault, so let's hold on this until there's been a team discussion.

codegen/templates.go
tyrannosaurus-becks: Same comments as above.

codegen/templates.go
tyrannosaurus-becks: Sure. When I do that, seems like I should move the sorting stuff too, and rename it to more of a "parseParameters" function. https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/76093acf660f8b36fa7456de6ffdc6d94cfcf960

tyrannosaurus-becks: I'm not sure what you mean here, can you elaborate? They're sorted just before this so birds of a feather should be together? Not sure if that covers what you're thinking.

codegen/templates_test.go
tyrannosaurus-becks: Sure, I added tests for all of those but one. https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/8c7130421beaa54b1d6f9be2081e3cd3521b8959

The fields being passed in here originate from the OpenAPI doc and they all look like this:
```
"/transit/hash/{urlalgorithm}"
"/transit/hmac/{name}"
"/transit/hmac/{name}/{urlalgorithm}"
"/transit/keys"
"/transit/keys/{name}"
"/transit/keys/{name}/config"
"/transit/keys/{name}/rotate"
"/transit/keys/{name}/trim"
"/transit/random"
"/transit/random/{urlbytes}"
"/transit/restore"
"/transit/restore/{name}"
"/transit/rewrap/{name}"
"/transit/sign/{name}"
"/transit/sign/{name}/{urlalgorithm}"
"/transit/verify/{name}"
"/transit/verify/{name}/{urlalgorithm}"
```
So, the algorithm and tests are more than robust enough for this. If ever there were a super unexpected input, that would be caught during testing the resource and the algorithm would be updated. It's hard to imagine, though, that we'd have an endpoint with a path parameter that started with an `_`, though, so I felt comfortable not accounting for that.

generated/datasources/transform/decode/role_name_test.go
tyrannosaurus-becks: Ah! Good idea! https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/8b3e3927a8532c525c932692981269112571c55c

codegen/endpoint_registry.go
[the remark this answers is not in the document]
tyrannosaurus-becks: There wasn't a finish on our discussion, but I don't want to hold this up. I think this aligns with the [Pass Values](https://github.com/golang/go/wiki/CodeReviewComments#pass-values) principle, though it's not exactly the same, so I'll go ahead and update it.

codegen/generate.go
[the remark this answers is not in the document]
tyrannosaurus-becks: In looking at the criteria [here](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type), I think it should remain a pointer, mainly because:

>If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.

In the method `fCreator.GenerateCode(endpoint, paths[endpoint], addedInfo)`, if we added 2 more arguments, the logger and template handler, that would feel large to me because it'd be 5 arguments, and we'd be forced to wrap the line. 

However, the word "feels" is by nature still rather subjective, and the guidance there is also:

>Finally, when in doubt, use a pointer receiver.

codegen/templates.go
tyrannosaurus-becks: Marking as resolved to consolidate the conversation to https://github.com/terraform-providers/terraform-provider-vault/pull/749#discussion_r428298305.

codegen/endpoint_registry.go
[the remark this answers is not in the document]
tyrannosaurus-becks: This is done in https://github.com/terraform-providers/terraform-provider-vault/pull/749/commits/f676d9f7cd401bd3f31eb32e991f8c4d6bbf6046

codegen/templates.go
pcman312: It does! I missed that it was doing a sort before the de-dupe.

website/docs/generated/datasources/transform/decode/role_name.md
pcman312: I'm confused - isn't a given doc file protected from being generated a second time if one already exists? https://github.com/terraform-providers/terraform-provider-vault/pull/749/files#diff-7b9765054e16f7d1b8cc96e6fca7e858R94-R95

PR #753, opened 2020-05-14
jwt_auth_backend: fix validation error with unknown values

### LINKED ISSUE #718: oidc_discovery_url fails when dependencies have not been resolved beforehand
Using Terraform version 0.12.20

Given the following code:

```
module "cognito" {
  source = "./modules/cognito"

  vault_account_id      = var.vault_account_id
  google_client_id      = var.google_client_id
  google_client_secret  = var.google_client_secret
  cognito_client_domain = var.cognito_client_domain
  callback_uris         = var.redirect_uris
  region                = var.region
}


resource "vault_jwt_auth_backend" "oidc" {
  description        = "OIDC backend for vault cognito"
  path               = "oidc"
  type               = "oidc"
  oidc_client_id     = module.cognito.client_id
  oidc_client_secret = module.cognito.client_secret
  oidc_discovery_url = module.cognito.discovery_url
  default_role       = "default"
  tune {
    listing_visibility = "unauth"
    max_lease_ttl      = "48h"
    default_lease_ttl  = "12h"
  }
}

```
I get the following error:

```
Error: exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided

  on main.tf line 41, in resource "vault_jwt_auth_backend" "oidc":
  41: resource "vault_jwt_auth_backend" "oidc" {

```
It doesn't seem to happen with any of the other values, just with this one. In case it helps any, this is the output from the `cognito` module:

```
output "discovery_url"    {
  value = aws_cognito_user_pool.main.endpoint
}

```

I am not sure if I did anything wrong, but otherwise this is a very frustrating bug as it renders all the work I did modularising Cognito useless.

bendrucker:
### Community Note

This PR resolves a bug with `vault_jwt_auth_backend`, where the validation implemented via `CustomDiff` that asserts "exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided" fails when one of these values is set in configuration but not known until apply time.

It handles this by first checking [`NewValueKnown(key)`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema?tab=doc#ResourceDiff.NewValueKnown) and returning early if a value is unknown.

This will happen any time these values are sourced from another resource that has not yet been created. We encountered this when pairing an [`okta_auth_server`](https://www.terraform.io/docs/providers/okta/r/auth_server.html) with Vault, but the effect would be the same for any other provider.

Closes #718

```release-note
`vault_jwt_auth_backend`: Fix plan error when `oidc_discovery_url`, `jwks_url`, or `jwt_validation_pubkeys` is set to a value that is not known until apply time
```

```
TESTARGS="-run TestAccJWTAuthBackend_missingMandatory" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccJWTAuthBackend_missingMandatory -timeout 120m
PASS
--- PASS: TestAccJWTAuthBackend_missingMandatory (0.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	1.237s
```

### DIFF
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -124,17 +124,23 @@ func jwtAuthBackendResource() *schema.Resource {
 }
 
 func jwtCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
-	var _ interface{}
-	var discUrlExists, jwksUrlExists, jwtPubKeysExists bool
-	_, discUrlExists = d.GetOk("oidc_discovery_url")
-	_, jwksUrlExists = d.GetOk("jwks_url")
-	_, jwtPubKeysExists = d.GetOk("jwt_validation_pubkeys")
-
-	if !(discUrlExists || jwksUrlExists || jwtPubKeysExists) {
-		return errors.New("exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided")
+	attributes := []string{
+		"oidc_discovery_url",
+		"jwks_url",
+		"jwt_validation_pubkeys",
 	}
 
-	return nil
+	for _, attr := range attributes {
+		if !d.NewValueKnown(attr) {
+			return nil
+		}
+
+		if _, ok := d.GetOk(attr); ok {
+			return nil
+		}
+	}
+
+	return errors.New("exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided")
 }
 
 var (
--- a/vault/resource_jwt_auth_backend_test.go
+++ b/vault/resource_jwt_auth_backend_test.go
@@ -220,6 +220,18 @@ func TestAccJWTAuthBackend_missingMandatory(t *testing.T) {
 				Destroy:     false,
 				ExpectError: regexp.MustCompile("exactly one of oidc_discovery_url, jwks_url or jwt_validation_pubkeys should be provided"),
 			},
+			{
+				// oidc_discovery_url will be unknown until apply time
+				Config: fmt.Sprintf(`
+				resource "vault_identity_oidc_key" "key" {
+					name = "key"
+				}
+
+				resource "vault_jwt_auth_backend" "unknown" {
+					path = "%s"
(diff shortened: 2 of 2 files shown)


bendrucker:
Very open to less hacky ways to test this if that's a concern! The "correct" way I can think of to do this:

* Create a second provider in the tests, call it "test" or something. Copy/append to `testAccProviders` where needed.
* Create a "test_resource" resource or similar
* In its schema, add two fields (both `map(string)`), `input` and `outputs`. Inputs is required and user-specifiable, outputs are computed and just a copy of the inputs.

If you're interested in that I'm happy to put it in a package somewhere, for now I've just gone with a resource where the `name` becomes the `id` after `Create()` as a simple but somewhat obscure hack.

tyrannosaurus-becks:
Hi! Thanks for this PR!

Question - I don't completely understand the use case here. Would simply using a `depends_on` parameter solve the issue where the desired value hasn't been created yet?

bendrucker:
Hi and thanks for responding! I'm 95% sure `depends_on` wouldn't suffice even as a workaround. As far as I understand, `depends_on` is only implemented during applies. From the [docs](https://www.terraform.io/docs/configuration/resources.html#depends_on-explicit-resource-dependencies):

> Explicitly specifying a dependency is only necessary when a resource relies on some other resource's behavior but doesn't access any of that resource's data in its arguments.

The `vault_jwt_auth_backend` is not depending on the behavior of another resource, it is referring to its attributes directly. Terraform would wait to create the `vault_jwt_auth_backend` until the other resource has been created such that all the attributes of the auth backend are fully known. By default, the provider would handle this dependency on an unknown value without issue:

```
+ oidc_discovery_url = (known after apply)
```

The bug comes at plan time when the `CustomizeDiff` function tries to implement "at least one attribute is required" logic. For comparison, if you had a `Required: true` field in the schema and that were set to an unknown value, Terraform would accept that and render as `(known after apply)` as shown above.

PR #763, opened 2020-05-19
Fix crash in vault_jwt_auth_backend_role when bound_audiences is nil

### LINKED ISSUE #595: Crash importing vault_jwt_auth_backend_role if bound_audiences is nil
### Terraform Version
```
Terraform v0.12.12
+ provider.vault v2.5.0
```

### Affected Resource(s)
- vault_jwt_auth_backend_role

```hcl
resource "vault_jwt_auth_backend_role" "azure_default_role" {
  backend         = "oidc"
  bound_audiences = []
  role_name       = "default"
  user_claim      = "email"
  groups_claim    = "groups"
  oidc_scopes = [
    "https://graph.microsoft.com/.default",
    "email"
  ]
  allowed_redirect_uris = [
    "${local.vault_addr}/ui/vault/auth/oidc/oidc/callback",
    "http://localhost:8250/oidc/callback"
  ]

  token_policies = []
  token_ttl      = 3600
  token_max_ttl  = 86400
}
```

### Debug Output
None

### Panic Output
```
vault_jwt_auth_backend_role.azuread_default_role: Importing from ID "auth/oidc/role/default"...
vault_jwt_auth_backend_role.azuread_default_role: Import prepared!
  Prepared vault_jwt_auth_backend_role for import
vault_jwt_auth_backend_role.azuread_default_role: Refreshing state... [id=auth/oidc/role/default]

Error: rpc error: code = Unavailable desc = transport is closing


panic: interface conversion: interface {} is nil, not []interface {}
2019-11-04T09:19:38.425-0600 [DEBUG] plugin.terraform-provider-vault_v2.5.0_x4:
2019-11-04T09:19:38.425-0600 [DEBUG] plugin.terraform-provider-vault_v2.5.0_x4: goroutine 82 [running]:
2019-11-04T09:19:38.425-0600 [DEBUG] plugin.terraform-provider-vault_v2.5.0_x4: github.com/terraform-providers/terraform-provider-vault/vault.jwtAuthBackendRoleRead(0xc0002a4000, 0x1ec5aa0, 0xc0002e05b0, 0xc0002a4000, 0x0)
2019-11-04T09:19:38.425-0600 [DEBUG] plugin.terraform-provider-vault_v2.5.0_x4: 	/opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-vault/vault/resource_jwt_auth_backend_role.go:312 +0x1ea6
```

https://gist.github.com/sean-nixon/82da78bc02dcd99bf939e28808f48e06

### Expected Behavior
Successfully import resource

### Actual Behavior
Crash

### Steps to Reproduce
Create a role using the vault CLI with no bound_audiences set as is specified in the configuration guide for OIDC (https://www.vaultproject.io/docs/auth/jwt.html#oidc-configuration-troubleshooting)

```
> vault write "auth/oidc/role/default" \
    user_claim="email" \
    groups_claim="groups" \
    oidc_scopes="https://graph.microsoft.com/.default,email" \
    allowed_redirect_uris="$VAULT_ADDR/ui/vault/auth/oidc/oidc/callback" \
    allowed_redirect_uris="http://localhost:8250/oidc/callback" \
    ttl=1h

> vault read auth/oidc/role/default
Key                        Value
---                        -----
allowed_redirect_uris      [<VAULT_ADDR>/ui/vault/auth/oidc/oidc/callback http://localhost:8250/oidc/callback]
bound_audiences            [c78a2df8-3bd2-4954-af76-47d1c3c7b777]
bound_claims               <nil>
bound_subject              n/a
claim_mappings             <nil>
clock_skew_leeway          0
expiration_leeway          0
groups_claim               groups
not_before_leeway          0
oidc_scopes                [https://graph.microsoft.com/.default email]
role_type                  oidc
token_bound_cidrs          []
token_explicit_max_ttl     0s
token_max_ttl              0s
token_no_default_policy    false
token_num_uses             0
token_period               0s
token_policies             []
token_ttl                  1h
token_type                 default
ttl                        1h
user_claim                 email
verbose_oidc_logging       false
```

Try to import the resource.

```
terraform import vault_jwt_auth_backend_role.azuread_default_role "auth/oidc/role/default"
```

Updating the role to set a value for `bound_audiences` fixes the crash.

dhduvall:
Closes #595

### Community Note

```release-note
NONE
```

### DIFF
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -333,10 +333,12 @@ func jwtAuthBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	boundAuds := util.JsonStringArrayToStringArray(resp.Data["bound_audiences"].([]interface{}))
-	err = d.Set("bound_audiences", boundAuds)
-	if err != nil {
-		return fmt.Errorf("error setting bound_audiences in state: %s", err)
+	if resp.Data["bound_audiences"] != nil {
+		boundAuds := util.JsonStringArrayToStringArray(resp.Data["bound_audiences"].([]interface{}))
+		err = d.Set("bound_audiences", boundAuds)
+		if err != nil {
+			return fmt.Errorf("error setting bound_audiences in state: %s", err)
+		}
 	}
 
 	d.Set("user_claim", resp.Data["user_claim"].(string))
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -339,6 +339,8 @@ func jwtAuthBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 		if err != nil {
 			return fmt.Errorf("error setting bound_audiences in state: %s", err)
 		}
+	} else {
+		d.Set("bound_audiences", make([]string, 0))
 	}
 
 	d.Set("user_claim", resp.Data["user_claim"].(string))
--- a/vault/resource_jwt_auth_backend_role_test.go
+++ b/vault/resource_jwt_auth_backend_role_test.go
@@ -288,10 +288,6 @@ func TestAccJWTAuthBackendRoleOIDC_full(t *testing.T) {
 					resource.TestCheckResourceAttr("vault_jwt_auth_backend_role.role",
 						"bound_subject", "sl29dlldsfj3uECzsU3Sbmh0F29Fios1@client"),
 					resource.TestCheckResourceAttr("vault_jwt_auth_backend_role.role",
-						"bound_audiences.#", "1"),
-					resource.TestCheckResourceAttr("vault_jwt_auth_backend_role.role",
-						"bound_audiences.2478800941", "https://myco.test"),
-					resource.TestCheckResourceAttr("vault_jwt_auth_backend_role.role",
 						"user_claim", "https://vault/user"),
 					resource.TestCheckResourceAttr("vault_jwt_auth_backend_role.role",
 						"groups_claim", "https://vault/groups"),
@@ -610,7 +606,7 @@ resource "vault_jwt_auth_backend" "jwt" {
   oidc_client_secret = "secret"
   lifecycle {
   ignore_changes = [
-     # Ignore changes to odic_clie_secret inside the tests
+     # Ignore changes to oidc_client_secret inside the tests
      "oidc_client_secret"
     ]
   }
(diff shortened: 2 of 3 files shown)


vault/resource_jwt_auth_backend_role.go
```diff
@@ -333,10 +333,12 @@ func jwtAuthBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	boundAuds := util.JsonStringArrayToStringArray(resp.Data["bound_audiences"].([]interface{}))
-	err = d.Set("bound_audiences", boundAuds)
-	if err != nil {
-		return fmt.Errorf("error setting bound_audiences in state: %s", err)
+	if resp.Data["bound_audiences"] != nil {
+		boundAuds := util.JsonStringArrayToStringArray(resp.Data["bound_audiences"].([]interface{}))
+		err = d.Set("bound_audiences", boundAuds)
```
catsby, line 338: `d.Set("bound_audiences")` should always be called when reading the Vault role to ensure we detect drift, including if the bound audiences were completely removed:

- https://www.terraform.io/docs/extend/best-practices/detecting-drift.html

If `resp.Data["bound_audiences"]` is `nil` we should probably call `d.Set("bound_audiences",nil)`.

I see this pattern of `if resp.Data["thing"] != nil` in a few places below this as well, which makes me think we may have other issues with possibly missing drift :/

dhduvall: Thanks for the review! I agree with your suggestion, but I'd submit that because the issue is present in existing code, it's outside the scope of this change. I can file an issue capturing that, though, if you'd like. Not sure if I'll have the bandwidth to fix it, though.

catsby: I opened https://github.com/terraform-providers/terraform-provider-vault/issues/767 for the other state issues. 

For this PR, we're potentially introducing drift though because we're no longer always calling `d.Set("bound_audiences", boundAuds)`. For this PR we should add an `else` similar to `oidc_scopes`:

https://github.com/terraform-providers/terraform-provider-vault/blob/7d07921e5bdc0bce7b70f5835dcd61d3b8ad290c/vault/resource_jwt_auth_backend_role.go#L355

What do you think?

dhduvall: My personal preference is to fix just the crash bug in this PR, and not the existing drift bug; then go back and fix all the drift problems for this resource (and do an audit for the same problem in other resources; I took a quick look, and most of them seem to do the right thing). But if you'd prefer to see the crash fix and the drift fix for this one attribute to come in together, that's easy enough, and I'm happy to do it.

dhduvall: Ah, I think I see your point. In the cases where the code didn't crash, there was no drift, and thus the drift here is a regression. I'll push a fix after lunch.

catsby: >  there was no drift, and thus the drift here is a regression

Correct! I should have worded my first message to be more clear as to say that was my concern, sorry about that 😄 The change look good!

Issue #368, opened 2019-03-27
Cannot create child token without root priviliges


tylersmith34:
With the release of 1.6, we're trying to use namespaces when we connect with the Vault provider.  Our user in Vault only has access in the namespace, it does not have any root level access.  When we execute a run in Terraform Enterprise and hit our Vault Enterprise instance, we get an error about creating child tokens.  It seems that namespace support did not get extended to that level.  If we add access to the root for our user, it works.  Our Vault admins do not want to do this and gave us a short window in which this is acceptable for our testing.

### Terraform Version
```
$ terraform --version
Terraform v0.11.13

+ provider.aws v2.3.0
+ provider.template v2.1.0
+ provider.vault v1.6.0
```

### Affected Resource(s)
-vault_aws_access_credentials

### Terraform Configuration Files
```hcl
provider "vault" {
  address   = "${var.vault_address}"
  namespace = "${var.vault_namespace}"
}

data "vault_aws_access_credentials" "creds" {
  backend = "${var.vault_aws_secret_backend_path}"
  role    = "${var.vault_aws_secret_backend_role}"
}

provider "aws" {
  access_key = "${data.vault_aws_access_credentials.creds.access_key}"
  secret_key = "${data.vault_aws_access_credentials.creds.secret_key}"
  region     = "${var.aws_region}"
}
```

### Debug Output
https://gist.github.com/tylersmith34/51611f2971e3032112465be926292b3c

```
Error: Error refreshing state: 1 error(s) occurred:

* provider.vault: failed to create limited child token: Error making API request.

URL: POST https://vaulturl.company.com/csg/processing/AIS/Vault/general/prod/v2/v1/auth/token/create
Code: 403. Errors:

* permission denied
```

### Panic Output
No panic output

### Expected Behavior
I expect the Vault provider to create a child token in the namespace declared in the provider's configuration (or set through `VAULT_NAMESPACE` env var).  After creating the token, the Vault provider and the `vault_aws_access_credentials` resource should provide us access to AWS in the declared Vault role.

### Actual Behavior
The call to create a child token fails with an authorization error.

### Steps to Reproduce
1. `terraform plan`

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
Not exactly the same thing as GH-289, we're not using an admin token in Vault E.

tyrannosaurus-becks:
[...]

Issue #192, opened 2018-09-25
Vault provider should not connect to server during initialization phase


CharlieC3:
In the event where a Terraform script is designed to first deploy a Vault server, then configure it using this Vault provider, the plan creation step will fail because the Vault provider attempts to connect to the Vault server when it does not exist yet.

The Vault provider should not attempt to establish a connection with a destination Vault server during the provider's initialization phase in order to avoid a failure in the scenario where the Vault server does not yet exist because Terraform hasn't deployed it yet.

The actual line causing the issue is here:
https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/provider.go#L193

### Terraform Version
Terraform v0.11.8

### Affected Resource(s)
- All vault resources

### Terraform Configuration Files
```hcl
provider "vault" {
  address = "https://127.0.0.1:8200"
  token   = "${data.external.vault_root_token.result.data}"
}

resource "helm_release" "vault" {
  depends_on = ["helm_release.consul"]
  name       = "vault"
  repository = "${helm_repository.incubator.name}"
  chart      = "incubator/vault"
  version    = "${var.vault_chart_version}"
  namespace  = "${var.k8s_namespace}"

  values = [<<-EOF
    fullnameOverride = vault
    image:
      tag: ${var.vault_version}
    consulAgent:
      tag: ${var.consul_version}
      pullPolicy: IfNotPresent
      join: ${local.consul_url}
      gossipKeySecretName: consul-gossip-key
    vault:
      dev: false
      customSecrets:
        - secretName: ${kubernetes_secret.tls_cert_and_key.metadata.0.name}
          mountPath: /vault/tls
      extraEnv:
        - name: VAULT_API_ADDR
          value: ${local.vault_url}
        - name: VAULT_CLUSTER_ADDR
          value: ${local.vault_url}
      config:
        listener:
          tcp:
            tls_disable: false
            tls_cert_file: /vault/tls/tls.crt
            tls_key_file: /vault/tls/tls.key
        storage:
          consul:
            address: ${local.consul_url}
            path: vault
    EOF
  ]

  // Initialize and unseal Vault
  provisioner "local-exec" {
    command = "${path.module}/tf-scripts/unseal_vault.sh -k ${path.module}/kube/kubeconfig_${local.cluster_name} -m ${var.vault_secret_name}"
  }
}

resource "null_resource" "vault_port_forward" {
  depends_on = ["helm_release.vault"]

  provisioner "local-exec" {
[...]

PR #775, opened 2020-05-25
Create optional provider config setting for using or disabling the child token

### LINKED ISSUE #29: Vault provider is unable to access data with a non-root token
_This issue was originally opened by @mperriere as hashicorp/terraform#16457. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

### Terraform Version
0.10.6

```
cat main.tf
provider "vault" {
  address = "http://${var.vault_host}:${var.vault_port}"
  skip_tls_verify = true
  token = "dc985ea7-57eb-77b5-17c9-ae86fe019c82"
}

data "vault_generic_secret" "key" {
  path = "secret/dev/onenext/db_user01"
}

output pwd {
 value = "${data.vault_generic_secret.key.data["pwd"]}"
}

cat variable.tf
variable vault_host {
 #default="consuldev.pipeline.aws"
 default="10.196.14.160"
}
variable vault_port {
 default="8200"
}
variable env {
 default="dev"
}
variable project_name {
 default="onenext"
}

```

### Debug Output
terraform-0.10.6 apply
data.external.read_token: Refreshing state...
data.external.fetch_token: Refreshing state...
Error refreshing state: 1 error(s) occurred:

* provider.vault: failed to create limited child token: Error making API request.

URL: POST http://10.196.14.160:8200/v1/auth/token/create
Code: 403. Errors:

* permission denied

### Expected Behavior
I'm trying to create a policy based for every project and env: we need a token for every project, in every environnment (DEV, PREP, PROD).
Then i attach a token to it, write a secret inside the path defined in the policy, and finally i'm trying to read it from Terraform. The token should allow read access to the credentials, but it does not.

the rules are defined in a ansible playbook (outside terraform scope)
      path "secret/{{env}}/{{project_name}}/*" {
        capabilities = ["create", "read", "update", "delete", "list"]
      }
      path "secret/{{env}}/{{project_name}}" {
        capabilities = ["list"]
      }
      path "auth/token/lookup-self" {
        capabilities = ["read"]
      }

Next step, i'm creating a token attached to that policy:
(sorry, ansible code too):
- hashivault_token_create:
    display_name: "{{project_name}}-{{env}}"
    policies: ["{{project_name}}-{{env}}"]
    renewable: True
    token: "{{root_token}}"

This token is available, and can be queried on the Vault server:
VAULT_TOKEN=dc985ea7-57eb-77b5-17c9-ae86fe019c82 vault token-lookup --address=http://127.0.0.1:8200
Key                     Value
---                     -----
accessor                ec4ccd86-c6b9-b0e5-b337-5504ced231cd
creation_time           1508927130
creation_ttl            2764800
display_name            token-onenext-dev
explicit_max_ttl        0
id                      dc985ea7-57eb-77b5-17c9-ae86fe019c82
meta                    <nil>
num_uses                0
orphan                  false
path                    auth/token/create
policies                [default onenext-dev]
renewable               true
ttl                     2761029

Next step. i'm writing a secret in the policy path. No issue there.

Now i'm trying to read it from the vault cli: this is ok, i can read the secret. This proves that the secret, policy and token are correctly defined:

 VAULT_TOKEN=dc985ea7-57eb-77b5-17c9-ae86fe019c82 vault read --address=http://127.0.0.1:8200/ secret/dev/onenext/db_user01
Key                     Value
---                     -----
refresh_interval        768h0m0s
pwd                     tai.s0dohM5r

Now i'm trying to do the same thing with terraform.
I get that error:
* provider.vault: failed to create limited child token: Error making API request.
URL: POST http://10.196.14.160:8200/v1/auth/token/create
Code: 403. Errors:

PS: if i switch back to the root token in the provider vault definition, i can access to the secret.

It  seems that terraform is trying to create a temporary token instead of using the one newly created to just read the secret key with the provided token.

Perhaps i did not understand the documentation? I did not find full usage examples on the internet, i don't know what's wrong here.

### Steps to Reproduce
Please list the full steps required to reproduce the issue, for example:
2. `terraform apply`

### LINKED ISSUE #722: Feature Request: add provider argument `use_child_token` to control child token creation

tstraley:
This adds a new (often requested) vault provider configuration setting which can allow users to disable the creation of the intermediate child token (strongly discouraged, but made available for the cases that need it).

```
provider "vault" {
    address = "https://vault.testdomain"
    create_intermediate_child_token = false
}
```

(Open to changing the name of this config item, or related env var, but I like to be verbose with settings that typically shouldn't be adjusted without reading the documentation).

### Community Note

Closes #29 
Closes #722 
Relates #550 
Relates #368 
Relates #192 

Relates to the following as well:
https://github.com/hashicorp/terraform/issues/16457
https://github.com/hashicorp/terraform/pull/14839
https://groups.google.com/forum/#!topic/terraform-tool/wtlLrKVQlAo

```release-note
Add `create_intermediate_child_token` optional provider configuration setting to control the use of ephemeral child tokens for use by terraform.
```

```
$ make testacc TESTARGS='-run TestAccChildToken'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccChildToken -timeout 120m
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.012s [no tests to run]
=== RUN   TestAccChildToken
--- PASS: TestAccChildToken (1.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	1.168s

```

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -138,6 +138,15 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_VERIFY", false),
 				Description: "Set this to true only if the target Vault server is an insecure development instance.",
 			},
+			"create_intermediate_child_token": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("TERRAFORM_VAULT_CREATE_CHILD_TOKEN", true),
+
+				// Setting to false will cause max_lease_ttl_seconds and token_name to be ignored (not used).
+				// Note that this is strongly discouraged due to the potential of exposing sensitive secret data.
+				Description: "Set this to false to prevent the creation and use of ephemeral child token used by this provider.",
+			},
 			"max_lease_ttl_seconds": {
 				Type:     schema.TypeInt,
 				Optional: true,
@@ -148,7 +157,7 @@ func Provider() *schema.Provider {
 				// after Terraform has finished running.
 				DefaultFunc: schema.EnvDefaultFunc("TERRAFORM_VAULT_MAX_TTL", 1200),
 
-				Description: "Maximum TTL for secret leases requested by this provider",
+				Description: "Maximum TTL for secret leases requested by this provider.",
 			},
 			"max_retries": {
 				Type:     schema.TypeInt,
@@ -161,7 +170,7 @@ func Provider() *schema.Provider {
 				Type:        schema.TypeString,
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_NAMESPACE", ""),
-				Description: "The namespace to use. Available only for Vault Enterprise",
+				Description: "The namespace to use. Available only for Vault Enterprise.",
 			},
 			"headers": {
 				Type:        schema.TypeList,
@@ -706,6 +715,62 @@ func providerToken(d *schema.ResourceData) (string, error) {
 	return strings.TrimSpace(token), nil
 }
 
+func providerChildToken(d *schema.ResourceData, c interface{}) error {
+	client := c.(*api.Client)
+
+	tokenName := d.Get("token_name").(string)
+	if tokenName == "" {
+		tokenName = "terraform"
+	}
+
+	// In order to enforce our relatively-short lease TTL, we derive a
+	// temporary child token that inherits all of the policies of the
+	// token we were given but expires after max_lease_ttl_seconds.
+	//
+	// The intent here is that Terraform will need to re-fetch any
+	// secrets on each run and so we limit the exposure risk of secrets
+	// that end up stored in the Terraform state, assuming that they are
+	// credentials that Vault is able to revoke.
+	//
+	// Caution is still required with state files since not all secrets
+	// can explicitly be revoked, and this limited scope won't apply to
+	// any secrets that are *written* by Terraform to Vault.
+
+	// Set the namespace to the token's namespace only for the
+	// child token creation
+	tokenInfo, err := client.Auth().Token().LookupSelf()
+	if err != nil {
+		return err
+	}
+	if tokenNamespaceRaw, ok := tokenInfo.Data["namespace_path"]; ok {
+		tokenNamespace := tokenNamespaceRaw.(string)
+		if tokenNamespace != "" {
+			client.SetNamespace(tokenNamespace)
+		}
+	}
+
+	renewable := false
+	childTokenLease, err := client.Auth().Token().Create(&api.TokenCreateRequest{
+		DisplayName:    tokenName,
+		TTL:            fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
+		ExplicitMaxTTL: fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
+		Renewable:      &renewable,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create limited child token: %s", err)
+	}
+
+	childToken := childTokenLease.Auth.ClientToken
+	policies := childTokenLease.Auth.Policies
+
+	log.Printf("[INFO] Using Vault token with the following policies: %s", strings.Join(policies, ", "))
+
+	// Set the token to the generated child token
+	client.SetToken(childToken)
+
+	return nil
+}
+
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	clientConfig := api.DefaultConfig()
 	addr := d.Get("address").(string)
@@ -810,55 +875,13 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, errors.New("no vault token found")
 	}
 
-	tokenName := d.Get("token_name").(string)
-	if tokenName == "" {
-		tokenName = "terraform"
-	}
-
-	// In order to enforce our relatively-short lease TTL, we derive a
-	// temporary child token that inherits all of the policies of the
-	// token we were given but expires after max_lease_ttl_seconds.
-	//
-	// The intent here is that Terraform will need to re-fetch any
-	// secrets on each run and so we limit the exposure risk of secrets
-	// that end up stored in the Terraform state, assuming that they are
-	// credentials that Vault is able to revoke.
-	//
-	// Caution is still required with state files since not all secrets
-	// can explicitly be revoked, and this limited scope won't apply to
-	// any secrets that are *written* by Terraform to Vault.
-
-	// Set the namespace to the token's namespace only for the
-	// child token creation
-	tokenInfo, err := client.Auth().Token().LookupSelf()
-	if err != nil {
-		return nil, err
-	}
-	if tokenNamespaceRaw, ok := tokenInfo.Data["namespace_path"]; ok {
-		tokenNamespace := tokenNamespaceRaw.(string)
-		if tokenNamespace != "" {
-			client.SetNamespace(tokenNamespace)
+	createChild := d.Get("create_intermediate_child_token").(bool)
+	if createChild {
+		err := providerChildToken(d, client)
+		if err != nil {
+			return nil, err
 		}
 	}
-
-	renewable := false
-	childTokenLease, err := client.Auth().Token().Create(&api.TokenCreateRequest{
-		DisplayName:    tokenName,
-		TTL:            fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		ExplicitMaxTTL: fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		Renewable:      &renewable,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create limited child token: %s", err)
-	}
-
-	childToken := childTokenLease.Auth.ClientToken
-	policies := childTokenLease.Auth.Policies
-
-	log.Printf("[INFO] Using Vault token with the following policies: %s", strings.Join(policies, ", "))
-
-	// Set the token to the generated child token
-	client.SetToken(childToken)
 
 	// Set the namespace to the requested namespace, if provided
 	namespace := d.Get("namespace").(string)
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -74,7 +74,7 @@ func Provider() *schema.Provider {
 			"skip_child_token": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_CHILD_TOKEN", false),
+				DefaultFunc: schema.EnvDefaultFunc("TERRAFORM_VAULT_SKIP_CHILD_TOKEN", false),
 
 				// Setting to true will cause max_lease_ttl_seconds and token_name to be ignored (not used).
 				// Note that this is strongly discouraged due to the potential of exposing sensitive secret data.
@@ -715,62 +715,6 @@ func providerToken(d *schema.ResourceData) (string, error) {
 	return strings.TrimSpace(token), nil
 }
 
-func providerChildToken(d *schema.ResourceData, c interface{}) error {
-	client := c.(*api.Client)
-
-	tokenName := d.Get("token_name").(string)
-	if tokenName == "" {
-		tokenName = "terraform"
-	}
-
-	// In order to enforce our relatively-short lease TTL, we derive a
-	// temporary child token that inherits all of the policies of the
-	// token we were given but expires after max_lease_ttl_seconds.
-	//
-	// The intent here is that Terraform will need to re-fetch any
-	// secrets on each run and so we limit the exposure risk of secrets
-	// that end up stored in the Terraform state, assuming that they are
-	// credentials that Vault is able to revoke.
-	//
-	// Caution is still required with state files since not all secrets
-	// can explicitly be revoked, and this limited scope won't apply to
-	// any secrets that are *written* by Terraform to Vault.
-
-	// Set the namespace to the token's namespace only for the
-	// child token creation
-	tokenInfo, err := client.Auth().Token().LookupSelf()
-	if err != nil {
-		return err
-	}
-	if tokenNamespaceRaw, ok := tokenInfo.Data["namespace_path"]; ok {
-		tokenNamespace := tokenNamespaceRaw.(string)
-		if tokenNamespace != "" {
-			client.SetNamespace(tokenNamespace)
-		}
-	}
-
-	renewable := false
-	childTokenLease, err := client.Auth().Token().Create(&api.TokenCreateRequest{
-		DisplayName:    tokenName,
-		TTL:            fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		ExplicitMaxTTL: fmt.Sprintf("%ds", d.Get("max_lease_ttl_seconds").(int)),
-		Renewable:      &renewable,
-	})
-	if err != nil {
-		return fmt.Errorf("failed to create limited child token: %s", err)
-	}
-
-	childToken := childTokenLease.Auth.ClientToken
-	policies := childTokenLease.Auth.Policies
-
-	log.Printf("[INFO] Using Vault token with the following policies: %s", strings.Join(policies, ", "))
-
-	// Set the token to the generated child token
-	client.SetToken(childToken)
-
-	return nil
-}
-
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	clientConfig := api.DefaultConfig()
 	addr := d.Get("address").(string)
@@ -877,7 +821,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	skipChildToken := d.Get("skip_child_token").(bool)
 	if !skipChildToken {
-		err := providerChildToken(d, client)
+		err := setChildToken(d, client)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


nikolaishields:
@tstraley Thankyou so much for the contribution! looks like the only thing holding this back is for you to sign the CLA as mentioned above by @hashicorp-cla. Don't want to be a bother, but me and my team could really use this once its merged upstream.

nikolaishields:
After further investigation it appears that @tstraley has been largely inactive on github for 6+ months, with the only blocker being that they have not accepted the hashicorp CLA. I will be reopening this pull request with no change in content other than accepting the CLA. @tstraley Thank you for your contributions, I in no way intend to take credit for the work you've performed, and only want this feature for the sake of the community.

tstraley:
@nikolaishields - thanks for the bumps, this had fallen way off my radar. I am now seeking approval from my employer's license review board before I can accept the hashicorp CLA and get this moved on.

tstraley:
Can I please get reviews from the maintainers of this project? I believe this is ready for merge (long overdue). Thanks :)

tstraley:
@catsby @tomhjp @tvoran @jasonodonnell - is this something any of you could look at?

matttrach:
I think this could resolve several other feature requests, what do you think @tyrannosaurus-becks ?

matttrach:
As for the security concern, in many use cases an encrypted Terraform backend could mitigate a lot of the risk.

tyrannosaurus-becks:
Hi @matttrach , I'm no longer involved with this project and have moved on from HashiCorp.

matttrach:
We will miss you @tyrannosaurus-becks !
@jasonodonnell would you mind taking a look?

surendarkumavat:
@tstraley Many Thanks for your contribution on this. With the remote disk encrypted backend becoming the norm, this has become a much awaited feature. @maintainers any updates on the ETA?

rmgpinto:
Can this me merged and released please?

rmgpinto:
Or is there a way to merge this change locally and replace the hashicorp provider?

rmgpinto:
@jasonodonnell can this be approved please? It's just a small change on 3 files.

jasonodonnell:
Hi @rmgpinto, we're still discussing the security implications of this change internally and haven't reached a decision yet. This PR is high quality and it looks good from a code perspective, but the exposure of Vault tokens via Terraform state files and logs are definitely concerning (even when the defaults are security focused).

rmgpinto:
Thanks for the quick feedback!

alan2wong:
@jasonodonnell, could you provide some insight as to what the security implications are for this particular PR?

tstraley:
@jasonodonnell giving this a bump.

I want to call out that all of the security considerations are well documented and this isn't something that users would use unless they need it and understand those risks. Having the option available has been a major request for many years now, and I'm not hearing any reason this should be held up further.

Please let me know if you'd like me to resolve the new merge conflicts and re-push so this can finally land.

janmaghuyop:
@tstraley thank you for this. Just commenting here for an update, so I could just `terraform apply --auto-approve` all the way when approved.

benashz:
@tstraley  we are taking another look at this PR. Would you mind rebasing it off of main?

Thanks,

Ben

tstraley:
@benashz - just rebased, resolved conflicts, and pushed updated version up. Please have a look and let me know if you'd like any changes.

Happy :turkey:

vault/provider_test.go
```diff
@@ -677,19 +754,38 @@ func testTokenName_check(expectedTokenName string) resource.TestCheckFunc {
...
-			return fmt.Errorf("token name [%s] expected, but got [%s]", expectedTokenName, tokenName)
+		if actualValue != expectedValue {
+			return fmt.Errorf("%s [%s] expected, but got [%s]", attrName, expectedValue, actualValue)
 		}
 
 		return nil
 	}
 }
 
+func checkUsedToken(expectChildToken bool) resource.TestCheckFunc {
```
benashz, line 770: I think this logic could probably be moved to the test, which would then just call `checkSelfToken()`.

website/docs/index.html.markdown
```diff
@@ -139,6 +139,19 @@ variables in order to keep credential information out of the configuration.
   that Terraform can be tricked into writing secrets to a server controlled
   by an intruder. May be set via the `VAULT_SKIP_VERIFY` environment variable.
 
+* `create_intermediate_child_token` - (Optional) Set this to `false` to disable
```
benashz, line 142: For extra clarity it may be a good idea to update the `token` documentation which discusses "child" token creation.

website/docs/index.html.markdown
```diff
@@ -139,6 +139,19 @@ variables in order to keep credential information out of the configuration.
...
+  creation of an intermediate ephemeral Vault token for Terraform to
+  use. This is strongly discouraged in most cases and environments because it
+  can result in the provided Vault token being exposed by Terraform's output.
```
benashz, line 145: This is only in the case when `TF_LOG` is set to `debug`. We should probably mention that here.

vault/provider_test.go
```diff
@@ -618,54 +618,131 @@ func TestAccTokenName(t *testing.T) {
...
+							if err != nil {
+								t.Fatal(err)
+							}
+						}
+					},
+					Config: testProviderConfig(test.TestChildTokenSchema, `create_intermediate_child_token = `+test.ChildTokenSchema),
+					Check:  checkUsedToken(test.ExpectChildToken),
+				},
+			},
+			CheckDestroy: testEnvCleanup("TERRAFORM_VAULT_CREATE_CHILD_TOKEN"),
```
benashz, line 708: It looks like `CheckDestroy` may be being misused here. According to the [docs](https://github.com/hashicorp/terraform-plugin-sdk/blob/76b3563a7afe40aa29d08d1661c5502b59e3811a/helper/resource/testing.go#L341-L342) it is used to ensure that the all the provisioned resources have been properly removed.

In addition `CheckDestroy` is called conditionally
[here](https://github.com/hashicorp/terraform-plugin-sdk/blob/76b3563a7afe40aa29d08d1661c5502b59e3811a/helper/resource/testing_new.go#L67-L72), so there is no guarantee that the environment will be properly cleaned. up. This could pollute the environment of other tests. I suggest adding
```
defer testEnvCleanup("TERRAFORM_VAULT_CREATE_CHILD_TOKEN")
```
 to the test function's scope.

vault/provider_test.go
```diff
@@ -618,54 +618,131 @@ func TestAccTokenName(t *testing.T) {
...
 				},
 			},
+			CheckDestroy: testEnvCleanup("VAULT_TOKEN_NAME"),
+		})
+	}
+}
+
+func TestAccChildToken(t *testing.T) {
+
+	tests := []struct {
```
benashz, line 632: I think that we probably only need the `ExpectChildToken` and `TestChildTokenEnv` fields, all of the others can be dropped.

You can use `strconv.FormatBool(test.ExpectChildToken)` for use in the string context.

See:  https://pkg.go.dev/strconv#FormatBool for more details

benashz (changes requested):
This is looking pretty good so far. I have provided some initial feedback.

Thanks,

Ben

vault/provider.go
```diff
@@ -138,6 +138,15 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_VERIFY", false),
 				Description: "Set this to true only if the target Vault server is an insecure development instance.",
 			},
+			"create_intermediate_child_token": {
```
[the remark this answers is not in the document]
tstraley: moved and renamed `skip_child_token`

vault/provider.go
```diff
@@ -138,6 +138,15 @@ func Provider() *schema.Provider {
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("TERRAFORM_VAULT_CREATE_CHILD_TOKEN", true),
```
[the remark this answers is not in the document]
tstraley: renamed to `VAULT_SKIP_CHILD_TOKEN`

vault/provider.go
```diff
@@ -706,6 +715,62 @@ func providerToken(d *schema.ResourceData) (string, error) {
 	return strings.TrimSpace(token), nil
 }
 
+func providerChildToken(d *schema.ResourceData, c interface{}) error {
```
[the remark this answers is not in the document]
tstraley: works for me. Done and done.

vault/provider_test.go
```diff
@@ -618,54 +618,131 @@ func TestAccTokenName(t *testing.T) {
...
-	}
 	`, headerName, headerValue)
-	return providerConfig
+	return testProviderConfig(true, providerConfig)
 }
 
 // Using the data lookup generic_secret to inspect used token
 // by terraform (this enables check of token name)
-func testTokenNameConfig(tokenNameSchema bool, tokenName string) string {
+func testProviderConfig(includeProviderConfig bool, config string) string {
```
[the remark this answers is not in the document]
tstraley: > Please update the documentation to match.

Can you be more clear about what documentation needs updating here? The comment above this function is still 100% accurate, and the change to make this more generic didn't affect anything else.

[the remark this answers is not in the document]
tstraley: as for your second comment:

There are several tests that call this function. One of those tests always includes a header provider config, the others will only want the provider configuration if the test is specifically testing the provider schema.

To be fair, I think I agree that simply passing in an empty provider config string would be more logical, and possibly nicer within this function, but it would require moving that specific logic (to determine if the provider config should be an empty string or actually contain the schema under test) elsewhere, and then we're back to this same conditional. If you would prefer I separate this into two functions, one for generating the Terraform config string and one for evaluating a conditional to decide if the provider portion of that Terraform config should including something or not, let me know, but I've left this as-is for now, since it doesn't really make sense to me.

In case the confusion you had was around the naming of this function (eg. why does `testProviderConfig` sometimes not include the "Provider" portion of the config and only a data resource?) I changed the name to `getTerraformConfig` which reduces that confusion at the expense of being a bit generic.

vault/provider_test.go
```diff
@@ -618,54 +618,131 @@ func TestAccTokenName(t *testing.T) {
...
+	if includeProviderConfig {
 		testConfig = providerConfig + dataConfig
 	} else {
 		testConfig = dataConfig
 	}
 	return testConfig
 }
 
-func testTokenName_check(expectedTokenName string) resource.TestCheckFunc {
+func checkTestSecretData(attrName string, expectedValue string) resource.TestCheckFunc {
```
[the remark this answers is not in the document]
tstraley: I'm not sure I agree with your reasoning (functions should be able to describe what they do with non-opaque names), but I changed to `checkSelfToken` as that seems reasonable to me.

vault/provider_test.go
tstraley: I'm not sure what you are asking for here. Can you elaborate, please?

In one interpretation I take your meaning to be that this logic can live in `CheckSelfToken` directly, but because that function is used by other tests that need to check their own specific attributes with their own expectations, this conditional based on whether we are using the child token or not doesn't apply to those tests. In another interpretation I think you may be asking for this to be an anonymous function within either the test case struct, or the `Check` of the test step, but I would argue using an anonymous function in that setting is less clear and readable.

Let me know if I'm missing your meaning.

website/docs/index.html.markdown
tstraley: updated :+1:

website/docs/index.html.markdown
tstraley: Done.

vault/provider_test.go
tstraley: Thanks, that's a fair call. Moved this into the test functions as a direct env var unset.

vault/provider_test.go
tstraley: I don't think so.
There are two configuration methods being tested in the parameterization: the schema (provider config block in .tf file) and/or the environment variables that manage the same configuration. Because we want to be able to test them independently as well as together, we need a way to state whether we want to **use** the schema, the env var, both, or neither.

I have updated the names of these fields to be more clear that they set whether to **use** a particular config method, and then assign the `skip_child_token` setting separately. Please note that each of these fields is independent and not always the same bool as any other field.

vault/provider_test.go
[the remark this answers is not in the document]
benashz: The extra boolean is argument is redundant with the config string.

`if config != "" {...}`

vault/provider_test.go
benashz: Rather than creating a new function call `checkSelfToken()` directly with the source attribute and the expected value.

tstraley: Could you provide some example code to illustrate how you expect this to look and work please? I must be missing something. Right now the attribute and expected value are based on whether we expect a child token should have been created or not; thus this if-else case. Since we do not know how to call `checkSelfToken` without switching on this condition, this logic needs to occur somewhere, rather than calling "directly" to the `checkSelfToken` function.

vault/provider_test.go
```diff
@@ -628,44 +710,34 @@ func TestAccTokenName(t *testing.T) {
...
 
 // Using the data lookup generic_secret to inspect used token
 // by terraform (this enables check of token name)
-func testTokenNameConfig(tokenNameSchema bool, tokenName string) string {
-	testConfig := ""
-	providerConfig := `
-provider "vault" {
-    token_name = "` + tokenName + `"
-}`
+func getTerraformConfig(includeProviderConfig bool, config string) string {
```
benashz, line 724:
```suggestion
func testProviderConfig(includeProviderConfig bool, config string) string {
```

benashz (changes requested):
Looking good! Just a few minor suggestions.

vault/provider_test.go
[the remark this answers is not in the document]
tstraley: Take a look at line 621: https://github.com/hashicorp/terraform-provider-vault/pull/775/files#diff-96e3c1e3f2b69ac273388431a9cebb277d7a5d596691beb0a32b5821c5d6b063R621
and 703: https://github.com/hashicorp/terraform-provider-vault/pull/775/files#diff-96e3c1e3f2b69ac273388431a9cebb277d7a5d596691beb0a32b5821c5d6b063R703

In both of those calls, config is never `""`. So this is NOT redundant.

I can refactor this (and the existing tests that use it) even further if you only want this function to have one argument, but to keep it clean and readable, I will probably need to add 2 additional functions then call this `getTerraformConfig` in order to keep each test's logic isolated. Is that what you are asking for?

vault/provider.go
```diff
@@ -71,6 +71,15 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_TOKEN_NAME", ""),
 				Description: "Token name to use for creating the Vault child token.",
 			},
+			"skip_child_token": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_CHILD_TOKEN", false),
```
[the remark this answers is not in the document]
tstraley: The other vars don't have `TERRAFORM_` prefix either. Should this really have it?

vault/provider_test.go
tstraley: This simply followed a pattern already set by other tests in here like the existing `TestAccTokenName`. I'm certainly not going to remove valid test cases at this point - more coverage is better.

vault/provider.go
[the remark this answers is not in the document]
benashz: Variable names with the `VAULT_` prefix are owned by Vault itself e.g. `VAULT_ADDR`, `VAULT_TOKEN`.  The variable in question is provider specific, and we wouldn't want it to potentially affect the Vault configuration in any way.

vault/provider_test.go
benashz: I was suggesting that we move `checkUsedToken` to be a closure within the related test and renaming it to `checkTokenUsed`:

```
@@ -628,6 +629,19 @@ func TestAccTokenName(t *testing.T) {
 
 func TestAccChildToken(t *testing.T) {
        defer os.Unsetenv("VAULT_SKIP_CHILD_TOKEN")
+
+       checkTokenUsed := func(expectChildToken bool) resource.TestCheckFunc {
+               if expectChildToken {
+                       // If the default child token was created, we expect the token
+                       // used by the provider was named the default "token-terraform"
+                       return checkSelfToken("display_name", "token-terraform")
+               } else {
+                       // If the child token setting was disabled, the used token
+                       // should match the user-provided VAULT_TOKEN
+                       return checkSelfToken("id", os.Getenv("VAULT_TOKEN"))
+               }
+       }
+
        tests := []struct {
                skipChildTokenEnv    string
                useChildTokenEnv     bool
@@ -701,7 +715,7 @@ func TestAccChildToken(t *testing.T) {
                                                }
                                        },
                                        Config: getTerraformConfig(test.useChildTokenSchema, `skip_child_token = `+test.skipChildTokenSchema),
-                                       Check:  checkUsedToken(test.expectChildToken),
+                                       Check:  checkTokenUsed(test.expectChildToken),
                                },
                        },
                })
@@ -762,18 +776,6 @@ func checkSelfToken(attrName string, expectedValue string) resource.TestCheckFun
        }
 }
 
-func checkUsedToken(expectChildToken bool) resource.TestCheckFunc {
-       if expectChildToken {
-               // If the default child token was created, we expect the token
-               // used by the provider was named the default "token-terraform"
-               return checkSelfToken("display_name", "token-terraform")
-       } else {
-               // If the child token setting was disabled, the used token
-               // should match the user-provided VAULT_TOKEN
-               return checkSelfToken("id", os.Getenv("VAULT_TOKEN"))
-       }
-}
-
```

vault/provider_test.go
benashz: Agreed, it's fine to keep it as-is for now.

vault/provider_test.go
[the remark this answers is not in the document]
benashz: Thanks, but I think it is okay to keep things as they are, but `getTerraformConfig` should be renamed as suggested earlier.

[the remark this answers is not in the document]
benashz: Thanks, but I think it is okay to keep things as they are, but `getTerraformConfig` should be renamed as suggested earlier.

vault/provider_test.go
tstraley: I'm happy with that. Done.

vault/provider.go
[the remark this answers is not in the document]
tstraley: Got it, thanks for the explanation. Fixed.

benashz (changes requested):

benashz (approved):
So we just need to address the documentation comment, then +1.

benashz:
Thanks again @tstraley for all the effort on this PR. It's much appreciated!

PR #780, opened 2020-05-27
Add resource that supports importing jwt config

### LINKED ISSUE #295: vault_jwt_auth_backend doesn't support importing
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
v0.11.11

### Affected Resource(s)
- vault_jwt_auth_backend

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "vault_jwt_auth_backend" "sl_iam_jwt_auth" {
  path               = "jwt"
  oidc_discovery_url = "https://iam.strln.net"
  bound_issuer       = "https://iam.strln.net"
}
```

### Expected Behavior
We should be able to do a terraform import for vault_jwt_auth_backend resources because otherwise we end up in a state where we can neither run "terraform apply" (it attempts and fails to create the backend again at the same path) nor can we import it to the terraform state easily, so an existing jwt backend puts us into a deadlock.

### Actual Behavior
terraform import fails saying that it's not supported by this resource type

Attempted to fix by adding 
```
},
```
to the vault/resource_jwt_auth_backend.go schema.Resource but this started failing with the following error on import attempts:
```
vault_jwt_auth_backend.sl_iam_jwt_auth: Importing from ID "jwt"...
vault_jwt_auth_backend.sl_iam_jwt_auth: Import complete!
  Imported vault_jwt_auth_backend (ID: jwt)
vault_jwt_auth_backend.sl_iam_jwt_auth: Refreshing state... (ID: jwt)

Error: vault_jwt_auth_backend.sl_iam_jwt_auth (import id: jwt): 1 error(s) occurred:

* import vault_jwt_auth_backend.sl_iam_jwt_auth result: jwt: import vault_jwt_auth_backend.sl_iam_jwt_auth (id: jwt): Terraform detected a resource with this ID doesn't
exist. Please verify the ID is correct. You cannot import non-existent
resources using Terraform import.
```
We've confirmed via "vault auth list" & "vault auth read /auth/jwt/config" that this backend does in fact exist but the import is failing

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
The idea to that Importer: blah code segment came from the (functional) generic auth backend resource:
https://github.com/terraform-providers/terraform-provider-vault/blob/master/vault/resource_auth_backend.go#L20-L22

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-1234

tyrannosaurus-becks:
Closes #295

This PR adds a resource that supports importing a pre-existing jwt config for the "auth/jwt/config" endpoint (as well as all other operations supported on that endpoint). Since the `vault_auth_backend` already supports importing, this means that any user will be able to import both a pre-existing jwt backend and its config.

The approach of creating a new resource was chosen because adding import support to `vault_jwt_auth_backend` was not trivial. That resource hits two combined endpoints - the auth mount endpoint and the jwt config endpoint. When adding import, it was not possible to support the "description" and "type" fields because they are not passed in with the data, nor are they both returned by any Vault endpoint. Perhaps there is some workaround that I don't know, but it seemed better to use a cleaner model of one resource per endpoint.

### DIFF
--- a/codegen/templates/resource.go.tpl
+++ b/codegen/templates/resource.go.tpl
@@ -61,7 +61,7 @@ func {{ .UpperCaseDifferentiator }}Resource() *schema.Resource {
 			{{- if .Schema.DisplayAttrs.Sensitive }}
 			Sensitive:   true,
 			{{- end }}
-			Description: "{{ .Description }}",
+			Description: `{{ .Description }}`,
 			{{- if .IsPathParam }}
 			ForceNew: true,
 			{{- end}}
@@ -77,7 +77,7 @@ func {{ .UpperCaseDifferentiator }}Resource() *schema.Resource {
 		Read:   read{{ .UpperCaseDifferentiator }}Resource,
 		Exists: resource{{ .UpperCaseDifferentiator }}Exists,
 		{{- end }}
-		{{- if .SupportsDelete }}
+		{{- if or .SupportsDelete .SupportsWrite  }}
 		Delete: delete{{ .UpperCaseDifferentiator }}Resource,
 		{{- end }}
 		Importer: &schema.ResourceImporter{
@@ -168,8 +168,8 @@ func update{{ .UpperCaseDifferentiator }}Resource(d *schema.ResourceData, meta i
 	{{- range .Parameters }}
 	{{- if not .IsPathParam }}
 	  {{- if not .Computed }}
-	if d.HasChange("{{ .Name }}") {
-		data["{{ .Name }}"] = d.Get("{{ .Name }}")
+	if raw, ok := d.GetOk("{{ .Name }}"); ok {
+		data["{{ .Name }}"] = raw
 	}
 	  {{- end }}
 	{{- end }}
@@ -196,6 +196,13 @@ func delete{{ .UpperCaseDifferentiator }}Resource(d *schema.ResourceData, meta i
 		return nil
 	}
 	log.Printf("[DEBUG] Deleted template auth backend role %q", vaultPath)
+	return nil
+}
+{{ else if .SupportsWrite }}
+func delete{{ .UpperCaseDifferentiator }}Resource(_ *schema.ResourceData, _ interface{}) error {
+	// Terraform requires the delete is implemented whenever create is implemented,
+	// but this endpoint doesn't support delete. Thus, we've simply stubbed out delete
+	// here.
 	return nil
 }
 {{ end }}
--- a/util/util.go
+++ b/util/util.go
@@ -202,13 +202,26 @@ func SliceRemoveIfPresent(list []interface{}, search interface{}) []interface{} 
 //   - parameters will include path parameters
 func ParsePath(userSuppliedPath, endpoint string, d *schema.ResourceData) string {
 	fields := strings.Split(endpoint, "/")
-	// The second field should be the one the user supplied rather
+	if fields[0] == "" {
+		// There was a leading slash that should be trimmed.
+		fields = fields[1:]
+	}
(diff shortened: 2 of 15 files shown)


codegen/templates/resource.go.tpl
```diff
@@ -168,8 +168,8 @@ func update{{ .UpperCaseDifferentiator }}Resource(d *schema.ResourceData, meta i
 	{{- range .Parameters }}
 	{{- if not .IsPathParam }}
 	  {{- if not .Computed }}
-	if d.HasChange("{{ .Name }}") {
-		data["{{ .Name }}"] = d.Get("{{ .Name }}")
+	if raw, ok := d.GetOk("{{ .Name }}"); ok {
```
tyrannosaurus-becks, line 171: This edit was due to a bug I found in the template when testing. If I sent up a field again and it hadn't changed, it wasn't passed through to Vault. This fix ensures that if it's been sent, it's passed through.

codegen/templates/resource.go.tpl
```diff
@@ -61,7 +61,7 @@ func {{ .UpperCaseDifferentiator }}Resource() *schema.Resource {
 			{{- if .Schema.DisplayAttrs.Sensitive }}
 			Sensitive:   true,
 			{{- end }}
-			Description: "{{ .Description }}",
+			Description: `{{ .Description }}`,
```
tyrannosaurus-becks, line 64: I changed the description to string literals because it turned out that some descriptions contain quotes, so if the description did as well, there were problems.

codegen/templates/resource.go.tpl
```diff
@@ -198,6 +198,13 @@ func delete{{ .UpperCaseDifferentiator }}Resource(d *schema.ResourceData, meta i
 	log.Printf("[DEBUG] Deleted template auth backend role %q", vaultPath)
 	return nil
 }
+{{ else if .SupportsWrite }}
+func delete{{ .UpperCaseDifferentiator }}Resource(_ *schema.ResourceData, _ interface{}) error {
```
tyrannosaurus-becks, line 202: I added this logic because with this resource was the first time I encountered an endpoint that supports Create and Update without supporting Delete.

util/util.go
```diff
@@ -202,13 +202,26 @@ func SliceRemoveIfPresent(list []interface{}, search interface{}) []interface{}
 //   - parameters will include path parameters
 func ParsePath(userSuppliedPath, endpoint string, d *schema.ResourceData) string {
 	fields := strings.Split(endpoint, "/")
-	// The second field should be the one the user supplied rather
+	if fields[0] == "" {
+		// There was a leading slash that should be trimmed.
+		fields = fields[1:]
+	}
+	isAuthPath := false
+	if fields[0] == "auth" {
```
tyrannosaurus-becks, line 210: The logic in this method needed to be changed to support paths that are prefixed with "auth".

vault/provider.go
```diff
@@ -249,7 +249,7 @@ var (
 			},
 		},
 		"vault_auth_backend": {
-			Resource:      authBackendResource(),
+			Resource:      AuthBackendResource(),
```
tyrannosaurus-becks, line 252: This needed to be exported so it could be used in testing.

tyrannosaurus-becks:
I'm going to go ahead and just merge this into the feature branch, then test that branch and do a PR into master to keep all this moving.

Issue #281, opened 2019-01-22
make connection url sensitive in vault_database_secret_backend_connection


eedwards-sk:
### Overview
Currently the connection string for vault_database_secret_backend_connection is printed in plaintext, which makes it very dangerous since its printed to logs by default and usually has root database credentials.

I run concourse in CI, so I make a concerted effort to minimize credential leaking, when possible. I can control access to state files, but this is currently out of my hands.

### Terraform Version
0.11.11
vault provider: 1.4.1

### Affected Resource(s)
Please list the resources as a list, for example:
- vault_database_secret_backend_connection

### Expected Behavior
```
  + vault_database_secret_backend_connection.concourse
      id:                                <computed>
      allowed_roles.#:                   "1"
      allowed_roles.0:                   "concourse"
      backend:                           "postgres"
      name:                              "concourse"
      postgresql.#:                      "1"
      postgresql.0.connection_url:       "<sensitive>"
      postgresql.0.max_open_connections: "2"
      verify_connection:                 "true"
```

### Actual Behavior
```
  + vault_database_secret_backend_connection.concourse
      id:                                <computed>
      allowed_roles.#:                   "1"
      allowed_roles.0:                   "concourse"
      backend:                           "postgres"
      name:                              "concourse"
      postgresql.#:                      "1"
      postgresql.0.connection_url:       "postgres://username-in-plaintext:password-in-plaintext@my-hostname:5432/my-db"
      postgresql.0.max_open_connections: "2"
      verify_connection:                 "true"
```

cvbarros:
Hi @eedwards-sk, there is a workaround you can use until this is addressed:

Use `vault_database_secret_backend_connection` resource with a `data` block, and inside this data block you reference variables. This trick is specially handy if `db_user` and `db_password` come instead from Vault itself as `generic_secret` (that's how we do it and they don't leak).

```hcl
variable "db_user" {}
variable "db_password" {}

locals {
  db_user       = "${var.db_user}"
  db_password     = "${var.db_password}"
}

resource "vault_database_secret_backend_connection" "mysql_aurora" {
  backend           = "${vault_mount.database_mount.path}"
  name              = "db"
  allowed_roles     = ["*"]
  verify_connection = "false"

  mysql_aurora {
    connection_url = "{{username}}:{{password}}@tcp(my-hostname)/"
  }

  data = {
    username = "${local.db_user}"
    password = "${local.db_password}"
  }
}

[...]

PR #794, opened 2020-06-12
database_secret_backend_connection: use data in example

bendrucker:
This PR emphasizes the use of the `data` attribute to substitute sensitive values. Since the example has a password in the `connection_url`, a reader might reasonably assume that it is a sensitive attribute and would be hidden in `terraform plan` and `terraform apply` output. This revises the example to make it clearer to users that they probably want to use `data`.

### Community Note

Relates to #281

```release-note
`vault_database_secret_backend_connection`: Revise example configuration in docs to avoid leaking a password in plan/apply output
```

### DIFF
--- a/website/docs/r/database_secret_backend_connection.md
+++ b/website/docs/r/database_secret_backend_connection.md
@@ -32,7 +32,13 @@ resource "vault_database_secret_backend_connection" "postgres" {
   allowed_roles = ["dev", "prod"]
 
   postgresql {
-    connection_url = "postgres://username:password@host:port/database"
+    connection_url = "postgres://{{username}}:{{password}}@host:port/database"
+  }
+
+  # persisted in state but hidden in plan/apply output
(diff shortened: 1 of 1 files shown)


reegnz:
The existance of these docs would have helped a lot before I implemented #860. :)

bendrucker:
Re-raising in hopes of a review here. I think documentation would be enough here as opposed to provider fields. But if there's no activity here after a few more weeks I'll just go ahead and close.

reegnz:
I would not close unless docs exist.

bendrucker:
Any interest in this? Since it's gone a year without acknowledgement I'm planning to close in a week so it doesn't sit in my PR list indefinitely.

reegnz:
I would definitely like to have this in the docs. @jasonodonnell @tvoran would you mind looking at this and merging this documentation improvement please?

PR #809, opened 2020-07-05
vault_ssh_secret_backend_role: add algorithm_signer argument

andreaso:
The `algorithm_signer` argument was introduced in Vault 1.4.3. See also https://github.com/hashicorp/vault/pull/9096.

### Community Note

###  Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
resource/vault_ssh_secret_backend_role: Support new `algorithm_signer` argument
```

```
$ TESTARGS="--run SSHSecretBackendRole" make testacc
...
=== RUN   TestAccSSHSecretBackendRole_basic
--- PASS: TestAccSSHSecretBackendRole_basic (0.27s)
=== RUN   TestAccSSHSecretBackendRoleOTP_basic
--- PASS: TestAccSSHSecretBackendRoleOTP_basic (0.16s)
=== RUN   TestAccSSHSecretBackendRole_import
--- PASS: TestAccSSHSecretBackendRole_import (0.18s)
...
```

### DIFF
--- a/vault/resource_ssh_secret_backend_role.go
+++ b/vault/resource_ssh_secret_backend_role.go
@@ -110,7 +110,7 @@ func sshSecretBackendRoleResource() *schema.Resource {
 			"algorithm_signer": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "ssh-rsa",
+				Computed: true,
 			},
 			"max_ttl": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


andreaso:
Hello @catsby. It was suggested in #vault-tool that I might want to ping you about this PR.

vault/resource_ssh_secret_backend_role.go
```diff
@@ -107,6 +107,11 @@ func sshSecretBackendRoleResource() *schema.Resource {
 				Type:     schema.TypeMap,
 				Optional: true,
 			},
+			"algorithm_signer": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "ssh-rsa",
```
catsby, line 113: I think we need to leave `algorithm_signer` as `Optional: true` and `Computed: true` and remove the default. The documentation on vaultproject.io says `ssh-rsa` is the default, but I believe that's not actually the case anymore. The API is returning `""` now and not returning any defaults

catsby (changes requested):
Sorry for the delay in review! I think we need to remove the `Default` at this time. Perhaps when this was originally opened the default was true, but I believe with https://github.com/hashicorp/vault/pull/9824 that's not longer the case (and the docs on vaultproject.io need to be updated, I think)

catsby:
Thank you for contributing! I made minor updates to account for the `ssh-rsa` no longer being a default, and I'm going to pull this in now. Thanks!

andreaso:
Thanks!

Any best guest on when there will be a new release of the provider?

PR #832, opened 2020-07-27
Add support for passing region information stored in vault backend to AWS Config

Valarissa:
This allows us to use non-inferrable regions with terraform and the vault provider.

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Relates OR Closes #679

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
Add support for passing region information stored in vault backend to AWS Config
```

Output from acceptance testing:
(Note: You need numerous environment variables to be set for these acceptance tests to run.
`TF_ACC=1`
`VAULT_TOKEN=????`

The following env vars must be set based on information provided by your AWS instance. In order to truly test these changes, you'll need to get Access Keys from the `us-gov-west-1` cloud, as those credentials were the ones that didn't work which prompted this change (You can also use `cn-northwest-1` to test this if you'd like). 
`AWS_ACCESS_KEY_ID=????`
`AWS_SECRET_ACCESS_KEY=????`
`AWS_DEFAULT_REGION=(cn-northwest-1|us-gov-west-1|us-east-1)`

```
$ go test -run="TestAccDataSourceAWSAccessCredentials" -v ./vault

=== RUN   TestAccDataSourceAWSAccessCredentials_basic
--- PASS: TestAccDataSourceAWSAccessCredentials_basic (54.20s)
=== RUN   TestAccDataSourceAWSAccessCredentials_sts
--- PASS: TestAccDataSourceAWSAccessCredentials_sts (1.49s)
=== RUN   TestAccDataSourceAWSAccessCredentials_sts/sts_without_role_arn
    --- PASS: TestAccDataSourceAWSAccessCredentials_sts/sts_without_role_arn (0.77s)
=== RUN   TestAccDataSourceAWSAccessCredentials_sts/sts_with_role_arn
    --- PASS: TestAccDataSourceAWSAccessCredentials_sts/sts_with_role_arn (0.72s)
PASS

Process finished with exit code 0
...
```

I'd suggest running the tests with both us-east-1 and us-gov-west-1 credentials to ensure it all works.

PR #837, opened 2020-07-31
Revert PR (#832)

catsby:
PR 832 inadvertently introduced issues when the token policy did not
have the required permissions to read the root configuration.

This reverts commit f8b83fbd30ae3934b79b1cbc833532ce99ab94e0.


BonzTM:
@catsby Even with correct permissions, the problem also exists when a region is not provided in root configuration.

catsby:
Hey @BonzTM - could you possibly share an example configuration that demonstrates what issue you're having? I'm trying to survey the types of issues present here, specifically in v2.12.0 or v2.12.2

BonzTM:
@catsby I'm sorry, I believe I was just echoing the comment in the open issue at the time.  v2.12.0 and v2.12.2 are fine.  v2.12.1 has issues when a region is not specified in the core config.  Such configurations exist when leveraging the assumed_role configuration of the auth method.

PR #855, opened 2020-08-27
Feature request 852

vsharma-ampion:
Added change to support additional parameter (local) for gcp secret engine mount. The change is related to issue 852 (https://github.com/terraform-providers/terraform-provider-vault/issues/852).


jasonodonnell (approved):
LGTM, thank you!

Issue #858, opened 2020-09-01
The gcp_auth_backend is using deprecated vault api client method.


vsharma-ampion:
Hi there,

### Terraform Version
v0.13.0

### Affected Resource(s)
- vault_gcp_auth_backend

### Expected Behavior
The source code for this resource must use the vault api method 'EnableAuthWithOptions' for enabling the gcp authentication.

### Actual Behavior
The source code for this resource is using the deprecated method 'EnableAuth' for enabling the gcp authentication.

kalafut:
Fixed in #861

PR #861, opened 2020-09-04
Feature request for issue #858

vsharma-ampion:
Added following changes:

- Added change to support additional parameter (local) for gcp auth engine mount.
-  Use non deprecated method EnableAuthWithOptions instead of the deprecated method EnableAuth on vault api.
- The change is related to issue #858


jasonodonnell (approved):
LGTM, thanks!

PR #871, opened 2020-09-25
Generated docs path update

catsby:
Update how/where our documentation generator generates the files for our generated documentation. 

The Vault Terraform provider documentation is now hosted at https://registry.terraform.io/providers/hashicorp/vault/latest/docs . The Terraform Registry no longer reads or uses `website/vault.erb` for finding and linking to documentation of resources or datasources; it now only reads the file structure and builds the sidebar itself. Because of this, our generated resources (only Transform resource/datasources at this time) are not displaying on the registry because they are in a non-standard location. 

There are two main changes here:

- Changing where **new** resources and datasources would be generated, and the naming scheme used. Primarily here we're generating things to `website/docs/r` for resources or `/website/docs/d` for datasources, instead of putting them in their own `website/docs/generated` tree. 
- Moves the existing documentation to the correct location(s)

The documentation generation does not re-generate the documentation; it only generates stubs and the actual documentation is done manually. If the file already exists, the generator does not changing anything. 

**Note:** The Terraform registry considers the `/website/docs` location for documentation to be a legacy format / location. New providers are expected to put all documentation in `/docs`, and uses `resources` and `data-sources` instead of the old style `r` and `d` folders. Migrating to the new style will come in a follow-up PR, but for now this PR only gets the transform resources and datasources into the correct location to render on https://registry.terraform.io/.

Output from codegen testing:

```
[terraform-provider-vault][generated-docs-path-update](4)$ make test TEST=./codegen TESTARGS="-count=1" 
go test -i ./codegen || exit 1
echo ./codegen | \
                xargs -t -n4 go test -count=1 -timeout=30s -parallel=4
go test -count=1 -timeout=30s -parallel=4 ./codegen
ok      github.com/terraform-providers/terraform-provider-vault/codegen 0.111s
...
```

Output from `make generate` (no changes is good, because they should already exist in the new location):

```
[terraform-provider-vault][generated-docs-path-update](4)$ make generate
result=
go run cmd/generate/main.go -openapi-doc=testdata/openapi.json
2020-09-25T17:07:04.379-0500 [INFO]  generated resource for /transform/alphabet/{name}
2020-09-25T17:07:04.379-0500 [INFO]  generated datasource for /transform/decode/{role_name}
2020-09-25T17:07:04.379-0500 [INFO]  generated datasource for /transform/encode/{role_name}
2020-09-25T17:07:04.379-0500 [INFO]  generated resource for /transform/role/{name}
2020-09-25T17:07:04.380-0500 [INFO]  generated resource for /transform/template/{name}
2020-09-25T17:07:04.380-0500 [INFO]  generated resource for /transform/transformation/{name}
2020-09-25T17:07:04.380-0500 [INFO]  generated 6 files
2020-09-25T17:07:04.380-0500 [INFO]  skipped generating 6 docs because they already existed
```


jasonodonnell (approved):
LGTM

PR #893, opened 2020-11-04
add/update docs for namespaces

catsby:
This PR expands on the existing provider documentation for [namespace][namespaces] support in Vault Enterprise. 

[namespaces]: https://www.vaultproject.io/docs/enterprise/namespaces#vault-enterprise-namespaces

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -3,13 +3,13 @@ layout: "vault"
 page_title: "Provider: Vault"
 sidebar_current: "docs-vault-index"
 description: |-
-  The Vault provider allows Terraform to read from, write to, and configure Hashicorp Vault
+  The Vault provider allows Terraform to read from, write to, and configure HashiCorp Vault
 ---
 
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -246,22 +246,181 @@ provider "vault" {
...
+default
+root
 ```
+
+### Nested Namespaces
+
+A more complex example of nested namespaces is possible using provider aliases
+and the `depends_on` configuration option. At time of writing Terraform (v0.13)
+does not support using the `depends_on` attribute in the `provider` block, which
+prevents the normal
```
yhyakuna, line 315: Incomplete sentence?

catsby: Good catch, thanks! I ended up refactoring/updating the example to not use `depends_on` at all, so this sentence kind of goes away anyway. Thank you!

Valarissa (approved):
LGTM!

PR #895, opened 2020-11-05
Add support for case_sensitive_names in the vault_ldap_auth_backend

### LINKED ISSUE #894: vault_ldap_auth_backend does not support setting case_sensitive_names
Vault resource `vault_ldap_auth_backend` does not support setting `case_sensitive_names` boolean

Now this needs to be done manual after applying

stokkie90:
Fixes #894

### DIFF
--- a/vault/resource_ldap_auth_backend.go
+++ b/vault/resource_ldap_auth_backend.go
@@ -54,6 +54,11 @@ func ldapAuthBackendResource() *schema.Resource {
 			Computed:  true,
 			Sensitive: true,
 		},
+		"case_sensitive_names": {
+			Type:     schema.TypeBool,
+			Optional: true,
+			Computed: true,
+		},
 		"userdn": {
 			Type:     schema.TypeString,
 			Optional: true,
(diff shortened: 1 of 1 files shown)


vault/resource_ldap_auth_backend.go
```diff
@@ -204,6 +209,10 @@ func ldapAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 		data["bindpass"] = v.(string)
 	}
 
+	if v, ok := d.GetOk("case_sensitive_names"); ok {
```
catsby, line 212: I _think_ this needs to be `d.GetOkExists` like `starttls` above, otherwise a value of `false` wont return `ok == true`, so we could be in a situation where we can enable this but never disable it. A default value seems in order as well, if possible

catsby (changes requested):
I think we may want to look at adding a default, using `GetOkExists`, and/or including a test that can verify we toggle this off and on again before merging. Does that sound reasonable?

vault/resource_ldap_auth_backend.go
stokkie90: I just created this mr with an idea this might solve it. I probably need some help with it. And how to test it. But will get back to this on Monday.

jasonodonnell:
```suggestion
	if v, ok := d.GetOkExists("case_sensitive_names"); ok {
```

benashz:
Closing with #1176

PR #902, opened 2020-11-10
Add Active Directory secret engine support

jasonodonnell:
This PR adds support for the Active Directory secret engine. The following endpoints are supported.

- `"/ad/config"` as a resource
- `"/ad/roles/<role name>"` as a resource
- `"/ad/library/<set name>"` as a resource
- `"/ad/creds/<role name>"` as a data source

Notably the rotate-root and check-in/check-out functionality aren't supported. These aren't really managed resources and don't work well with TF.

Currently automated acceptance testing isn't integrated with CircleCI because the Samba Docker image I'm using requires `privileged`, which is not possible with the `docker-compose` CircleCI service. This needs to be changed to use a machine instead so we can set the privileged flag. Integrated automated testing will be a follow up PR.

To test this, you can run the following: 

```bash
export VAULT_DEV_ROOT_TOKEN_ID="root"
export VAULT_TOKEN="root"
export VAULT_ADDR="http://0.0.0.0:8200"

export AD_BINDDN="CN=Administrator,CN=Users,DC=corp,DC=example,DC=net"
export AD_BINDPASS="SuperSecretPassw0rd"
export AD_URL="ldaps://ad"

docker network create --driver bridge vault

docker run \
  --name=vault \
  --hostname=vault \
  --network=vault \
  -p 8200:8200 \
  -e VAULT_DEV_ROOT_TOKEN_ID="root" \
  -e VAULT_ADDR="http://localhost:8200" \
  -e VAULT_DEV_LISTEN_ADDRESS="0.0.0.0:8200" \
  --privileged \
  --detach vault:latest

docker run \
  --name=ad \
  --hostname=ad \
  --network=vault \
  --privileged \
  -p 389:389 \
  -e SAMBA_DC_REALM="corp.example.net" \
  -e SAMBA_DC_DOMAIN="EXAMPLE" \
  -e SAMBA_DC_ADMIN_PASSWD="SuperSecretPassw0rd" \
  -e SAMBA_DC_DNS_BACKEND="SAMBA_INTERNAL" \
  --detach "laslabs/alpine-samba-dc" samba

sleep 30

cat <<'EOF' >> user.ldif
dn: CN=Bob,CN=Users,DC=corp,DC=example,DC=net
objectClass: top
objectClass: person
objectClass: organizationalPerson
objectClass: user
cn: Bob
description: test account
name: Bob
sAMAccountName: Bob
distinguishedName: CN=Bob,CN=Users,DC=corp,DC=example,DC=net
userPrincipalName: Bob

dn: CN=Mary,CN=Users,DC=corp,DC=example,DC=net
objectClass: top
objectClass: person
objectClass: organizationalPerson
objectClass: user
cn: Mary
description: test account
name: Mary
sAMAccountName: Mary
distinguishedName: CN=Mary,CN=Users,DC=corp,DC=example,DC=net
userPrincipalName: Mary
EOF

LDAPTLS_REQCERT=never ldapadd -h 127.0.0.1 -Z -p 389 -w "SuperSecretPassw0rd" -D "CN=Administrator,CN=Users,DC=corp,DC=example,DC=net" -f user.ldif

make testacc TESTARGS='-run TestAccDataSourceADAccessCredentials*'
make testacc TESTARGS='-run TestAccADSecretBackend*'
```

If you want to test this via `terraform` binary, here's an example resource:

```hcl
provider "vault" {
  address = "http://0.0.0.0:8200"
  token = "root"
}

resource "vault_ad_secret_backend" "config" {
    backend = "ad"
    binddn = "CN=Administrator,CN=Users,DC=corp,DC=example,DC=net"
    bindpass = "SuperSecretPassw0rd"
    url = "ldaps://ad"
    insecure_tls = "true"
    userdn = "CN=Users,DC=corp,DC=example,DC=net"
}

resource "vault_ad_secret_role" "bob" {
    backend = vault_ad_secret_backend.config.backend
    role = "bob"
    service_account_name = "Bob"
    ttl = 60
}

data "vault_ad_access_credentials" "bob_creds" {
  backend = vault_ad_secret_backend.config.backend
  role    = vault_ad_secret_role.bob.role
  depends_on = [vault_ad_secret_role.bob]
}

resource "vault_ad_secret_role" "mary" {
    backend = vault_ad_secret_backend.config.backend
    role = "mary"
    service_account_name = "Mary"
    ttl = 60
}

resource "vault_ad_secret_library" "qa" {
    backend = vault_ad_secret_backend.config.backend
    name = "qa"
    service_account_names = ["Bob", "Mary"]
    ttl = 60
    max_ttl = 120
    disable_check_in_enforcement = true
}

data "vault_ad_access_credentials" "mary_creds" {
  backend = vault_ad_secret_backend.config.backend
  role    = vault_ad_secret_role.mary.role
  depends_on = [vault_ad_secret_role.mary]
}

output "bob_password" {
  value = data.vault_ad_access_credentials.bob_creds.current_password
}

output "mary_password" {
  value = data.vault_ad_access_credentials.mary_creds.current_password
}
```

<details>
<summary>Related test output</summary>

```
[~/Git/terraform-provider-vault] make testacc TESTARGS='-run TestAccADSecretBackend*'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccADSecretBackend* -timeout 120m
?   	github.com/terraform-providers/terraform-provider-vault/cmd/generate	[no test files]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/codegen	(cached) [no tests to run]
?   	github.com/terraform-providers/terraform-provider-vault/generated	[no test files]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/decode	3.462s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/encode	2.830s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/alphabet	4.129s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/role	4.748s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/template	5.562s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/transformation	5.317s [no tests to run]
?   	github.com/terraform-providers/terraform-provider-vault/schema	[no test files]
PASS
=== RUN   TestAccADSecretBackendRole_basic
--- PASS: TestAccADSecretBackendRole_basic (0.47s)
=== RUN   TestAccADSecretBackendRole_import
--- PASS: TestAccADSecretBackendRole_import (0.25s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	6.046s

[~/Git/terraform-provider-vault] make testacc TESTARGS='-run TestAccDataSourceADAccessCredentials*'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccDataSourceADAccessCredentials* -timeout 120m
?   	github.com/terraform-providers/terraform-provider-vault/cmd/generate	[no test files]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/codegen	0.387s [no tests to run]
?   	github.com/terraform-providers/terraform-provider-vault/generated	[no test files]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/decode	0.993s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/encode	2.346s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/alphabet	2.087s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/role	1.245s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/template	1.831s [no tests to run]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/transformation	2.844s [no tests to run]
?   	github.com/terraform-providers/terraform-provider-vault/schema	[no test files]
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/util	0.586s [no tests to run]
=== RUN   TestAccDataSourceADAccessCredentials_basic
--- PASS: TestAccDataSourceADAccessCredentials_basic (0.45s)
PASS
ok  	github.com/terraform-providers/terraform-provider-vault/vault	3.068s
```
</details>


jeffsanicola:
Will `"ad/library/<set_name>"` resources also be supported in this release or is that expected in a future release?

jasonodonnell:
@jeffsanicola Not in this release but could happen in a future release.  See my note in the PR:

> Notably the rotate-root, library and check-in/check-out functionality aren't supported. These aren't really managed resources and don't work well with TF.

jasonodonnell:
@jeffsanicola Thinking about this more, library does make sense. I will add it to this PR.

jeffsanicola:
@jasonodonnell Awesome! Thank you for re-evaluating.

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,643 @@
...
+			Optional:    true,
+			Computed:    true,
+			Description: `The number of seconds after a Vault rotation where, if Active Directory shows a later rotation, it should be considered out-of-band.`,
+		},
+		"length": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: `The desired length of passwords that Vault generates.`,
+		},
```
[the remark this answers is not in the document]
jasonodonnell: Good catch! Done.

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,643 @@
...
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: `Use anonymous bind to discover the bind DN of a user (optional)`,
+		},
+		"formatter": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: `Text to insert the password into, ex. "customPrefix{{PASSWORD}}customSuffix".`,
+		},
```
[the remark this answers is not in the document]
jasonodonnell: Good catch! Done.

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,643 @@
...
+			Required:    false,
+			Optional:    true,
+			Description: "Mark the secrets engine as local-only. Local engines are not replicated or removed by replication.Tolerance duration to use when checking the last rotation time.",
+		},
+		"max_lease_ttl_seconds": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: "Maximum possible lease duration for secrets in seconds",
+		},
```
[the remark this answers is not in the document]
jasonodonnell: I just followed precedent set by the other backends in the project. Possibly in the future this should be cleaned up if possible but probably should keep it the same for now. Thoughts?

jasonodonnell:
@jeffsanicola `ad/library` support has been added.

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+			Optional:    true,
+			Description: `The mount path for a backend, for example, the path given in "$ vault auth enable -path=my-ad ad".`,
+			StateFunc: func(v interface{}) string {
+				return strings.Trim(v.(string), "/")
+			},
+		},
+		"anonymous_group_search": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: `Use anonymous binds when performing LDAP group searches (if true the initial credentials will still be used for the initial connection test).`,
```
tomhjp, line 27: Is the assumption that when default is not specified in the description for booleans, it defaults to false?

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+		},
+		"groupfilter": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: `Go template for querying group membership of user (optional) The template can access the following context variables: UserDN, Username Example: (&(objectClass=group)(member:1.2.840.113556.1.4.1941:={{.UserDN}})) Default: (|(memberUid={{.Username}})(member={{.UserDN}})(uniqueMember={{.UserDN}}))`,
+		},
+		"insecure_tls": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: `Skip LDAP server SSL Certificate verification - VERY insecure (optional)`,
```
tomhjp, line 108: Is the all-caps "very" a bit of a break from the usual tone in our docs? Maybe something like:

```suggestion
			Description: `Skip LDAP server SSL Certificate verification - insecure and not recommended for production use (optional)`,
```

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+		},
+		"last_rotation_tolerance": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: `The number of seconds after a Vault rotation where, if Active Directory shows a later rotation, it should be considered out-of-band.`,
```
tomhjp, line 114: It looks like some descriptions explicitly state (optional), and some omit it. I'm happy either way (especially as the documentation below consistently includes it, and that's the more user-facing part), but IMO we should stay consistent to one way or the other here.

vault/resource_ad_secret_backend.go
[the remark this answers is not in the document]
tomhjp: If this is very consistent elsewhere in the provider then SGTM

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: `Name of the password policy to use to generate passwords.`,
+		},
+		"request_timeout": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: `Timeout, in seconds, for the connection when making requests against the server before returning back an error.`,
+		},
+		"starttls": {
```
tomhjp, line 151: How come there's no underscore in this one?

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("error mounting to %q: %s", backend, err)
+	}
+
+	log.Printf("[DEBUG] Mounted AD backend at %q", backend)
+	d.SetId(backend)
+
+	data := map[string]interface{}{}
+	if v, ok := d.GetOkExists("anonymous_group_search"); ok {
```
tomhjp, line 246: Any reason not to use a loop over a slice of strings here?

Definitely not for this PR, but a wider point:
Also I know it's consistent with the style elsewhere, but it scares me that we duplicate all these strings between the schema and implementation. It seems like an easy win to just have some consts, although given everything is currently in the `vault` package, perhaps we would want to split into multiple packages to prevent massive top-level name noise.

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+		return fmt.Errorf("error reading %q: %s", configPath, err)
+	}
+	log.Printf("[DEBUG] Read %q", configPath)
+	if resp == nil {
+		log.Printf("[WARN] %q not found, removing from state", configPath)
+		d.SetId("")
+		return nil
+	}
+
+	if val, ok := resp.Data["anonymous_group_search"]; ok {
```
tomhjp, line 373: Same question on using a for loop here. I see there are some slightly more specific error messages, but I don't think they add much value compared to being confident that the code is doing the right thing.

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/vault/api"
+	"github.com/terraform-providers/terraform-provider-vault/util"
+)
+
+var (
+	adSecretBackendFromLibraryPathRegex = regexp.MustCompile("^(.+)/library/.+$")
```
tomhjp, line 15: I don't think so, but is it possible to defeat this by having a set name that has `library/` in it? If that is possible then you should be able to fix by making the `(.+)` group into a non-greedy wildcard `(.+?)`.

vault/resource_ad_secret_roles.go
```diff
@@ -0,0 +1,209 @@
...
+		d.SetId("")
+		return fmt.Errorf("invalid role ID %q: %s", rolePath, err)
+	}
+	d.Set("backend", backend)
+
+	resp, err := client.Logical().Read(rolePath)
+	if err != nil {
+		return fmt.Errorf("error reading %q: %s", rolePath, err)
+	}
+	log.Printf("[DEBUG] Read %q", rolePath)
```
tomhjp, line 120: Another comment for the wider code base rather than this PR - it would be lovely to get some helper functions for debug/warn/error etc logs

tomhjp (approved):
LGTM! A few small suggestions

vault/resource_ad_secret_library.go
jasonodonnell: I don't think this is possible. Library names can't be nested.

vault/resource_ad_secret_backend.go
jasonodonnell: I think this would make a good cleanup across the board. Something we could target in the future.

vault/resource_ad_secret_backend.go
jasonodonnell: ^ I think this would make a good cleanup across the board. Something we could target in the future.

vault/resource_ad_secret_backend.go
jasonodonnell: This configurable comes from Vault's LDAP client so we don't have a say here.

vault/resource_ad_secret_backend.go
jasonodonnell: Good catch, I removed optional from the description to match the rest.

vault/resource_ad_secret_backend.go
jasonodonnell: That's correct. Also remember that the backend itself has defaults, so it's not always proper for Terraform to duplicate it, I think.

vault/resource_ad_secret_roles.go
```diff
@@ -0,0 +1,209 @@
...
+		"backend": {
+			Type:        schema.TypeString,
+			Required:    true,
+			ForceNew:    true,
+			Description: "The mount path for the AD backend.",
+			StateFunc: func(v interface{}) string {
+				return strings.Trim(v.(string), "/")
+			},
+		},
+		"role": {
```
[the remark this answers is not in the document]
jasonodonnell: The use of `role` is precedent here. Agreed its strange but I suppose consistency is correct.

vault/resource_ad_secret_roles.go
```diff
@@ -0,0 +1,209 @@
...
+
+	roleName, err := adSecretBackendRoleNameFromPath(rolePath)
+	if err != nil {
+		log.Printf("[WARN] Removing AD role %q because its ID is invalid", rolePath)
+		d.SetId("")
+		return fmt.Errorf("invalid role ID %q: %s", rolePath, err)
+	}
+	d.Set("role", roleName)
+
+	backend, err := adSecretBackendFromPath(rolePath)
```
calvn, line 108: How come `createRoleResource` and `updateRoleResource` get the backend string from `d.Get("backend")`, but read infers it from the role path?

calvn: Ah, nevermind. This is the pattern on other role-related resources as well.

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+
+func readLibraryResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	setPath := d.Id()
+	log.Printf("[DEBUG] Reading %q", setPath)
+
+	setName, err := adSecretBackendSetNameFromPath(setPath)
+	if err != nil {
+		log.Printf("[WARN] Removing AD library %q because its ID is invalid", setPath)
+		d.SetId("")
```
catsby, line 115: I don't feel like we should remove from state here. My reasoning is that we're doing a name regex match and if it fails, we remove the thing from state. It's possible in the future the supported names / regex patterns in Vault could change, and we could inappropriately remove things from state because Terraform has an out of date regex. I'm not sure why we need to ensure our ID conforms to a regex at time of READ either 🤔 

For removing from state, we should typically reserve that for when Vault returns a 404 for that resource.

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+		return fmt.Errorf("invalid library ID %q: %s", setPath, err)
+	}
+	d.Set("name", setName)
+
+	backend, err := adSecretBackendFromLibraryPath(setPath)
+	if err != nil {
+		log.Printf("[WARN] Removing AD library %q because its ID is invalid", setPath)
+		d.SetId("")
+		return fmt.Errorf("invalid library ID %q: %s", setPath, err)
```
catsby, line 124: Same comment here as above in the regex matching, it doesn't seem appropriate in the READ method

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	d.Set("backend", backend)
+
+	resp, err := client.Logical().Read(setPath)
+	if err != nil {
+		return fmt.Errorf("error reading %q: %s", setPath, err)
```
catsby, line 130: If the Thing doesn't exist, do we get a 404 here or just `nil` for the error? If we get a 404 error we'd want to remove from state. I see we do that on line 136 but I'm curious if we never get to that part due to the error here

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+	return readLibraryResource(d, meta)
+}
+
+func deleteLibraryResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	setPath := d.Id()
+	log.Printf("[DEBUG] Deleting %q", setPath)
+
+	if _, err := client.Logical().Delete(setPath); err != nil && !util.Is404(err) {
+		return fmt.Errorf("error deleting %q: %s", setPath, err)
```
catsby, line 202: 404 check here?

catsby: Unless deleting a thing that doesn't exist doesn't return an error 🤔

vault/resource_ad_secret_roles.go
```diff
@@ -0,0 +1,209 @@
...
+func readRoleResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	rolePath := d.Id()
+	log.Printf("[DEBUG] Reading %q", rolePath)
+
+	roleName, err := adSecretBackendRoleNameFromPath(rolePath)
+	if err != nil {
+		log.Printf("[WARN] Removing AD role %q because its ID is invalid", rolePath)
+		d.SetId("")
+		return fmt.Errorf("invalid role ID %q: %s", rolePath, err)
```
catsby, line 104: Same thoughts here on validating the names/paths via regex and not by what Vault tells us

website/vault.erb
```diff
@@ -25,6 +25,10 @@
                             <a href="/docs/providers/vault/d/auth_backend.html">vault_auth_backend</a>
                         </li>
 
+                        <li<%= sidebar_current("docs-vault-datasource-ad-access-credentials") %>>
+                            <a href="/docs/providers/vault/d/ad_access_credentials.html">vault_ad_access_credentials</a>
+                        </li>
+
```
catsby, line 31: Unfortunately this file is no longer used at all, and should be removed in a separate PR

catsby (changes requested):
There are some minor things that I think need changing, and otherwise some questions I had or some nitpicks. 

I notice that there are a handful of generated resource documents that were moved? I believe those were moved in https://github.com/terraform-providers/terraform-provider-vault/pull/871 and should stay where they are. The Terraform registry now uses file structure as a means of document structure, and expects docs to be in specific locations. Let me know if you have questions or need help resolving those

calvn:
> By default, check-in must be called by the same entity or client token used for check-out.

Does TF use the same client token across runs? If not, this could be problematic for the library flow if a set was checked out by one run but then later checked in by another apply. @catsby do you see any potential issues with this?

jasonodonnell:
@calvn This provider does not support checkin/checkout functionality. We only allow you to create the library sets that can be used by checkin/checkout workflows.

vault/resource_ad_secret_backend.go
jasonodonnell: They don't always contain data and reading `ad/config` doesn't return them if they're null. Here's a read on the `ad/config` the bare minimum settings:

```
vault read ad/config
Key                             Value
---                             -----
binddn                          CN=Administrator,CN=Users,DC=corp,DC=example,DC=net
certificate                     n/a
formatter                       n/a
insecure_tls                    true
last_rotation_tolerance         5
length                          64
max_ttl                         768h
policy_name                     n/a
starttls                        false
tls_max_version                 tls12
tls_min_version                 tls12
ttl                             768h
upndomain                       n/a
url                             ldaps://ad
use_pre111_group_cn_behavior    false
userdn                          CN=Users,DC=corp,DC=example,DC=net
```

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+
+func deleteConfigResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	vaultPath := d.Id()
+	log.Printf("[DEBUG] Unmounting AD backend %q", vaultPath)
+
+	err := client.Sys().Unmount(vaultPath)
+	if err != nil {
+		return fmt.Errorf("error unmounting AD backend from %q: %s", vaultPath, err)
+	}
```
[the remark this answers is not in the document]
jasonodonnell: Good call, done!

vault/resource_ad_secret_library.go
jasonodonnell: Done

vault/resource_ad_secret_library.go
jasonodonnell: Done!

calvn:
Ah, we're good then!

vault/resource_ad_secret_library.go
jasonodonnell: It does, so I added it

vault/resource_ad_secret_library.go
jasonodonnell: It doesn't, just a 204:

```
curl -I \
    --header "X-Vault-Token: root" \
    --request DELETE \
    http://127.0.0.1:8200/v1/ad/library/clint
HTTP/1.1 204 No Content
Cache-Control: no-store
Date: Tue, 17 Nov 2020 00:13:09 GMT
```

vault/resource_ad_secret_roles.go
jasonodonnell: Done

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,645 @@
...
+func readConfigResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	path := d.Id()
+	log.Printf("[DEBUG] Reading %q", path)
+
+	mountResp, err := client.Sys().MountConfig(path)
+	if err != nil {
+		return fmt.Errorf("error reading %q: %s", path, err)
+	}
```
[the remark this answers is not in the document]
jasonodonnell: Done

jasonodonnell:
@catsby Good call on the transform docs being moved. Not sure how that happened but I fixed it.

vault/resource_ad_secret_backend.go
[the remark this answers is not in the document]
tvoran: I see what you mean, thanks for clarifying.

vault/resource_ad_secret_library.go
```diff
@@ -0,0 +1,232 @@
...
+			Optional:    true,
+			Computed:    true,
+			Description: `The maximum amount of time a check-out last with renewal before Vault automatically checks it back in. Defaults to 24 hours. Setting it to zero reflects an unlimited lending period.`,
+		},
+		"ttl": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: `The maximum amount of time a single check-out lasts before Vault automatically checks it back in. Defaults to 24 hours. Setting it to zero reflects an unlimited lending period.`,
+		},
```
tvoran, line 54: Are these the same parameter? They have almost identical descriptions.

And since these are TypeInt, might want to mention in the description that input should be in seconds (since the [API apparently accepts a duration string](https://www.vaultproject.io/api/secret/ad#library-management)).

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,654 @@
...
+		"discoverdn": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: `Use anonymous bind to discover the bind DN of a user.`,
+		},
+		"formatter": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Deprecated:  `Formatter is deprecated and password_policy should be used.`,
```
tvoran, line 88: May want to mention the version for deprecation too.

```suggestion
			Deprecated:  `Formatter is deprecated and password_policy should be used with Vault >= 1.5.`,
```

vault/resource_ad_secret_backend.go
```diff
@@ -0,0 +1,654 @@
...
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: `The number of seconds after a Vault rotation where, if Active Directory shows a later rotation, it should be considered out-of-band.`,
+		},
+		"length": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Deprecated:  `Length is deprecated and password_policy should be used.`,
```
tvoran, line 121: Same suggestion here:
```suggestion
			Deprecated:  `Length is deprecated and password_policy should be used with Vault >= 1.5.`,
```

catsby (approved):
👍

vault/resource_ad_secret_library.go
jasonodonnell: Good catch, done!

PR #919, opened 2020-11-24
update go mod to point to hashicorp/terraform-provider-vault

catsby:
This PR updates `go.mod` to reference `hashicorp/terraform-provider-vault`, from the old `terraform-providers/` org. Note the oddity in `go.mod` where we depend on ourselves; I believe that is due to the handful of sub-package in `generated/` (?)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,6 @@
 module github.com/hashicorp/terraform-provider-vault
 
+// This should ensure existing PRs are still valid
 replace github.com/terraform-providers/terraform-provider-vault => ./
 
 go 1.13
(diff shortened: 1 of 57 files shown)


jasonodonnell (approved):

go.mod
```diff
@@ -18,4 +20,5 @@ require (
 	github.com/hashicorp/vault/sdk v0.1.14-0.20191017173300-47a54ac8bc6c
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/rainycape/unidecode v0.0.0-20150907023854-cb7f23ec59be // indirect
+	github.com/terraform-providers/terraform-provider-vault v0.0.0-00010101000000-000000000000
```
calvn, line 23: This looks off, maybe being affected by the replace above?

catsby: As @jasonodonnell found, `main.go` was still using that. I've updated it in 56817de and `go mod tidy` removed that line

calvn (approved):

PR #831, opened 2020-07-26
WIP: Nomad secret backend

spuder:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Closes #640

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
*New Resource:* `vault_nomad_secret_backend` 
```

Output from acceptance testing:

<details><summary>make testacc TESTARGS='-run=testAccNomadSecretBackendRoleCheckDestroy'</summary>
<p>

```
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=testAccNomadSecretBackendRoleCheckDestroy -timeout 120m
?       github.com/terraform-providers/terraform-provider-vault [no test files]
?       github.com/terraform-providers/terraform-provider-vault/cmd/coverage    [no test files]
?       github.com/terraform-providers/terraform-provider-vault/cmd/generate    [no test files]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/codegen (cached) [no tests to run]
?       github.com/terraform-providers/terraform-provider-vault/generated       [no test files]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/decode  (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/datasources/transform/encode  (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/alphabet  (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/role      (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/template  (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/terraform-providers/terraform-provider-vault/generated/resources/transform/transformation    (cached) [no tests to run]
[...]

PR #923, opened 2020-11-30
Add Nomad secret backend

jasonodonnell:
This PR adds support for the Nomad secret engine. The following endpoints are supported.

- `"/nomad/config/access"` as a resource
- `"/nomad/config/lease"` as a resource
- `"/nomad/role/<role name>"` as a resource
- `"/nomad/creds/<role name>"` as a data source

Currently automated acceptance testing isn't integrated with CircleCI because to run Nomad in a container, it needs to run as `privileged`. When we update the CI for other integrated testing (such as AD) this should be added and is being tracked internally.

To manually test this, you can use the following script: https://gist.github.com/jasonodonnell/1a2cfaaa376592c170ad5fa2eab7e423. This script assumes you have `vault` and `nomad` installed.

To run the acceptance tests, here's a setup script you can run from the project root: https://gist.github.com/jasonodonnell/e40a9b63eb8e9cd3556d221fd2e16494. This script assumes you have `vault` and `nomad` installed.

If you want to test this via `terraform` binary, here's an example resource:

```hcl
variable "TOKEN" {}

provider "vault" {
  address = "http://0.0.0.0:8200"
  token = "root"
}

resource "vault_nomad_secret_backend" "config" {
    backend = "nomad"
    address = "http://0.0.0.0:4646"
    token = var.TOKEN
}

resource "vault_nomad_secret_lease" "lease" {
    backend = vault_nomad_secret_backend.config.backend
    max_ttl = 60
    ttl = 30
}

resource "vault_nomad_secret_role" "test" {
    backend = vault_nomad_secret_backend.config.backend
    role = "test"
    type = "client"
    policies = ["readonly"]
}

data "vault_nomad_access_token" "token" {
  backend = vault_nomad_secret_backend.config.backend
  role    = vault_nomad_secret_role.test.role
  depends_on = [vault_nomad_secret_role.test]
}

output "secret" {
  value = data.vault_nomad_access_token.token.secret_id
}

output "accessor" {
  value = data.vault_nomad_access_token.token.accessor_id
}
```

<details>
<summary>Related test output</summary>

```
[~/Git/terraform-provider-vault] make testacc TESTARGS='-run TestAccDataSourceNomadAccess*'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccDataSourceNomadAccess* -timeout 120m
PASS
ok  	github.com/hashicorp/terraform-provider-vault/codegen	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/decode	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/encode	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/alphabet	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/role	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/transformation	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/util	(cached) [no tests to run]
=== RUN   TestAccDataSourceNomadAccessCredentialsClientBasic
--- PASS: TestAccDataSourceNomadAccessCredentialsClientBasic (0.73s)
=== RUN   TestAccDataSourceNomadAccessCredentialsManagementBasic
--- PASS: TestAccDataSourceNomadAccessCredentialsManagementBasic (0.71s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	1.940s

[~/Git/terraform-provider-vault] make testacc TESTARGS='-run TestAccNomadSecret*'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccNomadSecret* -timeout 120m
PASS
ok  	github.com/hashicorp/terraform-provider-vault/codegen	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/decode	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/encode	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/alphabet	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/role	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/transformation	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/util	(cached) [no tests to run]
=== RUN   TestAccNomadSecretLease
--- PASS: TestAccNomadSecretLease (0.21s)
=== RUN   TestAccNomadSecretBackendRoleClientBasic
--- PASS: TestAccNomadSecretBackendRoleClientBasic (0.14s)
=== RUN   TestAccNomadSecretBackendRoleManagementBasic
--- PASS: TestAccNomadSecretBackendRoleManagementBasic (0.14s)
=== RUN   TestAccNomadSecretBackendRoleImport
--- PASS: TestAccNomadSecretBackendRoleImport (0.16s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	1.118s
```
</details>

Closes #831.


vault/provider.go
```diff
@@ -466,6 +470,18 @@ var (
 			Resource:      ldapAuthBackendGroupResource(),
 			PathInventory: []string{"/auth/ldap/groups/{name}"},
 		},
+		"vault_nomad_secret_backend": {
+			Resource:      nomadSecretAccessBackendResource(),
+			PathInventory: []string{"/nomad"},
```
calvn, line 475: Should `/nomad/config/access` be part of `PathInventory` here?

jasonodonnell: @calvn Good question! Throughout the provider the mount and the config are the same resource. We derive the full path `nomad/config/access` from the `backend`. Since we derive it, `/nomad` is the default.

jasonodonnell: `nomad/config/lease` is slightly different so I didn't add it to this resource. We _could_ but it felt different enough to have a separate resource.

jasonodonnell: @calvn Having thought about combining `nomad/config/lease` with the other configs, I decided to keep them as two separate resources. This is simply to reduce the amount of operations create/update/read need to do for all three paths (`/nomad`, `/nomad/config/access` and `/nomad/config/lease`).

tvoran: Is `PathInventory` used for routing? Or is is just for documentation purposes?

To Calvin's question this function does seem to cover the `/nomad/config/access` endpoint?

tvoran (commented):
The manual and acceptance tests worked for me locally, not seeing any blockers on the first pass here.

vault/provider.go
jasonodonnell: I can try in the morning having both but putting to the same resource handler, maybe that won't cause issues.

jasonodonnell: Update: both can't be present because the resource name is the key and you can't duplicate it.

I changed the PathInventory to be `/nomad/config/access`. This did not have any regressions on the resources and is probably a better self-documenting change.

tvoran: PathInventory is a list, so it could also be this:

```go
		"vault_nomad_secret_backend": {
			Resource:      nomadSecretAccessBackendResource(),
			PathInventory: []string{
				"/nomad",
				"/nomad/config/access",
			},
```

But I think setting it to just "/nomad/config/access" makes sense 👍

tvoran (approved):

vault/provider.go
jasonodonnell: Oh derp, yeah, that did work! Good call!

vault/provider.go
```diff
@@ -466,6 +470,21 @@ var (
...
 			PathInventory: []string{"/auth/ldap/groups/{name}"},
 		},
+		"vault_nomad_secret_backend": {
+			Resource: nomadSecretAccessBackendResource(),
+			PathInventory: []string{
+				"/nomad",
+				"/nomad/config/access",
+			},
+		},
+		"vault_nomad_secret_lease": {
```
catsby, line 480: I feel that the fields managed by this resource should be managed by `vault_nomad_secret_backend`, and this resource should be removed. From a user perspective, it makes more sense to have those values in a single resource as they pertain to just that resource.

vault/data_source_nomad_credentials.go
```diff
@@ -0,0 +1,71 @@
...
+	if accessorID == "" {
+		return fmt.Errorf("accessor_id is not set in response")
+	}
+
+	secretID := secret.Data["secret_id"].(string)
+	if secretID == "" {
+		return fmt.Errorf("secret_id is not set in response")
+	}
+
+	d.SetId(accessorID)
```
catsby, line 66: Why not use `path` here? Granted, we aren't using the ID at all that I can see

vault/resource_nomad_secret_backend.go
```diff
@@ -0,0 +1,304 @@
...
+		return fmt.Errorf("error reading %q: %s", configPath, err)
+	}
+	log.Printf("[DEBUG] Read %q", configPath)
+	if resp == nil {
+		log.Printf("[WARN] %q not found, removing from state", configPath)
+		d.SetId("")
+		return nil
+	}
+
+	if val, ok := resp.Data["address"]; ok {
```
catsby, line 189: I think these values should be read and update the state regardless if they are set in the response. If we use this resource to set them, and then out-of-band the backend is destroyed and recreated at the same address _but without these values_, we'll fail to detect the drift here.

vault/resource_nomad_secret_backend.go
```diff
@@ -0,0 +1,304 @@
...
+func updateNomadAccessConfigResource(d *schema.ResourceData, meta interface{}) error {
+	backend := d.Id()
+
+	client := meta.(*api.Client)
+	defaultTTL := d.Get("default_lease_ttl_seconds").(int)
+	maxTTL := d.Get("max_lease_ttl_seconds").(int)
+	tune := api.MountConfigInput{}
+	data := map[string]interface{}{}
+
+	if defaultTTL != 0 {
```
catsby, line 237: I believe this prevents us from initially setting a default lease to something that is not the default system lease, and then removing it, e.g. we can't set it to something non-standard and then try to later adapt the system defaults implicitly by removing our non-standard value. 

Example:

- System default is `5m`
- We use `default_lease_ttl_seconds` to set it to `6m` instead
- Later we decide we want to use the system defaults, so we remove our `default_lease_ttl_seconds`
- No change is detected here, because the new value would be the zero value, right? 
- No update happens, and the default least ttl remains at `6m`

Is my assumption correct here?

vault/provider.go
jasonodonnell: Merged

vault/provider.go
jasonodonnell: `/nomad`, `/nomad/config/access` and `/nomad/config/lease` are merged into a single resource `vault_nomad_secret_backend`.

vault/resource_nomad_secret_backend.go
jasonodonnell: Agreed, fixed.

vault/resource_nomad_secret_backend.go
jasonodonnell: This is `Required: true`, so you would get an error if you didn't set it. Thoughts?

vault/data_source_nomad_credentials.go
jasonodonnell: Each call to this endpoint generates a new token, so we want the ID to be the unique identifier of that specific generated token. Path would be too generic I think?

vault/resource_nomad_secret_backend_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/hashicorp/terraform-provider-vault/util"
+	"github.com/hashicorp/vault/api"
+)
+
+func TestAccNomadSecretBackend(t *testing.T) {
+	backend := acctest.RandomWithPrefix("tf-test-nomad")
+	address, token := util.GetTestNomadCreds(t)
```
tvoran, line 17: Does vault verify the connection to nomad with the address and token when set? Or could these be test values and run in CI? Something like:

```suggestion
	address := "https://test.address"
	token := "test-token"
```

Same question on the other nomad acceptance tests.

vault/resource_nomad_secret_backend.go
```diff
@@ -0,0 +1,368 @@
...
+		if err := d.Set("max_token_name_length", val); err != nil {
+			return fmt.Errorf("error setting state key 'max_token_name_length': %s", err)
+		}
+	}
+
+	// if val, ok := resp.Data["token"]; ok {
+	// 	if err := d.Set("token", val); err != nil {
+	// 		return fmt.Errorf("error setting state key 'token': %s", err)
+	// 	}
+	// }
```
tvoran, line 250: Still needed?

vault/resource_nomad_secret_backend_test.go
jasonodonnell: This should be the value of an actual token because we'll be running this against a real Nomad system. Nomad ACL bootstrap tokens must be generated, we can't assign it a dev static value like we do in Vault.

vault/resource_nomad_secret_backend.go
jasonodonnell: I removed this commented code. This is actually not needed because the read from Vault doesn't return a token because its a sensitive value.

PR #942, opened 2021-01-13
add new key_list resource which outputs list of keys at given path

kirecek:
Hello,

a long time ago somebody raised a request to add a resource for listing secret keys at given path - https://github.com/hashicorp/terraform-provider-vault/issues/57

I have created a quick, simple WIP patch that supports this.

Is this something that vault provider would consider to support? If yes, I can finish the MR (refactor, docs, tests) otherwise I'll just keep using simple patch.

Thank you

### DIFF
--- a/vault/data_source_key_list.go
+++ b/vault/data_source_key_list.go
@@ -1,0 +1,121 @@
+package vault
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"time"
+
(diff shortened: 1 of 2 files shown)


spacex:
For those tracking this, I have re-committed the code from this pull request at https://github.com/spacex/terraform-provider-vault/tree/add_key_list_data_source

See #57 for additional comments.

PR #959, opened 2021-01-29
Add Terraform Cloud/Enterprise credential support

Valarissa:
This PR seeks to introduce support for the upcoming Terraform Cloud Secrets Engine. It will enable users to create and manage the Terraform Secrets engine config, any associated roles, and generate credentials from those roles.

### Community Note

Relates OR Closes #0000

```release-note
Adding support for Terraform Cloud/Enterprise token generation for users, teams, and orgs.
```

```
/usr/local/Cellar/go/1.15.5/libexec/bin/go test -json ./... -run .*TerraformCloud.*
PASS
ok  	github.com/hashicorp/terraform-provider-vault/codegen	0.203s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/util	0.488s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/transformation	0.445s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/role	0.691s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/decode	0.951s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/alphabet	1.198s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/encode	1.427s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template	1.628s [no tests to run]
=== RUN   TestAccDataSourceTerraformCloudAccessCredentialsOrganizationClientBasic
--- PASS: TestAccDataSourceTerraformCloudAccessCredentialsOrganizationClientBasic (1.48s)
=== RUN   TestAccDataSourceTerraformCloudAccessCredentialsTeamClientBasic
--- PASS: TestAccDataSourceTerraformCloudAccessCredentialsTeamClientBasic (1.42s)
=== RUN   TestAccDataSourceTerraformCloudAccessCredentialsUserBasic
--- PASS: TestAccDataSourceTerraformCloudAccessCredentialsUserBasic (2.79s)
=== RUN   TestTerraformCloudSecretBackend
--- PASS: TestTerraformCloudSecretBackend (0.46s)
=== RUN   TestTerraformCloudSecretRole
--- PASS: TestTerraformCloudSecretRole (2.02s)
=== RUN   TestTerraformCloudSecretBackendRoleNameFromPath
--- PASS: TestTerraformCloudSecretBackendRoleNameFromPath (0.00s)
=== RUN   TestTerraformCloudSecretBackendRoleBackendFromPath
--- PASS: TestTerraformCloudSecretBackendRoleBackendFromPath (0.00s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	10.017s

Process finished with exit code 0
```


Valarissa:
Tests will continue to fail until tf secrets is bundled with vault, or I create a setup script for adding the plugin to vault.

vault/resource_terraform_cloud_secret_backend.go
```diff
@@ -0,0 +1,246 @@
...
+	log.Printf("[DEBUG] Mounting Terraform Cloud backend at %q", path)
+
+	if err := client.Sys().Mount(path, info); err != nil {
+		return fmt.Errorf("Error mounting to %q: %s", path, err)
+	}
+
+	log.Printf("[DEBUG] Mounted Terraform Cloud backend at %q", path)
+	d.SetId(path)
+
+	d.SetPartial("path")
```
catsby, line 106: `SetPartial` is deprecated, I believe we should just use `d.Set` here and set the values

- https://github.com/hashicorp/terraform-plugin-sdk/blob/9d2850173e33bf6e4b4fa0fd0819f602b45e23c8/helper/schema/resource_data.go#L206-L215

vault/resource_terraform_cloud_secret_backend.go
```diff
@@ -0,0 +1,246 @@
...
+	return nil
+}
+
+func terraformCloudSecretBackendUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	path := d.Id()
+	configPath := terraformCloudSecretBackendConfigPath(path)
+
+	d.Partial(true)
```
catsby, line 180: `Partial` is also deprecated and not necessary:

- https://github.com/hashicorp/terraform-plugin-sdk/blob/9d2850173e33bf6e4b4fa0fd0819f602b45e23c8/helper/schema/resource_data.go#L163

vault/resource_terraform_cloud_secret_backend.go
```diff
@@ -0,0 +1,246 @@
...
+
+func terraformCloudSecretBackendDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	path := d.Id()
+
+	log.Printf("[DEBUG] Unmounting Terraform Cloud backend %q", path)
+	err := client.Sys().Unmount(path)
+	if err != nil {
+		return fmt.Errorf("Error unmounting Terraform Cloud backend from %q: %s", path, err)
```
catsby, line 223: We should check for a 404 style error here; if so, remove from state and log that it wasn't found. Example: https://github.com/hashicorp/terraform-provider-vault/blob/3a348eac1c3cacbefbee3166e18f03f9c59824b5/vault/resource_ad_secret_backend.go#L645

vault/resource_terraform_cloud_secret_role.go
```diff
@@ -0,0 +1,229 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The name of an existing role against which to create this Terraform Cloud credential",
+			},
+			"path": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The path of the Terraform Cloud Secret Backend the role belongs to.",
+				Deprecated:    "use `backend` instead",
```
catsby, line 40: Since these are new resources, we should probably not include deprecated fields in the schema. Is there a reason we need them here? Possibly this is copy-paste or generated ?

vault/resource_terraform_cloud_secret_backend.go
catsby: I suppose that's assuming this isn't already accounted for because of the `Exists` method provided, in which case, we can ignore this suggestion :D

catsby (commented):
Looks great so far! Some minor nits that may or may not be accurate, depending on how good my Terraform knowledge has kept up. 

The comment on data source and resource will be had internally, I'll bring it up. 

The other thing is documentation missing, will that be opened/add separately?

vault/data_source_terraform_cloud_credentials.go
```diff
@@ -0,0 +1,85 @@
...
+	backend := d.Get("backend").(string)
+	role := d.Get("role").(string)
+	path := fmt.Sprintf("%s/creds/%s", backend, role)
+
+	secret, err := client.Logical().Read(path)
+	if err != nil {
+		return fmt.Errorf("error reading from Vault: %s", err)
+	}
+	log.Printf("[DEBUG] Read %q from Vault", path)
+	log.Printf("[DEBUG] Vault secret: %v", secret)
```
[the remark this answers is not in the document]
Valarissa: That was a leftover debug statement...

vault/resource_terraform_cloud_secret_role.go
Valarissa: Yeah, that was copy-pasta'd. I debated removing it, but figured I would leave it in for the sake of people unwilling to change? But, probably not the best call overall.

vault/resource_terraform_cloud_secret_backend.go
Valarissa: Removal of partial states will commence post-haste

vault/resource_terraform_cloud_secret_role.go
Valarissa: Removed

vault/resource_terraform_cloud_secret_backend.go
Valarissa: Partial states removed.

vault/resource_terraform_cloud_secret_backend.go
Valarissa: SetPartial removed

vault/resource_terraform_cloud_secret_backend.go
```diff
@@ -0,0 +1,246 @@
...
+				StateFunc: func(s interface{}) string {
+					return strings.Trim(s.(string), "/")
+				},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return old+"/" == new || new+"/" == old
+				},
+			},
+			"token": {
+				Type:        schema.TypeString,
+				Required:    true,
```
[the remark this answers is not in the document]
Valarissa: Made optional, all tests continue to run without issue.

vault/data_source_terraform_cloud_credentials.go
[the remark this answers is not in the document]
Valarissa: Leftover debug statement removed.

vault/resource_terraform_cloud_secret_backend.go
Valarissa: Suggestion ignored. :)

catsby (approved):
LGTM 👍 (data source -> resource discussion goes beyond this PR so we'll leave it as is)

vault/data_source_terraform_cloud_credentials.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+func terraformCloudAccessCredentialsDataSource() *schema.Resource {
```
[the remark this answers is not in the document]
Valarissa: This should now be updated to be a `resource` rather than `data`

vault/provider.go
```diff
@@ -629,6 +629,18 @@ var (
...
+			Resource:      terraformCloudSecretBackendResource(),
+			PathInventory: []string{"/terraform/config"},
+		},
+		"vault_terraform_cloud_secret_creds": {
+			Resource:      terraformCloudSecretCredsResource(),
+			PathInventory: []string{"/terraform/creds/{role}"},
+		},
+		"vault_terraform_cloud_secret_role": {
+			Resource:      terraformCloudSecretRoleResource(),
+			PathInventory: []string{"/terraform/roles/{name}"},
```
calvn, line 642: Should it be `]string{"/terraform/role/{name}"}` instead?

vault/resource_terraform_cloud_secret_role.go
```diff
@@ -0,0 +1,215 @@
...
+
+	path := d.Id()
+
+	log.Printf("[DEBUG] Checking Terraform Cloud secrets backend role at %q", path)
+
+	secret, err := client.Logical().Read(path)
+	if err != nil {
+		return false, fmt.Errorf("error reading role configuration for %q: %s", path, err)
+	}
+
```
calvn, line 187:
```suggestion
	log.Printf("[DEBUG] Checked if Terraform Cloud secrets backend role %q exists", path)

```

vault/resource_terraform_cloud_secret_role.go
```diff
@@ -0,0 +1,215 @@
...
+		return fmt.Errorf("invalid role ID %q: %s", path, err)
+	}
+
+	log.Printf("[DEBUG] Reading Terraform Cloud secrets backend role at %q", path)
+
+	secret, err := client.Logical().Read(path)
+	if err != nil {
+		return fmt.Errorf("error reading role configuration for %q: %s", path, err)
+	}
+
```
calvn, line 145:
```suggestion
	log.Printf("[DEBUG] Read Terraform Cloud secrets backend role at %q", path)

```

vault/resource_terraform_cloud_secret_role.go
```diff
@@ -0,0 +1,215 @@
...
+	if v, ok := d.GetOkExists("user_id"); ok {
+		payload["user_id"] = v
+	}
+
+	log.Printf("[DEBUG] Configuring Terraform Cloud secrets backend role at %q", path)
+
+	if _, err := client.Logical().Write(path, payload); err != nil {
+		return fmt.Errorf("error writing role configuration for %q: %s", path, err)
+	}
+
```
calvn, line 116:
```suggestion
	log.Printf("[DEBUG] Configured Terraform Cloud secrets backend role at %q", path)

```

calvn (approved):

vault/provider.go
Valarissa: Fixed.

husunal:
Great work thanks! We look forward to using this new TFE secret backend.

What happens to the Terraform resource when the generated token's TTL expires? Also, I didn't see any TTL argument in `resource_terraform_cloud_secret_creds.go` are you going to add this option as well so we can specify its TTL during secret creation.

Valarissa:
@husunal the creds' TTL should be based on the role used to create the creds.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,109 @@
...
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/vault/api"
+)
+
+func terraformCloudSecretCredsResource() *schema.Resource {
+	return &schema.Resource{
+		Create: readTerraformCloudSecretCredsResource,
+		Read:   readTerraformCloudSecretCredsResource,
```
catsby, line 14: I don't think `Create` and `Read` can be the same function here. For `Read` we need to read the token's lease ID to see if the token is still valid and exists. If not, then we will remove it from state with `d.SetId("")`. Alternatively we could do this by implementing an `Exists` function. 

As this stands, every `plan` (which does `Read`) is generating a token, and subsequent (if separate) `apply` also generates a token.

catsby (changes requested):
The secret resource cred needs to look up the current secret by lease ID to verify the token still exists, otherwise we generate a token with every read. I believe that for Team/Org tokens we wont have a lease ID so we can go ahead and do the full read again.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,148 @@
...
+	organization := d.Get("organization")
+	teamId := d.Get("team_id")
+
+	if organization == "" && teamId == "" {
+		data := map[string]interface{}{
+			"lease_id": id,
+		}
+		creds, err := client.Logical().Write("/sys/leases/lookup", data)
+		if err != nil {
+			if strings.Contains(err.Error(), "lease not found") {
```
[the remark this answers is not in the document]
catsby: Also, are we sure it's `lease not found`? I tried this locally and got `invalid lease`. This error came from creating a user token then coming back later, after the lease expired, and running `terraform plan`. Am I doing something different to get a different error message?

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,148 @@
...
+	d.Set("team_id", teamId)
+
+	return nil
+}
+
+func deleteTerraformCloudSecretCredsResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	id := d.Id()
+
+	err := client.Sys().Revoke(id)
```
catsby, line 99: I don't believe this actually revokes team or org tokens, as their IDs are not lease IDs. It turns out this endpoint returns a `204 no content` regardless of whatever lease id you give it

```console
$ curl -i -s --header "X-Vault-Token: $(vault print token)" -X PUT -d '{"lease_id":"ihavethehighground"}' http://127.0.0.1:8200/v1/sys/leases/revoke 
HTTP/1.1 204 No Content
Cache-Control: no-store
Date: Fri, 12 Mar 2021 21:28:19 GMT
```

If you try to revoke a team or org token, it appears to succeed, but instead the token remains (still visible on app.terraform.io)

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		if creds.LeaseDuration <= 0 {
+			return nil
+		}
+
+		d.SetId(id)
+		d.Set("token", data["token"])
+		d.Set("token_id", data["token_id"])
+		d.Set("organization", data["organization"])
+		d.Set("team_id", data["team_id"])
```
catsby, line 142: in this block starting around line 123 we define `data` to be a `map[string]interface{}` with only a `lease_id` key, so all of these `d.Set` calls would empty this values out, if we got here.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		creds, err := client.Logical().Write("/sys/leases/lookup", data)
+		if err != nil {
+			if strings.Contains(err.Error(), "lease not found") {
+				return nil
+			} else {
+				return err
+			}
+		}
+		if creds.LeaseDuration <= 0 {
```
catsby, line 134: `LeaseDuration` is part of the standard `api.Secret` but it's not populated by the leases lookup call, so the only meaningful data we get back from `lookup` would be in `creds.Data` here. See the example output:

- https://www.vaultproject.io/api-docs/system/leases#sample-response

catsby: As a result, `creds.LeaseDuration` is always zero here, and we return `nil` and don't update anything. Returning `nil` does not remove this from state either, I'm not sure if that was the intent or not

catsby (commented):
We're getting close, but the handling of the lease look up across orgs/teams and users is not quite right. Let me know if you have any questions here or if I need to clarify anything

vault/resource_terraform_cloud_secret_creds.go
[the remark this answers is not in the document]
Valarissa: It might be invalid lease, this was based off of a different time when I think the wrong thing was being used for the lease id.

vault/resource_terraform_cloud_secret_creds.go
Valarissa: My understanding is that we don't want the token to be destroyed on the site? IIRC, the token is created externally and then referenced through ENV vars. Because it's created outside of Terraform, my instinct was to leave it in place.

vault/resource_terraform_cloud_secret_creds.go
Valarissa: Welp... my brain thought, for some reason that it was gonna do some kind of mutation on the data, not thinking the fact that it's a write call so it's providing the data... *sigh* My bad.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -78,7 +80,11 @@ func readTerraformCloudSecretCredsResource(d *schema.ResourceData, meta interfac
 	organization := secret.Data["organization"]
 	teamId := secret.Data["team_id"]
 
-	d.SetId(tokenId)
+	if secret.LeaseID != "" {
+		d.SetId(secret.LeaseID)
+	} else {
+		d.SetId(tokenId)
+	}
```
calvn, line 87: Would this mean that any reference to the lease where it's assumed that the value is the lease ID won't work? For instance `deleteTerraformCloudSecretCredsResource` assumes that `d.Id()` is a lease ID and will call revoke using this value, which doesn't work if `d.SetId(tokenId)`.

Valarissa: The failure of the lease lookup for Org/Team tokens is fine as there is no dynamic cred to revoke/delete.

vault/resource_terraform_cloud_secret_creds.go
catsby: Follow-up: we discussed this synchronously and decided to leave as is. Resolving this.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if organization == "" && teamId == "" {
+		data := map[string]interface{}{
+			"lease_id": id,
+		}
+		creds, err := client.Logical().Write("/sys/leases/lookup", data)
+		if err != nil {
+			if strings.Contains(err.Error(), "lease not found") ||
+				strings.Contains(err.Error(), "invalid lease") {
+				log.Printf("User token %s lease expired, removing user token from schema", d.Get("token_id"))
```
catsby, line 133: We should prefix log messages with a `DEBUG` or similar, see https://www.terraform.io/docs/internals/debugging.html for options

catsby, line 133:
```suggestion
				log.Printf("User token %s lease expired, removing user token from state", d.Get("token_id"))
```

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,152 @@
...
+func deleteTerraformCloudSecretCredsResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	id := d.Id()
+
+	err := client.Sys().Revoke(id)
+	if err != nil {
+		return fmt.Errorf("error revoking token from Vault: %s", err)
+	}
+	log.Printf("[DEBUG] Revoked lease: %q from Vault", id)
+	log.Printf("Lease for token %s, removing user token from schema", d.Get("token_id"))
```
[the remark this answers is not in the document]
catsby: Also we should refer to the data we save as `state`, instead of `schema`

catsby (commented):
A few more minor edits and I think we'll be good to go here. Please let me know when you have a chance to look!

vault/resource_terraform_cloud_secret_creds.go
calvn: Would it make sense to always set the ID to `tokenId`, and only `d.Set("lease_id", secret.LeasedID)` if that's non-empty. That way you can call lease revoke/lookup if the lease_id is non-empty. @catsby thoughts on this approach?

Valarissa: I went through and added the `lease_id` attribute. Ran into an issue where `lease_id` needed to be set for every token. I moved it to an empty string if no lease is created.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func deleteTerraformCloudSecretCredsResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	leaseId := d.Get("lease_id").(string)
+
+	organization := d.Get("organization")
+	teamId := d.Get("team_id")
+
+	if organization == "" && teamId == "" {
```
calvn, line 109: You can also check for `leaseId != ""` now that it's an attribute.

vault/resource_terraform_cloud_secret_creds.go
```diff
@@ -0,0 +1,154 @@
...
+	return err
+}
+
+func readTerraformCloudSecretCredsResource(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	leaseId := d.Get("lease_id")
+	organization := d.Get("organization")
+	teamId := d.Get("team_id")
+
+	if organization == "" && teamId == "" {
```
calvn, line 137: Same here, you can now lease perform the lookup if `leaseId != ""` (maybe even get rid of `organization == "" && teamId == ""` since it doesn't really depend on that).

Valarissa: Fixed in latest commit.

vault/resource_terraform_cloud_secret_creds.go
Valarissa: Good idea, fixed in latest commit.

catsby (approved):
👍

Issue #880, opened 2020-10-16
vault_gcp_auth_backend_role create fails when backend uses nested mount path.


lsmith130:
### Terraform Version
```
Terraform v0.13.4
+ provider registry.terraform.io/hashicorp/google v3.43.0
+ provider registry.terraform.io/hashicorp/nomad v1.4.9
+ provider registry.terraform.io/hashicorp/vault v2.14.0
```

### Affected Resource(s)
- vault_gcp_auth_backend_role

### Terraform Configuration Files
```hcl
resource "vault_auth_backend" "gcp" {
  type = "gcp"
  path = "${var.env}/${var.region}/gpc"
}

resource "vault_gcp_auth_backend_role" "master" {
  role = "${var.playback_region}-${var.env}-master"
  type = "iam"

  backend                = vault_auth_backend.gcp.path
  bound_service_accounts = [var.master_service_account]
  token_policies         = [vault_policy.nomad_server.name]
}
```

### Debug Output
https://gist.github.com/lsmith130/d763a5a0cd5b9ccf9a11da438e31613e

### Panic Output
No crash

### Expected Behavior
Should create a vault_gcp_auth_backend_role when backend uses a nested path.

### Actual Behavior
`terraform apply` gave the error:
```
Error: invalid path "auth/test/usc1/gpc/role/usc1-test-master" for GCP auth backend role: Expected 4 parts in path 'auth/test/usc1/gpc/role/usc1-test-master'
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
None

### References
- #459

ls-brentsmith:
Also issue https://github.com/hashicorp/terraform-provider-vault/issues/532

Issue #532, opened 2019-09-09
vault_gcp_auth_backend_role:  "Expecdted 4 parts in path 'auth/gcp/foo/role/bar'"


zx8:
Caused by https://github.com/terraform-providers/terraform-provider-vault/pull/517

```
Terraform v0.12.8
+ provider.vault v2.3.0
```

The linked PR introduced a bug with multiple `gcp` backends:

```
Error: invalid path "auth/gcp/foo/role/bar" for GCP auth backend role: Expecdted 4 parts in path 'auth/gcp/foo/role/bar'
```

The PR incorrectly assumes there will be 4 `/`-separated components to the path. This is correct if you use a single auth backend (e.g. `auth/gcp`), but breaks when using multiple GCP auth backends (e.g. `auth/gcp/project1` and `auth/gcp/project2`).

[Nit: Also `Expecdted` is spelled wrong]

zx8:
Ping @julianvmodesto

KeisukeYamashita:
@julianvmodesto I have several GCP project and having the same problem.

Is there a reason that auth backed path has to be in 4 parts?

PR #461, opened 2019-07-08
aws_auth_backend_client supports nested paths

megan07:
Addresses #459 where the aws_auth_backend_client currently expects that the backend path for the aws auth backend client will not be nested.

Before Fix:
```
--- FAIL: TestAccAWSAuthBackendClient_nested (0.05s)
    testing.go:568: Step 0 error: errors during apply:

        Error: expected "auth/aws-8097158027947488229/nested/config/client" to have 4 pieces, has 5

          on /var/folders/nj/n5hr70hd2dx45qb__6wl10nc0000gp/T/tf-test019218845/main.tf line 8:
          (source code not available)


    testing.go:629: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.

        Error: errors during refresh: expected "auth/aws-8097158027947488229/nested/config/client" to have 4 pieces, has 5

        State: <nil>
FAIL
```
After Fix:
```=== RUN   TestAccAWSAuthBackendClient_nested
--- PASS: TestAccAWSAuthBackendClient_nested (0.15s)
PASS
```

PR #970, opened 2021-02-03
Updated gcp_auth_backend_role to fix the provider issue 880

vsharma-ampion:
This pull request cover fix for the issue 880 (https://github.com/hashicorp/terraform-provider-vault/issues/880).

### DIFF
--- a/vault/resource_gcp_auth_backend_role.go
+++ b/vault/resource_gcp_auth_backend_role.go
@@ -3,11 +3,17 @@ package vault
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


ls-brentsmith:
also would resolve issue https://github.com/hashicorp/terraform-provider-vault/issues/880

ls-brentsmith:
also would resolve issue https://github.com/hashicorp/terraform-provider-vault/issues/532

calvn:
Thanks for working on this! Can you add an acceptance test for this change, similar to the one in https://github.com/hashicorp/terraform-provider-vault/pull/461 ?

PR #984, opened 2021-02-26
add Import section to vault_ssh_secret_backend_ca docs

djcrabhat:
Was having a stubborn problem with an ssh backend I was importing, found that it _is_ possible to import a `vault_ssh_secret_backend_ca`, it was just an undocumented feature.  

`vault_ssh_secret_backend_ca` using `schema.ImportStatePassthrough`: https://github.com/hashicorp/terraform-provider-vault/blob/3a348eac1c3cacbefbee3166e18f03f9c59824b5/vault/resource_ssh_secret_backend_ca.go#L17


djcrabhat:
Oh actually, this caused my CA to get trashed and be recreated on next terraform apply (thought generate_keys going from nil->true was a change).  Maybe instead of adding the import section, should add a warning against doing what I just did 😆

PR #1054, opened 2021-05-17
Fix bug where entity values not removed

jasonodonnell:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
resource/vault_identity_entity: Fix bug where values are not removed if removed from file
```

If values are removed from the terraform file, update sees the change but sends no values to Vault. This results in no value changes on the Vault side but Terraform says it succeeded. I've updated the code slightly to use `HasChange`.

Output from acceptance testing:

```
$ VAULT_TOKEN=root TF_ACC=true go test -v -run TestAccIdentityEntityUpdate
=== RUN   TestAccIdentityEntityUpdate
vault_identity_entity.entity:
  ID = 880aae6c-8253-d7ed-ca74-1a51cc875126
  provider = provider.vault
  disabled = false
  external_policies = false
  metadata.% = 1
  metadata.version = 2
  name = test-entity-495886142902770503-2
  policies.# = 2
  policies.1785148924 = test
  policies.326271447 = dev
vault_identity_entity.entity:
  ID = 880aae6c-8253-d7ed-ca74-1a51cc875126
  provider = provider.vault
  disabled = false
  external_policies = false
  metadata.% = 1
  metadata.version = 2
  name = test-entity-495886142902770503-2
  policies.# = 2
  policies.1785148924 = test
  policies.326271447 = dev
vault_identity_entity.entity:
  ID = 880aae6c-8253-d7ed-ca74-1a51cc875126
  provider = provider.vault
  disabled = false
  external_policies = false
  metadata.% = 1
  metadata.version = 2
  name = test-entity-495886142902770503-2
  policies.# = 2
  policies.1785148924 = test
  policies.326271447 = dev
vault_identity_entity.entity:
  ID = 880aae6c-8253-d7ed-ca74-1a51cc875126
  provider = provider.vault
  disabled = false
  external_policies = false
  metadata.% = 1
  metadata.version = 2
  name = test-entity-495886142902770503-2
  policies.# = 2
  policies.1785148924 = test
  policies.326271447 = dev
vault_identity_entity.entity:
[...]

PR #1061, opened 2021-05-25
Fix bug where group values not removed

basingh:
### Community Note

Relates
This change is similar to PR (https://github.com/hashicorp/terraform-provider-vault/pull/1054) but just for `vault_identity_group`. 

```release-note
resource/vault_identity_group: Fix bug where metadata values are not removed if removed from file
```
If values are removed from the terraform file, update sees the change but sends no values to Vault. This results in no value changes on the Vault side but Terraform says it succeeded. Updated the code slightly to use HasChange.

### DIFF
--- a/vault/resource_identity_group.go
+++ b/vault/resource_identity_group.go
@@ -110,26 +110,41 @@ func identityGroupResource() *schema.Resource {
 	}
 }
 
-func identityGroupUpdateFields(d *schema.ResourceData, data map[string]interface{}) error {
-	if name, ok := d.GetOk("name"); ok {
-		data["name"] = name
-	}
-
-	if externalPolicies, ok := d.GetOk("external_policies"); !(ok && externalPolicies.(bool)) {
-		data["policies"] = d.Get("policies").(*schema.Set).List()
-	}
-
-	// Member groups and entities can't be set for external groups
-	if d.Get("type").(string) == "internal" {
-		data["member_group_ids"] = d.Get("member_group_ids").(*schema.Set).List()
-
-		if externalMemberEntityIds, ok := d.GetOk("external_member_entity_ids"); !(ok && externalMemberEntityIds.(bool)) {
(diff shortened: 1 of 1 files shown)


vault/resource_identity_group.go
```diff
@@ -110,26 +110,41 @@ func identityGroupResource() *schema.Resource {
...
 		}
-	}
 
-	if metadata, ok := d.GetOk("metadata"); ok {
-		data["metadata"] = metadata
+		if metadata, ok := d.GetOk("metadata"); ok {
+			data["metadata"] = metadata
+		}
+	} else {
+		if d.HasChanges("name", "external_policies", "policies", "metadata") {
```
jasonodonnell, line 136: Looks like `member_entity_ids` and `member_group_ids` are missing from this function call. They will also need to be set below.

vault/resource_identity_group.go
```diff
@@ -110,26 +110,41 @@ func identityGroupResource() *schema.Resource {
...
+			data["name"] = d.Get("name")
+			data["metadata"] = d.Get("metadata")
+			data["policies"] = d.Get("policies").(*schema.Set).List()
+
+			// Edge case where if external_policies is true, no policies
+			// should be configured on the entity.
+			data["external_policies"] = d.Get("external_policies").(bool)
+			if data["external_policies"].(bool) {
+				data["policies"] = nil
+			}
```
jasonodonnell, line 146: If `external_member_entity_ids` is set, we want to also do something similar like we did for `policies`. Basically if `external_member_entity_ids` is true, `member_entity_ids` should be nil.

jasonodonnell: Doc: https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/identity_group

basingh: thanks @jasonodonnell , this has been updated.

jasonodonnell (approved):
LGTM, thanks!

PR #1065, opened 2021-06-01
Migrate most GitHub HashiBot behaviors to GitHub Actions

bflad:
GitHub HashiBot is being deprecated and this replaces behaviors with equivalent workflows in GitHub Actions. The `pull_request_size_labeler` behavior will be handled once an upstream enhancement is merged (or if we decide to use a fork with the change).

### Community Note

```release-note
NONE
```

Output from acceptance testing: N/A (hashibot configuration)
```

### DIFF
--- a/.github/labeler-pull-request-triage.yml
+++ b/.github/labeler-pull-request-triage.yml
@@ -1,0 +1,6 @@
+dependencies:
+  - .github/dependabot.yml
+  - go.mod
+  - go.sum
+documentation:
+  - website/**/*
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: .github/labeler-pull-request-triage.yml [hunk 1]
```diff
@@ -0,0 +1,6 @@
+dependencies:
+  - .github/dependabot.yml
```
tvoran, line 2: Does it matter that this file doesn't exist in this repo currently?

tvoran (approved):

.github/labeler-pull-request-triage.yml
bflad: Nope! It's just glob matching for if/when that file might be created.

PR #1107, opened 2021-07-21
fix: propagate the provider's namespace throughout

benashz:
* ensure that all Vault resources are provisioned under the namespace of
  the provider.

### Community Note

Relates OR Closes #0000

```release-note
NONE
```

```
$ make dev testacc TF_ACC_ENTERPRISE=1 TESTARGS='-run \^TestAccAppRoleAuthBackendRoleSecretI'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run \^TestAccAppRoleAuthBackendRoleSecretI -timeout 120m

[...]

=== RUN   TestAccAppRoleAuthBackendRoleSecretID_basic
--- PASS: TestAccAppRoleAuthBackendRoleSecretID_basic (0.24s)
=== RUN   TestAccAppRoleAuthBackendRoleSecretID_wrapped
--- PASS: TestAccAppRoleAuthBackendRoleSecretID_wrapped (0.23s)
=== RUN   TestAccAppRoleAuthBackendRoleSecretID_wrapped_namespace
--- PASS: TestAccAppRoleAuthBackendRoleSecretID_wrapped_namespace (0.34s)
=== RUN   TestAccAppRoleAuthBackendRoleSecretID_full
--- PASS: TestAccAppRoleAuthBackendRoleSecretID_full (0.22s)
PASS

[...]
```


calvn (approved):

vault/resource_approle_auth_backend_role_secret_id_test.go
```diff
@@ -55,6 +57,39 @@ func TestAccAppRoleAuthBackendRoleSecretID_wrapped(t *testing.T) {
...
+
+	namespacePath := acctest.RandomWithPrefix("test-namespace")
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { util.TestEntPreCheck(t) },
+		Providers: testProviders,
+		CheckDestroy: func(s *terraform.State) error {
+			if err := testAccCheckAppRoleAuthBackendRoleSecretIDDestroy(s); err != nil {
+				return err
+			}
+			return testNamespaceDestroy(namespacePath)(s)
```
tvoran, line 72: Does this actually delete the namespace created for this test? I'm seeing leftover test-namespace-* namespaces after running this test:

```
$ vault namespace list                                      
No namespaces found

$ make dev testacc TF_ACC_ENTERPRISE=1 TESTARGS='-run \^TestAccAppRoleAuthBackendRoleSecretI -count=1'
[...]

$ vault namespace list                                                                                
Keys
----
test-namespace-2112513148285852065/
```

tvoran (approved):
Looks good, just a question in the test.

vault/resource_approle_auth_backend_role_secret_id_test.go
benashz: Yeah I noticed the same thing, it does seem to `queue` the namespace for deletion, at least that's what the Vault response says when `TF_LOG=DEBUG` is set.

benashz: I confirmed with core that it is an async operation, and it is possible to list `queued for delete` namespaces.

benashz: I believe the `destroy` is handled  in `resource.Test()`.

tvoran: Hmm, the test namespace hasn't been deleted from my vault yet. And looking at the debug output, I wonder if `testNamespaceDestroy()` is passing because the X-Vault-Namespace header is included in the GET request to sys/namespaces, so it's looking for namespace `foo` inside namespace `foo`, essentially?

benashz: So that means we probably have an issue in `destroy` when the provider is configured with a `namespace`.

tvoran: Yep, I think you're right, and I also see a namespace leak when running `TestNamespace_basic` on master as well. So something we'll want to fix in the future, but not worth holding this PR up over.

PR #1110, opened 2021-07-22
Benashz/vault 1065 drop vendor

benashz:
In order to lessen the burden during code reviews we have decided to no longer `vendor` this project.


tomhjp (approved):
I guess this will cause conflicts for a good number of open PRs but they should be easy to resolve.

tvoran (approved):

PR #960, opened 2021-01-29
Fix bug where provider_config only used strings

jasonodonnell:
A bug was recently introduced with the new `provider_config` parameter because the `TypeMap` was casting everything to a string. This parameter can have mixed types, like when configuring GSuite.

This updates the `provider_config` to support all known `provider_config` settings.

Currently the tests pass but when testing gsuite, the backend actually tries to open and verify the JSON service account file. Since we're using a remote docker host which we don't have access to, its currently not possible to check this without auth engine changes or switching to a machine executor. I'll circle back to this next week and wrap up the testing changes.

Fixes #957.

PR #1117, opened 2021-07-26
Add state upgrade for provider_config

### LINKED ISSUE #1112: resource/vault_jwt_auth_backend: Error: missing expected [
v2.22.0 included a fix for `resource/vault_jwt_auth_backend`( #960 ) where the schema type for `provider_config` was not sufficient for non-string types. This schema was converted to a TypeSet which allows for more flexibility in types.

Due to this change, a bug was introduced where this resource is no longer backwards compatible with `vault_jwt_auth_backend` prior to v2.22.0. This is due to the state file not having the correct type for this value.

We're looking into migration fixes for this resource to automatically convert the types.

jasonodonnell:
When changing types of a schema resource in #960 I introduced a small bug where Terraform would complain about unexpected state if this resource was being used. When changing the types of schema, a state upgrade is necessary to update the state file during the plan phase. This PR adds the StateUpgrader for the `provider_config` to automatically update the state file for the user.

Fixes #1112.

### DIFF
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
@@ -22,6 +23,7 @@ func jwtAuthBackendResource() *schema.Resource {
 
 		CustomizeDiff: jwtCustomizeDiff,
 
+		SchemaVersion: 1,
 		Schema: map[string]*schema.Schema{
 
 			"path": {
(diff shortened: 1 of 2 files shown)


vault/resource_jwt_auth_backend.go
```diff
@@ -131,7 +133,176 @@ func jwtAuthBackendResource() *schema.Resource {
...
+	}
+}
+
+func resourceJwtAuthStateUpgradeV0(rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+
+	// The provider_config schema changed types from TypeMap to
+	// TypeSet, so we require a state upgrade. The values in this
+	// schema are already strings, so we just need to convert the
+	// non-string values.
+	if rawConfig, ok := rawState["provider_config"]; ok {
```
calvn, line 265: Nit: Can you flip the check here to `!ok` and return early so we don't wrap the rest of the logic within an indented block?

vault/resource_jwt_auth_backend.go
```diff
@@ -131,7 +133,176 @@ func jwtAuthBackendResource() *schema.Resource {
...
+		if rawConfig == nil {
+			rawState["provider_config"] = make([]map[string]interface{}, 0)
```
calvn, line 267: Do you have to return a zero length slice here, or can it be nil?

PR #1118, opened 2021-07-26
Add support for all provider_config types

### LINKED ISSUE #1112: resource/vault_jwt_auth_backend: Error: missing expected [

jasonodonnell:
Originally we decided to change `provider_config` to a TypeSet, to support different types of values for this config. This unfortunately leaves users to migrate their TF to use the new type and requires the provider to do state upgrades. Additionally issues were uncovered using TypeSet where the provider was sending default type values to Vault, which resulted in configuring the auth method incorrectly.

This PR proposes a different way of configuring `provider_config` by keeping the `TypeMap` but adding conversion code when certain parameters are found in the configuration. The UX is a little weird but a lot of issues are solved by using this method and users won't need to migrate to the new type.

I found some weird issues when testing this with Terraform itself and in Vault. It's difficult to test this auth method with the full range of `provider_config` values due to credential files being required when configuring it. Vault will try to open the credentials file and will result in an error if it's not there. It's possible to send the escaped JSON credentials file to Vault but after trying for a while, I could to get it to work correctly.

There's also an issue with a race condition when deleting test cases in the testing framework. Please see the comment on the skipped test describing what I found.

Fixes #1112.

### DIFF
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
@@ -280,6 +281,29 @@ func jwtAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
 
 }
 
+func convertProviderConfigValues(input map[string]interface{}) (map[string]interface{}, error) {
+	newConfig := make(map[string]interface{})
+	for k, v := range input {
+		switch k {
+		case "fetch_groups", "fetch_user_info":
+			valBool, err := strconv.ParseBool(v.(string))
+			if err != nil {
+				return nil, fmt.Errorf("could not convert %s to bool: %s", k, err)
+			}
+			newConfig[k] = valBool
+		case "groups_recurse_max_depth":
+			valInt, err := strconv.ParseInt(v.(string), 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("could not convert %s to int: %s", k, err)
+			}
+			newConfig[k] = valInt
+		default:
+			newConfig[k] = v.(string)
+		}
+	}
+	return newConfig, nil
+}
+
 func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*api.Client)
 
@@ -288,9 +312,19 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	configuration := map[string]interface{}{}
 	for _, configOption := range matchingJwtMountConfigOptions {
-		// Set the configuration if the user has specified it, or the attribute is in the Diff
-		if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
-			configuration[configOption] = d.Get(configOption)
+		if configOption != "provider_config" {
+			// Set the configuration if the user has specified it, or the attribute is in the Diff
+			if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
+				configuration[configOption] = d.Get(configOption)
+			}
+		} else {
+			if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
+				newConfig, err := convertProviderConfigValues(d.Get(configOption).(map[string]interface{}))
+				if err != nil {
+					return err
+				}
+				configuration[configOption] = newConfig
+			}
 		}
 	}
 
--- a/vault/resource_jwt_auth_backend.go
+++ b/vault/resource_jwt_auth_backend.go
@@ -312,17 +312,15 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	configuration := map[string]interface{}{}
 	for _, configOption := range matchingJwtMountConfigOptions {
-		if configOption != "provider_config" {
-			// Set the configuration if the user has specified it, or the attribute is in the Diff
-			if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
-				configuration[configOption] = d.Get(configOption)
-			}
-		} else {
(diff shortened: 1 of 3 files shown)


vault/resource_jwt_auth_backend.go
```diff
@@ -288,9 +312,19 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
...
+			if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
+				configuration[configOption] = d.Get(configOption)
+			}
+		} else {
+			if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
+				newConfig, err := convertProviderConfigValues(d.Get(configOption).(map[string]interface{}))
+				if err != nil {
+					return err
+				}
+				configuration[configOption] = newConfig
```
jasonodonnell, line 326: This can probably be simplified. I'll see if I can do that.

vault/resource_jwt_auth_backend.go
```diff
@@ -288,9 +312,19 @@ func jwtAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	configuration := map[string]interface{}{}
 	for _, configOption := range matchingJwtMountConfigOptions {
-		// Set the configuration if the user has specified it, or the attribute is in the Diff
-		if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
-			configuration[configOption] = d.Get(configOption)
+		if configOption != "provider_config" {
```
benashz, line 315: I feel like a switch statement here might make the flow a little bit clearer.

vault/resource_jwt_auth_backend_test.go
```diff
@@ -230,3 +266,134 @@ func TestAccJWTAuthBackend_missingMandatory(t *testing.T) {
...
+		{name: "fetch_user_info", value: "foo", err: true, want: ""},
+	}
+
+	for _, tc := range tests {
+		config := map[string]interface{}{
+			tc.name: tc.value,
+		}
+		actual, err := convertProviderConfigValues(config)
+		if tc.err && err == nil {
+			t.Fatalf("exepcted error, got none for key: %s, value: %s", tc.name, tc.value)
```
benashz, line 298:
```suggestion
			t.Fatalf("expected error, got none for key: %s, value: %s", tc.name, tc.value)
```

vault/resource_jwt_auth_backend_test.go
```diff
@@ -230,3 +266,134 @@ func TestAccJWTAuthBackend_missingMandatory(t *testing.T) {
...
+		} else if !tc.err && err != nil {
+			t.Fatalf("expected no error, got one: %s", err)
+		} else if !tc.err {
+			if actual[tc.name] != tc.want {
+				t.Fatalf("exepcted %s, got %s", tc.want, actual[tc.name])
```
benashz, line 303:
```suggestion
				t.Fatalf("expected %s, got %s", tc.want, actual[tc.name])
```

vault/resource_jwt_auth_backend.go
jasonodonnell: I simplified this in general. Thanks!

website/docs/r/jwt_auth_backend.html.md
```diff
@@ -71,7 +89,7 @@ The following arguments are supported:
 
 * `default_role` - (Optional) The default role to use if none is provided during login
 
-* `provider_config` - (Optional) Provider specific handling configuration
+* `provider_config` - (Optional) Provider specific handling configuration. All values should be a string, but the provider will convert to the appropriate type when configuring Vault.
```
tvoran, line 92: Since it seems like setting these parameters as non-strings does work, we could soften this language if we like:

```suggestion
* `provider_config` - (Optional) Provider specific handling configuration. All values may be strings, and the provider will convert to the appropriate type when configuring Vault.
```

vault/resource_jwt_auth_backend.go
```diff
@@ -280,6 +281,29 @@ func jwtAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
 
 }
 
+func convertProviderConfigValues(input map[string]interface{}) (map[string]interface{}, error) {
+	newConfig := make(map[string]interface{})
+	for k, v := range input {
+		switch k {
```
calvn, line 287: Nit: you can set (and check) against `v` outside of the switch to avoid type checking inside it via `val, ok := v.(string)`.

calvn (approved):
Looks good!

benashz (approved):
LGTM!

vault/resource_jwt_auth_backend.go
jasonodonnell: Done!

tvoran (approved):

Keramblock:
I see, that this was fixed 14 days ago and last release was 18 days ago. So I assume that was not fixed in it. Could you please provide deadline for next release?

tvoran:
Hi @Keramblock, it'll be included in this month's release. Basically everything in the Unreleased section of the changelog will go into the release: https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md

Keramblock:
Got it, but when it will be released?

Keramblock:
Hi, seems that now it see diff every apply even if there is nothing to change:
Screenshot 2021-08-25 at 15 10 29

sdzhinovski:
I have the same problem as @Keramblock above. I can add that the results are inconsistent. Roughly 1 of 5 is "No changes", most of the rest are for change of 3 attributes, some are for 4 attributes, and some for 2.

tvoran:
Hey @sdzhinovski, that behavior sounds odd, so I'd suggest opening a new issue with as much detail as you can provide (easier to track than on closed PRs).

PR #1123, opened 2021-07-29
fix: support importing okta_auth_backend resource

### LINKED ISSUE #796: support import for vault_okta_auth_backend
Several other auth backends support importing, would like to have OKTA auth backend support imports as well

### Terraform Version

0.12.26

### Affected Resource(s)

- vault_okta_auth_backend

```hcl
terraform {
  required_version = "~> 0.12.23"

  required_providers {
    vault = "~> 2.11.0"
  }

#  backend remote {}
}

provider vault {
  address               = var.address
  token                 = var.token
  token_name            = var.token_name
  ca_cert_file          = var.ca_cert_file
  ca_cert_dir           = var.ca_cert_dir
  skip_tls_verify       = var.skip_tls_verify
  max_lease_ttl_seconds = var.max_lease_ttl_seconds
  max_retries           = var.max_retries
  namespace             = var.namespace
}

variable address {
  type        = string
  description = "Origin URL of the Vault server"
  default     = null
}

variable token {
  type        = string
  description = "Vault token that will be used by Terraform to authenticate"
  default     = null
}

variable token_name {
  type        = string
  description = "Token name that will be used by Terraform when creating the child token"
  default     = null
}

variable ca_cert_file {
  type        = string
  description = "Path to a file on local disk that will be used to validate the certificate presented by the Vault server"
  default     = null
}

variable ca_cert_dir {
  type        = string
  description = "Path to a directory on local disk that contains one or more certificate files that will be used to validate the certificate presented by the Vault server"
  default     = null
}

variable skip_tls_verify {
  type        = bool
  description = "Set this to `true` to disable verification of the Vault server's TLS certificate"
  default     = false
}

variable max_lease_ttl_seconds {
  type        = number
  description = "Used as the duration for the intermediate Vault token Terraform issues itself, which in turn limits the duration of secret leases issued by Vault"
  default     = 1200
}

variable max_retries {
  type        = number
  description = "Used as the maximum number of retries when a 5xx error code is encountered"
  default     = 2
}

variable namespace {
  type        = string
  description = "Set the namespace to use"
  default     = null
}

variable okta {
  type = object({
    path         = string
    organization = string
    token        = string
    description  = string
  })
  description = "Okta auth method configuration"
  default     = null
}

resource "vault_okta_auth_backend" "okta" {
  count = var.okta != null ? 1 : 0

  path         = var.okta.path
  organization = var.okta.organization
  token        = var.okta.token
  description  = var.okta.description
}
```

### Debug Output

https://gist.github.com/mengesb/12c2a9ac63eb69d7cb491cdce670cc6b#file-vault_okta_auth_backend

### Panic Output

none

### Expected Behavior

import

### Actual Behavior

resource doesn't support importing

### Steps to Reproduce

1. `terraform import 'vault_okta_auth_backend.gcp[0]' okta`

### Important Factoids

I tried using the `auth_okta_<id>` as the import source resource to replace `okta` but that didn't work either

### References

- #683 
- #696 
- https://www.terraform.io/docs/providers/vault/r/okta_auth_backend.html

benashz:
### Community Note

Closes #796

```release-note
resource/okta_auth_backend`: Support importing preexisting resource 
```

```
make dev testacc TESTARGS='-run \^TestAccOktaAuthBackend_'
TF_ACC=1 go test $(go list ./...) -v -run \^TestAccOktaAuthBackend_ -timeout 120m

[...]

=== RUN   TestAccOktaAuthBackend_basic
--- PASS: TestAccOktaAuthBackend_basic (0.42s)
=== RUN   TestAccOktaAuthBackend_import
--- PASS: TestAccOktaAuthBackend_import (0.33s)
=== RUN   TestAccOktaAuthBackend_invalid_ttl
--- PASS: TestAccOktaAuthBackend_invalid_ttl (0.02s)
=== RUN   TestAccOktaAuthBackend_invalid_max_ttl
--- PASS: TestAccOktaAuthBackend_invalid_max_ttl (0.02s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     2.393s
```


vault/resource_okta_auth_backend.go
```diff
@@ -79,14 +83,18 @@ func oktaAuthBackendResource() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    false,
 				Optional:    true,
+				Default:     "0",
```
tomhjp, line 86: What does 0 mean for ttl and max_ttl?

vault/resource_okta_auth_backend.go
```diff
@@ -204,6 +212,14 @@ func oktaAuthBackendResource() *schema.Resource {
 	}
 }
 
+func parseTTLForState(i interface{}) string {
+	d, err := parseutil.ParseDurationSecond(i)
+	if err != nil {
+		return i.(string)
```
tomhjp, line 218: Is this definitely the right behaviour? I'm always very wary about throwing away `err` values.

vault/resource_okta_auth_backend.go
```diff
@@ -263,12 +283,15 @@ func oktaAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
+	d.Set("path", path)
```
tomhjp, line 286: Should we be checking the error return values for these `Set` calls?

benashz: Good question, taking a look now.

vault/resource_okta_auth_backend.go
benashz: Yeah, I agree with you, but unfortunately `SchemaStateFunc` signature does not include returning `error`.  I suppose we could at least log the error, not sure if a panic is warranted here.

ref signature:
```
type SchemaStateFunc func(interface{}) string
```

vault/resource_okta_auth_backend.go
benashz: Fixed in [8c19db6](https://github.com/hashicorp/terraform-provider-vault/pull/1123/commits/8c19db6d726cbbf8156ba2105bd84ef0dd293f6a)

vault/resource_okta_auth_backend.go
benashz: It's not completely clear from the docs what a value of 0 means, only that it is the default https://www.vaultproject.io/api-docs/auth/okta#token_ttl

vault/resource_okta_auth_backend.go
benashz: I think adding a `ValidateFunc` should mitigate this issue. Changes forthcoming.

benashz: addressed in [641159f](https://github.com/hashicorp/terraform-provider-vault/pull/1123/commits/641159fbcad86ace9bcca6d7d0816a20f46a09fe)

vault/resource_okta_auth_backend.go
```diff
@@ -76,17 +80,23 @@ func oktaAuthBackendResource() *schema.Resource {
...
 
 			"ttl": {
-				Type:        schema.TypeString,
-				Required:    false,
-				Optional:    true,
-				Description: "Duration after which authentication will be expired",
+				Type:         schema.TypeString,
+				Required:     false,
+				Optional:     true,
+				Default:      "0",
```
tvoran, line 86: If `ttl` and `max_ttl` are optional, should they even have a default value here on the terraform side?

benashz: Good question, without the defaults Terraform will detect a difference between the configured auth backend after an `apply`.  I will investigate further to see if there is a better approach that satisfies this use case.

vault/resource_okta_auth_backend_test.go
```diff
@@ -3,43 +3,124 @@ package vault
...
-		CheckDestroy: testAccOktaAuthBackend_Destroyed(path),
-		Steps: []resource.TestStep{
-			{
-				Config: testAccOktaAuthConfig_basic(path, organization),
-				Check: resource.ComposeTestCheckFunc(
-					testAccOktaAuthBackend_InitialCheck,
-					testAccOktaAuthBackend_GroupsCheck(path, "dummy", []string{"one", "two", "default"}),
-					testAccOktaAuthBackend_UsersCheck(path, "foo", []string{"dummy"}, []string{}),
-				),
+	tests := []struct {
```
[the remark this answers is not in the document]
benashz: Yeah, I totally feel that tension as well. I'm happy to go to separate test cases.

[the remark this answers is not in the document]
benashz: Fixed in [8caeb94](https://github.com/hashicorp/terraform-provider-vault/pull/1123/commits/8caeb9403cc06b84c17d5bae7da664276afe1426)

vault/resource_okta_auth_backend.go
benashz: Fixed in [185d8311](https://github.com/hashicorp/terraform-provider-vault/pull/1123/commits/185d8311c0c99e9fcff665c128f310271bdddaae)

vault/resource_okta_auth_backend.go
benashz: Used a computed value from Vault instead of the default.

vault/resource_okta_auth_backend.go
```diff
@@ -288,8 +340,38 @@ func oktaAuthBackendRead(d *schema.ResourceData, meta interface{}) error {
...
+	config, err := client.Logical().Read(oktaConfigEndpoint(path))
+	if err != nil {
+		return err
+	}
 
+	// map schema config to okta auth params.
+	fieldMap := map[string]string{
+		"organization": "organization",
+		"ttl":          "token_ttl",
+		"max_ttl":      "token_max_ttl",
```
tvoran, line 361: Do we need to add the other parameters here? Like `base_url`, `bypass_okta_mfa`, and `token`? Those come up as blank when I import an existing okta mount.

benashz: Fixed in [4ce4298](https://github.com/hashicorp/terraform-provider-vault/pull/1123/commits/4ce429803284ec3d38a76fcaa1d165ee69ed0e28)

vault/resource_okta_auth_backend.go
```diff
@@ -76,17 +80,23 @@ func oktaAuthBackendResource() *schema.Resource {
 			},
 
 			"ttl": {
-				Type:        schema.TypeString,
-				Required:    false,
-				Optional:    true,
-				Description: "Duration after which authentication will be expired",
+				Type:         schema.TypeString,
+				Computed:     true,
```
tvoran, line 84: So one behavior I'm noticing with these set to `Computed: true` is that if `ttl` or `max_ttl` is removed from the terraform manifest, those values are never unset in Vault. I'm not sure that's expected terraform behavior or not, so I'd be fine with setting a `Default: "0"` like you had it before instead.

tvoran (approved):

vault/resource_okta_auth_backend.go
benashz: Interesting, I guess it makes sense for  `computed` values to never be persisted to the TF state. We can go with defaults for now, once we deprecate the `ttl` and `max_ttl` string inputs we should be in line with the other `auth_backend` resources that do not suffer from this issue.

PR #1145, opened 2021-08-11
feat: support for VAULT_TLS_SERVER_NAME

### LINKED ISSUE #300: feature request: support VAULT_TLS_SERVER_NAME in provider config
### Terraform Version
0.11.11

### Overview

to connect to a vault server with TLS, currently one may require setting VAULT_TLS_SERVER_NAME

currently, there is not an equivalent option to set this using the provider config block

I would prefer to not have to control my environment, as currently doing a plan and apply strategy according to best practices would ideally capture this configuration in the plan, but without it being a variable, it cannot be, and thus has to be tracked across orchestration specifically.

### Example

```hcl
provider "vault" {
  version = "~> 1.4"
  address = "https://${data.consul_service.vault.service.0.address}:${data.consul_service.vault.service.0.port}"
  token = "${var.vault_token}"
  ca_cert_file = "${var.vault_ca_cert_file}"
  client_auth {
    cert_file = "${var.vault_client_cert_file}"
    key_file = "${var.vault_client_key_file}"
  }
  tls_server_name = "vault.service.consul"
}
```

see https://github.com/terraform-providers/terraform-provider-vault/blob/be535c17c58e0954df84a9e204952f8526a3767a/vault/provider.go

greut:
### Community Note

Closes #300

```release-note
FEATURES:
* `provider`: Add support for `tls_server_name` to use as the SNI host when connecting via TLS.
```

```

...
```

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -157,6 +157,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_VERIFY", false),
 				Description: "Set this to true only if the target Vault server is an insecure development instance.",
 			},
+			"tls_server_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VAULT_TLS_SERVER_NAME", false),
+				Description: "Name to use as the SNI host when connecting via TLS.",
+			},
 			"max_lease_ttl_seconds": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


sfc-gh-swinkler:
i was about to make a PR for this as well but seeing how you already made one but it never got approved is troubling...

vault/provider.go
```diff
@@ -157,6 +157,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("VAULT_SKIP_VERIFY", false),
 				Description: "Set this to true only if the target Vault server is an insecure development instance.",
 			},
+			"tls_server_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VAULT_TLS_SERVER_NAME", false),
```
benashz, line 163: Since `tls_server_name`'s type is a string, the default value should probably be `""` rather than a boolean; also we can use the constant from the `api` package for the env variable name.

```suggestion
				DefaultFunc: schema.EnvDefaultFunc(api.EnvVaultTLSServerName, ""),
```

greut: @benashz Thanks for the fix and merge!

Issue #1129, opened 2021-08-02
terraform_cloud_secret_backend documentation error


czembower:
The usage example sites a "path" parameter that throws an error when a run is attempted:

`An argument named "path" is not expected here.`

Perhaps the intention was to use the "backend" parameter in the example code.

tvoran:
Fixed in #1158

PR #1158, opened 2021-08-25
Fix sample code in documentation for `vault_terraform_cloud_secret_backend`

mechastorm:
This was reported in #1129.
The sample code uses `path` argument which does not exist for the resource `vault_terraform_cloud_secret_backend`. It should be instead be using the `backend` argument


tvoran (approved):
Thanks!

PR #1170, opened 2021-09-14
[VAULT-3157] Add X-Vault-Index header functionality

vinay-gopalan:
### Community Note

Relates OR Closes #0000
https://hashicorp.atlassian.net/browse/VAULT-3157

```release-note
Add X-Vault-Index header to ensure that the state resulting from a write request that modifies storage is visible to a subsequent request.
```

```

...
```

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -764,6 +764,10 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	client.SetMaxRetries(d.Get("max_retries").(int))
 
+	var newState string
+	//Leverage Vault helpers for eventual consistency on login
+	client = client.WithResponseCallbacks(api.RecordState(&newState))
+
 	// Try an get the token from the config or token helper
 	token, err := providerToken(d)
 	if err != nil {
@@ -855,6 +859,8 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	// Set the token to the generated child token
 	client.SetToken(childToken)
+
+	client = client.WithRequestCallbacks(api.RequireState(newState)).WithResponseCallbacks()
 
 	// Set the namespace to the requested namespace, if provided
 	namespace := d.Get("namespace").(string)


vault/provider.go
```diff
@@ -705,6 +706,13 @@ func providerToken(d *schema.ResourceData) (string, error) {
 	return strings.TrimSpace(token), nil
 }
 
+func RecordMultipleStates(lastStates ...string) func(*api.Response) {
+	return func(resp *api.Response) {
+		newState := resp.Header.Get("X-Vault-Index")
+		lastStates = vault.MergeStates(lastStates, newState)
```
ncabatoff, line 712: May need locking here if the client is used by multiple goroutines.

vault/provider.go
```diff
@@ -764,6 +772,10 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 
 	client.SetMaxRetries(d.Get("max_retries").(int))
 
+	var lastStates []string
+	//Leverage Vault helpers for eventual consistency on login
+	client = client.WithResponseCallbacks(RecordMultipleStates(lastStates...))
```
ncabatoff, line 777: Remember that client.Clone won't preserve callbacks.  (Might not matter for your PR, just calling it out in case.)

vault/provider.go
vinay-gopalan: Included locking with a newly initialized mutex

vault/provider.go
```diff
@@ -705,6 +706,16 @@ func providerToken(d *schema.ResourceData) (string, error) {
 	return strings.TrimSpace(token), nil
 }
 
+func RecordMultipleStates(lastStates ...string) func(*api.Response) {
+	return func(resp *api.Response) {
+		var m sync.Mutex
```
tomhjp, line 711: I think we need to move this variable declaration outside the anonymous function, otherwise all invocations of the function will have their own mutex variable and not actually synchronise.

vault/provider.go
```diff
@@ -705,6 +706,16 @@ func providerToken(d *schema.ResourceData) (string, error) {
...
+		newState := resp.Header.Get("X-Vault-Index")
+		m.Lock()
+		lastStates = api.MergeStates(lastStates, newState)
```
tomhjp, line 714: I don't think this will actually update the `lastStates` slice in `providerConfigure`'s scope. Maybe worth adding a test to verify either way.

vault/provider.go
calvn: It looks like both `WithResponseCallbacks` and `WithRequestCallbacks` makes a shallow clone of the client, so I'm not sure that it will carry over things like headers set right above.

vault/provider.go
calvn: Yeah, `lastState` will not have the updated value outside of this func (`RecordState ` accepts a string pointer which is why it works there).

vinay-gopalan: @calvn is there a way to make this work with a string slice? Or are string pointers our best bet here? (can we pass arrays by pointers in Go?)

vault/provider.go
vinay-gopalan: @calvn and @ncabatoff I updated the `providerConfigure` method to return the client plus the underlying state variable referenced in the callbacks, so if we need to Clone, we can easily give the clone callbacks that make use of the same state reference. That should take care of the cloning headers aspect.

Syncing with Ben today at 1PM PDT to move forward with that

vault/provider.go
vinay-gopalan: Good call! Updated and moved the declaration outside the anonymous function but still under `RecordMultipleStates`

vault/provider.go
calvn: Though this is not common, we could have as the argument type `*[]string` and we pass in the pointer to the slice for `lastStates` whenever this is called.

vault/provider.go
```diff
@@ -856,12 +871,19 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	// Set the token to the generated child token
 	client.SetToken(childToken)
 
+	client = client.WithRequestCallbacks(api.RequireState(lastStates...)).WithResponseCallbacks()
```
calvn, line 874: Why is this chained like so in there? If this is necessary, can you leave a comment around this line?

vinay-gopalan:
Closing this branch because this work is incorporated in this PR that adds client control support in the form of a `ClientFactory`: https://github.com/hashicorp/terraform-provider-vault/pull/1188

PR #1174, opened 2021-09-21
Avoid token generation if max_ttl is zero

jorgemarey:
### Community Note

```release-note
Avoids token generation at provider configuration time if max_lease_ttl_seconds is set to zero
```

Hi, I'm making a terraform that deploys vault and configures it, just for testing purposes, my idea is to run this every time we need to test something and then destroy it.

I found out that vault makes this difficult. On this configuration:

```
module "bootstrap1" {
  source = "../bootstrap"

  ...
}

resource "vault_mount" "kv" {
  path        = "secret"
  type        = "kv-v1"
}

resource "vault_generic_endpoint" "test" {
  path                 = "${vault_mount.kv.path}/foo"
  ignore_absent_fields = true
  disable_delete       = true

  data_json = jsonencode({
    value = "bar"
  })
}

provider "vault" {
  address               = format("https://%s:8200", module.bootstrap1.ip)
  token                 = module.bootstrap1.vault_token
  skip_tls_verify       = true
}
```

Terraform fails because it want's to create a token previous to apply anything.

I changed the code so adding `  max_lease_ttl_seconds = 0` to the provider config would avoid creating a short live token.

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -811,50 +811,53 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		tokenName = "terraform"
 	}
 
-	// In order to enforce our relatively-short lease TTL, we derive a
-	// temporary child token that inherits all of the policies of the
-	// token we were given but expires after max_lease_ttl_seconds.
-	//
-	// The intent here is that Terraform will need to re-fetch any
-	// secrets on each run and so we limit the exposure risk of secrets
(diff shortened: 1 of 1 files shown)


benashz:
This can now be accomplished by setting `skip_child_token` to `true` #775 which be included in the upcoming `3.1.0` release. 

Thank you for contributing to HashiCorp!

PR #1176, opened 2021-09-21
 Add support for case_sensitive_names in the vault_ldap_auth_backend

### LINKED ISSUE #894: vault_ldap_auth_backend does not support setting case_sensitive_names

gothrek22:
Fixes #894

This PR is an extension of @stokkie90 fix that was proposed in https://github.com/hashicorp/terraform-provider-vault/pull/895

I've updated the docs and tests, plus added switched to GetOkExists when checking `case_sensitive_names` as recommended by @catsby.

I've also set the default to `false` to be consistent with Vault's behaviour.

### DIFF
--- a/vault/resource_ldap_auth_backend.go
+++ b/vault/resource_ldap_auth_backend.go
@@ -58,6 +58,7 @@ func ldapAuthBackendResource() *schema.Resource {
 			Type:     schema.TypeBool,
 			Optional: true,
 			Computed: true,
+			Default: false,
 		},
 		"userdn": {
 			Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


gothrek22:
@vinay-gopalan @benashz  I see you did some work in the repo recently, could you guys review this and merge it?

vault/resource_ldap_auth_backend.go
```diff
@@ -204,6 +209,10 @@ func ldapAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 		data["bindpass"] = v.(string)
 	}
 
+	if v, ok := d.GetOkExists("case_sensitive_names"); ok {
+		data["case_sensitive_names"] = v.(bool)
+	}
+
```
benashz, line 215:
```suggestion

```

PR #1212, opened 2021-11-01
Update Changelog to include #1200 changes

vinay-gopalan:
This PR updates the Changelog to include changes from https://github.com/hashicorp/terraform-provider-vault/pull/1200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 IMPROVEMENTS:
 * Upgrade Terraform Plugin SDK to v2
 * Add support for client controlled consistency on Vault Enterprise ([#1188](https://github.com/hashicorp/terraform-provider-vault/pull/1188))
-* `resource_jwt_auth_backend_role`: Add field `disable_bound_claims_parsing` to disable bound claim value parsing, which is useful when values contain commas ([#1200](https://github.com/hashicorp/terraform-provider-vault/pull/1200))
+* `resource/jwt_auth_backend_role`: Add field `disable_bound_claims_parsing` to disable bound claim value parsing, which is useful when values contain commas ([#1200](https://github.com/hashicorp/terraform-provider-vault/pull/1200))
 
 BUGS:
 * `data/gcp_auth_backend_role`: Report an error when attempting to access a nonexistent role. ([#1184](https://github.com/hashicorp/terraform-provider-vault/pull/1184))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 IMPROVEMENTS:
 * Upgrade Terraform Plugin SDK to v2
 * Add support for client controlled consistency on Vault Enterprise ([#1188](https://github.com/hashicorp/terraform-provider-vault/pull/1188))
+* `resource_jwt_auth_backend_role`: Add field to disable `bound_claims` comma separation ([#1200](https://github.com/hashicorp/terraform-provider-vault/pull/1200))
```
benashz, line 5: I wonder if we should add the name of the new field? Also I think we should use the same language  that is used in the `disable_bound_claims_parsing` rather than the bit about comma separation (comma delimiting).

vinay-gopalan: Updated to `Add field `disable_bound_claims_parsing` to disable bound claim value parsing, which is useful when values contain commas`

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 IMPROVEMENTS:
 * Upgrade Terraform Plugin SDK to v2
 * Add support for client controlled consistency on Vault Enterprise ([#1188](https://github.com/hashicorp/terraform-provider-vault/pull/1188))
+* `resource_jwt_auth_backend_role`: Add field `disable_bound_claims_parsing` to disable bound claim value parsing, which is useful when values contain commas ([#1200](https://github.com/hashicorp/terraform-provider-vault/pull/1200))
```
tvoran, line 5: The changelog formatting isn't super consistent, but it looks like more recent releases format entries with `resource/`:

```suggestion
* `resource/jwt_auth_backend_role`: Add field `disable_bound_claims_parsing` to disable bound claim value parsing, which is useful when values contain commas ([#1200](https://github.com/hashicorp/terraform-provider-vault/pull/1200))
```

tvoran (approved):
👍 

(Also a friendly reminder to update the PR description on https://github.com/hashicorp/terraform-provider-vault/pull/1200 with the `disable_bound_claims_parsing` name, since it's still referring to it as `disable_claims_comma_separation`.)

vinay-gopalan:
> 👍
> 
> (Also a friendly reminder to update the PR description on #1200 with the `disable_bound_claims_parsing` name, since it's still referring to it as `disable_claims_comma_separation`.)

Good call! Updated the description in that PR as well :)

benashz:
> > 👍
> > (Also a friendly reminder to update the PR description on #1200 with the `disable_bound_claims_parsing` name, since it's still referring to it as `disable_claims_comma_separation`.)
> 
> Good call! Updated the description in that PR as well :)

Looks like the example usage still has the old field name.

PR #1214, opened 2021-11-02
transform/template: add encode_format and decode_formats options.

benashz:
- add encode_format and decode_formats fields to
  vault_transform_template
- temporarily disable broken resource generation code
- ensure true resource name is used in template tests.

### Community Note

Relates OR Closes #0000

```release-note
resource/transform_template:  Add add encode_format and decode_formats fields
```

```

make dev testacc TESTARGS='-v -test.run TestTemplateName'
TF_ACC=1 go test $(go list ./...) -v -v -test.run TestTemplateName -timeout 20m

[...]

=== RUN   TestTemplateName
--- PASS: TestTemplateName (4.07s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template 
...
```

### DIFF
--- a/generated/resources/transform/template/name_test.go
+++ b/generated/resources/transform/template/name_test.go
@@ -47,14 +47,16 @@ func TestTemplateName(t *testing.T) {
 					resource.TestCheckResourceAttr("vault_transform_template_name.test", "type", "regex"),
 					resource.TestCheckResourceAttr("vault_transform_template_name.test", "pattern", `(\d{4})-(\d{4})-(\d{4})-(\d{4})`),
 					resource.TestCheckResourceAttr("vault_transform_template_name.test", "alphabet", "numerics"),
+					resource.TestCheckResourceAttr("vault_transform_template_name.test", "encode_format", "$1-$2-$3-$4"),
+					resource.TestCheckResourceAttr("vault_transform_template_name.test", "decode_formats.last-four", "$4"),
 				),
 			},
 			{
 				Config: basicConfig(path, "regex",
 					`(\\d{9})`,
 					"builtin/numeric",
-					"$1",
-					`{ "only" = "$1" }`,
+					"",
+					"",
 				),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("vault_transform_template_name.test", "path", path),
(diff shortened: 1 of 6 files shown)


generated/resources/transform/template/name_test.go
```diff
@@ -35,7 +35,12 @@ func TestTemplateName(t *testing.T) {
 		CheckDestroy: destroy,
 		Steps: []resource.TestStep{
 			{
-				Config: basicConfig(path, "regex", `(\\d{4})-(\\d{4})-(\\d{4})-(\\d{4})`, "numerics"),
+				Config: basicConfig(path, "regex",
+					`(\\d{4})-(\\d{4})-(\\d{4})-(\\d{4})`,
+					"numerics",
+					"$1-$2-$3-$4",
```
victorr, line 41: Shouldn't there be a couple of `resource.TestCheckResourceAttr()` calls for the new fields? 

Would it also be useful to test the cases where the fields are not present, since they are optional?

benashz: > Shouldn't there be a couple of `resource.TestCheckResourceAttr()` calls for the new fields?

Good call, the import test was catching those. I can make them explicit.

> 
> Would it also be useful to test the cases where the fields are not present, since they are optional?

The only required fields are `path` and `name`,  but yeah it's a good idea any how.

victorr (approved):
Looks good to me.

website/vault.erb
```diff
@@ -422,7 +422,7 @@
                         </li>
 
                         <li<%= sidebar_current("docs-vault-resource-transform-template") %>>
-                            <a href="/docs/providers/vault/generated/resources/transform/template/name.html">vault_transform_template</a>
+                            <a href="/docs/providers/vault/r/transform_template.html">vault_transform_template</a>
```
tvoran, line 425: I wonder how the other links here are working since they're set to /vault/generated/resources/transform/...

tvoran (approved):

tvoran:
Might as well include the changelog update in this PR too.

website/vault.erb
benashz: Good question, I guess that when the release is deployed to the registry there is some path munging going on. We should probably update all of the links at some point to be as the change above.

Of note, the links under `generated` are broken when testing the website with `make website-test`

PR #1222, opened 2021-11-10
Add a `TestRaftPrecheck` in util to enable Raft tests using env var

vinay-gopalan:
Relates OR Closes #0000

```
release-note: Add a PreCheck to enable Raft tests using an environment variable
```

```
$ make testacc TESTARGS='-v -run TestAccRaftAutopilotConfig_basic'
RUN   TestAccRaftAutopilotConfig_basic
    util.go:126: TF_RAFT_ENABLED is not set, test is applicable only with a Raft Cluster setup
--- SKIP: TestAccRaftAutopilotConfig_basic (0.00s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     (cached)
...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,7 @@ jobs:
             export MYSQL_CONNECTION_USERNAME="root"
             export MYSQL_CONNECTION_PASSWORD="mysql"
             export MONGODB_URL="mongodb://root:mongodb@localhost:27017/admin?ssl=false"
+            export SKIP_RAFT_TESTS="true"
             make testacc TESTARGS='-v'
       - run:
           name: "Run Build"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,8 @@ jobs:
             export MYSQL_CONNECTION_USERNAME="root"
             export MYSQL_CONNECTION_PASSWORD="mysql"
             export MONGODB_URL="mongodb://root:mongodb@localhost:27017/admin?ssl=false"
-            export SKIP_RAFT_TESTS="true"
+            # This will be removed after VAULT-4324 is fixed
+            export SKIP_RAFT_TESTS='true'
             make testacc TESTARGS='-v'
       - run:
           name: "Run Build"
--- a/util/util.go
+++ b/util/util.go
@@ -119,20 +119,6 @@ func TestEntPreCheck(t *testing.T) {
 	}
 }
 
-func TestRaftPreCheck(t *testing.T) {
-	isRaft := os.Getenv("TF_RAFT_ENABLED")
-
-	if isRaft == "" {
-		t.Skip("TF_RAFT_ENABLED is not set, test is applicable only with a Raft Cluster setup")
-	}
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: util/util.go [hunk 1]
```diff
@@ -119,6 +119,20 @@ func TestEntPreCheck(t *testing.T) {
 	}
 }
 
+func TestRaftPreCheck(t *testing.T) {
```
benashz, line 122: I'd like to switch the logic up a bit here by adding a `SKIP_RAFT_TESTS` variable instead. This means we only skip of the variable is set, and that the test will always be run otherwise.

It looks like you might be duplicating code that we are already have:

https://github.com/hashicorp/terraform-provider-vault/blob/66602ebb29019a52613818947a7b7f64b29d7dbe/util/util.go#L100-L107

vault/resource_raft_autopilot_test.go
```diff
@@ -5,15 +5,17 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/hashicorp/terraform-provider-vault/util"
 	"github.com/hashicorp/vault/api"
 )
 
 func TestAccRaftAutopilotConfig_basic(t *testing.T) {
 	resource.Test(t, resource.TestCase{
 		Providers:    testProviders,
-		PreCheck:     func() { testAccPreCheck(t) },
+		PreCheck:     func() { util.TestRaftPreCheck(t) },
```
benashz, line 18: Think this might work, although it has not been tested.
```suggestion
		PreCheck:     func() { 
				util.TestAccPreCheck(t)
				if _, ok ; os.LookupEnv("SKIP_RAFT_TESTS"); ok {
					t.Skip("Warning: SKIP_RAFT_TESTS set, skipping test")
			},
		},
```

benashz (changes requested):

vault/resource_raft_autopilot_test.go
vinay-gopalan: Tested that it works and fixed!

benashz (approved):
LGTM!

benashz (changes requested):
CI not updated

.circleci/config.yml
```diff
@@ -45,6 +45,7 @@ jobs:
             export MYSQL_CONNECTION_USERNAME="root"
             export MYSQL_CONNECTION_PASSWORD="mysql"
             export MONGODB_URL="mongodb://root:mongodb@localhost:27017/admin?ssl=false"
+            export SKIP_RAFT_TESTS="true"
```
benashz, line 48: no need for double quotes here, single quotes are safer in shell if there is no need for variable interpolation, etc.

benashz: Add a TODO comment referencing the Jira `VAULT-...` that we need to drop this when the related Jira is fixed.

benashz (changes requested):

.circleci/config.yml
vinay-gopalan: Added a comment stating `This will be removed after VAULT-4324 is fixed`

benashz (approved):

PR #1320, opened 2022-01-26
Add support for configuring a DB secret's plugin_name

### LINKED ISSUE #1288: database_secret_backend_connection resource missing `plugin_name` field.
Hello,

 I need to create the oracle database engine in Vault as mentionned in https://www.vaultproject.io/api/secret/databases/oracle
I plan to use Terraform and so :  resource "vault_database_secret_backend_connection"

but I can not use :
=> plugin_name  as a configuration parameter (and it's mandatory for Oracle) as it's not defined in the documentation  https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/database_secret_backend_connection

=> And under oracle arguments I also need to define username/password in order for Vault to define resources in the Oracle database but it's the same, the arguments don't exist  

do I have to use =>  resource "vault_generic_endpoint"  ?
Or Did I miss something?
Thanks for your help

benashz:
This PR adds support for setting the `plugin_name` in the `vault_database_secret_backend_connection` resource. Previously all related resources could only be provisioned using a well known default value from Vault.

### Community Note

Closes #1288 

```release-note

```

```
make testacc TESTARGS='-v -test.run TestAccDatabaseSecretBackendConnection_mssql'
TF_ACC=1 go test -v -v -test.run TestAccDatabaseSecretBackendConnection_mssql -timeout 20m ./...

[...]

=== RUN   TestAccDatabaseSecretBackendConnection_mssql
--- PASS: TestAccDatabaseSecretBackendConnection_mssql (17.70s)

```

### DIFF
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -44,25 +44,48 @@ const (
 var (
 	databaseSecretBackendConnectionBackendFromPathRegex = regexp.MustCompile("^(.+)/config/.+$")
 	databaseSecretBackendConnectionNameFromPathRegex    = regexp.MustCompile("^.+/config/(.+$)")
-	dbBackendTypes                                      = []string{
-		dbBackendCassandra,
-		dbBackendCouchbase,
-		dbBackendElasticSearch,
-		dbBackendHana,
-		dbBackendInfluxDB,
-		dbBackendMSSQL,
-		dbBackendMongoDB,
-		dbBackendMongoDBAtlas,
-		dbBackendMySQL,
-		dbBackendMySQLAurora,
-		dbBackendMySQLLegacy,
-		dbBackendMySQLRDS,
-		dbBackendPostgres,
-		dbBackendOracle,
-		dbBackendSnowflake,
-		dbBackendRedshift,
+
+	// map all dbBackends to their corresponding dbEngineInfo's
+	// XXX: This map must be updated whenever a new database engine is added.
+	dbEngines = map[string]*dbEngineInfo{
+		dbBackendCassandra:     {name: dbBackendCassandra},
+		dbBackendCouchbase:     {name: dbBackendCouchbase},
+		dbBackendElasticSearch: {name: dbBackendElasticSearch},
+		dbBackendHana:          {name: dbBackendHana},
+		dbBackendInfluxDB:      {name: dbBackendInfluxDB},
+		dbBackendMSSQL:         {name: dbBackendMSSQL},
+		dbBackendMongoDB:       {name: dbBackendMongoDB},
+		dbBackendMongoDBAtlas:  {name: dbBackendMongoDBAtlas},
+		dbBackendMySQL:         {name: dbBackendMySQL},
+		dbBackendMySQLAurora:   {name: dbBackendMySQLAurora},
+		dbBackendMySQLLegacy:   {name: dbBackendMySQLLegacy},
+		dbBackendMySQLRDS:      {name: dbBackendMySQLRDS},
+		dbBackendPostgres:      {name: dbBackendPostgres},
+		dbBackendOracle:        {name: dbBackendOracle},
+		dbBackendSnowflake:     {name: dbBackendSnowflake},
+		dbBackendRedshift:      {name: dbBackendRedshift},
 	}
+	// dbEngines key values, initialized in init()
+	dbBackendTypes []string
 )
+
+func init() {
+	for k := range dbEngines {
+		dbBackendTypes = append(dbBackendTypes, k)
+	}
+}
+
+type dbEngineInfo struct {
+	name string
+}
+
+func (i *dbEngineInfo) getPluginName(d *schema.ResourceData) string {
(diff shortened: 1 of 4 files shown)


vault/resource_database_secret_backend_connection_test.go
```diff
@@ -956,20 +967,23 @@ func testAccDatabaseSecretBackendConnectionConfig_mssql(name, path, connURL stri
...
-	return fmt.Sprintf(`
+	result := fmt.Sprintf(`
 resource "vault_mount" "db" {
   path = "%s"
   type = "database"
 }
 
 resource "vault_database_secret_backend_connection" "test" {
   backend = vault_mount.db.path
+  plugin_name = "%s"
```
vinay-gopalan, line 978: Since we are explicitly setting `plugin_name` for MSSQL configs, would it be worth testing with a custom plugin name other than the default ones we are passing in wherever `testAccDatabaseSecretBackendConnectionConfig_mssql` is called? 

From an initial glance it looks like adding this field explicitly in MSSQL doesn't have a benefit since it always uses the default, but I could be wrong

vinay-gopalan (commented):
Looking great! Had 1 query about a test with a custom plugin name, but I'm good with the rest

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -19,6 +22,12 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/testutil"
 )
 
+const testDefaultDatabaseSecretBackendResource = "vault_database_secret_backend_connection.test"
+
+// TODO: add support for automating tests for plugin_name
+// Currently we have to configure the Vault server with a plugin_directory,
+// copy/build a db plugin and install it with a unique name, then register it in vault.
+
```
benashz, line 30: @vinay-gopalan, in regards to https://github.com/hashicorp/terraform-provider-vault/pull/1320#discussion_r794094203 - do you think this sufficient for now, or would you rather I document the procedure somewhere else?

vinay-gopalan: Ah I think I now understand what you meant by this comment. I didn't fully take into account the Vault server setup required for custom plugin names. I believe this comment is sufficient, and resolves my query

vault/resource_database_secret_backend_connection_test.go
vinay-gopalan: Resolved with https://github.com/hashicorp/terraform-provider-vault/pull/1320#discussion_r794779496

vinay-gopalan (approved):

vault/resource_database_secret_backend_connection.go
```diff
@@ -44,27 +44,47 @@ const (
...
-		dbBackendMySQLLegacy,
-		dbBackendMySQLRDS,
-		dbBackendPostgres,
-		dbBackendOracle,
-		dbBackendSnowflake,
-		dbBackendRedshift,
+
+	// map all dbBackends to their corresponding dbEngine's
+	// XXX: This map must be updated whenever a new database engine is added.
+	dbEngines = map[string]*dbEngine{
```
calvn, line 50: Minor struct nit: Does this have to be a map, or can it be `[]*dbEngine{}`? The range loops where this is referenced would still work by changing the map key `k` to object's name `e.name`.

For instance:
```
for _, e := range dbEngines {
	dbBackendTypes = append(dbBackendTypes, e.name)
}
```

vault/resource_database_secret_backend_connection.go
```diff
@@ -1257,12 +1260,30 @@ func databaseSecretBackendConnectionRead(d *schema.ResourceData, meta interface{
 		roles = append(roles, role.(string))
 	}
 
-	d.Set("allowed_roles", roles)
-	d.Set("backend", backend)
-	d.Set("name", name)
-	d.Set("root_rotation_statements", resp.Data["root_credentials_rotate_statements"])
+	if err := d.Set("allowed_roles", roles); err != nil {
```
calvn, line 1263: When is it okay to skip this error check? I see that we've added them in here, but still have them left out on call within the switch case above.

calvn (approved):
Couple of minor nits, otherwise looks good! There's a bit of redundancy between the consts, the map, and the slice that we build, but we've discussed online that you plan on refactoring those bits separately.

vault/resource_database_secret_backend_connection.go
benashz: Good question, I would say never. We have been trying update unhandled errors when we see them. Perhaps we should do that across the entire code base...

vault/resource_database_secret_backend_connection.go
benashz: I believe we still need the map in order to get the `dbEngine` from the Schema's field name , for all callers of `getDBEngine()`

benashz: I suppose we could iterate through `[]*dbEgine{}` instead of relying on the map.

benashz: refactored in https://github.com/hashicorp/terraform-provider-vault/pull/1326

PR #1326, opened 2022-01-28
Use dbEngine instances in place of constants.

benashz:
- replace the use of map for looking up dbEngine instances by name with
  a slice
- factor out dbBackend* constants to dbEngine instances.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -122,7 +122,10 @@ type dbEngine struct {
 	defaultPluginName string
 }
 
-func (i *dbEngine) getPluginName(d *schema.ResourceData) (string, error) {
+// GetPluginName from the schema.ResourceData if it is configured,
+// otherwise return the default plugin name.
+// Return an error if no plugin name can be found.
+func (i *dbEngine) GetPluginName(d *schema.ResourceData) (string, error) {
 	if val, ok := d.GetOk("plugin_name"); ok {
 		return val.(string), nil
 	}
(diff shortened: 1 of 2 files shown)


vault/resource_database_secret_backend_connection.go
```diff
@@ -23,66 +24,124 @@ type connectionStringConfig struct {
...
+		dbEngineMySQLLegacy,
+		dbEngineMySQLRDS,
+		dbEnginePostgres,
+		dbEngineOracle,
+		dbEngineSnowflake,
+		dbEngineRedshift,
 	}
 )
 
 type dbEngine struct {
```
calvn, line 120: Small nit: It shouldn't really matter too much since this referenced within the same package (thus fields will technically still be accessible), but since these fields are never meant to be changed, I wonder if we should expose getters, i.e. `Name()` and `DefaultPluginName()`, so that we don't end up accidentally overriding them down the road on future references.

calvn (approved):
Thanks for the tidying it up! One small suggestion, totally optional.

vinay-gopalan (approved):

vault/resource_database_secret_backend_connection.go
benashz: Updated in https://github.com/hashicorp/terraform-provider-vault/pull/1326/commits/aac2869670f393b557d9c949363b8a7e10a5fd7b

calvn (approved):

PR #1342, opened 2022-02-11
Release for vault-1.10

vinay-gopalan:
Release notes for changelog:

```
`resource/database_secret_backend_connection`: Add `disable_escaping` parameter support to Redshift, HanaDB, Postgres and MSSQL
`resource/transit_secret_backend_key`: Add `auto_rotate_interval` parameter support to Transit Key Backend
```


benashz:
Since this is a staging release branch it should not be squashed before merging.

vault/resource_consul_secret_backend_role.go
```diff
@@ -98,27 +106,33 @@ func consulSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) erro
 	path := consulSecretBackendRolePath(backend, name)
 
 	policies := d.Get("policies").([]interface{})
+	roles := d.Get("consul_roles").(*schema.Set).List()
 
-	payload := map[string]interface{}{
-		"policies": policies,
+	if len(policies) == 0 && len(roles) == 0 {
+		return fmt.Errorf("policies or consul_roles must be set")
```
dekimsey, line 126: I think this test is only valid if if the token_type != 'management'. Based on the vault docs for [creating roles](https://www.vaultproject.io/api/secret/consul#create-update-role) and the corresponding validation code in [vault](https://github.com/hashicorp/vault/blob/main/builtin/logical/consul/path_roles.go#L161) itself.

benashz: Thanks @dekimsey , we'll take a look!

dekimsey: Okay, so mixed bag on this on as I'm testing this right now by hand to verify things. 

If you try to use a role that uses a `token_type= 'management'` on Consul 1.11, it errors because of the ACL changes.
```
$ vault write consul/roles/demo-admin token_type=management
$ vault read -field=token consul/creds/demo-admin
...
 * Unexpected response code: 410 (Endpoint /v1/acl/create for the legacy ACL system was removed in Consul 1.11.)
```

The correct process is to set a policy that gives the token root (the built-in global-management for instance)
```
$ vault write consul/roles/demo-admin token_type=client policies=global-management
$ vault read -field=token consul/creds/demo-admin
deadbeef-uuid
```


So while the provider test should align with how Vault is testing this. It wasn't immediately obvious to me how to use the API here. I'm not sure if that's worth a documentation blurb (perhaps an example?) on the TF side or not.

benashz (approved):
👍

PR #1366, opened 2022-03-02
Add support for Consul roles

benashz:
Update the `consul_secret_backend_role` resource to support Consul roles.

```
$ make testacc TESTARGS='-v -test.run TestConsul'
TF_ACC=1 go test -v -v -test.run TestConsul -timeout 30m ./...
=== RUN   TestConsulSecretBackendRole
--- PASS: TestConsulSecretBackendRole (3.37s)
=== RUN   TestConsulSecretBackendRoleNameFromPath
--- PASS: TestConsulSecretBackendRoleNameFromPath (0.00s)
=== RUN   TestConsulSecretBackendRoleBackendFromPath
--- PASS: TestConsulSecretBackendRoleBackendFromPath (0.00s)
=== RUN   TestConsulSecretBackend
--- PASS: TestConsulSecretBackend (6.77s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     13.025s
```

### DIFF
--- a/vault/resource_consul_secret_backend_role.go
+++ b/vault/resource_consul_secret_backend_role.go
@@ -47,6 +47,14 @@ func consulSecretBackendRoleResource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"roles": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: `Set of Consul roles to attach to the token. Applicable for Vault 1.10+ with Consul 1.5+`,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 			"max_ttl": {
 				Type:        schema.TypeInt,
 				Optional:    true,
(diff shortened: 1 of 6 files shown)


vault/resource_consul_secret_backend_role.go
```diff
@@ -158,17 +170,38 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+		"max_ttl":      "max_ttl",
+		"ttl":          "ttl",
+		"token_type":   "token_type",
+		"local":        "local",
+		"consul_roles": "roles",
+	}
+
+	for k, v := range params {
+		val, ok := data[k]
+		if k == "consul_roles" && !ok {
```
vinay-gopalan, line 198: Wondering if we should have a similar case for `policies`? Since according to the Vault docs `Either "policies"
or "consul_roles" are required for Consul 1.5 and above`. Are we only targeting Consul 1.4 for now (in which case this would make sense since only `policies` is required)

benashz: I don't  believe so. This case is only for vault 1.9 and below, and is not tied a specific version of Consul. I suppose Vault will figure that out for us. Unless I am misunderstanding your question?

vinay-gopalan: What if in the future a user would like to only set `roles` and not `policies`? This is technically possible in Vault, but the user wouldn't be able to do that with the current implementation of the provider if I understand it correctly

vinay-gopalan: Actually never mind, I think in that case the Read function should just set the `policies` field to be empty?

vinay-gopalan (approved):
LGTM!

vinay-gopalan (approved):

PR #1381, opened 2022-03-24
Rework the credential validation to use azidentity

benashz:
Credential validation now relies on listing of the resource principles
to determine success. If the client credentials can successfully list
the principles once then the credentials provided from Vault are deemed
to be valid. In the case where the 401 HTTP status code is returned from
the Azure API, the validation check is terminated since this means that
the credentials were not authorized to perform any API requests.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```


vault/data_source_azure_access_credentials.go
```diff
@@ -125,87 +128,84 @@ func azureAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface
...
 
-	secret, err = client.Logical().Read(configPath)
+	config, err := client.Logical().Read(configPath)
 	if err != nil {
 		return fmt.Errorf("error reading from Vault: %s", err)
 	}
-	log.Printf("[DEBUG] Read %q from Vault", configPath)
+	log.Printf("[DEBUG] Successfully read %q from Vault", configPath)
 
 	subscriptionID := ""
```
mb290, line 137: Instead of leveraging the subscriptionId and tenantId returned from the Vault configuration, can these be optionally defined in the data source? Otherwise it'll require all Azure SPs in Vault to be granted access to a centralized Subscription that is defined in the Vault configuration, which is likely not needed by the SP that's generated for anything else.

vault/data_source_azure_access_credentials.go
```diff
@@ -125,87 +128,84 @@ func azureAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface
...
+			}
+			log.Printf("[DEBUG] Provider Client List response %+v", pr)
 		}
-		// If the creds have propagated to the server where we're checking, we receive no error from the above Prepare call.
-		sequentialSuccesses++
-		if sequentialSuccesses == sequentialSuccessesRequired {
-			overallSuccess = true
+
+		if pager.Err() == nil {
+			log.Printf("[DEBUG] Credential validation succeeded")
```
mb290, line 198: With num_sequential_successes being removed this will mean the moment a single Azure AD endpoint returns success then it'll assume that the credential is valid, however there can be a propogation delay between endpoints. with num_sequential_successes this allows for the provider to initiate multiple attempts to ensure credentials are replicated.

vault/data_source_azure_access_credentials.go
benashz: Sure, we can add that, but I think that it is out of scope for this PR. I can open another PR, but it probably won't make it into 3.4.0. Sound okay?

mb290: I'm concerned it'll cause the provider to break for any existing vault roles that don't have access to the subscription that's defined in the Azure Secrets engine config in Vault.

vinay-gopalan (approved):

vault/data_source_azure_access_credentials.go
benashz: Interesting, I was not able to reproduce that in my testing with a 5 minute poll. I'm happy to add it back.

benashz: Is there a way to know which endpoint we hit?

mb290: I don't believe there is. Replication time seems to take longer when using an the existing SP functionality as documented here [https://www.vaultproject.io/docs/secrets/azure#application-object-ids](https://www.vaultproject.io/docs/secrets/azure#application-object-ids)

benashz: Fix in https://github.com/hashicorp/terraform-provider-vault/pull/1381/commits/e71e3efc4f2c7e3a715208ac9c37e407620189c6

vault/data_source_azure_access_credentials.go
benashz: PR forthcoming

vinay-gopalan (approved):

vault/data_source_azure_access_credentials.go
benashz: Addressed in #1384

robison:
I'm happy to raise a separate issue for this, but we ran headlong directly into the concern that @mb290 raised yesterday: https://github.com/hashicorp/terraform-provider-vault/pull/1381#discussion_r834672881

This seems to be a semi-breaking change for existing roles - could something be added to CHANGELOG.md?

benashz:
> This seems to be a semi-breaking change for existing roles - could something be added to CHANGELOG.md?

Hi @robison I don't believe we introduced any breaking changes with this fix. The `num_sequential_successes` behaviour remains the same.

Would you mind  unpacking the `semi-breaking` change for us?

Thanks,

Ben

robison:
Per @mb290's comment https://github.com/hashicorp/terraform-provider-vault/pull/1381#discussion_r834672881, we had an existing Vault role that did not have access to the subscription that's defined in the Azure Secrets engine config in Vault. Thankfully ran into this in a dev environment, where the `azure` provider version wasn't pinned. We first pinned it back to 3.3.0, then realized that we would need to specify the `subscription_id` and `tenant_id` in 3.4.0+ - which we couldn't do until https://github.com/hashicorp/terraform-provider-vault/pull/1391 landed.

benashz:
> Per @mb290's comment [#1381 (comment)](https://github.com/hashicorp/terraform-provider-vault/pull/1381#discussion_r834672881), we had an existing Vault role that did not have access to the subscription that's defined in the Azure Secrets engine config in Vault. Thankfully ran into this in a dev environment, where the `azure` provider version wasn't pinned. We first pinned it back to 3.3.0, then realized that we would need to specify the `subscription_id` and `tenant_id` in 3.4.0+ - which we couldn't do until #1391 landed.

Ah, I see now. Thanks for the clarification. #1391 should going out on Thursday in a patch release `3.4.1`.

robison:
Excellent. Could we also get a update/note in CHANGELOG.md for v3.4.1 indicating the potential necessity of specifying `subscription_id` & `tenant_id` when using a pre-existing Vault role that doesn't have access to the subscription?

PR #1384, opened 2022-03-24
Allow for overriding the tenant and subscription IDs.

benashz:
The best description for this PR can be found here: https://github.com/hashicorp/terraform-provider-vault/pull/1381#discussion_r834665429

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/vault/data_source_azure_access_credentials.go
+++ b/vault/data_source_azure_access_credentials.go
@@ -107,7 +107,6 @@ func azureAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface
 	backend := d.Get("backend").(string)
 	role := d.Get("role").(string)
 
-	configPath := backend + "/config"
 	credsPath := backend + "/creds/" + role
 
 	secret, err := client.Logical().Read(credsPath)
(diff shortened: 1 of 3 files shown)


vault/data_source_azure_access_credentials.go
```diff
@@ -134,19 +146,25 @@ func azureAccessCredentialsDataSourceRead(d *schema.ResourceData, meta interface
 	log.Printf("[DEBUG] Successfully read %q from Vault", configPath)
 
 	subscriptionID := ""
-	if subscriptionIDIfc, ok := config.Data["subscription_id"]; ok {
+	if v, ok := d.GetOk("subscription_id"); ok {
```
mb290, line 149: Can you set line 142 that if subscription id and tenant id are defined in the provider then it doesn't need to do the vault config read at all? I don't believe it's used anywhere else in the code and would mean the token that is being used by terraform can have fewer permissions in vault as it no longer needs to read the config.

vinay-gopalan (approved):

vault/data_source_azure_access_credentials.go
benashz: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/1384/commits/bcb6464a0fbda175162d7cb9b9cbde4bb83cf781

vinay-gopalan (approved):

PR #1407, opened 2022-04-08
Update SSH Role resource with Vault 1.10 API changes

vinay-gopalan:


benashz:
Worked  done under #1413

PR #1418, opened 2022-04-19
changes to prep for releases-api migration

modrake:
This reflects the changes that acccompany the releases-api anticipated to be launched in early May. Included is:
*  Version bump of the terraform provider release gh action
*  Variables added for interfacing with releases-api
*  hc-releases syntax modified for releases-api use
*  metadata file added for ingesting with releases-api

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,3 +47,5 @@ jobs:
     with:
       release-notes: true
       setup-go-version: '${{ needs.go-version.outputs.version }}'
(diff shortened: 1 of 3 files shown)


jeanneryan (changes requested):
Not sure why I can't comment on the actual line, but in `.github/workflows/release.yml` after line 49, it needs `product-version: '${{ github.ref_name }}'`

jeanneryan (approved):

PR #1426, opened 2022-04-22
Add support to perform semantic version comparisons against Vault's server version

vinay-gopalan:
This PR adds a mechanism that performs semantic version comparison against Vault’s target version to determine whether certain features are available in the current Vault client. This mechanism allows the TFVP to have knowledge of Vault's version, supported paths and version-specific schemas, allowing us to further improve upon the codebase's overall architecture.

In this PR, we demonstrate how the semantic version comparison mechanism can be used to determine whether Auth and Secret Engines can be remounted. Mount Migration is a newly available feature as of Vault 1.10.0, and hence this feature is only enabled for clients with the required version or above.

This PR also includes wrapper functions that can be used with new resources to determine whether said resources are available within the current version of Vault. If the user attempts to create a resource using this wrapper function and it does not meet the necessary version requirements, the resource shall fail to create. 

Features that this PR adds semantic version comparison to:
- Kubernetes Secret Backend resource (1.11)
- Kubernetes Secret Backend Role resource (1.11)
- Managed Keys resource (1.10)
- Mount migration for AWS (1.10)

```
$ make testacc TESTARGS='-run=TestAccAWSSecretBackend_remount'
=== RUN   TestAccAWSSecretBackend_remount
--- PASS: TestAccAWSSecretBackend_remount (2.90s)
PASS

=== RUN   TestAccKubernetesSecretBackend
--- PASS: TestAccKubernetesSecretBackend (4.38s)
PASS

=== RUN   TestAccKubernetesSecretBackendRole
--- PASS: TestAccKubernetesSecretBackendRole (4.52s)
PASS

=== RUN   TestManagedKeys
--- PASS: TestManagedKeys (4.92s)
PASS
```


vault/provider.go
```diff
@@ -864,6 +865,48 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
...
+		}
+
+		featureEnabled, currentVersion, err := semver.SemanticVersionComparison(minVersion, client)
+		if err != nil {
+			return err
+		}
+
+		if !featureEnabled {
+			return fmt.Errorf("feature not enabled on current Vault version. min version required=%s"+
+				"Vault version=%s", minVersion, currentVersion)
```
swenson, line 882: I think you want a space before `Vault` or after the `%s` above.

swenson (approved):
Awesome! This is super useful.

internal/semver/semver.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/hashicorp/vault/api"
+
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+)
+
+func getTargetVaultVersion(client *api.Client) (string, error) {
+	// This endpoint only returns data upon an HTTP request
+	// Reads using the Client do not return data
+	path := "/v1/sys/seal-status"
+	r := client.NewRequest("GET", path)
```
tvoran, line 18: It looks like there's a [SealStatus()/SealStatusWithContext() client function](https://github.com/hashicorp/vault/blob/8de51d4ae6006504227c064d19901830b9e20736/api/sys_seal.go#L8) in vault/api that may simplify some of this a bit.

vault/provider.go
```diff
@@ -864,6 +865,48 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
 	}
 }
 
+func VersionCheckWrapper(f schema.CreateFunc, d *schema.ResourceDiff, meta interface{}, minVersion string) schema.CreateFunc {
```
tvoran, line 868: Is this being used?

vault/provider.go
```diff
@@ -864,6 +865,48 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
...
+		if !featureEnabled {
+			return fmt.Errorf("feature not enabled on current Vault version. min version required=%s "+
+				"current vault version=%s", minVersion, currentVersion)
+		}
+
+		return f(d, meta)
+	}
+}
+
+func VersionCheckContextWrapper(f schema.CreateContextFunc, minVersion string) schema.CreateContextFunc {
```
tvoran, line 889: Just another thought on naming:

```suggestion
func MinVersionCheckContextWrapper(f schema.CreateContextFunc, minVersion string) schema.CreateContextFunc {
```

internal/semver/semver.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"io"
+
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/vault/api"
+
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+)
+
+func getTargetVaultVersion(client *api.Client) (string, error) {
```
tvoran, line 14: It would be great if this took a context and used the context-aware vault client functions, now that they're available.

internal/semver/semver.go
```diff
@@ -0,0 +1,70 @@
...
+	if v, ok := data[consts.FieldVersion]; ok {
+		version = v.(string)
+	} else {
+		return "", fmt.Errorf("key %q not found in response", consts.FieldVersion)
+	}
+
+	return version, nil
+}
+
+func SemanticVersionComparison(minVersionString string, client *api.Client) (bool, string, error) {
```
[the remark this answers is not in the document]
vinay-gopalan: Good call! I wasn't sure if I should use a conflicting name with the `go-version` package, but sounds like it should be fine

internal/semver/semver.go
vinay-gopalan: Thanks for pointing that one out! Definitely helped simplify a lot of the logic. I should've probably done more digging! :)

vault/provider.go
vinay-gopalan: Updated in latest commit

internal/semver/semver.go
vinay-gopalan: Great idea! I refactored the code to use the context-aware functions

vault/resource_aws_secret_backend_test.go
```diff
@@ -103,6 +103,38 @@ func TestAccAWSSecretBackend_usernameTempl(t *testing.T) {
...
+func TestAccAWSSecretBackend_remount(t *testing.T) {
+	path := acctest.RandomWithPrefix("tf-test-aws")
+	updatedPath := acctest.RandomWithPrefix("tf-test-aws-updated")
+
+	resourceName := "vault_aws_secret_backend.test"
+	accessKey, secretKey := testutil.GetTestAWSCreds(t)
+	resource.Test(t, resource.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testutil.TestAccPreCheck(t) },
+		Steps: []resource.TestStep{
```
robmonte, line 115: I was looking in the testing framework to see if there was anything to support checking whether a resource is being destroyed or not between test steps, but I didn't find anything. Do you know if there might be a way to accomplish that? It may be a useful ability to have going forward as we utilize the feature.

vault/provider.go
vinay-gopalan: I thought we might need a wrapper for the resources using the deprecated functions, but now that I think more about it, semantic version comparisons will mostly be used for new features which won't use this wrapper function and any old resources that need semantic version comparison should probably be updated to not use deprecated code anyway. Removed this in the latest commits

vault/resource_aws_secret_backend_test.go
vinay-gopalan: Good point. I've been meaning to add this check in the Managed keys tests as well, but I think one way would be if Vault has a UUID for the resource then we can check if the UUIDs are different after re-creation. Other than that, I haven't come up with a better solution as of yet unfortunately

fairclothjm (approved):
👍

maxcoulombe (approved):

vault/resource_aws_secret_backend.go
```diff
@@ -97,6 +99,33 @@ func awsSecretBackendResource() *schema.Resource {
...
+				// Mount migration not available
+				// Destroy and recreate resource
+				if err := diff.ForceNew("path"); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
vinay-gopalan: Ah nice, I wasn't explicitly aware of that! Updated in the latest commit, thanks! :)

vault/resource_aws_secret_backend_test.go
vinay-gopalan: For this test, I had to check my Vault logs to confirm that a mount migration had occurred successfully

tvoran: There is a `CheckDestroy` field in resource.TestCase that I've seen used elsewhere, though I think that's after the test and not between steps.

You may be able to use `TestCheckNoResourceAttr` to check that the old mount doesn't exist in the state at least?

internal/semver/semver.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if resp.Version == "" {
+		return "", fmt.Errorf("key %q not found in response", consts.FieldVersion)
+	}
+
+	return resp.Version, nil
+}
+
+func GreaterThanOrEqual(ctx context.Context, client *api.Client, minVersionString string) (bool, string, error) {
```
tvoran, line 30: It might be nice to include a doc string for this function since it's exported, especially since it's not immediately apparent what all the return values are.

tvoran (approved):

vault/resource_aws_secret_backend_test.go
vinay-gopalan: You are correct! `CheckDestroy` runs after the test completes, so unfortunately we can't use that in this case

I looked into trying to make `TestCheckNoResourceAttr` work. The function checks whether certain state keys exist in the TF state for a particular resource. In this case, while the mount does change (expressed in this resource as a `path` change), the `path` key isn't being deleted or removed, so it still exists in the TF state (albeit with a different value). In this case unfortunately, I don't think that would work exactly the way we want it to

I'll definitely keep looking into this as an improvement to our testing since this is also needed in a few other places!

In the future I think we also plan to refactor all secret and auth engines be mounted solely via the `vault_mount` resource (currently some resources mount the engine as well as the config; including the resource we are talking about: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_aws_secret_backend.go#L119). This might help check the TF state for deleted mounts in a cleaner way.

internal/semver/semver.go
vinay-gopalan: Good call! Added a description for the function

internal/semver/semver.go
```diff
@@ -0,0 +1,64 @@
...
+// returns:
+//    - a boolean describing whether the Vault
+//      server version was above the minimum version
+//    - the current Vault server version as a string
+//    - errors captured during operation, if any
+//
+// This function can be used to perform semantic version comparisons
+// to conditionally enable features, or to resolve any diffs in the TF
+// state based on the Vault version.
+func GreaterThanOrEqual(ctx context.Context, client *api.Client, minVersionString string) (bool, string, error) {
```
benashz, line 45: I wonder if we should be doing the initial version check in `provider.NewProviderMeta` instead. This would allow the provider to abort early if it encounters an error getting the Vault version. It would also make it easier to use in the resource code since we would no longer have to handle the error. We would store the `version.Version` in the `ProviderMeta` so there would be no need to return the vault version string.

If we make `minVersionString` a `version.Version` the caller would be responsible for handling any parse errors.

internal/consts/consts.go
```diff
@@ -82,6 +82,13 @@ const (
 	MountTypeNomad      = "nomad"
 	MountTypeKubernetes = "kubernetes"
 
+	/*
+		Vault version constants
+	*/
+	VaultVersion11 = "1.11.0"
```
benashz, line 88: Suggest making these `version.Version` variables instead, then there is no need to reparse them for every API request.

internal/semver/semver.go
benashz, line 45: It looks like we will be calling the API for every Version check, this seems like a unnecessary increase in the overall the number of requests per operation. It is very unlikely that the vault server version will change during a run of Terraform.

vinay-gopalan: @benashz I'm glad you brought this up! I initially thought of this approach as well, and the main drawback that came to my mind (which I remember we briefly chatted about) was in the case where users use the same TF instance to configure multiple versions of Vault. In that case, the approach of storing the current Vault version to the `ProviderMeta` would explicitly disallow the use of the TFVP in those kind of scenarios, and I wanted to get your take on the larger implications/architecture before going down that route. I do agree that the current implementation adds an overhead of multiple calls to the same endpoint

benashz: I think we are okay here since each instance of the provider corresponds to a single Vault server instance.

benashz: By instance of the provider, I am referring to each `provider{}` block defined; terraform will spawn a separate process for each block defined.

PR #1461, opened 2022-05-17
CI: Test against vault-enterprise 1.10.3-ent

### LINKED ISSUE #1463: Argument named "disable_escaping" is not expected
Argument named "disable_escaping" is not expected in vault_database_secret_backend_connection config

### Terraform Version
Terraform v1.1.9
on linux_amd64
+ provider registry.terraform.io/hashicorp/vault v3.6.0

### Affected Resource(s)
vault_database_secret_backend_connection

```hcl
resource "vault_database_secret_backend_connection" "db" {
  backend = "test"
  name    = "test"

  postgresql {
    disable_escaping = false
    username = "test"
    password = "test"
    connection_url = "postgres://test"
  }
}
```

### Debug Output
```
│ Error: Unsupported argument
│ 
│   on _bootstrap.tf line 35, in resource "vault_database_secret_backend_connection" "db":
│   35:     disable_escaping = false
│ 
│ An argument named "disable_escaping" is not expected here.
```

### Expected Behavior
According to the docs this should be a valid argument
https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/database_secret_backend_connection#postgresql-configuration-options

### Actual Behavior
Apply normally

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
I think the tests that would catch this are being skipped.

benashz:
This PR moves us from CircleCI to GitHub Actions and allows us to run all acceptance tests against Vault Enterprise

Additional fixes:
- Fix up broken enterprise tests
- Add support for testing against Postgres
- Fix a regression introduced in v3.5.0 which inadvertently removed the the `disable_escaping` feature from some db engines
- Fix bogus quota test that caused the tests to fail

Closes #1463


vinay-gopalan (approved):

vault/resource_kmip_secret_backend.go
```diff
@@ -133,17 +134,39 @@ func kmipSecretBackendUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		err := client.Sys().Remount(src, dest)
 		if err != nil {
 			return fmt.Errorf("error remounting in Vault: %s", err)
 		}
 
-		d.SetId(newPath)
-		path = newPath
+		// There is something similar in resource_mount.go, but in the call to TuneMount().
+		var tries int
+		for {
```
vinay-gopalan, line 149: Is this  mount check loop something we can/should refactor to a function? I think this might be a useful check to add to other similar resources that also setup their own mounts; since we plan to support remounts for resources such as `vault_aws_secret_backend` in the future via adaptive migration

vault/resource_quota_lease_count_test.go
```diff
@@ -14,41 +14,44 @@ import (
...
+	whole := acctest.RandIntRange(50000, 60000)
 	return strconv.Itoa(whole + 1000)
 }
 
 func TestQuotaLeaseCount(t *testing.T) {
 	name := acctest.RandomWithPrefix("tf-test")
-	leaseCount := randomQuotaLeaseString()
-	newLeaseCount := randomQuotaLeaseString()
+	ns := "ns-" + name
+	leaseCount := "2"    // randomQuotaLeaseString()
```
vinay-gopalan, line 24: Are the comments left over and meant to be removed? The way it reads it looks like we got those numbers using the random function?

vinay-gopalan (commented):
PR looks great! Couple of nits and questions :)

codegen/generate.go
```diff
@@ -14,7 +16,7 @@ import (
 
 // generatedDirPerms uses 0775 because it is the same as for
 // the "vault" directory, which is at "drwxrwxr-x".
-const generatedDirPerms os.FileMode = 0775
+const generatedDirPerms os.FileMode = 0o775
```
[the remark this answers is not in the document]
benashz: I think the syntax has changed for the FileMode mask, it should still be the same `chmod 0755`. I think that the formatter made this change.

vault/resource_kmip_secret_backend.go
benashz: Nice, was hoping you would catch that! Yeah, for sure. I thought you might do that in your secrets remount work.

vault/resource_quota_lease_count_test.go
benashz: ah good catch, I will restore this

codegen/generate.go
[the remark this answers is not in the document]
vinay-gopalan: Ah okay, thanks for clarifying!

vault/resource_kmip_secret_backend.go
vinay-gopalan: Nice! Sounds good, I will make a note to make this a part of that work :)

vault/resource_quota_lease_count_test.go
benashz: So the random bit is no longer necessary since we run the tests in a unique namespace.

Fixed in 99fa2897f8a55c5dbe56989e24965f9c46f19cfe

PR #1478, opened 2022-06-02
auth/jwt: adds user_claim_json_pointer and max_age to roles

austingebauer:
### Community Note

Relates OR Closes: N/A

```release-note
* `resource/vault_jwt_auth_backend`: Adds support for `user_claim_json_pointer` and `max_age` (#1478)
```

```
$ make testacc TESTARGS='-count=1 -run=TestAccJWTAuth'
TF_ACC=1 go test -v -count=1 -run=TestAccJWTAuth -timeout 30m ./...
=== RUN   TestAccJWTAuthBackendRole_import
--- PASS: TestAccJWTAuthBackendRole_import (1.62s)
=== RUN   TestAccJWTAuthBackendRole_basic
--- PASS: TestAccJWTAuthBackendRole_basic (1.07s)
=== RUN   TestAccJWTAuthBackendRole_update
--- PASS: TestAccJWTAuthBackendRole_update (1.91s)
=== RUN   TestAccJWTAuthBackendRole_full
--- PASS: TestAccJWTAuthBackendRole_full (1.10s)
=== RUN   TestAccJWTAuthBackendRoleOIDC_full
--- PASS: TestAccJWTAuthBackendRoleOIDC_full (1.57s)
=== RUN   TestAccJWTAuthBackendRoleOIDC_disableParsing
--- PASS: TestAccJWTAuthBackendRoleOIDC_disableParsing (1.24s)
=== RUN   TestAccJWTAuthBackendRole_fullUpdate
--- PASS: TestAccJWTAuthBackendRole_fullUpdate (2.81s)
=== RUN   TestAccJWTAuthBackend
--- PASS: TestAccJWTAuthBackend (3.89s)
=== RUN   TestAccJWTAuthBackendProviderConfig
--- PASS: TestAccJWTAuthBackendProviderConfig (1.11s)
=== RUN   TestAccJWTAuthBackend_OIDC
--- PASS: TestAccJWTAuthBackend_OIDC (1.17s)
=== RUN   TestAccJWTAuthBackend_invalid
--- PASS: TestAccJWTAuthBackend_invalid (0.30s)
--- PASS: TestAccJWTAuthBackend_missingMandatory (1.91s)
=== RUN   TestAccJWTAuthBackendProviderConfigConversionBool
--- PASS: TestAccJWTAuthBackendProviderConfigConversionBool (0.00s)
=== RUN   TestAccJWTAuthBackendProviderConfigConversionInt
--- PASS: TestAccJWTAuthBackendProviderConfigConversionInt (0.00s)
=== RUN   TestAccJWTAuthBackendProviderConfig_negative
    resource_jwt_auth_backend_test.go:380: true
--- SKIP: TestAccJWTAuthBackendProviderConfig_negative (0.00s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     21.384s
```

### DIFF
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -385,11 +385,11 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
 	data["user_claim"] = d.Get("user_claim").(string)
 
 	if v, ok := d.GetOkExists("user_claim_json_pointer"); ok {
-		data["user_claim_json_pointer"] = v.(bool)
+		data["user_claim_json_pointer"] = v
 	}
 
 	if v, ok := d.GetOk("max_age"); ok {
-		data["max_age"] = v.(int)
+		data["max_age"] = v
 	}
 
 	if dataList := util.TerraformSetToStringArray(d.Get("allowed_redirect_uris")); len(dataList) > 0 {
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -119,6 +119,17 @@ func jwtAuthBackendRoleResource() *schema.Resource {
 			Default:     false,
 			Description: "Log received OIDC tokens and claims when debug-level logging is active. Not recommended in production since sensitive information may be present in OIDC responses.",
 		},
+		"user_claim_json_pointer": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Specifies if the user_claim value uses JSON pointer syntax for referencing claims. By default, the user_claim value will not use JSON pointer.",
+		},
+		"max_age": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "Specifies the allowable elapsed time in seconds since the last time the user was actively authenticated.",
+		},
 		"backend": {
 			Type:        schema.TypeString,
 			Optional:    true,
@@ -277,6 +288,12 @@ func jwtAuthBackendRoleRead(_ context.Context, d *schema.ResourceData, meta inte
 	if v, ok := resp.Data["verbose_oidc_logging"]; ok {
 		d.Set("verbose_oidc_logging", v)
 	}
+	if v, ok := resp.Data["user_claim_json_pointer"]; ok {
+		d.Set("user_claim_json_pointer", v)
+	}
+	if v, ok := resp.Data["max_age"]; ok {
+		d.Set("max_age", v)
+	}
 
 	d.Set("backend", backend)
 	d.Set("role_name", role)
@@ -366,6 +383,14 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
 
 	data["bound_audiences"] = util.TerraformSetToStringArray(d.Get("bound_audiences"))
 	data["user_claim"] = d.Get("user_claim").(string)
+
+	if v, ok := d.GetOkExists("user_claim_json_pointer"); ok {
+		data["user_claim_json_pointer"] = v.(bool)
+	}
+
+	if v, ok := d.GetOk("max_age"); ok {
+		data["max_age"] = v.(int)
+	}
 
 	if dataList := util.TerraformSetToStringArray(d.Get("allowed_redirect_uris")); len(dataList) > 0 {
 		data["allowed_redirect_uris"] = dataList
--- a/vault/resource_jwt_auth_backend_role.go
+++ b/vault/resource_jwt_auth_backend_role.go
@@ -382,11 +382,8 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
 	updateTokenFields(d, data, create)
 
 	data["bound_audiences"] = util.TerraformSetToStringArray(d.Get("bound_audiences"))
-	data["user_claim"] = d.Get("user_claim").(string)
-
-	if v, ok := d.GetOkExists("user_claim_json_pointer"); ok {
-		data["user_claim_json_pointer"] = v
-	}
+	data["user_claim"] = d.Get("user_claim")
+	data["user_claim_json_pointer"] = d.Get("user_claim_json_pointer")
 
 	if v, ok := d.GetOk("max_age"); ok {
 		data["max_age"] = v
--- a/vault/resource_jwt_auth_backend_role_test.go
+++ b/vault/resource_jwt_auth_backend_role_test.go
@@ -16,7 +16,10 @@ func TestAccJWTAuthBackendRole_import(t *testing.T) {
 	backend := acctest.RandomWithPrefix("jwt")
 	role := acctest.RandomWithPrefix("test-role")
 	resource.Test(t, resource.TestCase{
-		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)
+		},
 		Providers:    testProviders,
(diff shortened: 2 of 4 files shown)


austingebauer:
Looks like the `TestAccJWTAuthBackendRole_import` test failed in CI. This is likely because the test is using Vault 1.10 and the new role parameter is not released yet. The test passed when ran locally.

vault/resource_jwt_auth_backend_role.go
```diff
@@ -367,6 +384,14 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
 	data["bound_audiences"] = util.TerraformSetToStringArray(d.Get("bound_audiences"))
 	data["user_claim"] = d.Get("user_claim").(string)
 
+	if v, ok := d.GetOkExists("user_claim_json_pointer"); ok {
```
austingebauer, line 387: I couldn't find a way around using `GetOkExists`. Looks like it's deprecated but there isn't a good solution to the problem yet from my search. Let me know if there is something you all are typically doing for these optional boolean params.

vinay-gopalan: We have used `GetOkExists` in the past due to `GetOk` not behaving exactly the way we'd like when it comes to boolean parameters. I think for this case it is fine to go with `GetOkExists` similar to other fields within this file. `d.Get` would also work I believe

vinay-gopalan:
> Looks like the `TestAccJWTAuthBackendRole_import` test failed in CI. This is likely because the test is using Vault 1.10 and the new role parameter is not released yet. The test passed when ran locally.

Ah, yes. This is expected until we have access to Vault builds built off of the `main` branch running in containers for our CI! 

For 1.10, we made use of the `testutil.SkipTestEnvSet` util function to skip tests and then populated an environment variable in CI. I believe we'll be doing something similar for 1.11 tests :)

vault/resource_jwt_auth_backend_role.go
benashz: If your field type is a boolean and has a default then using `GetOk()` should work the same as `GetOkExists()`

vault/resource_jwt_auth_backend_role.go
```diff
@@ -367,6 +384,14 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
...
+		data["user_claim_json_pointer"] = v.(bool)
```
benashz, line 388: No need for the type assert here.

vault/resource_jwt_auth_backend_role.go
```diff
@@ -367,6 +384,14 @@ func jwtAuthBackendRoleDataToWrite(d *schema.ResourceData, create bool) map[stri
...
+	}
+
+	if v, ok := d.GetOk("max_age"); ok {
+		data["max_age"] = v.(int)
```
benashz, line 392: Same

vault/resource_jwt_auth_backend_role_test.go
```diff
@@ -675,5 +688,6 @@ resource "vault_jwt_auth_backend_role" "role" {
     department = "engineering-*-admin"
     sector = "7g"
   }
+  user_claim_json_pointer = false
```
benashz, line 691: Might be a good idea test the handling of the default value with a test config that does not specify this field.

benashz (approved):
Looking good. A few minor suggestions/comments. Then +1

vault/resource_jwt_auth_backend_role.go
austingebauer: The `user_claim_json_pointer` currently has a default. With that, my experience is that `GetOk()` and `GetOkExists()` don't actually behave the same. For example, if `user_claim_json_pointer=true` on create, then we switch it to `false` for an update, the `false` update is not applied when using `GetOk()` but is applied when using `GetOkExists()`. We want the `GetOkExists()` behavior in this case.

I tried a number of different things and wasn't able to find a workaround that allows me to not use `GetOkExists()`.

austingebauer: Perhaps we could just use `Get()`.

vault/resource_jwt_auth_backend_role_test.go
austingebauer: Good idea. Done in [f9b1c00](https://github.com/hashicorp/terraform-provider-vault/pull/1478/commits/f9b1c004d509e226479b1b3d78b0d55de385fab0).

austingebauer:
@vinay-gopalan - Thanks for the explanation about the test skipping. Makes sense 👍 Any immediate suggestion? I could also hold on merging this until things are set up / ready.

vault/resource_jwt_auth_backend_role.go
benashz: Yeah, the behavior of `GetOk()` and `GetOkExists()` are a bit odd IMO. Since you have a default then using `Get()` should work fine.

benashz:
> @vinay-gopalan - Thanks for the explanation about the test skipping. Makes sense 👍 Any immediate suggestion? I could also hold on merging this until things are set up / ready.

We use this env var to work around the lack of a vault dev instance to test against: https://github.com/hashicorp/terraform-provider-vault/blob/7205d9939d8e1afb1cc2b6a48e2819c8499e4cb7/testutil/testutil.go#L23

austingebauer:
@benashz - Thanks, that's what I was looking for. I've updated the tests that depend on a Vault 1.11 build to use that env var in [2fd9a5b](https://github.com/hashicorp/terraform-provider-vault/pull/1478/commits/2fd9a5b39316582cce412b1b8feb13cb1578f86c).

vinay-gopalan (approved):

vault/resource_jwt_auth_backend_role.go
austingebauer: I changed this to use `Get()` in [749ca8b](https://github.com/hashicorp/terraform-provider-vault/pull/1478/commits/749ca8bd694f3d27a497158201166c6a1d934abf). That works as expected with the default.

vault/resource_jwt_auth_backend_role_test.go
```diff
@@ -16,7 +16,10 @@ func TestAccJWTAuthBackendRole_import(t *testing.T) {
 	backend := acctest.RandomWithPrefix("jwt")
 	role := acctest.RandomWithPrefix("test-role")
 	resource.Test(t, resource.TestCase{
-		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)
```
benashz, line 21: Normally we put the `SkipTestEnv..*` at the top of the test rather than in the step's `PreCheck`. Avoids failing a test that we are going to skip.

austingebauer: Sounds good. I can move it. I think I saw some env checks like that elsewhere so just copied it.

benashz: Think you also might need to update https://github.com/hashicorp/terraform-provider-vault/blob/24044c48b6aed5d76da9d8856aa348a13a727c9b/.github/workflows/build.yml#L81

benashz: > Sounds good. I can move it. I think I saw some env checks like that elsewhere so just copied it.

Yeah, those could be bogus depending...

austingebauer: @benashz - Moved the env check out of `PreCheck` in [7020928](https://github.com/hashicorp/terraform-provider-vault/pull/1478/commits/7020928ce4ff454fe16fd13a8d0b3013312b2ca7). Also updated that `make testacc-ent ...` for CI. Let me know if that all looks okay 👍

benashz (approved):
LGTM!

PR #1495, opened 2022-06-09
PKI: Add support for CPS URL in custom policy identifiers

swenson:
Update the `vault_resource_pki_secret_backend_role` to support
specifying the CPS URL when specifying policy identifiers in line with
the recent changes to the PKI Secrets Engine in Vault 1.11:
https://github.com/hashicorp/vault/pull/15751

We do this by deprecating the existing `policy_identifiers` argument and
creating a new block, `policy_identifier`, which can be specified
multiple times.

If both `policy_identifiers` and `policy_identifier` blocks are present,
then `policy_identifier` is ignored. (Otherwise, refreshing would delete
one or the other, and the state wouldn't have round trip stability.)

This was also tested locally with, for example, a terraform file like:

```hcl
provider "vault" {
}

resource "vault_mount" "pki" {
  path                      = "pki"
  type                      = "pki"
  default_lease_ttl_seconds = 3600
  max_lease_ttl_seconds     = 86400
}

resource "vault_pki_secret_backend_role" "role" {
  name = "example-dot-com"
  backend = vault_mount.pki.path
  allowed_domains = ["example.com"]
  allow_subdomains = true
  allow_bare_domains = true
  allow_glob_domains = true
  allow_ip_sans = true
  allow_localhost = "true"
  generate_lease = true
  organization = ["Hashi test"]
  country = ["USA"]
  locality = ["Area 51"]
  province = ["NV"]
  max_ttl = "720h"
  policy_identifiers = ["2.5.29.32","1.2.3"]
  // or
  policy_identifier {
    oid = "2.5.29.32"
    cps = "https://example.com/cps"
    notice = "Some notice"
  }
  policy_identifier {
    oid = "1.2.3"
  }
}
```

### Community Note

```release-note
`resource/pki_secret_backend_role`: Add support for `policy_identifier` blocks, deprecate `policy_identifiers`
```

```
$ make testacc TESTARGS="-v ./vault -run .*TestPkiSecretBackendRole.*"
TF_ACC=1 go test -v -v ./vault -run .*TestPkiSecretBackendRole.* -timeout 30m ./...
=== RUN   TestPkiSecretBackendRole_basic
--- PASS: TestPkiSecretBackendRole_basic (4.75s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	5.093s
...
```


vault/resource_pki_secret_backend_role.go
```diff
@@ -286,11 +286,38 @@ func pkiSecretBackendRoleResource() *schema.Resource {
 				Type:        schema.TypeList,
 				Required:    false,
 				Optional:    true,
-				Description: "Specify the list of allowed policies IODs.",
+				Description: "Specify the list of allowed policies OIDs.",
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
 			},
+			"policy_identifier": {
```
vinay-gopalan, line 294: Would adding `ConflictsWith` to this schema field help in this situation? If we specify that this new field conflicts with the legacy `policy_identifiers` field, then TF would log an error to the user if they try to use both fields within the same config. ([here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_jwt_auth_backend.go#L59) is an example of using `ConflictsWith`)

I see we currently log a warning to the console and ignore the legacy field; maybe explicitly erroring might be better?

vault/resource_pki_secret_backend_role_test.go
```diff
@@ -67,7 +76,36 @@ func TestPkiSecretBackendRole_basic(t *testing.T) {
 		CheckDestroy: testPkiSecretBackendRoleCheckDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testPkiSecretBackendRoleConfig_basic(name, backend, 3600, 7200),
+				Config: testPkiSecretBackendRoleConfig_basic(name, backend, 3600, 7200, legacyPolicyIdentifiers),
```
vinay-gopalan, line 79: Since we currently don't have access to the latest Vault builds within CI, the acceptance tests for 1.11 fields will unfortunately fail within GitHub Actions. At present, we're skipping 1.11 field tests until we can get the latest version of Vault running in CI :) 

We use `testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)` to do so. (see example [here](https://github.com/hashicorp/terraform-provider-vault/blob/release/vault-next/vault/resource_gcp_auth_backend_test.go#L30))

vinay-gopalan (commented):
PR is looking great! Had some comments/questions

website/docs/r/pki_secret_backend_role.html.md
```diff
@@ -112,7 +112,14 @@ The following arguments are supported:
 
 * `require_cn` - (Optional) Flag to force CN usage
 
-* `policy_identifiers` - (Optional) Specify the list of allowed policies IODs
+* `policy_identifiers` - (Optional) Specify the list of allowed policies OIDs; Deprecated: use `policy_identifier` blocks instead
```
benashz, line 115: I don't believe we have deprecated this field yet. We probably just want to say that it is used for Vault 1.10 and below, or something like that.

vault/resource_pki_secret_backend_role_test.go
```diff
@@ -12,6 +12,17 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/testutil"
 )
 
+var legacyPolicyIdentifiers = `policy_identifiers = ["1.2.3.4"]`
```
benashz, line 15: Since the `vault` project is already so massive, I'd suggest that we either move theses vars within the test's scope, or include `test` in their names.

benashz (commented):
Looking good! I left some initial feedback/suggestions.

vault/resource_pki_secret_backend_role.go
swenson: Awesome, I did not know about this.

vault/pki.go
```diff
@@ -0,0 +1,79 @@
+package vault
```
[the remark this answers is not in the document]
swenson: Can do. I was mostly following the pattern I saw of other utilities, e.g., `gcp.go`

vault/pki.go
```diff
@@ -0,0 +1,79 @@
...
+	} else {
+		return nil
+	}
+}
+
+// makePkiPolicyIdentifiersListOrSet converts the Vault "policy_identifiers" response
+// into either a list of OIDs, i.e., ["1.2.3","4.5.6"], or a set to represent
+// `policy_identifier` blocks. We return either of these so that round-tripping is stable,
+// and to preserve backwards compatibility with previous versions of Vault.
+func makePkiPolicyIdentifiersListOrSet(rawPolicyIdentifiers []interface{}) ([]string, *schema.Set, error) {
```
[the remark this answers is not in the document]
swenson: I don't think this is true for this function. Here, we have to inspect the response from Vault to see if an entry is JSON or a plain string, and we won't necessarily know which ahead of time, since they come in on the same field in `secret.Data`.

vault/pki.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/hashicorp/terraform-provider-vault/helper"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// readPolicyIdentifiers converts the `policy_identifiers` list and `policy_identifier` blocks
+// into a list of strings (the OIDs) or the JSON serialization of the `policy_identifier` blocks,
+// respectively.
+func readPolicyIdentifiers(d *schema.ResourceData) interface{} {
+	policyIdentifiersList := d.Get("policy_identifiers").([]interface{})
```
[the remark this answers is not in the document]
swenson: :+1:

vault/resource_pki_secret_backend_role_test.go
swenson: :+1:

vault/resource_pki_secret_backend_role_test.go
swenson: :+1:

internal/pki.go
```diff
@@ -0,0 +1,64 @@
+package internal
```
vinay-gopalan, line 1: I think it might be better to instead make this live under `internal/pki` and call it something like `package pki`. We did something similar for packages like `internal/consts` and `internal/identity/entity`. Importing might make things more explictly clear as well with things like `pki.ReadPolicyIdentifierBlocks`

swenson: :+1: makes sense

vault/resource_pki_secret_backend_role.go
```diff
@@ -631,8 +659,10 @@ func pkiSecretBackendRoleUpdate(d *schema.ResourceData, meta interface{}) error
 		data["ext_key_usage"] = extKeyUsage
 	}
 
-	if len(policyIdentifiers) > 0 {
-		data["policy_identifiers"] = policyIdentifiers
+	if len(policyIdentifiersList) > 0 {
+		data["policy_identifiers"] = policyIdentifiersList
+	} else if policyIdentifierBlocks != "" {
+		data["policy_identifiers"] = policyIdentifierBlocks
```
benashz, line 665: Same suggestion as in  the `Read()` method. Ideally we would merge the two and rely on `d.HasChange()`, but I think that is probably out of scope for this PR.

```suggestion
	if v, ok := d.GetOk("policy_identifiers"); ok {
		data["policy_identifiers"] = v
	} else if v, ok := d.GetOk("policy_identifier"); ok {
		data["policy_identifiers"] = internal.ReadPolicyIdentifierBlocks(v.(*schema.Set))
	}
```

vault/resource_pki_secret_backend_role.go
```diff
@@ -572,11 +603,8 @@ func pkiSecretBackendRoleUpdate(d *schema.ResourceData, meta interface{}) error
 		extKeyUsage = append(extKeyUsage, iUsage.(string))
 	}
 
-	iPolicyIdentifiers := d.Get("policy_identifiers").([]interface{})
-	policyIdentifiers := make([]string, 0, len(iPolicyIdentifiers))
-	for _, iIdentifier := range iPolicyIdentifiers {
-		policyIdentifiers = append(policyIdentifiers, iIdentifier.(string))
-	}
+	policyIdentifiersList := d.Get("policy_identifiers").([]interface{})
+	policyIdentifierBlocks := internal.ReadPolicyIdentifierBlocks(d.Get("policy_identifier").(*schema.Set))
```
benashz:
```suggestion
```

website/docs/r/pki_secret_backend_role.html.md
```diff
@@ -112,7 +112,15 @@ The following arguments are supported:
 
 * `require_cn` - (Optional) Flag to force CN usage
 
-* `policy_identifiers` - (Optional) Specify the list of allowed policies IODs
+* `policy_identifiers` - (Optional) Specify the list of allowed policies OIDs. Use with Vault 1.10 or before. For Vault 1.11+, use `policy_identifier` blocks instead
+
+* `policy_identifier` - (Optional) (Vault 1.11+ only) A block for specifying policy identifers. The `policy_identifier` block can be repeated, and supports the following arguments:
```
benashz, line 117: It might be worth adding a new example for this new feature.

vault/resource_pki_secret_backend_role_test.go
```diff
@@ -7,11 +7,98 @@ import (
...
+				Config: testPkiSecretBackendRoleConfig_basic(name, backend, 3600, 7200, newPolicyIdentifiers),
+				Check: resource.ComposeTestCheckFunc(
+					append(checks,
+						resource.TestCheckResourceAttr(resourceName, "policy_identifier.#", "2"),
+						resource.TestCheckTypeSetElemNestedAttrs(resourceName, "policy_identifier.*", map[string]string{"oid": "1.2.3.4.5", "cps": "https://example.com/cps", "notice": "Some notice"}),
+						resource.TestCheckTypeSetElemNestedAttrs(resourceName, "policy_identifier.*", map[string]string{"oid": "1.2.3.4.5.6"}),
+					)...,
+				),
+			},
+		},
```
benashz, line 98: We might want test that the two fields are mutually exclusive by adding a `ExpectError`  `TestStep` similar to:

https://github.com/hashicorp/terraform-provider-vault/blob/70a187215f580d6dae538d9eb3c0e17540ac337a/vault/data_source_gcp_auth_backend_role_test.go#L99-L104

benashz (commented):
Nice update! I have a few more comments/suggestions.

vault/pki.go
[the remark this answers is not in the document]
benashz: Ah, that's right the field has a variable data type from Vault.

vault/resource_pki_secret_backend_role_test.go
```diff
@@ -7,11 +7,98 @@ import (
...
+			{
+				Config: testPkiSecretBackendRoleConfig_basic(name, backend, 3600, 7200, newPolicyIdentifiers),
+				Check: resource.ComposeTestCheckFunc(
+					append(checks,
+						resource.TestCheckResourceAttr(resourceName, "policy_identifier.#", "2"),
+						resource.TestCheckTypeSetElemNestedAttrs(resourceName, "policy_identifier.*", map[string]string{"oid": "1.2.3.4.5", "cps": "https://example.com/cps", "notice": "Some notice"}),
+						resource.TestCheckTypeSetElemNestedAttrs(resourceName, "policy_identifier.*", map[string]string{"oid": "1.2.3.4.5.6"}),
+					)...,
+				),
+			},
```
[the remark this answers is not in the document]
swenson: :+1:

vault/resource_pki_secret_backend_role.go
```diff
@@ -488,10 +516,9 @@ func pkiSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
 		extKeyUsage = append(extKeyUsage, iUsage.(string))
 	}
 
-	iPolicyIdentifiers := secret.Data["policy_identifiers"].([]interface{})
-	policyIdentifiers := make([]string, 0, len(iPolicyIdentifiers))
-	for _, iIdentifier := range iPolicyIdentifiers {
-		policyIdentifiers = append(policyIdentifiers, iIdentifier.(string))
+	legacyPolicyIdentifiers, newPolicyIdentifiers, err := internal.MakePkiPolicyIdentifiersListOrSet(secret.Data["policy_identifiers"].([]interface{}))
```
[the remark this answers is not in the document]
swenson: :+1:

website/docs/r/pki_secret_backend_role.html.md
swenson: :+1:

vault/resource_pki_secret_backend_role_test.go
swenson: Ah, perfect. I had wanted to do that, but didn't quite know how. That helps.

vault/resource_pki_secret_backend_role.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-
+	"github.com/hashicorp/terraform-provider-vault/internal/pki"
```
benashz, line 12: 👍

benashz (approved):
LGTM!

swenson:
Thanks for all of the helpful feedback!

PR #1499, opened 2022-06-13
WIP: Transit BYOK

swenson:
Using Transit BYOK is kind of weird with Terraform, and I'm not sure if this an approach we like.

I tried to create a separate resource, but it doesn't quite work -- importing a key into transit creates a normal key, so the path doesn't quite line up.

I compromised on this approach: if `ciphertext` is present, then we will try to import the key (or `import_version` the key if it changes).

Let me know if that makes sense, or what other suggestions you might have. If this is good, I can work on adding tests.

## Testing
Testing this is tricky.
 
 First, setup Vault with transit:
 
 ```hcl
 provider "vault" {
}

resource "vault_mount" "transit" {
  path                      = "transit"
  type                      = "transit"
}
```

After that is setup, you need to read the wrapping key in order to import keys:

```sh
vault read -field=public_key transit/wrapping_key > wrapping_key.pem
```

You'll need SoftHSM installed, and also [pkcs11-key-wrap](https://github.com/smallstep/pkcs11-key-wrap)
```sh
go install github.com/smallstep/pkcs11-key-wrap@latest
```

Here is how to create and wrap a key:

```sh
softhsm2-util --init-token --slot 0 --label "My token 1" --so-pin 1234 --pin 1234
pkcs11-tool --module /opt/homebrew/Cellar/softhsm/2.6.1/lib/softhsm/libsofthsm2.so --slot 0 -l -p 1234 -k --key-type rsa:2048 -a somelabel --extractable
pkcs11-key-wrap -module /opt/homebrew/Cellar/softhsm/2.6.1/lib/softhsm/libsofthsm2.so -slot 0 -pin 1234 -wrapping-key wrapping_key.pem -label somelabel > wrapped.bin
```

Now we can create Terraform resource to import it:

```hcl
resource "vault_transit_secret_backend_key" "some_imported_key" {
  backend = vault_mount.transit.path
  name = "some_imported_key"

  // import
  ciphertext = filebase64("wrapped.bin")
  type = "rsa-2048"
  hash_function = "SHA1"
}
```

Updating the ciphertext will create a new version of the key as well.

### DIFF
--- a/vault/resource_transit_secret_backend_key.go
+++ b/vault/resource_transit_secret_backend_key.go
@@ -156,6 +156,22 @@ func transitSecretBackendKeyResource() *schema.Resource {
 				Computed:    true,
 				Description: "Whether or not the key supports signing, based on key type.",
 			},
+			// import only options
+			"ciphertext": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The base64-encoded ciphertext of the keys. The AES key should be encrypted using OAEP with the wrapping key and then concatenated with the import key, wrapped by the AES key.",
+				Sensitive:     true, // not actually sensitive, but it's a big blob and people often think it is sensitive
+				ConflictsWith: []string{"min_encryption_version", "min_decryption_version", "convergent_encryption", "deletion_allowed"},
+			},
+			"hash_function": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Default:       "SHA256",
+				Description:   "The hash function used as a random oracle in the OAEP wrapping of the user-generated, ephemeral AES key. Can be one of SHA1, SHA224, SHA256 (default), SHA384, or SHA512",
+				ValidateFunc:  validation.StringInSlice([]string{"SHA1", "SHA224", "SHA256", "SHA384", "SHA512"}, false),
+				ConflictsWith: []string{"min_encryption_version", "min_decryption_version", "convergent_encryption", "deletion_allowed"},
+			},
 		},
 		CustomizeDiff: customdiff.All(
 			customdiff.ValidateChange("exportable", func(_ context.Context, old, new, meta interface{}) error {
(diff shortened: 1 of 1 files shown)


benashz:
I wonder if adding a new `vault_transit_secret_backend_import_key` resource might be a better alternative to updating the current resource. That approach seems to align with the API: https://www.vaultproject.io/api-docs/secret/transit#import-key. Thoughts?

swenson:
I tried that route, but I couldn't quite make it work because the path wouldn't match -- the import command still creates a `vault_transit_secret_backend_key` resource, after all.

vault/resource_transit_secret_backend_key.go
```diff
@@ -156,6 +156,22 @@ func transitSecretBackendKeyResource() *schema.Resource {
 				Computed:    true,
 				Description: "Whether or not the key supports signing, based on key type.",
 			},
+			// import only options
+			"ciphertext": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The base64-encoded ciphertext of the keys. The AES key should be encrypted using OAEP with the wrapping key and then concatenated with the import key, wrapped by the AES key.",
+				Sensitive:     true, // not actually sensitive, but it's a big blob and people often think it is sensitive
+				ConflictsWith: []string{"min_encryption_version", "min_decryption_version", "convergent_encryption", "deletion_allowed"},
```
benashz, line 165: Should any of these not already `Computed` fields  be?

vault/resource_transit_secret_backend_key.go
```diff
@@ -156,6 +156,22 @@ func transitSecretBackendKeyResource() *schema.Resource {
 				Computed:    true,
 				Description: "Whether or not the key supports signing, based on key type.",
 			},
+			// import only options
+			"ciphertext": {
```
benashz, line 160: I wonder if it would make sense to also add a `bool` field that denotes that this `key` is an import/foreign? Could be overkill,  but it would be more explicit.

vault/resource_transit_secret_backend_key.go
```diff
@@ -409,11 +491,21 @@ func transitSecretBackendKeyUpdate(d *schema.ResourceData, meta interface{}) err
 		"auto_rotate_period":     getTransitAutoRotatePeriod(d),
 	}
 
-	_, err := client.Logical().Write(path+"/config", data)
-	if err != nil {
-		return fmt.Errorf("error updating transit secret backend key %q: %s", path, err)
+	// check for import_version
+	ciphertext := d.Get("ciphertext").(string)
```
benashz, line 495: update for related suggestion in `transitSecretBackendKeyUpdate()`

vault/resource_transit_secret_backend_key.go
```diff
@@ -236,6 +263,61 @@ func transitSecretBackendKeyCreate(d *schema.ResourceData, meta interface{}) err
...
+func transitSecretBackendKeyImport(d *schema.ResourceData, meta interface{}) error {
+	client, e := provider.GetClient(d, meta)
+	if e != nil {
+		return e
+	}
+
+	backend := d.Get("backend").(string)
+	name := d.Get("name").(string)
+	autoRotatePeriod := getTransitAutoRotatePeriod(d)
+	data := map[string]interface{}{
```
benashz, line 275: This data map is now duplicated in 3 places here,  `transitSecretBackendKeyCreate()`, and `transitSecretBackendKeyUpdate()`.  One suggestion would be to have `Create()`, call `Update()`, another would be to merge the two functions.

benashz (commented):
Looking good. I added some initial comments/suggestions.

I do wonder how we would transition between imported and non-imported keys?  We will definitely want test that out. I made a suggestion about adding an explicit `import` boolean field, we could possible make that a `ForceNew`. That would trigger resource recreation, yet still allow the `ciphertext` to be mutated.

swenson:
Picking this back up and hoping to have it finished for 3.11.0

PR #1505, opened 2022-06-16
add namespace fields in KV docs

vinay-gopalan:
Updates KV resource and datasource docs with the `namespace` field.


benashz (approved):
Looks good! Would be great if there was a way to include doc fragments. There might be with the newer TF docs layout?

vinay-gopalan:
> Looks good! Would be great if there was a way to include doc fragments. There might be with the newer TF docs layout?

+1 would definitely be nice to autogenerate docs for fields like these! will look into fragments and the new TF docs layout 👍🏼

PR #1508, opened 2022-06-23
Add Managed Keys support to TFVP

vinay-gopalan:
This PR adds support for Managed keys to the PKI Secret Engine and adds a new resource `vault_managed_keys` to the TFVP.

```
$ make testacc TESTARGS='-v -run TestResourceMountManagedKeys'
=== RUN   TestResourceMountMangedKeys
--- PASS: TestResourceMountMangedKeys (1.78s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     3.683s

$ make testacc TESTARGS='-v -run TestManagedKeys'
=== RUN   TestManagedKeys
--- PASS: TestManagedKeys (1.72s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     2.221s


```


vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+		path := getManagedKeysPath(KMSTypeAzure, azureKeyName)
+
+		log.Printf("[DEBUG] Deleting managed key %s", path)
+		_, err := client.Logical().Delete(path)
+		if err != nil {
+			return diag.Errorf("error deleting managed key %s", path)
+		}
+		log.Printf("[DEBUG] Deleted managed key %q", path)
+	}
+
```
benashz, line 471: Do we support deleting the `PKCS` type?

vault/resource_pki_secret_backend_config_ca.go
```diff
@@ -25,45 +26,64 @@ func pkiSecretBackendConfigCAResource() *schema.Resource {
 			},
 			"pem_bundle": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
```
benashz, line 29: So if we are using KMS then the `pem_bundle` is not required? Not sure why this is now optional.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+func managedKeysAzureConfigSchema() schemaMap {
+	s := schemaMap{
+		"name": {
+			Type:     schema.TypeString,
+			Required: true,
+			ForceNew: true,
+			Description: "A unique lowercase name that serves as " +
+				"identifying the key",
+		},
+		"tenant_id": {
```
benashz, line 238: Since this is new code, I would recommend pulling all of these out to constants, adding them to the `consts` package should be fine.

benashz (commented):
Looking good. Looks like we are missing the website docs though.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+	}
+
+	// @TODO figure out what the ID should be
+	d.SetId("sys/managed-keys")
+
+	return managedKeysRead(ctx, d, meta)
+}
+
+func managedKeysRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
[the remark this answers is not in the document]
vinay-gopalan: Split into separate functions in latest commits

vault/resource_managed_keys.go
vinay-gopalan: Yup! Happened to miss deletion for PKCS in previous commit, thanks for the catch! Added in latest commits

vault/resource_managed_keys.go
vinay-gopalan: Agreed. Factored out the fields into the `internal/consts` package

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "The AWS access key to use. This can also " +
+				"be provided with the 'AWS_ACCESS_KEY_ID' env variable",
+		},
+		"secret_key": {
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "The AWS secret key to use. This can also " +
+				"be provided with the 'AWS_SECRET_ACCESS_KEY' env variable",
```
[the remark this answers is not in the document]
vinay-gopalan: Bad copypasta on description. Fixed in latest commits

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+)
+
+const (
+	KMSTypePKCS  = "pkcs11"
```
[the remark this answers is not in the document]
vinay-gopalan: Good idea. Made these fields as private constants

vault/resource_mount.go
```diff
@@ -109,6 +109,14 @@ func getMountSchema(excludes ...string) schemaMap {
 			ForceNew:    true,
 			Description: "Enable the secrets engine to access Vault's external entropy source",
 		},
+
+		"allowed_managed_keys": {
+			Type:        schema.TypeList,
```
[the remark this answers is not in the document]
vinay-gopalan: Updated as a `TypeSet` in latest commits

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,473 @@
...
+		awsKeyPath := getManagedKeysPath(KMSTypeAWS, awsKeyName)
+		resp, err := client.Logical().Read(awsKeyPath)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		data := map[string]interface{}{}
+		for k := range managedKeysAWSConfigSchema() {
+			if v, ok := resp.Data[k]; ok {
+				data[k] = v
```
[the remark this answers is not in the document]
vinay-gopalan: The fields such as `access_key` and `secret_key` are in fact returned by Vault, but instead they are being returned as `"redacted"`. In this case I don't think using `continue` would help, since we will always be able to get our value using `resp.Data[k]`; if we skip the logic below that specifically handles those fields, then we will be in a perpetual drift state for the `"redacted"` value fields

vault/resource_mount_test.go
```diff
@@ -217,6 +217,56 @@ func TestResourceMount_ExternalEntropyAccess(t *testing.T) {
 	})
 }
 
+func TestResourceMountMangedKeys(t *testing.T) {
```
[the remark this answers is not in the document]
vinay-gopalan: Since it's quite a different config and setup, and since the test cases are also slightly tailored, I figured it might be worth it to have a separate test for it; the other mount tests use config functions and checks which are shared between multiple tests, and I didn't want to break any of the existing code with my work or trample over it. Thoughts?

vault/resource_pki_secret_backend_config_ca.go
vinay-gopalan: I had initially gotten this from the RFC:

```
we will also make pem_bundle optional, to allow updating the managed key name parameter after a key has already been established, either via the API or directly in the HSM.  This allows for operators to change the configuration in the future.
```

Looking more into the Vault docs however, I realized that this was no longer a necessary change. Reverted in latest commits

website/docs/r/managed_keys.html.md
```diff
@@ -0,0 +1,151 @@
...
+* `force_rw_session` - (Optional) Force all operations to open up a read-write session to
+  the HSM.
+
+
+## Import
+
+Mounts can be imported using the `id`, e.g.
+
+```
+$ terraform import vault_managed_keys.my_keys sys/managed-keys
```
vinay-gopalan, line 149: @benashz I'm unsure how the Import of this resource should work

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,690 @@
...
+	// set ID to 'default'
+	d.SetId("default")
+
+	return readManagedKeys(ctx, d, meta)
+}
+
+func getRedactedFields(d *schema.ResourceData, providerType, name string, fields []string) map[string]interface{} {
+	m := map[string]interface{}{}
+
+	if mKeysBlocks, ok := d.GetOk(providerType); ok {
```
benashz, line 432: Is there a specific reason why we need to iterate over all config resources for every `name`? I  wonder if we could fetch the  values directly from `d` e.g: `d.GetOk(providerType+"."+hash(name)+"."+field)` ?

benashz (commented):
Some additional comments.

vault/resource_managed_keys.go
vinay-gopalan: Ah that's a good point! I didn't think to use `hash(name)` to achieve that. I'll make the necessary changes

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,690 @@
...
+					m[k] = v
+				}
+			}
+
+			// set these from TF config since they are
+			// returned as "redacted"
+			redactedFields := []string{"access_key", "secret_key"}
+			redactedData := getRedactedFields(d, consts.FieldAWS, name.(string), redactedFields)
+
+			for k, v := range redactedData {
```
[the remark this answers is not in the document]
vinay-gopalan: Those are great ideas, thanks! I definitely agree with the idea of avoiding the extra `for` loop, so I'm now passing in `m` to `getRedactedFields`

vault/resource_managed_keys.go
vinay-gopalan: Fixed in latest commit

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,655 @@
...
+					}
+				}
+			}
+		}
+	}
+	return m
+}
+
+func readAndSetManagedKeys(d *schema.ResourceData, meta interface{}, providerType string,
+	sm map[string]interface{}, redacted []string) error {
```
benashz: `sm` looks to be  `map[string]string`, we could probably be more explicit here

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,655 @@
...
+
+			if resp == nil {
+				continue
+			}
+
+			for k := range handlers[keyType]() {
+				// Map TF schema fields to Vault API
+				vaultKey := k
+				if v, ok := sm[k]; ok {
+					vaultKey = v.(string)
```
benashz, line 491: if `sm` is `map[string]string` then we could skip the type assert here

vault/resource_managed_keys.go
benashz: maybe rename `redacted` to `redactedFields`

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,655 @@
...
+				}
+
+				if v, ok := resp.Data[vaultKey]; ok {
+					m[k] = v
+				}
+			}
+
+			// set these from TF config since they are
+			// returned as "redacted"
+			redactedData := getRedactedFields(d, providerType, name.(string), redacted)
```
benashz: maybe compute this before the `for` loop above, then you can `get()` the override value within a single `for` loop?

vault/resource_pki_secret_backend_config_ca_test.go
```diff
@@ -41,9 +43,9 @@ resource "vault_mount" "test" {
 }
 
 resource "vault_pki_secret_backend_config_ca" "test" {
-  depends_on = [ "vault_mount.test" ]
-  backend = vault_mount.test.path
-  pem_bundle = <<EOT
+  depends_on       = ["vault_mount.test"]
```
benashz, line 46: Since backend references vault_mount.test there's no need to declare the extra dependency here.

```suggestion
```

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,646 @@
...
+		consts.FieldUUID: {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "ID of the managed key read from Vault",
+		},
+	}
+}
+
+func setCommonManagedKeysSchema(s schemaMap) schemaMap {
+	for k, v := range getCommonManagedKeysSchema() {
```
benashz, line 119: I think you probably want  https://github.com/hashicorp/terraform-provider-vault/blob/aaa7b1a5ba8caeba1ed1c8fd05a49ef0f695ba26/vault/provider.go#L822, which will panic on duplicate fields. Or add your own check for `key` duplication.

benashz (commented):
Getting there :)

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,646 @@
...
+	d.SetId("default")
+
+	return readManagedKeys(ctx, d, meta)
+}
+
+func getRedactedFields(d *schema.ResourceData, providerType, name string,
+	fields []string, m map[string]interface{},
+) map[string]interface{} {
+	for _, field := range fields {
+		k := fmt.Sprintf("%s.%d.%s", providerType, helper.HashCodeString(name), field)
```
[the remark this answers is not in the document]
vinay-gopalan: Updated in [3d04163](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/3d0416345363e61a932b93acbbeed257b1cbee9b)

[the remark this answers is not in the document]
vinay-gopalan: Updated in [3d04163](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/3d0416345363e61a932b93acbbeed257b1cbee9b)

vault/resource_pki_secret_backend_intermediate_cert_request_test.go
```diff
@@ -35,6 +35,36 @@ func TestPkiSecretBackendIntermediateCertRequest_basic(t *testing.T) {
 	})
 }
 
+func TestPkiSecretBackendIntermediateCertRequest_managedKeys(t *testing.T) {
+	// Remove once we are running Ent HSM binaries in CI
+	testutil.SkipTestEnvUnset(t, "TF_ACC_ENT_HSM")
```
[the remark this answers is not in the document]
vinay-gopalan: Added a JIRA to the backlog and updated the comment

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,655 @@
...
+	switch providerType {
+	case consts.FieldAWS:
+		keyType = kmsTypeAWS
+
+	case consts.FieldPKCS:
+		keyType = kmsTypePKCS
+
+	case consts.FieldAzure:
+		keyType = kmsTypeAzure
+	}
```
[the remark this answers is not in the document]
vinay-gopalan: Good call, added in [3d04163](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/3d0416345363e61a932b93acbbeed257b1cbee9b)

vault/resource_managed_keys.go
vinay-gopalan: Ah nice catch! I added a key duplication check in my function; I didn't want to change all the schema function signatures to match the `mustAddSchema` signature, but the duplication was a quick fix :)

vault/resource_managed_keys.go
vinay-gopalan: @benashz I was a bit unsure what you meant by this part; how would we consolidate everything into one loop? Wouldn't we still need to loop over all redacted fields within the `getRedactedFields` function?

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+				}
+
+				if v, ok := resp.Data[vaultKey]; ok {
+					m[k] = v
+				}
+			}
+
+			// get these from TF config since they are
+			// returned as "redacted" from Vault
+			m = getRedactedFields(d, providerType, name.(string), redactedFields, m)
```
benashz, line 516: nit: since `getRedactedFields()` is mutating `m` , a better name might be `updateRedactedFields` instead. Not sure it needs to return the map it just mutated either.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+	if err := d.Set(providerType, s); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readAWSManagedKeys(d *schema.ResourceData, meta interface{}) error {
+	redacted := []string{"access_key", "secret_key"}
+	sm := map[string]string{
```
benashz, line 533: nit: probably could inline this slice.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+	if err := readAndSetManagedKeys(d, meta, consts.FieldAWS, sm, redacted); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readAzureManagedKeys(d *schema.ResourceData, meta interface{}) error {
+	var redacted []string
+	sm := map[string]string{
```
benashz, line 545: nit: suggest inlining this slice.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+	if err := readAndSetManagedKeys(d, meta, consts.FieldAzure, sm, redacted); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readPKCSManagedKeys(d *schema.ResourceData, meta interface{}) error {
+	redacted := []string{"pin"}
+	sm := map[string]string{
```
benashz, line 557: nit: suggest inlining this slice.

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+
+	err := readAWSManagedKeys(d, meta)
+	if err != nil {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  fmt.Sprintf("Failed to read AWS Managed Keys, err=%s", err),
+		})
+	}
+
+	err = readPKCSManagedKeys(d, meta)
```
benashz, line 578: nit: we can avoid the shadowed `err`  if we change these to be of the form `if err := .... ; err != nil {...}`

benashz (commented):
Looking good!

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+		}
+	}
+
+	return result
+}
+
+func getHashFromName(name string) int {
+	return helper.HashCodeString(name)
+}
+
```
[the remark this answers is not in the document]
vinay-gopalan: Fixed in [2347120](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/2347120bc674a335eb4ca5a02a893c098bd2cb4e)

vault/resource_managed_keys.go
```diff
@@ -0,0 +1,661 @@
...
+	case consts.FieldPKCS:
+		keyType = kmsTypePKCS
+
+	case consts.FieldAzure:
+		keyType = kmsTypeAzure
+
+	default:
+		return fmt.Errorf("received unexpected provider type %s", providerType)
+	}
+
```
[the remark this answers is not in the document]
vinay-gopalan: Refactored in [2347120](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/2347120bc674a335eb4ca5a02a893c098bd2cb4e)

vault/resource_managed_keys.go
vinay-gopalan: Renamed in [2347120](https://github.com/hashicorp/terraform-provider-vault/pull/1508/commits/2347120bc674a335eb4ca5a02a893c098bd2cb4e)

vault/resource_mount.go
```diff
@@ -108,6 +108,14 @@ func getMountSchema(excludes ...string) schemaMap {
 			ForceNew:    true,
 			Description: "Enable the secrets engine to access Vault's external entropy source",
 		},
+
+		"allowed_managed_keys": {
```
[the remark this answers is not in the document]
vinay-gopalan: Updated docs for resources in latest commit

vault/resource_managed_keys_test.go
```diff
@@ -0,0 +1,194 @@
...
+					}
+
+					_, err = client.Logical().Write(p, data)
+					if err != nil {
+						t.Fatalf("could not create managed key to Vault")
+					}
+				},
+				Config: testManagedKeysConfig_updated(name0),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "aws.#", "1"),
```
benashz, line 132: This doesn't really prove that the a delete/create operation took place. In the end you should end up with a new UUID that is different from the original created by the provider and the one created manually.

vault/resource_managed_keys_test.go
```diff
@@ -0,0 +1,194 @@
...
+							consts.FieldKMSKey:       "alias/test_identifier_string_2",
+							consts.FieldAWSAccessKey: "ASIAKBASDADA09090",
+							consts.FieldAWSSecretKey: "8C7THtrIigh2rPZQMbguugt8IUftWhMRCOBzbuyz",
+						},
+					),
+				),
+			},
+			// This test removes one of the managed keys from the set
+			// and also updates the name for the other remaining key
+			// Tests: ForceNew on Name + Deletion of removed keys
```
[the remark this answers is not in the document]
vinay-gopalan: Updated

benashz (approved):
Ship it!

vault/resource_managed_keys_test.go
vinay-gopalan: Tracking this task in a separate Jira. Thanks for the callout!

PR #1515, opened 2022-06-27
Adds support for kubernetes secrets engine

austingebauer:
### Community Note

This PR adds resources and a data source for Vault [Kubernetes Secrets Engine](https://www.vaultproject.io/docs/secrets/kubernetes).

Closes: https://github.com/hashicorp/terraform-provider-vault/issues/1522

```release-note
* Adds support for Vault [Kubernetes Secrets Engine](https://www.vaultproject.io/docs/secrets/kubernetes) (#1515)
```

```
$ make testacc TESTARGS='-count=1 -run=TestAccKubernetesSecret'
TF_ACC=1 go test -v -count=1 -run=TestAccKubernetesSecret -timeout 30m ./...
=== RUN   TestAccKubernetesSecretCredentialsDataSource
--- PASS: TestAccKubernetesSecretCredentialsDataSource (1.50s)
=== RUN   TestAccKubernetesSecretBackendRole
--- PASS: TestAccKubernetesSecretBackendRole (3.12s)
=== RUN   TestAccKubernetesSecretBackend
--- PASS: TestAccKubernetesSecretBackend (3.14s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     9.856s
```


vault/resource_kubernetes_secret_backend.go
```diff
@@ -0,0 +1,178 @@
...
+		DeleteContext: kubernetesSecretBackendDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: getKubernetesSecretsMountSchema(),
+	}
+}
+
+func getKubernetesSecretsMountSchema() map[string]*schema.Schema {
+	s := getMountSchema("type")
```
austingebauer, line 36: This ended up being an easy way to get all of the mount config and tuning options without rewriting code. Happy to consider a different approach if you all would prefer that. It would be nice to have some common mount config code to share for secrets engines.

vault/resource_kubernetes_secret_backend_role.go
```diff
@@ -0,0 +1,257 @@
...
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "The list of Kubernetes namespaces this role can generate " +
+					"credentials for. If set to '*' all namespaces are allowed.",
+				Required: true,
+			},
+			fieldTokenMaxTTL: {
+				Type: schema.TypeString,
```
austingebauer, line 68: I thought that we could take in Go duration strings as config, but it turns out this is a bit difficult to get right. The API takes a Go duration string and returns integer seconds. I thought I could use [`flattenVaultDuration`](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/structures.go#L113) to achieve going back from integer seconds to duration string, but it doesn't work correctly in some situations.

Is anyone aware of examples of how to do this properly? If not, I'll change these to take in integer seconds instead.

austingebauer:
Documentation is still in progress. Will update this PR when they're ready for review.

tvoran (commented):
Looks good so far, and I tested it a bit locally too.

vault/resource_kubernetes_secret_backend.go
```diff
@@ -0,0 +1,178 @@
...
+				"local pod’s CA if found, or otherwise the host's root CA set.",
+			Optional: true,
+			Computed: true,
+		},
+		fieldServiceAccountJWT: {
+			Type: schema.TypeString,
+			Description: "The JSON web token of the service account used by the " +
+				"secrets engine to manage Kubernetes credentials. Defaults to the " +
+				"local pod’s JWT if found.",
+			Optional: true,
```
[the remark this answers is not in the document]
austingebauer: I think you're right. I marked it as sensitive in 4660948. Thanks!

vault/resource_kubernetes_secret_backend_role.go
austingebauer: Hmm, I tried to play around with some different options here and didn't find anything I liked. I'm going to switch these over to seconds and add a `_seconds` suffix to the fields. Thanks, all!

benashz: I think it is fine to forgo the extra `_seconds` suffix. We have standardized on seconds (int) for all TTL type fields, since it better matches the read from Vault.

vault/data_source_kubernetes_credentials.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+const (
+	fieldKubernetesNamespace     = "kubernetes_namespace"
+	fieldClusterRoleBinding      = "cluster_role_binding"
+	fieldServiceAccountNamespace = "service_account_namespace"
+	fieldServiceAccountToken     = "service_account_token"
+)
+
+func kubernetesServiceAccountTokenDataSource() *schema.Resource {
```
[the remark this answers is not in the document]
benashz: @vinay-gopalan I think it is fine to continue as-is, since we already have a precedent set here.

internal/consts/consts.go
```diff
@@ -32,12 +42,13 @@ const (
...
-	MountTypeKMIP     = "kmip"
-	MountTypeRabbitMQ = "rabbitmq"
-	MountTypeNomad    = "nomad"
+	MountTypeDatabase   = "database"
+	MountTypePKI        = "pki"
+	MountTypeAWS        = "aws"
+	MountTypeKMIP       = "kmip"
+	MountTypeRabbitMQ   = "rabbitmq"
+	MountTypeNomad      = "nomad"
+	MountTypeKubernetes = "kubernetes"
```
benashz, line 51: 👍

vault/resource_kubernetes_secret_backend.go
```diff
@@ -0,0 +1,179 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: getKubernetesSecretsMountSchema(),
+	}
+}
+
+func getKubernetesSecretsMountSchema() map[string]*schema.Schema {
+	s := getMountSchema("type")
+	for k, v := range getKubernetesSecretsSchema() {
```
benashz, line 37: Suggest replacing with `mustAddSchema()`  https://github.com/hashicorp/terraform-provider-vault/blob/b914b1333431cda7c29180818359411c16145c0e/vault/provider.go#L810

vault/resource_kubernetes_secret_backend.go
```diff
@@ -0,0 +1,179 @@
...
+			Optional:    true,
+			Computed:    true,
+		},
+		fieldKubernetesCACert: {
+			Type: schema.TypeString,
+			Description: "A PEM-encoded CA certificate used by the secret engine to " +
+				"verify the Kubernetes API server certificate. Defaults to the " +
+				"local pod’s CA if found, or otherwise the host's root CA set.",
+			Optional: true,
+			Computed: true,
```
benashz, line 57: Will Vault return the pod's CA if we no CA is specified?

vault/resource_mount.go
```diff
@@ -176,6 +176,10 @@ func createMount(d *schema.ResourceData, client *api.Client, path string, mountT
 }
 
 func mountUpdate(d *schema.ResourceData, meta interface{}) error {
+	return updateMount(d, meta, false)
+}
+
+func updateMount(d *schema.ResourceData, meta interface{}, excludeType bool) error {
```
benashz, line 182: 👍

vault/resource_kubernetes_secret_backend_test.go
```diff
@@ -0,0 +1,136 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldAuditNonHMACRequestKeys+".#", "0"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldAuditNonHMACResponseKeys+".#", "0"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldSealWrap, "false"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldExternalEntropyAccess, "false"),
+					resource.TestCheckResourceAttr(resourceName, fieldKubernetesHost, "https://127.0.0.1:63247"),
+					resource.TestCheckResourceAttr(resourceName, fieldKubernetesCACert, "test_ca_cert"),
+					resource.TestCheckResourceAttr(resourceName, fieldDisableLocalCAJWT, "true"),
+				),
+			},
+			{
```
benashz, line 77: nit: I usually like to include an import test step after each `Config` step.

vault/resource_kubernetes_secret_backend_test.go
```diff
@@ -0,0 +1,136 @@
...
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{fieldServiceAccountJWT},
+			},
+		},
+	})
+}
+
+func testAccKubernetesSecretBackendCheckDestroy(path string) resource.TestCheckFunc {
```
benashz, line 87: Suggest replacing with https://github.com/hashicorp/terraform-provider-vault/blob/c520061319a277600ea367adacdcf7e17d2820ee/vault/testhelpers_test.go#L14

vault/resource_kubernetes_secret_backend.go
benashz: I like that we are including the mount creation with this new resource, normally when provisioning a secret's engine the operator would need to provide `vault_mount` resource.

vault/resource_kubernetes_secret_backend.go
```diff
@@ -0,0 +1,179 @@
...
+		fieldDisableLocalCAJWT,
+	}
+	for _, k := range fields {
+		if d.HasChange(k) {
+			data[k] = d.Get(k)
+		}
+	}
+
+	// kubernetes_host always needs to be provided on configuration updates.
+	// Otherwise, an error will occur if the KUBERNETES_SERVICE_HOST and
```
benashz, line 110: This would be in the context of a k8s  pod?

vault/resource_kubernetes_secret_backend_role.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+)
+
+var (
+	kubernetesSecretBackendFromPathRegex = regexp.MustCompile("^(.+)/roles/.+$")
+)
+
+const (
+	fieldAllowedKubernetesNamespaces = "allowed_kubernetes_namespaces"
+	fieldTokenMaxTTL                 = "token_max_ttl"
```
benashz, line 22: Suggest using: https://github.com/hashicorp/terraform-provider-vault/blob/d97e63ca7f3752779e1616e96547abdd3b719521/vault/auth_token.go#L15

vault/resource_kubernetes_secret_backend_role_test.go
```diff
@@ -0,0 +1,167 @@
...
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{fieldServiceAccountJWT},
+			},
+		},
+	})
+}
+
+func testAccKubernetesSecretBackendRoleCheckDestroy(s *terraform.State) error {
```
benashz, line 87: Suggest replacing with https://github.com/hashicorp/terraform-provider-vault/blob/c520061319a277600ea367adacdcf7e17d2820ee/vault/testhelpers_test.go#L14

benashz (commented):
Looks great! I add some initial comments/suggestions.

internal/consts/consts.go
```diff
@@ -4,24 +4,34 @@ const (
...
+	FieldNames                    = "names"
+	FieldLeaseID                  = "lease_id"
+	FieldLeaseDuration            = "lease_duration"
+	FieldLeaseRenewable           = "lease_renewable"
+	FieldDepth                    = "depth"
+	FieldDataJSON                 = "data_json"
+	FieldRole                     = "role"
+	FieldDescription              = "description"
+	FieldTTL                      = "ttl"
+	FieldDefaultLeaseTTL          = "default_lease_ttl_seconds"
```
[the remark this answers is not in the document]
austingebauer: I created this constant to refer to the field in [resource_mount.go#L38](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_mount.go#L38) which is defined with the `_seconds` suffix. I don't think I can remove the suffix, since I'm using the same mount schema directly for the new k8s secrets resource. 

I could just remove this constant completely. I added it here because it seemed like something other secrets engine resources could use (if they follow a similar pattern using the mount schema).

internal/consts/consts.go
```diff
@@ -4,24 +4,34 @@ const (
...
+	FieldMaxLeaseTTL              = "max_lease_ttl_seconds"
```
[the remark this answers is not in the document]
austingebauer: Same comment from above.

vault/resource_kubernetes_secret_backend.go
austingebauer: I actually noticed the opposite on the secrets engines that I looked at as a reference. For example, the following secret engine resources include mount creation/tuning directly (i.e., do not require a `vault_mount` resource): aws, azure, consul, gcp, nomad, tfc. 

I do see a couple of them that require a `vault_mount` resource though.

benashz (commented):
Ah, I noticed that we did not include the website docs. If you could also add/update those, that would be great!

internal/consts/consts.go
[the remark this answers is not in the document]
benashz: Committing the above will probably break the `gofumpt` check...

vault/resource_kubernetes_secret_backend.go
benashz: ah, maybe I was thinking of the database secrets engines.

vault/resource_kubernetes_secret_backend.go
austingebauer: Nice, didn't know about that function. Done in 57179fd.

vault/resource_kubernetes_secret_backend.go
austingebauer: Ah, no it will not. This probably shouldn't be computed then, right?

vault/resource_kubernetes_secret_backend_test.go
austingebauer: Cool! Done in ea2dc71.

vault/resource_kubernetes_secret_backend_role.go
austingebauer: Changed this to use integer seconds in cc3b1aa. I left the `_seconds` suffix off of the field name.

austingebauer:
Documentation has been added in fb6dbe4.

vault/resource_kubernetes_secret_backend.go
austingebauer: If Vault isn't running in k8s, then `kubernetes_host` always needs to be provided (even if it's just the empty string). Otherwise, we'll get an error when we try to make the update. You can see this from the logic at https://github.com/hashicorp/vault-plugin-secrets-kubernetes/blob/main/path_config.go#L126-L130.

internal/consts/consts.go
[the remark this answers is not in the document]
austingebauer: Wondering if I should commit the above? Any idea on why it might break the gofumpt check?

vault/resource_kubernetes_secret_backend_role_test.go
benashz: Oops, yeah you are right, I think this comment was meant for the backend resource.

benashz (approved):
Looks great!

vinay-gopalan (approved):

internal/consts/consts.go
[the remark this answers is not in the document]
benashz: whitespace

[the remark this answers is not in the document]
benashz: actually not sure....

PR #1518, opened 2022-06-29
Add support for Consul secrets engine enhancements

robmonte:
This adds supports for service identities and node identities in Consul and for the ability to bootstrap the Consul ACL system by writing the secrets engine config without a token.

Bootstrapping example:
```
resource "vault_consul_secret_backend" "config" {
  path    = "consul"
  address = "http://127.0.0.1:8500"
}
```

Identities example:
```
resource "vault_consul_secret_backend_role" "role" {
  name               = "management"
  backend            = vault_consul_secret_backend.config.path
  node_identities    = ["server-1:dc1"]
  service_identities = ["db-service:dc1"]
}
```

### Community Note

```release-note
IMPROVEMENTS
* resource/vault_consul_secret_backend: Try automatically bootstrapping the Consul ACL system when a token is not provided
* resource/vault_consul_secret_backend_role: Add support for service identities and node identities
```

```
$ make testacc TESTARGS='-run=TestConsulSecretBackendRole'
=== RUN   TestConsulSecretBackendRole
--- PASS: TestConsulSecretBackendRole (2.17s)
=== RUN   TestConsulSecretBackendRoleNameFromPath
--- PASS: TestConsulSecretBackendRoleNameFromPath (0.00s)
=== RUN   TestConsulSecretBackendRoleBackendFromPath
--- PASS: TestConsulSecretBackendRoleBackendFromPath (0.00s)
PASS
```

### DIFF
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -183,10 +183,11 @@ func GetTestNomadCreds(t *testing.T) (string, string) {
 	return v[0], v[1]
 }
 
-func CheckTestVaultVersion(t *testing.T) bool {
+// Returns true if TF_VAULT_VERSION is greater than or equal to the given Vault version
+func CheckTestVaultVersion(t *testing.T, cutoff string) bool {
 	v := SkipTestEnvUnset(t, "TF_VAULT_VERSION")
 
-	cutoffVersion, _ := goversion.NewVersion("1.11")
+	cutoffVersion, _ := goversion.NewVersion(cutoff)
 	envVersion, err := goversion.NewVersion(v[0])
 	if err != nil {
 		t.Fatalf("error parsing vault version from TF_VAULT_VERSION environment variable: %v", err)
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -181,6 +181,11 @@ func GetTestNomadCreds(t *testing.T) (string, string) {
 	return v[0], v[1]
 }
 
+func GetTestVaultVersion(t *testing.T) string {
+	v := SkipTestEnvUnset(t, "TF_VAULT_VERSION")
+	return v[0]
+}
+
 func TestCheckResourceAttrJSON(name, key, expectedValue string) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		resourceState, ok := s.RootModule().Resources[name]
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -19,6 +19,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/hashicorp/vault/api"
 	"github.com/mitchellh/go-homedir"
+
+	goversion "github.com/hashicorp/go-version"
 )
 
 const (
@@ -181,9 +183,18 @@ func GetTestNomadCreds(t *testing.T) (string, string) {
 	return v[0], v[1]
 }
 
-func GetTestVaultVersion(t *testing.T) string {
+func CheckTestVaultVersion(t *testing.T) bool {
 	v := SkipTestEnvUnset(t, "TF_VAULT_VERSION")
-	return v[0]
+
+	cutoffVersion, _ := goversion.NewVersion("1.11")
+	envVersion, err := goversion.NewVersion(v[0])
+	if err != nil {
+		t.Fatalf("error parsing vault version from TF_VAULT_VERSION environment variable: %v", err)
+	} else {
+		return envVersion.GreaterThanOrEqual(cutoffVersion)
+	}
+
+	return false
 }
 
 func TestCheckResourceAttrJSON(name, key, expectedValue string) resource.TestCheckFunc {
--- a/vault/resource_consul_secret_backend_role.go
+++ b/vault/resource_consul_secret_backend_role.go
@@ -48,7 +48,6 @@ func consulSecretBackendRoleResource() *schema.Resource {
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
-				Deprecated:    `Deprecated in favor of "consul_policies" as of Vault 1.11`,
 				ConflictsWith: []string{"consul_policies"},
 			},
 			"consul_policies": {
@@ -117,6 +116,7 @@ func consulSecretBackendRoleResource() *schema.Resource {
 				Optional:    true,
 				Description: "Specifies the type of token to create when using this role. Valid values are \"client\" or \"management\".",
 				Default:     "client",
+				Deprecated:  "Consul 1.11 and later removed the legacy ACL system which supported this field.",
 			},
 			"local": {
 				Type:        schema.TypeBool,
@@ -153,15 +153,20 @@ func consulSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) erro
 
 	path := consulSecretBackendRolePath(backend, name)
 
-	policies := d.Get("policies").([]interface{})
-	consulPolicies := d.Get("consul_policies").(*schema.Set).List()
+	var policies []interface{}
+	if v, ok := d.GetOk("policies"); ok {
+		policies = v.([]interface{})
+	} else if v, ok := d.GetOk("consul_policies"); ok {
+		policies = v.(*schema.Set).List()
+	}
+
 	roles := d.Get("consul_roles").(*schema.Set).List()
 	serviceIdentities := d.Get("service_identities").(*schema.Set).List()
 	nodeIdentities := d.Get("node_identities").(*schema.Set).List()
 
 	data := map[string]interface{}{
 		"policies":           policies,
-		"consul_policies":    consulPolicies,
+		"consul_policies":    policies,
 		"consul_roles":       roles,
 		"service_identities": serviceIdentities,
 		"node_identities":    nodeIdentities,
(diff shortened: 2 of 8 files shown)


vault/resource_consul_secret_backend_role.go
```diff
@@ -125,15 +155,22 @@ func consulSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) erro
...
 
 	policies := d.Get("policies").([]interface{})
+	consulPolicies := d.Get("consul_policies").([]interface{})
 	roles := d.Get("consul_roles").(*schema.Set).List()
+	serviceIdentities := d.Get("service_identities").(*schema.Set).List()
+	nodeIdentities := d.Get("node_identities").(*schema.Set).List()
 
-	if len(policies) == 0 && len(roles) == 0 {
-		return fmt.Errorf("policies or consul_roles must be set")
+	if len(consulPolicies) == 0 && len(roles) == 0 &&
```
austingebauer, line 163: I think you can use [`AtLeastOneOf`](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/schema.go#L286-L298) to have Terraform perform this validation before an apply? Might be worth a try.

austingebauer (approved):
LGTM otherwise.

benashz (commented):
Looking good. I added a few comments/suggestions. Also I think we need to update the website docs.

benashz:
Looks like the build is failing when testing against Vault-1.10, we will probably need special case some of the tests to account for feature differences between 1.10 and 1.11.

vault/resource_consul_secret_backend_role.go
austingebauer: @benashz - Do you happen to know when one would reach for `AtLeastOneOf` if not in this case? It seems like a good thing to catch known configuration issues before an apply and before sending a request to Vault (since we know it fill fail). I agree though -- probably use `AtLeastOneOf` or let Vault take care of validation.

vault/resource_consul_secret_backend_role.go
```diff
@@ -210,14 +247,22 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+		"consul_namespace":   "consul_namespace",
+		"partition":          "partition",
+		"service_identities": "service_identities",
+		"node_identities":    "node_identities",
+	}
+
+	if _, ok := d.GetOk("policies"); ok {
+		params["policies"] = "policies"
+	}
+	if _, ok := d.GetOk("consul_policies"); ok {
```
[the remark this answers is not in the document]
robmonte: These if statements are currently causing the testing to fail, but the CI isn't reaching that test at the moment. I had `else if` originally which actually passes the testing, but I had a concern that me and Vinay discussed so I tentatively changed it to this.

I'm having a hard time reasoning out whether the logic could ever reach certain states involving both values. If Vault currently has roles that use `policies`, but after updating begins using `consul_policies`, would there be a situation where we need to check for both values to return?

vault/resource_consul_secret_backend_role.go
robmonte: Yes, Vault returns an error when none of the above are provided so I think it could be changed to rely on Vault's handling of this situation instead.

vault/resource_consul_secret_backend_role.go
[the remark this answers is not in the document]
vinay-gopalan: The `else if` causes the Import test to fail. `d.GetOk` will return `ok = false` when running the import (`d` doesn't have any attributes in that case other than the ID of the resource), so we never end up setting `consul_policies` to the state during an Import. If we're conditionally setting these state fields only if we see them in `d`, then that might cause failures in the cases where users are trying to import via an ID

vault/resource_consul_secret_backend_role.go
```diff
@@ -210,14 +247,22 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+		params["consul_policies"] = "consul_policies"
```
benashz, line 265: I think this be what you are after?

```suggestion
		"service_identities": "service_identities",
		"node_identities":    "node_identities",
		"policies":           "policies",
		"consul_policies":    "consul_policies",
```

vault/resource_consul_secret_backend_role.go
```diff
@@ -88,6 +117,7 @@ func consulSecretBackendRoleResource() *schema.Resource {
 				Optional:    true,
 				Description: "Specifies the type of token to create when using this role. Valid values are \"client\" or \"management\".",
 				Default:     "client",
+				Deprecated:  "Deprecated in Vault as of 1.11",
```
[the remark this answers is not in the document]
robmonte: The `token_type` field isn't strictly ignored. Even when using Vault 1.11, if the version of Consul used is very old (pre-1.4), then the field will still be needed. It was deprecated in Vault because the most recent versions of Consul completely removed the old ACL system which is the only place where `token_type` gets used.

Would you still prefer purely documentation over a deprecation warning for this?

vault/resource_consul_secret_backend_role.go
robmonte: I opted to fully remove this check instead of using `AtLeastOneOf` because this is logic specifically handled by Vault, which will return a very specific error about missing all of the parameters. Checking in the provider is essentially duplicate code.
https://github.com/hashicorp/terraform-provider-vault/pull/1518/commits/0283556698a7355551d8ee4a2637d696ffc9eb55

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -38,40 +45,155 @@ func TestConsulSecretBackendRole(t *testing.T) {
...
+			}
+		}
+	} else {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func checkVaultVersionEnvConsulPolicies() (bool, error) {
```
vinay-gopalan, line 180: Isn't this the same as the `checkVaultVersionEnvPolicies` above? Can we consolidate some of the code into one function?

robmonte: Oh, I forgot to delete those functions. They are leftover code I forgot to remove I was first trying to use SkipFunc in the TestSteps. However, I couldn't find a way to implement SkipFunc with the tests in a way that worked because it needs to accept a parameter which breaks the expected function definition type.

testutil/testutil.go
```diff
@@ -181,6 +183,20 @@ func GetTestNomadCreds(t *testing.T) (string, string) {
 	return v[0], v[1]
 }
 
+func CheckTestVaultVersion(t *testing.T) bool {
+	v := SkipTestEnvUnset(t, "TF_VAULT_VERSION")
+
+	cutoffVersion, _ := goversion.NewVersion("1.11")
```
vinay-gopalan, line 189: Wondering if it makes sense to add a constant for the latest Vault version as opposed to hard-coding. Might make it easier to update after new version are released?

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -96,7 +141,7 @@ func testAccConsulSecretBackendRoleCheckDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testConsulSecretBackendRole_initialConfig(backend, name, token string, withPolicies, withRoles bool) string {
+func testConsulSecretBackendRole_initialConfig(backend, name, token string, withPolicies, isVersion111OrNewer bool) string {
```
vinay-gopalan, line 144: nit: Should we rename this variable to be more generic, maybe something like `isAboveVersionThreshold` or something?

robmonte: Yeah that sounds better, changing it now.

website/docs/r/consul_secret_backend_role.html.md
```diff
@@ -43,23 +43,30 @@ The following arguments are supported:
...
 * `name` - (Required) The name of the Consul secrets engine role to create.
- 
-* `consul_namespace` - (Optional) The Consul namespace that the token will be created in.
-   Applicable for Vault 1.10+ and Consul 1.7+",
-
-* `partition` - (Optional) The admin partition that the token will be created in.
-   Applicable for Vault 1.10+ and Consul 1.11+",
 
-* `policies` - (Required when `consul_roles` is unset) The list of Consul ACL policies to associate with these roles.
+* `consul_policies` - (Required)<sup><a href="#note-about-required-arguments">SEE NOTE</a></sup> The list of Consul ACL policies to associate with these roles.
```
vinay-gopalan, line 47: I think these fields should still be documented as `Optional`, but we should also make users refer to the note on how at least one of these needs to be provided like you are doing. They are technically all optional but there is a caveat to that

vinay-gopalan (approved):
Awesome work! Couple of nits and questions but PR is looking great

testutil/testutil.go
robmonte: Ah yeah that makes sense. In that case, do you think the `CheckTestVaultVersion` function should accept a second parameter, the Vault version to use as `cutoffVersion`? 
```
func CheckTestVaultVersion(t *testing.T, string cutoff) bool {
   ⋮
   cutoffVersion, _ := goversion.NewVersion(cutoff)
```

vinay-gopalan: That sounds like a great idea!

robmonte: Updated https://github.com/hashicorp/terraform-provider-vault/pull/1518/commits/f7bde28d4de4f4f657c25f16fef9daea43bb5a44

benashz (approved):
LGTM!

robmonte:
Thanks for all of the help!!

PR #1556, opened 2022-07-28
Add test case for Consul ACL bootstrapping

robmonte:
This adds test coverage for automatic Consul ACL Bootstrapping, which is performed when Consul is configured without a token like so:
```
resource "vault_consul_secret_backend" "backend" {
  path        = "consul"
  address     = "http://127.0.0.1:8500"
}
```

The test requires a Consul instance to be available so bootstrapping may actually occur, and thus adds a new required environment variable for the test to run: `TF_CONSUL_ADDR`.

&nbsp;

```release-note
NONE
```

```
$ make testacc TESTARGS='-run=TestConsulSecretBackend_Bootstrap'
=== RUN   TestConsulSecretBackend_Bootstrap
--- PASS: TestConsulSecretBackend_Bootstrap (3.82s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     4.219s
...
```

&nbsp;

### Community Note


robmonte:
The Vault 1.11 test is failing from what appears to be Vault being unable to reach Consul. Looking into if I added the container to GHA incorrectly.

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.0
 	github.com/hashicorp/go-secure-stdlib/awsutil v0.1.6
 	github.com/hashicorp/go-secure-stdlib/parseutil v0.1.6
-	github.com/hashicorp/go-version v1.4.0 // indirect
+	github.com/hashicorp/go-version v1.4.0
```
benashz, line 27: +1

vault/resource_consul_secret_backend_test.go
```diff
@@ -157,6 +196,26 @@ resource "vault_consul_secret_backend" "test" {
 }`, path, token)
 }
 
+func testConsulSecretBackend_bootstrapConfig(path, addr string, double bool) string {
```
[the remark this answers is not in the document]
robmonte: Sounds like a cleaner solution, I'll look into it!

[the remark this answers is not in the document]
robmonte: After a little refactoring it turns out the duplicate resource was unneeded!

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +27,123 @@ func TestConsulSecretBackend(t *testing.T) {
...
+	if testutil.CheckTestVaultVersion(t, "1.11") {
+		resource.Test(t, resource.TestCase{
+			Providers:    testProviders,
+			PreCheck:     func() { testutil.TestAccPreCheck(t) },
+			CheckDestroy: testAccConsulSecretBackendCheckDestroy,
+			Steps: []resource.TestStep{
+				{
+					Config: testConsulSecretBackend_bootstrapConfig(path, consulAddr, false),
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(resourcePath, "path", path),
```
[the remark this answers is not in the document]
robmonte: I've gone ahead and replaced all of the string constants in both `resource_consul_secret_backend_role_test.go` and `resource_consul_secret_backend_test.go` that I could find. Please let me know if you see any missing.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +27,123 @@ func TestConsulSecretBackend(t *testing.T) {
...
 				),
 			},
 		},
 	})
 }
 
+func TestConsulSecretBackend_Bootstrap(t *testing.T) {
+	path := acctest.RandomWithPrefix("tf-test-consul")
+	resourcePath := "vault_consul_secret_backend.test"
+	bootstrapPath := "vault_consul_secret_backend_role.test"
```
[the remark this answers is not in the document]
robmonte: Agreed, changed.

.github/workflows/build.yml
```diff
@@ -82,6 +82,11 @@ jobs:
           --health-interval 1s
           --health-timeout 5s
           --health-retries 60
+      consul:
+        image: docker.mirror.hashicorp.services/consul:latest
+        ports:
+        - 8500:8500
+        - 8600:8600
```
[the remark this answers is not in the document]
robmonte: I added what I think is the right way to do it based on the way the Vault container's health check is written - could you please double-check my logic?

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +27,123 @@ func TestConsulSecretBackend(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourcePath, "ca_cert", "FAKE-CERT-MATERIAL"),
+					resource.TestCheckResourceAttr(resourcePath, "client_cert", "UPDATED-FAKE-CLIENT-CERT-MATERIAL"),
+					resource.TestCheckResourceAttr(resourcePath, "client_key", "UPDATED-FAKE-CLIENT-CERT-KEY-MATERIAL"),
 				),
 			},
 		},
 	})
 }
 
+func TestConsulSecretBackend_Bootstrap(t *testing.T) {
```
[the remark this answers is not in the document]
robmonte: Do any other tests do something similar that I can reference? I'm not quite sure how to go about accomplishing this.

.github/workflows/build.yml
```diff
@@ -82,6 +82,18 @@ jobs:
...
           --health-retries 60
+      consul:
+        image: docker.mirror.hashicorp.services/consul:latest
+        env:
+          CONSUL_LOCAL_CONFIG: "{\"acl\":{\"enabled\":true}}"
+        ports:
+        - 8500:8500
+        - 8600:8600
+        options: >-
+          --health-cmd "curl 'http://127.0.0.1:8500/v1/health/node/server-1'"
```
benashz, line 93: Think you want `curl -f` or `curl --fail` here since we want curl exit with a non-zero status whenever the response returns an error.

````suggestion
          --health-cmd "curl -f 'http://127.0.0.1:8500/v1/health/node/server-1'"
````

File: .github/workflows/build.yml [hunk 8]
```diff
@@ -82,18 +82,6 @@ jobs:
           --health-interval 1s
           --health-timeout 5s
           --health-retries 60
-      consul:
```
benashz: Probably want to revert this since the build is running in a docker container.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +30,132 @@ func TestConsulSecretBackend(t *testing.T) {
...
+	backend := acctest.RandomWithPrefix("tf-test-backend")
+	resourceName := "vault_consul_secret_backend.test"
+	resourceRoleName := "vault_consul_secret_backend_role.test"
+
+	if testutil.CheckTestVaultVersion(t, "1.11") {
+		var consulAddr string
+		if envAddr := os.Getenv("CONSUL_ADDR"); envAddr != "" {
```
benashz, line 121: I think we can drop this conditional in favor of https://github.com/hashicorp/vault/blob/ca804e5596e87b732eb7321b3d624d1883b29c5f/helper/testhelpers/consul/consulhelper.go#L33

We will probably just need to set the `CONSUL_HTTP_ADDR` env var instead.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +30,132 @@ func TestConsulSecretBackend(t *testing.T) {
...
+			consulAddr = envAddr
+		} else {
+			cleanup, consulConfig := consulhelper.PrepareTestContainer(t, "latest", false, false)
```
benashz, line 124: I prefer what you had before, where you were pinning the Consul version. You can allow for an override from an env var with the default being the version we are testing against.

Pinning the version provides more deterministic test results.

benashz (commented):
Looking good! Just a few suggestions.

vault/resource_consul_secret_backend_test.go
robmonte: Makes sense, I'll revert back to 1.12.3

.github/workflows/build.yml
robmonte: Thanks for explaining, fixed

vault/resource_consul_secret_backend_test.go
robmonte: Oh nice catch! That's a neat, cleaner way to do it

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +29,127 @@ func TestConsulSecretBackend(t *testing.T) {
...
 		},
 	})
 }
 
+func TestConsulSecretBackend_Bootstrap(t *testing.T) {
+	backend := acctest.RandomWithPrefix("tf-test-backend")
+	resourceName := "vault_consul_secret_backend.test"
+	resourceRoleName := "vault_consul_secret_backend_role.test"
+
+	if testutil.CheckTestVaultVersion(t, "1.11") {
```
benashz, line 118: Ideally we would call `t.Skip("Requires Vault version >=  1.11)"` if ! vault >= 1.11. Then you can drop the whole `if{}` statement.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,87 +29,127 @@ func TestConsulSecretBackend(t *testing.T) {
...
+			PreCheck:     func() { testutil.TestAccPreCheck(t) },
+			CheckDestroy: testAccConsulSecretBackendCheckDestroy,
+			Steps: []resource.TestStep{
+				{
+					Config: testConsulSecretBackend_bootstrapConfig(backend, consulAddr),
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(resourceName, consts.FieldPath, backend),
+						resource.TestCheckResourceAttr(resourceName, "address", consulAddr),
+					),
+				},
```
benashz, line 134: It would be good to interleave import test steps between each config test. You can use https://github.com/hashicorp/terraform-provider-vault/blob/9df0538f67bb3b26d9b44d7720893bc146a06afe/testutil/testutil.go#L656 to get the import step.

vault/import_consul_secret_backend_test.go
```diff
@@ -11,7 +11,9 @@ import (
 
 func TestAccConsulSecretBackend_import(t *testing.T) {
 	path := acctest.RandomWithPrefix("tf-test-consul")
+	resourceName := "vault_consul_secret_backend.test"
 	token := "aea34d3f-17e4-4387-801c-5e41be678e46"
+
```
benashz, line 16: This test is sort of bogus, we could probably drop it all together after we add `import ` test steps to the other `ConsulSecretBackend*` tests.

website/docs/r/consul_secret_backend.html.md
```diff
@@ -38,7 +38,7 @@ The following arguments are supported:
   The `namespace` is always relative to the provider's configured [namespace](../index.html#namespace).
    *Available only for Vault Enterprise*.
 
-* `token` - (Required) The Consul management token this backend should use to issue new tokens.
+* `token` - (Optional) The Consul management token this backend should use to issue new tokens. If unspecified, Vault will attempt to perform an ACL Bootstrap automatically on the Consul node.
```
benashz, line 41: This is still required with Vault-1.10 and below right?

If you could break on the 100 column mark that would great as well.

benashz (approved):
A few more nits/suggestions then :+1:

vault/resource_consul_secret_backend_test.go
robmonte: @benashz It looks like Consul doesn't like this naming format. Is it okay to use 127.0.0.1 in this case?
CI error here: https://github.com/hashicorp/terraform-provider-vault/runs/7660137651?check_suite_focus=true#step:4:692

website/docs/r/consul_secret_backend.html.md
robmonte: That is right, I shouldn't have forgotten that.

vault/resource_consul_secret_backend_test.go
benashz: I think the issue is that `CONSUL_HTTP_ADDR` includes the scheme, if you remove that then it should work fine.

benashz: It looks like the scheme should be included: https://www.consul.io/commands#consul_http_addr

robmonte: 🤦 Thank you for pointing it out

robmonte:
This last commit is going to fail the final import test.

When `TestConsulSecretBackend_Bootstrap` attempts the second config write with Consul while it is already bootstrapped and fails, it successfully catches the error but the resourceName is still  changing to the new resourceName. Not sure how to prevent this yet.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,83 +29,128 @@ func TestConsulSecretBackend(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceRoleName, "consul_policies.#", "1"),
+					resource.TestCheckTypeSetElemAttr(resourceRoleName, "consul_policies.*", "global-management"),
 				),
 			},
+			testutil.GetImportTestStep(resourceName, false),
+			{
+				Config:      testConsulSecretBackend_bootstrapConfig(backend+"-new", consulAddr),
+				ExpectError: regexp.MustCompile(`Token not provided and failed to bootstrap ACLs`),
+			},
+			testutil.GetImportTestStep(resourceName, false),
```
benashz, line 153: Suggest moving this to come after the first config test step.

benashz (approved):
Minor suggestion. Should be good to go after that!

vault/resource_consul_secret_backend_test.go
robmonte: I actually have it here twice. This one and another a couple lines up after the checkBootstrapConfig, which is after the first config is made. Are you saying it should specifically be checked before the role is made in step 2?

Second question - when attempting to write the resource config the second time, it fails because Consul is already bootstrapped which is what the test wants. But the import afterwards is looking for that new resource path that didn't succeed. I would have guessed the error should stop Terraform from changing the state to the new path. Is this expected behavior though?

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,83 +29,128 @@ func TestConsulSecretBackend(t *testing.T) {
...
+		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy: testAccConsulSecretBackendCheckDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testConsulSecretBackend_bootstrapConfig(backend, consulAddr),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, backend),
+					resource.TestCheckResourceAttr(resourceName, "address", consulAddr),
+				),
+			},
```
benashz, line 138: suggest testing import after each config test step:
```suggestion
			},
			testutil.GetImportTestStep(resourceName, false),

File: vault/resource_consul_secret_backend_test.go (hunk 15 above)
Comment (benashz) [inline vault/resource_consul_secret_backend_test.go:153]: Probably no need for the extra import test here. You could perhaps move the expected failure step to be between the two valid test steps above. That would mimic more what a user would experience, I believe. 
```suggestion
```

benashz: > I actually have it here twice. This one and another a couple lines up after the checkBootstrapConfig, which is after the first config is made. Are you saying it should specifically be checked before the role is made in step 2?

Essentially yes, see suggestion above.

> 
> Second question - when attempting to write the resource config the second time, it fails because Consul is already bootstrapped which is what the test wants. But the import afterwards is looking for that new resource path that didn't succeed. I would have guessed the error should stop Terraform from changing the state to the new path. Is this expected behavior though?

Yes. since the "new" path was never  created the import test step should fail since the Vault is not provisioned with that path.

vault/resource_consul_secret_backend_test.go
```diff
@@ -24,83 +29,128 @@ func TestConsulSecretBackend(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDescription, "test description"),
+					resource.TestCheckResourceAttr(resourceName, "default_lease_ttl_seconds", "3600"),
+					resource.TestCheckResourceAttr(resourceName, "max_lease_ttl_seconds", "86400"),
+					resource.TestCheckResourceAttr(resourceName, "address", "127.0.0.1:8500"),
+					resource.TestCheckResourceAttr(resourceName, "token", token),
+					resource.TestCheckResourceAttr(resourceName, "scheme", "http"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldLocal, "false"),
+					resource.TestCheckNoResourceAttr(resourceName, "ca_cert"),
+					resource.TestCheckNoResourceAttr(resourceName, "client_cert"),
+					resource.TestCheckNoResourceAttr(resourceName, "client_key"),
```
benashz, line 42: Probably want to add some import test steps in this test, since we dropped the dedicated import test case.

benashz (approved):
LGTM!

PR #1567, opened 2022-08-05
fix identityGroupExists

### LINKED ISSUE #1525: vault_identity_group resource fails to recreate group if missing from Vault
### Terraform Version
1.2.4

### Affected Resource(s)
- vault_identity_group

```hcl
resource "vault_identity_group" "this" {
  name     = "test"
  type     = "internal"
  policies = ["this"]

  member_entity_ids = null
}
```

### Debug Output
Full gist example: [https://gist.github.com/czembower/57288bdc796fca9e301abb50ac48ba05](url)
```
Error: error checking if IdentityGroup "this" exists: entity not found: "/identity/group/id/a86ebf5a-56c1-ba33-a926-20675ab65ab6"
with vault_identity_group.this
on vault_oidc.tf line 5, in resource "vault_identity_group" "this":
resource "vault_identity_group" "this" {
```

### Expected Behavior
Provider should create the missing resource

### Actual Behavior
Provider errors, does not create missing resource. After removing the resource from the state file, the plan/apply succeeds. 

### Steps to Reproduce
1. Create vault_identity_group via terraform
2. Delete vault identity group via some means other than terraform
3. `terraform plan`

czembower:
### Community Note

Closes #1525 

```
fixed condition where missing/deleted entity group caused Terraform to error instead of recreate the resource
```

### DIFF
--- a/vault/resource_identity_group.go
+++ b/vault/resource_identity_group.go
@@ -291,16 +291,18 @@ func identityGroupExists(d *schema.ResourceData, meta interface{}) (bool, error)
 	}
 
 	id := d.Id()
+
+	path := identityGroupIDPath(id)
 	key := id
 
+	// use the name if no ID is set
 	if len(id) == 0 {
-		return false, nil
-	} else {
 		key = d.Get("name").(string)
+		path = identityGroupNamePath(key)
 	}
 
 	log.Printf("[DEBUG] Checking if IdentityGroup %q exists", key)
-	resp, err := readIdentityGroup(client, id, true)
+	resp, err := client.Logical().Read(path)
 	if err != nil {
 		return true, fmt.Errorf("error checking if IdentityGroup %q exists: %s", key, err)
 	}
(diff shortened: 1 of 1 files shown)


vault/resource_identity_group.go
```diff
@@ -291,16 +291,18 @@ func identityGroupExists(d *schema.ResourceData, meta interface{}) (bool, error)
...
 	if len(id) == 0 {
-		return false, nil
-	} else {
 		key = d.Get("name").(string)
+		path = identityGroupNamePath(key)
 	}
 
 	log.Printf("[DEBUG] Checking if IdentityGroup %q exists", key)
-	resp, err := readIdentityGroup(client, id, true)
+	resp, err := client.Logical().Read(path)
```
benashz, line 305: This change will break the client controlled consistency (CCC) bits. We may need to take a closer look at the issue.

benashz: Since we actually do no longer are supporting `schema.Resource.Exists`, we can just call `d.SetId("")` in `identityGroupRead()` when `resp` is nil.

czembower: Done! Thanks @benashz

vinay-gopalan (approved):
Looks good to me and seems to resolve the underlying issue. Thank you for your contribution to HashiCorp! :)

PR #1568, opened 2022-08-05
Update jira sync

benashz:


swenson (approved):
LGTM, but is this part of foundations too?

benashz:
> LGTM, but is this part of foundations too?

Yes for now. We are relying on it to populate the TFVP  bucket of work.

PR #1570, opened 2022-08-07
fix: remove unnecessary nesting of secret data for KV-V1 secrets

### LINKED ISSUE #1549: Using terraform to create vault_kv_secret resources results in json_data stored in a single key
Hello,

I believe we've discovered a bug in how `vault_kv_secret` has been implemented in that it stores the entire `data_json` in a single key named `data`.  As indicated in the [referenced discussion below](https://discuss.hashicorp.com/t/using-terraform-to-create-vault-kv-secret-resources-results-in-json-data-stored-in-a-single-key/42360/2?u=kris-runzer), it appears to be treating KV1 as KV2.

### Terraform Version
Terraform v1.1.8 and higher (multiple workspaces).

### Affected Resource(s)
- vault_kv_secret

```hcl
resource "vault_mount" "hello" {
  path        = "hello"
  type        = "kv"
  options     = { version = "1" }
  description = "hello"
}

resource "vault_kv_secret" "hi" {
  path = "${vault_mount.hello.path}/redis"
  data_json = jsonencode(
    {
      endpoint   = "127.0.0.1"
      ttl        = "60m"
    }
  )
}
```

### Debug Output
n/a

### Panic Output
n/a

### Expected Behavior
We would expect to have the following at /hello/redis, as JSON:
```json
{
  "endpoint" = "127.0.0.1",
  "ttl" = "60m"
}
```

### Actual Behavior
```
Instead, we end up with everything nested under a single key, named data:

{
  "data": {
    "endpoint" = "127.0.0.1",
    "ttl" = "60m"
  }
}
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
We’re using version `3.7.0` of the Vault provider against an enterprise Vault cluster running version `1.10.3+ent`.

### References
You can see the original discussion here:  https://discuss.hashicorp.com/t/using-terraform-to-create-vault-kv-secret-resources-results-in-json-data-stored-in-a-single-key/42360/1.

l-with:
### Community Note

Closes #1549

```release-note
fix #1549
```

```
$ make testacc TESTARGS='-run=TestAccKVSecret'
TF_ACC=1 go test -v -run=TestAccKVSecret -timeout 30m ./...
PASS
ok  	github.com/hashicorp/terraform-provider-vault/codegen	0.277s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/decode	0.276s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/encode	1.362s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/alphabet	0.926s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/role	0.483s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template	1.155s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/generated/resources/transform/transformation	0.699s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/internal/identity/entity	0.341s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/internal/provider	1.429s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/testutil	1.335s [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-vault/util	1.247s [no tests to run]
2022/08/07 09:57:30 [INFO] Using Vault token with the following policies: root
=== RUN   TestAccKVSecretBackendV2
--- PASS: TestAccKVSecretBackendV2 (1.86s)
=== RUN   TestAccKVSecret
--- PASS: TestAccKVSecret (1.71s)
=== RUN   TestAccKVSecretV2
--- PASS: TestAccKVSecretV2 (1.11s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	5.369s
```

### DIFF
--- a/vault/resource_kv_secret_test.go
+++ b/vault/resource_kv_secret_test.go
@@ -123,7 +123,6 @@ func testResourceKVSecret_apiAcessCheck(s *terraform.State, want string) error {
 	}
 
 	return nil
-
 }
 
 func testResourceKVSecret_basic_apiAcessCheck(s *terraform.State) error {
--- a/vault/resource_kv_secret_test.go
+++ b/vault/resource_kv_secret_test.go
@@ -28,7 +28,7 @@ func TestAccKVSecret(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zap"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
-					testResourceKVSecret_basic_apiAcessCheck,
+					testResourceKVSecret_apiAcessCheck("zap"),
 				),
 			},
 			{
@@ -37,7 +37,7 @@ func TestAccKVSecret(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zoop"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
-					testResourceKVSecret_updated_apiAcessCheck,
+					testResourceKVSecret_apiAcessCheck("zoop"),
 				),
 			},
 			{
@@ -102,33 +102,27 @@ resource "vault_kv_secret" "test" {
 	return ret
 }
 
-func testResourceKVSecret_apiAcessCheck(s *terraform.State, want string) error {
-	resourceState := s.Modules[0].Resources["vault_kv_secret.test"]
-	state := resourceState.Primary
+func testResourceKVSecret_apiAcessCheck(want string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		resourceState := s.Modules[0].Resources["vault_kv_secret.test"]
+		state := resourceState.Primary
 
-	path := state.ID
+		path := state.ID
 
-	client, err := provider.GetClient(state, testProvider.Meta())
-	if err != nil {
-		return err
+		client, err := provider.GetClient(state, testProvider.Meta())
+		if err != nil {
+			return err
+		}
+
+		secret, err := client.Logical().Read(path)
+		if err != nil {
+			return fmt.Errorf("error reading back secret: %s", err)
+		}
+
+		if got := secret.Data["zip"]; got != want {
+			return fmt.Errorf("'zip' data is %q; want %q", got, want)
+		}
+
+		return nil
 	}
-
-	secret, err := client.Logical().Read(path)
-	if err != nil {
-		return fmt.Errorf("error reading back secret: %s", err)
-	}
-
-	if got := secret.Data["zip"]; got != want {
-		return fmt.Errorf("'zip' data is %q; want %q", got, want)
-	}
-
-	return nil
-}
-
-func testResourceKVSecret_basic_apiAcessCheck(s *terraform.State) error {
-	return testResourceKVSecret_apiAcessCheck(s, "zap")
-}
-
-func testResourceKVSecret_updated_apiAcessCheck(s *terraform.State) error {
-	return testResourceKVSecret_apiAcessCheck(s, "zoop")
 }
--- a/vault/resource_kv_secret_test.go
+++ b/vault/resource_kv_secret_test.go
@@ -2,6 +2,8 @@ package vault
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
@@ -27,6 +29,10 @@ func TestAccKVSecret(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zap"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
 				),
+			},
+			{
+				Config: testKVSecretConfigUpdate(mount, name),
+				Check:  testResourceKVSecret_updateCheck,
 			},
 			{
 				ResourceName:            resourceName,
@@ -68,4 +74,47 @@ resource "vault_kv_secret" "test" {
 }`, name)
 
 	return ret
+}
+
+func testKVSecretConfigUpdate(mount, name string) string {
+	ret := fmt.Sprintf(`
+%s
+
+`, kvV1MountConfig(mount))
+
+	ret += fmt.Sprintf(`
+resource "vault_kv_secret" "test" {
+  path = "${vault_mount.kvv1.path}/%s"
+  data_json = jsonencode(
+    {
+      zip = "zoop",
+      foo = "bar"
+    }
+  )
+}`, name)
+
+	return ret
+}
+
+func testResourceKVSecret_updateCheck(s *terraform.State) error {
+	resourceState := s.Modules[0].Resources["vault_kv_secret.test"]
+	state := resourceState.Primary
+
+	path := state.ID
+
+	client, err := provider.GetClient(state, testProvider.Meta())
+	if err != nil {
+		return err
+	}
+
+	secret, err := client.Logical().Read(path)
+	if err != nil {
+		return fmt.Errorf("error reading back secret: %s", err)
+	}
+
+	if got, want := secret.Data["zip"], "zoop"; got != want {
+		return fmt.Errorf("'zip' data is %q; want %q", got, want)
+	}
+
+	return nil
 }
--- a/vault/resource_kv_secret_test.go
+++ b/vault/resource_kv_secret_test.go
@@ -28,7 +28,7 @@ func TestAccKVSecret(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zap"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
-					testResourceKVSecret_apiAcessCheck,
+					testResourceKVSecret_basic_apiAcessCheck,
 				),
 			},
 			{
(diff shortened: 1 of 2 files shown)


vault/resource_kv_secret_test.go
```diff
@@ -69,3 +75,46 @@ resource "vault_kv_secret" "test" {
 
 	return ret
 }
+
+func testKVSecretConfigUpdate(mount, name string) string {
```
vinay-gopalan, line 79: We can rename this to `testKVSecretConfig_updated` and also rename  `testKVSecretConfig` -> `testKVSecretConfig_basic` on L59; there is precedent in the codebase for naming config functions in that style :)

vault/resource_kv_secret_test.go
```diff
@@ -69,3 +75,46 @@ resource "vault_kv_secret" "test" {
...
+      zip = "zoop",
+      foo = "bar"
+    }
+  )
+}`, name)
+
+	return ret
+}
+
+func testResourceKVSecret_updateCheck(s *terraform.State) error {
```
vinay-gopalan, line 99: We prefer using the `resource.ComposeTestCheckFunc` test code style whenever possible, as done in L27; something like 

```
resource.ComposeTestCheckFunc(
    resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
    resource.TestCheckResourceAttr(resourceName, "data.zip", "zoop"),					 
    resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
),
```

vinay-gopalan (changes requested):
Thanks for the PR! The code's looking great, had a couple of comments :)

vault/resource_kv_secret_test.go
l-with: I committed the renaming.

vault/resource_kv_secret_test.go
l-with: With the suggested implementation of the ComposeTestCheckFunc the test with the previous implementation in resource_kv_secret.go will not fail, because the extra "data" in the path is implemented in kvSecretWrite as well as in kvSecretRead.

This is why I made the test fail with the first commit and fixed the implementation with the second commit. I had to use the api client directly to make the test fail, because the read context compensated the failure in the create context. It is somehow that the terraform provider is not a tool for itself but for configuring the application to be used (here for retrieving values by api or cli).

vinay-gopalan: Since we have the Import test on L38, we are already take care of checking the TF config values against what is configured in Vault. The composition of test funcs would also make comparing the two maps simpler, as opposed to only checking a single key. I'd still suggest removing this extra check func and adding the precedent style of `ComposeTestCheckFunc`  :)

l-with: I refactored the update test to composition.

I added an extra test step for the api access check.

I think that the import test also uses the state read by the provider. Thus the test does not fail. You can verify this by checking out https://github.com/l-with/terraform-provider-vault/tree/does-kv-v1-test-fail and run the test. It will not fail. It will fail with the added api access check function (https://github.com/l-with/terraform-provider-vault/blob/kv-v1-test-fail/vault/resource_kv_secret_test.go).

benashz: > I refactored the update test to composition.

Thanks!

> 
> I added an extra test step for the api access check.
> 
> I think that the import test also uses the state read by the provider. Thus the test does not fail. You can verify this by checking out https://github.com/l-with/terraform-provider-vault/tree/does-kv-v1-test-fail and run the test. It will not fail. It will fail with the added api access check function (https://github.com/l-with/terraform-provider-vault/blob/kv-v1-test-fail/vault/resource_kv_secret_test.go).

The import does a complete read of the resource from Vault into a new state, and verifies that there are no differences between the TF code and that state. So I think we are covered here without the extra direct API check.

l-with: I have tried it out, only the direct API check fails with the implementation of the current release that nests the kv v1 secrets in extra path component "data". Also the test in the current implementation does not fail, the added update test makes no difference. 

Surely this is a rare case, but since the the current release has the issue, I think it is reasonable that the fix contains the test that reveals the issue together with the fix of the issue.

benashz: I think I agree with you, we can keep the regression check in this case.

vault/resource_kv_secret_test.go
```diff
@@ -69,3 +83,46 @@ resource "vault_kv_secret" "test" {
...
+      zip = "zoop",
+      foo = "bar"
+    }
+  )
+}`, name)
+
+	return ret
+}
+
+func testResourceKVSecret_apiAcessCheck(s *terraform.State) error {
```
benashz, line 107: Please adapt this check to follow the pattern set in: https://github.com/hashicorp/terraform-provider-vault/blob/c520061319a277600ea367adacdcf7e17d2820ee/vault/resource_identity_entity_test.go#L143

vault/resource_kv_secret_test.go
```diff
@@ -21,13 +23,25 @@ func TestAccKVSecret(t *testing.T) {
...
-				Config: testKVSecretConfig(mount, name),
+				Config: testKVSecretConfig_basic(mount, name),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zap"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
 				),
 			},
+			{
+				Config: testKVSecretConfig_basic(mount, name),
```
benashz, line 34: Should not be needed, you can include the extra API check function in the step above.

vault/resource_kv_secret_test.go
```diff
@@ -21,13 +23,25 @@ func TestAccKVSecret(t *testing.T) {
...
+				Check:  testResourceKVSecret_apiAcessCheck,
+			},
+			{
+				Config: testKVSecretConfig_updated(mount, name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
+					resource.TestCheckResourceAttr(resourceName, "data.zip", "zoop"),
+					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
```
benashz, line 42: Add extra API check function.

l-with: I hope, I understand this correctly: I also added an extra API check test to the updated test.

benashz: yes, that is correct.

vault/resource_kv_secret_test.go
l-with: nice pattern!

l-with: done with
[refactor api access check to functional pattern](https://github.com/hashicorp/terraform-provider-vault/pull/1570/commits/e23506b34a074546367933bec992886108f4e196)

vault/resource_kv_secret_test.go
l-with: done with [add extra api check to update test](https://github.com/hashicorp/terraform-provider-vault/pull/1570/commits/4347da9e0ac956801a41c0fc82a890042fe58041)

vault/resource_kv_secret_test.go
l-with: done with [move extra api check to basic test](https://github.com/hashicorp/terraform-provider-vault/pull/1570/commits/13b2fee3fa69c11b6e553bdaf2b33070710c9a9c)

l-with:
I take the liberty to ask if there is any requested change still open?

vault/resource_kv_secret_test.go
```diff
@@ -69,3 +81,48 @@ resource "vault_kv_secret" "test" {
...
+      zip = "zoop",
+      foo = "bar"
+    }
+  )
+}`, name)
+
+	return ret
+}
+
+func testResourceKVSecret_apiAcessCheck(want string) resource.TestCheckFunc {
```
benashz, line 105:
```suggestion
func testResourceKVSecret_apiAccessCheck(want string) resource.TestCheckFunc {
```

vault/resource_kv_secret_test.go
```diff
@@ -21,11 +23,21 @@ func TestAccKVSecret(t *testing.T) {
 		PreCheck:  func() { testutil.TestAccPreCheck(t) },
 		Steps: []resource.TestStep{
 			{
-				Config: testKVSecretConfig(mount, name),
+				Config: testKVSecretConfig_basic(mount, name),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
 					resource.TestCheckResourceAttr(resourceName, "data.zip", "zap"),
 					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
+					testResourceKVSecret_apiAcessCheck("zap"),
```
benashz, line 31:
```suggestion
					testResourceKVSecret_apiAccessCheck("zap"),
```

vault/resource_kv_secret_test.go
```diff
@@ -21,11 +23,21 @@ func TestAccKVSecret(t *testing.T) {
...
+				),
+			},
+			{
+				Config: testKVSecretConfig_updated(mount, name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/%s", mount, name)),
+					resource.TestCheckResourceAttr(resourceName, "data.zip", "zoop"),
+					resource.TestCheckResourceAttr(resourceName, "data.foo", "bar"),
+					testResourceKVSecret_apiAcessCheck("zoop"),
```
benashz, line 40:
```suggestion
					testResourceKVSecret_apiAccessCheck("zoop"),
```

benashz (approved):
Thank your for your contribution to HashiCorp!

vinay-gopalan (approved):

PR #1579, opened 2022-08-11
add kv-v2 write retry

### LINKED ISSUE #677: vault_generic_secret type=kv-v2 first write fails status 400 Upgrading from non-versioned to versioned data. "This backend will be unavailable for a brief period and will resume service shortly."
# Terraform Version
v0.12.20
### Affected Resource(s)
- vault_generic_secret

```hcl
variable "users" {
  description = "list of users to set up"
  type        = map(string)
  default = {
    "user"  = "vault-user"
    "admin" = "vault-admin"
  }
}

resource "vault_mount" "secret" {
  path        = "mysecrets"
  type        = "kv-v2"
  description = "store key/value secret data"
}

resource "random_password" "user_pwd" {
  for_each    = var.users
  length      = 16
  special     = false
  min_upper   = 1
  min_lower   = 1
  min_numeric = 1
}

resource "vault_generic_secret" "logins" {
  for_each   = var.users
  path       = "mysecrets/realm/users/${each.key}"
  data_json  = <<-EOT
    {
	"username":   "${each.key}",
	"initialpwd": "random_password.user_pwd[each.key].result",
	"memberof":   "${each.value}"
    }
    EOT
  depends_on = [vault_mount.secret]
}
```

### Debug Output
```
  vault_generic_secret.logins["vault-user"]: Creating...
  vault_generic_secret.logins["vault-admin"]: Creating...
  vault_generic_secret.logins["vault-admin"]: Creation complete after 1s [id=mysecrets/realm/users/vault-admin]

 Error: error writing to Vault: Error making API request.
 URL: PUT https://vault.gitlab-ci.mydomain.net:8200/v1/mysecrets/data/realm/users/vault-user
 Code: 400. Errors:
 * Upgrading from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service shortly.
   on vault_config/terraform/users.tf line 53, in resource "vault_generic_secret" "logins":
   53: resource "vault_generic_secret" "logins" {
```

### Expected Behavior
First insert to kv-v2 should have succeeded along with the following ones.

### Actual Behavior
First insert failed with status 400, second insert succeeded. This does not occur on every run. I'd speculate it depends on the speed at which the vault server is running.

### Steps to Reproduce
1. create spec files that enable a new kv-v2 endpoint then insert  secrets into it.
2. `terraform apply`

### Discussion
Its not clear why vault engages in a "upgrade from non-versioned to versioned data" when the endpoint has just been enabled with type=kv-v2. You'd think it would be set up as a versioned data endpoint from the get-go. Given that it does engage in this behavior, terraform-provider-vault should internally retry the secret write when this error is returned from vault.

czembower:
An error condition can occur if attempting to create a kv-v2 mount and write secrets to it in the same terraform run, resulting in the error:

```Upgrading from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service shortly```

It appears that this error is common on Vault running in a cluster, and less so when standalone. 

This PR adds retry with backoff on writes to vault_kv_secret_v2 as well as vault_generic_secret resources. Followed precedent set for retry in namespaceDelete().

Closes #677 

```
Fixed failed writes to newly-created kv-v2 mount by adding retry/backoff
```

```
TESTARGS="--run TestAccKVSecretV2 --run TestResourceGenericSecret" make testacc-ent
make testacc TF_ACC_ENTERPRISE=1
TF_ACC=1 go test -v --run TestAccKVSecretV2 --run TestResourceGenericSecret -timeout 30m ./...

2022/08/11 11:24:20 [INFO] Using Vault token with the following policies: root
--- PASS: TestResourceGenericSecret (2.41s)
=== RUN   TestResourceGenericSecretNS
--- PASS: TestResourceGenericSecretNS (3.14s)
--- PASS: TestResourceGenericSecret_deleted (1.99s)
=== RUN   TestResourceGenericSecret_deleteAllVersions
--- PASS: TestResourceGenericSecret_deleteAllVersions (2.03s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	9.999s
```

### Community Note

### DIFF
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
@@ -146,9 +148,15 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error 
 	}
 
 	log.Printf("[DEBUG] Writing generic Vault secret to %s", path)
-	_, err = client.Logical().Write(path, data)
-	if err != nil {
-		return fmt.Errorf("error writing to Vault: %s", err)
+	if err := backoff.RetryNotify(func() error {
+		_, err := client.Logical().Write(path, data)
+		return err
+	}, backoff.WithMaxRetries(
+		backoff.NewConstantBackOff(time.Millisecond*500), 10),
+		func(err error, duration time.Duration) {
+			log.Printf("[WARN] create generic secret %q failed, retrying in %s", path, duration)
+		}); err != nil {
+		return fmt.Errorf("error creating generic secret: %s", err)
 	}
 
 	d.SetId(originalPath)
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -147,15 +147,19 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error 
 
 	}
 
-	log.Printf("[DEBUG] Writing generic Vault secret to %s", path)
-	if err := backoff.RetryNotify(func() error {
-		_, err := client.Logical().Write(path, data)
-		return err
-	}, backoff.WithMaxRetries(
-		backoff.NewConstantBackOff(time.Millisecond*500), 10),
-		func(err error, duration time.Duration) {
-			log.Printf("[WARN] create generic secret %q failed, retrying in %s", path, duration)
-		}); err != nil {
+	writeData := func() error {
+		if _, err := client.Logical().Write(path, data); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	bo := backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Millisecond*500), 10)
+
+	log.Printf("[DEBUG] Writing generic Vault secret to  %s", path)
+	if err := backoff.RetryNotify(writeData, bo, func(err error, duration time.Duration) {
+		log.Printf("[WARN] create generic secret %q failed, retrying in %s", path, duration)
+	}); err != nil {
 		return fmt.Errorf("error creating generic secret: %s", err)
 	}
 
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -4,10 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/vault/api"
 
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 	"github.com/hashicorp/terraform-provider-vault/internal/provider"
@@ -149,7 +151,11 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error 
 
 	writeData := func() error {
 		if _, err := client.Logical().Write(path, data); err != nil {
-			return err
+			if respErr, ok := err.(*api.ResponseError); ok && (respErr.StatusCode == http.StatusBadRequest) {
+				return err
+			} else {
+				return backoff.Permanent(err)
+			}
 		}
 		return nil
 	}
--- a/vault/resource_kv_secret_v2.go
+++ b/vault/resource_kv_secret_v2.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: vault/resource_kv_secret_v2.go [hunk 1]
```diff
@@ -134,8 +136,20 @@ func kvSecretV2Write(ctx context.Context, d *schema.ResourceData, meta interface
 		data[k] = d.Get(k)
 	}
 
-	if _, err := client.Logical().Write(path, data); err != nil {
-		return diag.Errorf("error writing secret data to %s, err=%s", path, err)
+	writeData := func() error {
```
benashz, line 139: I don't believe we need this for the dedicated kvv2 resource. We shouldn't  encounter the migration issue going from kv1 to kv2.

vault/resource_generic_secret.go
```diff
@@ -145,10 +147,20 @@ func genericSecretResourceWrite(d *schema.ResourceData, meta interface{}) error
 
 	}
 
-	log.Printf("[DEBUG] Writing generic Vault secret to %s", path)
-	_, err = client.Logical().Write(path, data)
-	if err != nil {
-		return fmt.Errorf("error writing to Vault: %s", err)
+	writeData := func() error {
+		if _, err := client.Logical().Write(path, data); err != nil {
+			return err
```
benashz, line 152: This will retry on any error, which may cause unnecessary delays or perhaps introduce undesirable side-effects. It would be best to only retry on certain errors.

See https://github.com/hashicorp/terraform-provider-vault/blob/eae35c228b08cdd6636864535efcf2c5e2778d0e/vault/resource_namespace.go#L84 for a proper example of this.

benashz (commented):
Thanks for the PR! I added some initial comments/suggestions, also It be good to include some tests for this case.

vault/resource_kv_secret_v2.go
czembower: Thanks for taking a look, appreciate the feedback! I confirmed that the condition does indeed occur for both generic_secret and kv_secret_v2. In vault core, it looks like a kv_v2 mount is created as v1, then tuned to v2 and reloaded via reloadBackendCommon(), which returns without waiting for the upgrade to complete. I assume this is by design, so better to just retry on the terraform provider side...

vault/resource_generic_secret.go
czembower: Thanks! Added this error validation.

bfiics (approved):
It makes the code better for sure. Good luck

vault/resource_kv_secret_v2.go
benashz: But in the case of new kv_v2 mount there would no data to migrate, no?

vault/resource_kv_secret_v2.go
```diff
@@ -134,8 +138,24 @@ func kvSecretV2Write(ctx context.Context, d *schema.ResourceData, meta interface
...
+	}
+
+	bo := backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Millisecond*500), 10)
+
+	log.Printf("[DEBUG] creating KVV2 secret %s", path)
+	if err := backoff.RetryNotify(writeData, bo, func(err error, duration time.Duration) {
+		log.Printf("[WARN] create KVV2 %q failed, retrying in %s", path, duration)
+	}); err != nil {
+		return diag.Errorf("error creating KVV2 secret: %s", err)
 	}
```
benashz, line 159: I would prefer to remove the duplicated retry logic here since this resource will only ever support kv-v2. If this becomes an issue in the future we can revisit it then.

```suggestion
	if _, err := client.Logical().Write(path, data); err != nil {
		return diag.Errorf("error writing secret data to %s, err=%s", path, err)
```

vault/resource_kv_secret_v2.go
```diff
@@ -5,9 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
+	"time"
```
benashz, line 9:
```suggestion
```

vault/resource_kv_secret_v2.go
```diff
@@ -134,8 +138,24 @@ func kvSecretV2Write(ctx context.Context, d *schema.ResourceData, meta interface
...
+		return nil
+	}
+
+	bo := backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Millisecond*500), 10)
+
+	log.Printf("[DEBUG] creating KVV2 secret %s", path)
+	if err := backoff.RetryNotify(writeData, bo, func(err error, duration time.Duration) {
+		log.Printf("[WARN] create KVV2 %q failed, retrying in %s", path, duration)
+	}); err != nil {
+		return diag.Errorf("error creating KVV2 secret: %s", err)
```
benashz:
```suggestion
	if _, err := client.Logical().Write(path, data); err != nil {
		return diag.Errorf("error writing secret data to %s, err=%s", path, err)
```

benashz:
```suggestion
	if _, err := client.Logical().Write(path, data); err != nil {
		return diag.Errorf("error writing secret data to %s, err=%s", path, err)
```

vault/resource_kv_secret_v2.go
```diff
@@ -5,9 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
```
benashz, line 11:
```suggestion
```

vault/resource_kv_secret_v2.go
```diff
@@ -5,9 +5,13 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/vault/api"
```
benashz, line 14:
```suggestion
```

benashz (approved):
LGTM!

czembower:
@benashz it looks like you reverted this PR's changes to kv_secret_v2. Could you explain this? Several people have informed me that this is still broken for that resource, and looking at the final merge, that is clearly the reason. The retry/back-off is necessary here.

eldengates:
any updates on this ?

benashz:
Hi @eldengates, we should the fix out in the next release [3.16.0](https://github.com/hashicorp/terraform-provider-vault/milestone/31)

edupo:
I still see this in `v3.24.0 (signed by HashiCorp)`.

```
│ * Waiting for the primary to upgrade from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service when the primary is finished.
```

Any idea why? Workaround?

PR #1593, opened 2022-08-30
Use new semantic version checking for Consul secrets backend logic

robmonte:
Previously we used a workaround to check which Consul policies field was needed by peeking at the Vault response. With semantic version checking we can more simply get which version of Vault is being used, then reconcile that with the field provided by the user.

For backwards compatibility, both `policies` and `consul_policies` work regardless of Vault version, 1.11 being where the field name changed in Vault, but the provider handles mapping the data given by the user to the correct parameter name for Vault depending on version.

```release-note
NONE
```

```shell
❯ make testacc TESTARGS='-run=TestConsulSecretBackend'
TF_ACC=1 go test -run=TestConsulSecretBackend -timeout 30m ./...

ok      github.com/hashicorp/terraform-provider-vault/vault     11.553s
```

&nbsp;

### Community Note

### DIFF
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -223,15 +223,14 @@ func docFilePath(tfTp tfType, endpoint string) (string, error) {
 // normalizeDocEndpoint changes the raw endpoint into the format we expect for
 // using in generated documentation structure on registry.terraform.io.
 // Example:
+//  endpoint: /transform/alphabet/{name}
+//  normalized: transform_alphabet
 //
-//	endpoint: /transform/alphabet/{name}
-//	normalized: transform_alphabet
+//  endpoint: /transform/decode/{role_name}
+//  normalized: transform_decode
 //
-//	endpoint: /transform/decode/{role_name}
-//	normalized: transform_decode
-//
-//	endpoint: /transform/encode/{role_name}
-//	normalized: transform_encode
+//  endpoint: /transform/encode/{role_name}
+//  normalized: transform_encode
 func normalizeDocEndpoint(endpoint string) string {
 	endpoint = stripCurlyBraces(endpoint)
 	endpoint = strings.TrimRight(endpoint, "name")
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -223,14 +223,15 @@ func docFilePath(tfTp tfType, endpoint string) (string, error) {
 // normalizeDocEndpoint changes the raw endpoint into the format we expect for
 // using in generated documentation structure on registry.terraform.io.
 // Example:
-//  endpoint: /transform/alphabet/{name}
-//  normalized: transform_alphabet
 //
-//  endpoint: /transform/decode/{role_name}
-//  normalized: transform_decode
+//	endpoint: /transform/alphabet/{name}
+//	normalized: transform_alphabet
 //
-//  endpoint: /transform/encode/{role_name}
-//  normalized: transform_encode
+//	endpoint: /transform/decode/{role_name}
+//	normalized: transform_decode
+//
+//	endpoint: /transform/encode/{role_name}
+//	normalized: transform_encode
 func normalizeDocEndpoint(endpoint string) string {
 	endpoint = stripCurlyBraces(endpoint)
 	endpoint = strings.TrimRight(endpoint, "name")
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -34,10 +34,10 @@ func getTargetVaultVersion(ctx context.Context, client *api.Client) (string, err
 // It uses the go-version package
 // to perform a semantic version comparison, and
 // returns:
-//    - a boolean describing whether the Vault
-//      server version was above the minimum version
-//    - the current Vault server version as a string
-//    - errors captured during operation, if any
+//   - a boolean describing whether the Vault
+//     server version was above the minimum version
+//   - the current Vault server version as a string
+//   - errors captured during operation, if any
 //
 // This function can be used to perform semantic version comparisons
 // to conditionally enable features, or to resolve any diffs in the TF
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -34,10 +34,10 @@ func getTargetVaultVersion(ctx context.Context, client *api.Client) (string, err
 // It uses the go-version package
 // to perform a semantic version comparison, and
 // returns:
-//   - a boolean describing whether the Vault
-//     server version was above the minimum version
-//   - the current Vault server version as a string
-//   - errors captured during operation, if any
+//    - a boolean describing whether the Vault
+//      server version was above the minimum version
+//    - the current Vault server version as a string
+//    - errors captured during operation, if any
 //
 // This function can be used to perform semantic version comparisons
 // to conditionally enable features, or to resolve any diffs in the TF
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -682,16 +682,23 @@ func CheckJSONData(resourceName, attr, expected string) resource.TestCheckFunc {
 	}
 }
 
-// GetImportTestStep for resource name. Optionally include field names that should be ignored during the import
-// verification, typically ignore fields should only be provided for values that are not returned from the
-// provisioning API.
-func GetImportTestStep(resourceName string, skipVerify bool, ignoreFields ...string) resource.TestStep {
-	return resource.TestStep{
+// GetImportTestStep for resource name. If a custom ImportStateCheck function is not desired, pass
+// a nil value. Optionally include field names that should be ignored during the import
+// verification, typically ignore fields should only be provided for values that are not returned
+// from the provisioning API.
+func GetImportTestStep(resourceName string, skipVerify bool, check resource.ImportStateCheckFunc, ignoreFields ...string) resource.TestStep {
+	ts := resource.TestStep{
 		ResourceName:            resourceName,
 		ImportState:             true,
 		ImportStateVerify:       !skipVerify,
 		ImportStateVerifyIgnore: ignoreFields,
 	}
+
+	if check != nil {
+		ts.ImportStateCheck = check
+	}
+
+	return ts
 }
 
 // GetNamespaceImportStateCheck checks that the namespace was properly imported into the state.
--- a/vault/gcp.go
+++ b/vault/gcp.go
@@ -10,9 +10,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-/// GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
-/// `Resource` is the self-link of a GCP resource
-/// Roles is a list of IAM roles to be assigned to an entity for that resource.
+// / GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
+// / `Resource` is the self-link of a GCP resource
+// / Roles is a list of IAM roles to be assigned to an entity for that resource.
 type GCPBinding struct {
 	Resource string
 	Roles    []string
--- a/vault/gcp.go
+++ b/vault/gcp.go
@@ -10,9 +10,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-// / GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
-// / `Resource` is the self-link of a GCP resource
-// / Roles is a list of IAM roles to be assigned to an entity for that resource.
+/// GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
+/// `Resource` is the self-link of a GCP resource
+/// Roles is a list of IAM roles to be assigned to an entity for that resource.
 type GCPBinding struct {
 	Resource string
 	Roles    []string
--- a/vault/resource_consul_secret_backend.go
+++ b/vault/resource_consul_secret_backend.go
@@ -142,18 +142,19 @@ func consulSecretBackendCreate(ctx context.Context, d *schema.ResourceData, meta
 	d.Partial(true)
 	log.Printf("[DEBUG] Mounting Consul backend at %q", path)
 
-	// If a token isn't provided and the Vault version is less than 1.11, fail before mounting the path in Vault.
-	bootstrapSupport, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
+	// If a token isn't provided and the Vault version is less than 1.11, fail before
+	// mounting the path in Vault.
+	useAPIVer1, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
 	if err != nil {
-		return diag.Errorf("Failed to read Vault client version: %s", err)
+		return diag.Errorf("failed to read Vault client version: %s", err)
 	}
-	if token == "" && !bootstrapSupport {
-		return diag.Errorf(`Error writing Consul configuration: no token provided and the Vault client
-version does not meet the minimum requirement for this feature (Vault 1.11+)`)
+	if token == "" && !useAPIVer1 {
+		return diag.Errorf(`error writing Consul configuration: no token provided and the 
+Vault client version does not meet the minimum requirement for this feature (Vault 1.11+)`)
 	}
 
 	if err := client.Sys().Mount(path, info); err != nil {
-		return diag.Errorf("Error mounting to %q: %s", path, err)
+		return diag.Errorf("error mounting to %q: %s", path, err)
 	}
 
 	log.Printf("[DEBUG] Mounted Consul backend at %q", path)
@@ -172,7 +173,7 @@ version does not meet the minimum requirement for this feature (Vault 1.11+)`)
 	}
 
 	if _, err := client.Logical().Write(configPath, data); err != nil {
-		return diag.Errorf("Error writing Consul configuration for %q: %s", path, err)
+		return diag.Errorf("error writing Consul configuration for %q: %s", path, err)
 	}
 	log.Printf("[DEBUG] Wrote Consul configuration to %q", configPath)
 	d.Partial(false)
@@ -193,7 +194,7 @@ func consulSecretBackendRead(_ context.Context, d *schema.ResourceData, meta int
 
 	mounts, err := client.Sys().ListMounts()
 	if err != nil {
-		return diag.Errorf("Error reading mount %q: %s", path, err)
+		return diag.Errorf("error reading mount %q: %s", path, err)
 	}
 
 	// path can have a trailing slash, but doesn't need to have one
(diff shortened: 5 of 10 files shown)


robmonte:
Note that I have a gofmt commit before my feature commits.

vault/resource_consul_secret_backend_role.go
```diff
@@ -19,11 +22,11 @@ var (
...
-		Create: consulSecretBackendRoleWrite,
-		Read:   ReadWrapper(consulSecretBackendRoleRead),
-		Update: consulSecretBackendRoleWrite,
-		Delete: consulSecretBackendRoleDelete,
-		Exists: consulSecretBackendRoleExists,
+		CreateContext: consulSecretBackendRoleWrite,
+		ReadContext:   ReadContextWrapper(consulSecretBackendRoleRead),
+		UpdateContext: consulSecretBackendRoleWrite,
+		Delete:        consulSecretBackendRoleDelete,
+		Exists:        consulSecretBackendRoleExists,
```
robmonte, line 29: These were updated to Context functions to provide a context to the semver lookup. I opted to only update the ones affected by the need for this code change, but I could see the argument for updating `Delete` and `Exists` as well, to not leave any un-updated portions.

benashz: Please go ahead and update all of them.

robmonte: Do the underlying logical operations needs to be updated to use the Context version?

For example on line 205 in write:
`client.Logical().Write(path, data)`
to
`client.Logical().WriteWithContext(ctx, path, data)`

benashz: Shouldn't be necessary.

internal/semver/semver.go
```diff
@@ -34,10 +34,10 @@ func getTargetVaultVersion(ctx context.Context, client *api.Client) (string, err
...
 // to perform a semantic version comparison, and
 // returns:
-//    - a boolean describing whether the Vault
-//      server version was above the minimum version
-//    - the current Vault server version as a string
-//    - errors captured during operation, if any
+//   - a boolean describing whether the Vault
+//     server version was above the minimum version
+//   - the current Vault server version as a string
+//   - errors captured during operation, if any
```
benashz, line 40: not sure if these changes were intended.

vault/gcp.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-/// GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
-/// `Resource` is the self-link of a GCP resource
-/// Roles is a list of IAM roles to be assigned to an entity for that resource.
+// / GCPBinding is used to generate the HCL binding format that GCP Secret Engine Requires
+// / `Resource` is the self-link of a GCP resource
+// / Roles is a list of IAM roles to be assigned to an entity for that resource.
```
benashz, line 15: Strange update here, probably best to just use `//`

vault/resource_consul_secret_backend_role.go
```diff
@@ -165,13 +170,22 @@ func consulSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) erro
...
 
 	data := map[string]interface{}{
-		"policies":           policies,
-		"consul_policies":    policies,
 		"consul_roles":       roles,
 		"service_identities": serviceIdentities,
 		"node_identities":    nodeIdentities,
 	}
 
+	checkConsulPolicies, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
```
benashz, line 178: nit: the `checkConsulPolicies` variable name seems to imply that we are checking some policies but it seems to only be used for selecting which field to use when configuring the consul policies. Wonder if we could come up with a more generic variable name.  The way that the semver check is implemented makes it tricky since we have to handle the error each time.

vault/resource_consul_secret_backend_role.go
```diff
@@ -189,17 +203,17 @@ func consulSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) erro
...
-	return consulSecretBackendRoleRead(d, meta)
+	return consulSecretBackendRoleRead(ctx, d, meta)
 }
 
-func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
-	client, e := provider.GetClient(d, meta)
-	if e != nil {
-		return e
+func consulSecretBackendRoleRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := provider.GetClient(d, meta)
```
benashz, line 214: Would you mind reverting this to be `client, e` ? It will make it consistent with all the other functions.

vault/resource_consul_secret_backend_role.go
```diff
@@ -257,17 +271,21 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
-		if _, ok := d.GetOk("consul_policies"); ok {
-			params["policies"] = "consul_policies"
-		}
+	// Check whether Vault will return consul_policies or policies based on its version.
+	hasNewPolicies, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	returnedPoliciesVal := "consul_policies"
```
benashz, line 280: This is the target field to set in the in the state, rather than a value.

vault/resource_consul_secret_backend_role.go
```diff
@@ -257,17 +271,21 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
 	}
 
-	_, hasLegacyPolicies := data["policies"]
-	if hasLegacyPolicies {
-		params["policies"] = "policies"
-		if _, ok := d.GetOk("consul_policies"); ok {
-			params["policies"] = "consul_policies"
-		}
+	// Check whether Vault will return consul_policies or policies based on its version.
+	hasNewPolicies, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
```
benashz, line 275: Maybe rename to `useAPIVer1` or something like that since we we are adapting our response handler based on the Vault version

vault/resource_consul_secret_backend_role.go
```diff
@@ -257,17 +271,21 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+	}
+
+	if hasNewPolicies {
+		params["consul_policies"] = returnedPoliciesVal
 	} else {
-		params["consul_policies"] = "consul_policies"
-		if _, ok := d.GetOk("policies"); ok {
-			params["consul_policies"] = "policies"
-		}
+		params["policies"] = returnedPoliciesVal
```
benashz, line 288: The remap value will always be `policies`

codegen/generate.go
```diff
@@ -223,14 +223,15 @@ func docFilePath(tfTp tfType, endpoint string) (string, error) {
 // normalizeDocEndpoint changes the raw endpoint into the format we expect for
 // using in generated documentation structure on registry.terraform.io.
 // Example:
-//  endpoint: /transform/alphabet/{name}
-//  normalized: transform_alphabet
 //
-//  endpoint: /transform/decode/{role_name}
-//  normalized: transform_decode
+//	endpoint: /transform/alphabet/{name}
```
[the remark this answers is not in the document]
robmonte: That's correct, make had complained about format so I had run `make fmt` before working on my changes. I can remove that commit after reviews are complete.

vault/resource_consul_secret_backend_role.go
robmonte: Sure I don't mind. I thought it was just an overlooked variable name since both e and err were used.

vault/resource_consul_secret_backend_role.go
vinay-gopalan: That's correct. `Exists` is deprecated and may be removed in favor of the Context functions :)

vault/resource_consul_secret_backend_role.go
vinay-gopalan: +1, would suggest renaming to `stateKey` or `field` as done in other resources for explicitness

vault/resource_consul_secret_backend_role.go
```diff
@@ -138,21 +141,23 @@ func consulSecretBackendRoleGetBackend(d *schema.ResourceData) string {
...
 	backend := consulSecretBackendRoleGetBackend(d)
 	if backend == "" {
-		return fmt.Errorf("No backend specified for Consul secret backend role %s", name)
+		return diag.Errorf("No backend specified for Consul secret backend role %s", name)
 	}
 
 	path := consulSecretBackendRolePath(backend, name)
 
+	// This loads either the consul_policies or policies field, depending on which the
+	// user provided, and then stores it under both keys in the data.
```
vinay-gopalan, line 160: Is this comment still valid? Looks like we are no longer storing the `policies` variable under both keys, and are conditionally only setting one of them

codegen/generate.go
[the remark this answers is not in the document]
robmonte: I've removed the commit with these gofmt changes

vault/resource_consul_secret_backend_role.go
robmonte: 👍

vault/resource_consul_secret_backend_role.go
robmonte: I tentatively put `useAPIVer2` to match the below suggestion.

vinay-gopalan (approved):
LGTM!

robmonte:
I'm not sure what is causing the CI failures now, is it from the fmt issues?

vault/resource_consul_secret_backend_role.go
benashz: Since Vault version checks were only recently added I would go with `useAPIVer1`, since it is a new epoch.

robmonte: Ah that makes sense, I don't mind changing it 👍

vault/resource_consul_secret_backend.go
```diff
@@ -140,8 +142,18 @@ func consulSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
 	d.Partial(true)
 	log.Printf("[DEBUG] Mounting Consul backend at %q", path)
 
+	// If a token isn't provided and the Vault version is less than 1.11, fail before mounting the path in Vault.
+	bootstrapSupport, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion11)
```
benashz, line 146: nit: `bootstrapSupport` reads like a verb, you may want to rename it to `bootstrapSupported` or `canBootstrap` or something generic like `useAPIVer1` or `isAPIVer1`

vault/resource_consul_secret_backend.go
```diff
@@ -140,8 +142,18 @@ func consulSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
...
+	if err != nil {
+		return diag.Errorf("Failed to read Vault client version: %s", err)
+	}
+	if token == "" && !bootstrapSupport {
+		return diag.Errorf(`Error writing Consul configuration: no token provided and the Vault client
```
benashz, line 151: nit: this line looks to be bit too  wide, mind folding it

vault/resource_consul_secret_backend.go
```diff
@@ -160,18 +172,18 @@ func consulSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if _, err := client.Logical().Write(configPath, data); err != nil {
-		return fmt.Errorf("Error writing Consul configuration for %q: %s", path, err)
+		return diag.Errorf("Error writing Consul configuration for %q: %s", path, err)
```
benashz, line 175: nit: errors typically start with a lowercase

vault/resource_consul_secret_backend.go
```diff
@@ -140,8 +142,18 @@ func consulSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
...
+		return diag.Errorf("Failed to read Vault client version: %s", err)
+	}
+	if token == "" && !bootstrapSupport {
+		return diag.Errorf(`Error writing Consul configuration: no token provided and the Vault client
+version does not meet the minimum requirement for this feature (Vault 1.11+)`)
+	}
+
 	if err := client.Sys().Mount(path, info); err != nil {
-		return fmt.Errorf("Error mounting to %q: %s", path, err)
+		return diag.Errorf("Error mounting to %q: %s", path, err)
```
benashz, line 156: nit: errors typically start with a lowercase

vault/resource_consul_secret_backend.go
```diff
@@ -181,7 +193,7 @@ func consulSecretBackendRead(d *schema.ResourceData, meta interface{}) error {
 
 	mounts, err := client.Sys().ListMounts()
 	if err != nil {
-		return fmt.Errorf("Error reading mount %q: %s", path, err)
+		return diag.Errorf("Error reading mount %q: %s", path, err)
```
benashz, line 196: nit: errors typically start with a lowercase

benashz (approved):
a few minor nits to address then :+1:

vault/resource_consul_secret_backend.go
robmonte: Missed this one before pushing the extra change, thanks for catching.

vault/resource_consul_secret_backend_role.go
```diff
@@ -257,17 +270,24 @@ func consulSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error
...
+
+	// Return either policies or consul_policies depending on the following criteria:
+	// * Vault version < 1.11: Always use policies
+	// * Vault version >= 1.11: Default consul_policies; use policies if the user specified it
+	policiesField := "consul_policies"
+	if !useAPIVer1 {
+		policiesField = "policies"
+	}
+
+	if _, ok := d.GetOk("policies"); ok {
```
benashz, line 287: Might be worth adding a comment documenting why these checks done here.

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -38,6 +37,8 @@ func TestConsulSecretBackendRole(t *testing.T) {
 		resource.TestCheckResourceAttr(resourceName, "consul_namespace", "consul-ns-1"),
 		resource.TestCheckResourceAttr(resourceName, "partition", "partition-1"),
 	}
+	var updateTestImportCheckFuncs []resource.TestCheckFunc
+	var ignoreFields []string
```
benashz, line 41:
```suggestion

	var ignoreFields []string
```

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -157,20 +170,24 @@ resource "vault_consul_secret_backend_role" "test" {
   partition = "partition-0"
 `, path, token, name)
 
-	if withPolicies {
+	if usePolicies {
```
benashz, line 173: Maybe rename to `withLegacyPoliciesField`

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -157,20 +170,24 @@ resource "vault_consul_secret_backend_role" "test" {
...
-  policies = [
+policies = [
     "boo",
-  ]
+]
 `
 	}
 
-	if isAboveVersionThreshold {
+	if useCPolicies {
```
benashz, line 181: Do we really need to two booleans for this? I think we could do it in an`if/else` instead

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -216,22 +233,26 @@ resource "vault_consul_secret_backend_role" "test" {
...
+policies = [
     "boo",
 	 "far",
-  ]
+]
 `
 	}
 
-	if isAboveVersionThreshold {
+	if useCPolicies {
```
benashz, line 245: similar to the related comment above

vault/resource_consul_secret_backend_role_test.go
robmonte: I don't think that will work because of the combination of values used to test things. Some tests use both at the same time. I don't see any way to limit it down to 2 booleans current.

I think re-writing the test into 2 separate test functions like we discussed will make it able to be simplified a little bit, though.

testutil/testutil.go
```diff
@@ -682,16 +682,23 @@ func CheckJSONData(resourceName, attr, expected string) resource.TestCheckFunc {
...
+// verification, typically ignore fields should only be provided for values that are not returned
+// from the provisioning API.
+func GetImportTestStep(resourceName string, skipVerify bool, check resource.ImportStateCheckFunc, ignoreFields ...string) resource.TestStep {
+	ts := resource.TestStep{
 		ResourceName:            resourceName,
 		ImportState:             true,
 		ImportStateVerify:       !skipVerify,
 		ImportStateVerifyIgnore: ignoreFields,
 	}
+
```
benashz, line 696: :+1:

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -37,82 +48,183 @@ func TestConsulSecretBackendRole(t *testing.T) {
...
+	// Imported policies will always be in the new field consul_policies, so the import test ignores
+	// both fields but has a custom ImportStateCheck function for those values.
+	resource.Test(t, resource.TestCase{
+		Providers:    testProviders,
+		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy: testAccConsulSecretBackendRoleCheckDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testConsulSecretBackendRole_initialConfig(path, name, token, true, false),
+				Check: resource.ComposeTestCheckFunc(
```
benashz, line 112: It would be good to include any checks that are common to the various phases create/upgrade etc.

vault/resource_consul_secret_backend_role_test.go
```diff
@@ -37,82 +48,183 @@ func TestConsulSecretBackendRole(t *testing.T) {
...
+	// This separate test is used to check the functionality when using the new consul_policies
+	// field in the provider when using an older version of Vault (versions 1.10 and below).
+	resource.Test(t, resource.TestCase{
+		Providers:    testProviders,
+		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy: testAccConsulSecretBackendRoleCheckDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testConsulSecretBackendRole_initialConfig(path, name, token, false, true),
+				Check: resource.ComposeTestCheckFunc(
```
benashz, line 211: Same comment as for the 1.11 test case above.

benashz (approved):
A few minor suggestions then :+1:

PR #1594, opened 2022-09-02
feat: add Mount Migration support to all secrets/auth backends

vinay-gopalan:
Adds mount migration support to all secrets/auth backends.

This PR removes the previously added `semver` package and instead moves those functions to `meta.go` instead. The `ProviderMeta` object now stores the Vault Version after making a call to the `client.Sys().SealStatus()` endpoint.

For testing, the best way I found to confirm that Mount Migration for secrets/auth mounts had successfully completed was by running a Remount test and then viewing the Vault logs to confirm that a Migration ID was dispatched, and the mount was moved accordingly:

```
2022-09-08T10:13:34.835-0700 [INFO]  core: successful mount: namespace="" path=tf-test-terraform-cloud-5084382457978848863/ type=terraform

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Starting to update the mount table and revoke leases: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Removing the source mount from filtered paths on secondaries: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Updating quotas associated with the source mount: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Completed mount move operations: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:38.301-0700 [INFO]  core: successfully unmounted: path=tf-test-terraform-cloud-updated-8145651656808162423/ namespace=""
```


vault/resource_ad_secret_backend_test.go
```diff
@@ -39,7 +39,7 @@ func TestADSecretBackend(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "userdn", "CN=Users,DC=corp,DC=example,DC=net"),
 				),
 			},
-			testutil.GetImportTestStep(resourceName, false, "bindpass", "description"),
+			//testutil.GetImportTestStep(resourceName, false, "bindpass", "description"),
```
benashz, line 42: Is this still needed?

vault/resource_ad_secret_backend_test.go
```diff
@@ -49,13 +49,13 @@ func TestADSecretBackend(t *testing.T) {
...
 			//	PlanOnly: true,
 			//},
-			{
-				Config: testADSecretBackendConflictsConfig(
-					resourceName, bindDN, bindPass, url, "formatter", "{{foo}}"),
-				ExpectError: regexp.MustCompile(`.*"formatter": conflicts with password_policy.*`),
-
-				PlanOnly: true,
-			},
+			//{
```
benashz, line 52: Is this still needed?

vault/resource_ad_secret_backend_test.go
```diff
@@ -74,6 +74,49 @@ func TestADSecretBackend(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "description", "test description"),
+					resource.TestCheckResourceAttr(resourceName, "default_lease_ttl_seconds", "3600"),
+					resource.TestCheckResourceAttr(resourceName, "max_lease_ttl_seconds", "7200"),
+					resource.TestCheckResourceAttr(resourceName, "binddn", bindDN),
+					resource.TestCheckResourceAttr(resourceName, "bindpass", bindPass),
+					resource.TestCheckResourceAttr(resourceName, "url", url),
+					resource.TestCheckResourceAttr(resourceName, "insecure_tls", "true"),
+					resource.TestCheckResourceAttr(resourceName, "userdn", "CN=Users,DC=corp,DC=example,DC=net"),
+				),
+			},
```
benashz, line 101: Should we be testing import?

vault/resource_ad_secret_backend_test.go
```diff
@@ -74,6 +74,49 @@ func TestADSecretBackend(t *testing.T) {
...
+			{
```
benashz, line 102: Does this really test the remount? There are two possible outcomes for the update: a new resource is created, or a graceful remount is done.

vault/resource_aws_secret_backend.go
```diff
@@ -99,31 +99,34 @@ func awsSecretBackendResource() *schema.Resource {
 	}
 }
 
-func mountMigrationCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
-	if !diff.HasChange("path") {
+func mountMigrationCustomizeDiff_FieldPath(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
```
benashz, line 102: See my comments for `backend` field

vault/resource_aws_secret_backend.go
```diff
@@ -99,31 +99,34 @@ func awsSecretBackendResource() *schema.Resource {
...
+	o, _ := diff.GetChange(mountField)
 	if o == "" {
 		return nil
 	}
 
 	// Mount Migration is only available for versions >= 1.10
-	client, e := provider.GetClient(diff, meta)
-	if e != nil {
-		return e
+	minVersion, err := version.NewVersion(consts.VaultVersion10)
```
benashz, line 117: Can we pull these out to to `meta.go` instead? The versions are a known quantity.

We can add an `init()`  that sets some global variables up for us:

```go

var VaultVersion110

func init() {
    VaultVersion110 = version.Must(version.NewSemVer(consts.VaultVersion10))
}
```

vault/resource_aws_secret_backend.go
```diff
@@ -99,31 +99,34 @@ func awsSecretBackendResource() *schema.Resource {
...
+	if err != nil {
+		return err
 	}
-
-	remountEnabled, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion10)
+	remountEnabled, _, err := provider.GreaterThanOrEqual(meta, minVersion)
```
benashz, line 121: nit: suggest renaming `remountSupported`

benashz, line 121: Since we are making a behavioral change here, I wonder if it would make sense to provide a boolean field for the user to opt out of the automatic remount, in the case where they are relying on the previous `ForceNew` behavior?

vault/resource_consul_secret_backend.go
```diff
@@ -226,6 +226,19 @@ func consulSecretBackendUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	d.Partial(true)
 
+	if d.HasChange(consts.FieldPath) {
```
benashz, line 229: Similar code seems to be repeated throughout, suggest generalizing this.

internal/provider/meta.go
```diff
@@ -76,6 +78,18 @@ func (p *ProviderMeta) GetNSClient(ns string) (*api.Client, error) {
 	return c, nil
 }
 
+func (p *ProviderMeta) GreaterThanOrEqual(minVersion *version.Version) (bool, string, error) {
```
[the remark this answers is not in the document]
vinay-gopalan: Updated to new names

internal/provider/meta.go
```diff
@@ -281,6 +307,54 @@ func GetClient(i interface{}, meta interface{}) (*api.Client, error) {
...
+
+	comparison, versionString, err := p.GreaterThanOrEqual(minVersion)
+	if err != nil {
+		return false, "", err
+	}
+
+	return comparison, versionString, nil
+}
+
+func getTargetVaultVersion(client *api.Client) (string, error) {
```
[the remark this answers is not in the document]
vinay-gopalan: Updated

vault/resource_ad_secret_backend.go
```diff
@@ -531,6 +537,19 @@ func updateConfigResource(d *schema.ResourceData, meta interface{}) error {
 		return e
 	}
 
+	if d.HasChange(consts.FieldBackend) {
+		// semantic version check completed in CustomizeDiff
```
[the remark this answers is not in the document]
vinay-gopalan: Updated the comment to be more informative and relevant

vault/resource_ad_secret_backend_test.go
```diff
@@ -2,7 +2,7 @@ package vault
 
 import (
 	"fmt"
-	"regexp"
+	//"regexp"
```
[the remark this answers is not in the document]
vinay-gopalan: Removed

vault/resource_ad_secret_backend_test.go
vinay-gopalan: Yup, accidentally got removed 😅 fixed

vault/resource_ad_secret_backend_test.go
vinay-gopalan: Added import tests for all new tests for secret/auth backends

vault/resource_consul_secret_backend.go
vinay-gopalan: Refactored to helper method `remountToNewPath`

vault/resource_ad_secret_backend_test.go
vinay-gopalan: That's a great point. So far the only definite way I found to confirm that mount migration had successfully completed was by manually checking the Vault logs and confirming that I saw the proper logging:

```
2022-09-08T10:13:34.835-0700 [INFO]  core: successful mount: namespace="" path=tf-test-terraform-cloud-5084382457978848863/ type=terraform

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Starting to update the mount table and revoke leases: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Removing the source mount from filtered paths on secondaries: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Updating quotas associated with the source mount: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:36.012-0700 [INFO]  core.mounts.migration: Completed mount move operations: from_path=tf-test-terraform-cloud-5084382457978848863 migration_id=21d4beff-97c0-afac-974c-b18daf982bfd namespace="" to_path=tf-test-terraform-cloud-updated-8145651656808162423

2022-09-08T10:13:38.301-0700 [INFO]  core: successfully unmounted: path=tf-test-terraform-cloud-updated-8145651656808162423/ namespace=""
```

vinay-gopalan: This test passes for versions < 1.10.0 as well, and currently the only way to know what sort of mount move happened in Vault is to check the logs. For v1.9.0 say, this test will pass in CI, but when we manually check the logs we will see that in fact a new mount was created and the previous mount was destroyed: 

```
2022-09-08T10:28:18.426-0700 [INFO]  core: successful mount: namespace="" path=tf-test-terraform-cloud-7726032830502065044/ type=terraform

2022-09-08T10:28:19.595-0700 [INFO]  core: successfully unmounted: path=tf-test-terraform-cloud-7726032830502065044/ namespace=""

2022-09-08T10:28:19.620-0700 [INFO]  core: successful mount: namespace="" path=tf-test-terraform-cloud-updated-6295110747372747410/ type=terraform

2022-09-08T10:28:20.891-0700 [INFO]  core: successfully unmounted: path=tf-test-terraform-cloud-updated-6295110747372747410/ namespace=""
```

vinay-gopalan: Unfortunately, I wasn't able to find a way for the TF SDK to know whether Vault had created a new mount and destroyed the old one, or whether Vault had remounted with a new name. Open to any suggestions :)

vault/resource_aws_secret_backend.go
vinay-gopalan: Interesting, I think it might make sense if we can determine a case that the user wouldn't want their mounts to be moved and instead would want them to be destroyed and recreated.

I think what might help inform our decision is the fact that Vault itself provides no such feature/boolean field for users to opt out of remounting, and mount migration is the default and only option available for >= v1.10.0 ([docs for mount migration](https://www.vaultproject.io/api-docs/system/remount#move-backend)). This leads me to believe that we might be okay without the boolean flag field

internal/provider/meta.go
```diff
@@ -76,6 +78,14 @@ func (p *ProviderMeta) GetNSClient(ns string) (*api.Client, error) {
 	return c, nil
 }
 
+func (p *ProviderMeta) IsAPISupported(minVersion *version.Version) bool {
```
benashz, line 81: Probably want to document this.

internal/provider/meta.go
```diff
@@ -76,6 +78,14 @@ func (p *ProviderMeta) GetNSClient(ns string) (*api.Client, error) {
...
+	return p.vaultVersion.GreaterThanOrEqual(minVersion)
+}
+
+func (p *ProviderMeta) GetVaultVersion() *version.Version {
```
benashz, line 85: Probably want to document this as well.

internal/provider/meta.go
```diff
@@ -222,6 +232,17 @@ func NewProviderMeta(d *schema.ResourceData) (interface{}, error) {
 		}
 	}
 
+	// Set the Vault version to *ProviderMeta object
+	vaultVersionString, err := getVaultVersion(client)
```
benashz, line 236: Would it be simpler or more consistent if `getVaultVersion`  whould return a `version.Version`?

internal/provider/meta.go
```diff
@@ -281,6 +303,47 @@ func GetClient(i interface{}, meta interface{}) (*api.Client, error) {
...
+// It uses the go-version package
+// to perform a semantic version comparison, and
+// returns:
+//    - a boolean describing whether the Vault
+//      server version was above the minimum version
+//
+// This function can be used to perform semantic version comparisons
+// to conditionally enable features, or to resolve any diffs in the TF
+// state based on the Vault version.
+func IsAPISupported(meta interface{}, minVersion *version.Version) bool {
```
benashz, line 318: Since this function only supports `*ProviderMeta` we may want to panic here on the type assert

internal/provider/meta.go
```diff
@@ -281,6 +303,47 @@ func GetClient(i interface{}, meta interface{}) (*api.Client, error) {
...
+	case *ProviderMeta:
+		p = v
+	default:
+		return false
+	}
+
+	return p.IsAPISupported(minVersion)
+}
+
+func getVaultVersion(client *api.Client) (string, error) {
```
benashz, line 330: Instead of returning the stringified version you can return a `version.Version` which implements `String()` interface

ref: https://pkg.go.dev/github.com/hashicorp/go-version#Version.String

vault/provider.go
```diff
@@ -869,15 +869,14 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
...
+		mv, e := version.NewVersion(minVersion)
 		if e != nil {
 			return diag.FromErr(e)
 		}
 
-		featureEnabled, currentVersion, err := semver.GreaterThanOrEqual(ctx, client, minVersion)
-		if err != nil {
-			return diag.FromErr(err)
-		}
+		featureEnabled := provider.IsAPISupported(meta, mv)
```
benashz, line 877: nit: this is more at the api level, rather than the feature level, perhaps rename to to something like `apiSupported` or something like that.

vault/provider.go
```diff
@@ -869,15 +869,14 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
...
+
+		currentVersion := meta.(*provider.ProviderMeta).GetVaultVersion().String()
```
benashz, line 879: you can pass `currentVersion` directly to `diag.Errorf()` since it is a stringer  i.e. no need to call its `String()` method

vault/resource_ad_secret_backend.go
```diff
@@ -531,6 +537,11 @@ func updateConfigResource(d *schema.ResourceData, meta interface{}) error {
 		return e
 	}
 
+	backend, err := remountToNewPath(d, client, consts.FieldBackend, false)
```
benashz, line 540: `backend` is being redeclared here?

vault/resource_ad_secret_backend_test.go
benashz: Would it be possible to leverage the `uuid` of the mount, it seems that is available in the response: https://www.vaultproject.io/api-docs/system/mounts#sample-response

vault/resource_aws_secret_backend.go
```diff
@@ -99,38 +99,65 @@ func awsSecretBackendResource() *schema.Resource {
...
 	}
 
 	// Mount Migration is only available for versions >= 1.10
-	client, e := provider.GetClient(diff, meta)
-	if e != nil {
-		return e
-	}
-
-	remountEnabled, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion10)
+	minVersion, err := version.NewVersion(consts.VaultVersion10)
```
benashz, line 117: `consts.VaultVesrion10` seems to be missing the major, shouldn't that be `consts.VaultVersion110`? That would handle the a major version bump.

vault/resource_aws_secret_backend.go
```diff
@@ -99,38 +99,65 @@ func awsSecretBackendResource() *schema.Resource {
...
-		if err := diff.ForceNew("path"); err != nil {
+		if err := diff.ForceNew(mountField); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
+func remountToNewPath(d *schema.ResourceData, client *api.Client, mountField string, isAuthMount bool) (string, error) {
```
benashz, line 134: nit: the `Path` portion in the function name may be redundant. A remount implies a source and destination. Suggest renaming to `remount()` instead.

benashz, line 134: This seems like a general purpose function, suggest moving it to the `util` package.

vault/resource_aws_secret_backend.go
```diff
@@ -99,38 +99,65 @@ func awsSecretBackendResource() *schema.Resource {
...
+
+	if d.HasChange(mountField) {
+		// since this function is only called within Update
+		// we know that remount is enabled
+		o, n := d.GetChange(mountField)
+		oldPath := o.(string)
+		newPath := n.(string)
+
+		if isAuthMount {
+			oldPath = "auth/" + oldPath
```
benashz, line 145: any danger of `oldPath` being having the `auth` parent?

vault/resource_consul_secret_backend.go
```diff
@@ -3,12 +3,13 @@ package vault
...
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-provider-vault/internal/consts"
-	"github.com/hashicorp/terraform-provider-vault/internal/semver"
 	"log"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
```
benashz, line 11: Formatting seems a bit off, local package imports should be grouped together.

vault/resource_consul_secret_backend_role.go
```diff
@@ -3,12 +3,12 @@ package vault
...
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-provider-vault/internal/semver"
 	"log"
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+
```
benashz, line 11: nit: gratuitous new line.

vault/resource_ad_secret_backend_test.go
```diff
@@ -74,6 +74,58 @@ func TestADSecretBackend(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "insecure_tls", "true"),
+					resource.TestCheckResourceAttr(resourceName, "userdn", "CN=Users,DC=corp,DC=example,DC=net"),
+				),
+			},
+			{
+				Config: testADSecretBackend_initialConfig(updatedBackend, bindDN, bindPass, url),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "backend", updatedBackend),
+					resource.TestCheckResourceAttr(resourceName, "description", "test description"),
+					resource.TestCheckResourceAttr(resourceName, "default_lease_ttl_seconds", "3600"),
```
benashz, line 107: nit:  mind using `testutil.GetImportTestStep()` instead.

vault/provider.go
```diff
@@ -41,6 +41,16 @@ const (
 // The key of the mutex should be the path in Vault.
 var vaultMutexKV = helper.NewMutexKV()
 
+var (
+	VaultVersion110 *version.Version
+	VaultVersion111 *version.Version
+)
+
+func init() {
+	VaultVersion110 = version.Must(version.NewSemver(consts.VaultVersion10))
```
benashz, line 50: nit: update constant name to match. Include the major.

vault/resource_aws_secret_backend.go
benashz: The remount is an explicit action, here it is happening behind the scenes. One consideration is that a remount operation may not migrate everything related to the mount. I believe that is a known limitation, so having the extra escape hatch might make sense here. The default is to perform a graceful remount, rather than a destroy before create operation.

vinay-gopalan: Ah, that is a good point. I think implementing a boolean field wouldn't be too bad. Are you okay with me implementing a `mustAddMountMigrationSchema` util? I could add that field to all backends that enable mount migration

benashz: sure, sounds good to me!

internal/provider/meta.go
vinay-gopalan: Updated with documentation

internal/provider/meta.go
vinay-gopalan: Updated by returning a `version.Version`

vault/provider.go
vinay-gopalan: Nice! Didn't know that, thanks :)

vault/resource_ad_secret_backend.go
vinay-gopalan: Nice catch! Updated

vault/resource_aws_secret_backend.go
vinay-gopalan: Added the major version to the constants

vault/provider.go
vinay-gopalan: Updated

internal/provider/meta.go
vinay-gopalan: Good call. Added a panic statement

vault/resource_aws_secret_backend.go
vinay-gopalan: Added new variables and an `init` method

vault/resource_ad_secret_backend_test.go
vinay-gopalan: Updated

vault/resource_aws_secret_backend.go
vinay-gopalan: All Auth backends set the `id` and the `path` fields to the mount values, so I think we are good here!

internal/consts/consts.go
```diff
@@ -85,9 +85,9 @@ const (
 	/*
 		Vault version constants
 	*/
-	VaultVersion11 = "1.11.0"
-	VaultVersion10 = "1.10.0"
-	VaultVersion9  = "1.9.0"
+	VaultVersion111 = "1.11.0"
+	VaultVersion110 = "1.10.0"
+	VaultVersion19  = "1.9.0"
```
benashz, line 90:
```suggestion
	VaultVersion190  = "1.9.0"
```

vault/provider.go
```diff
@@ -41,6 +41,16 @@ const (
...
 var vaultMutexKV = helper.NewMutexKV()
 
+var (
+	VaultVersion110 *version.Version
+	VaultVersion111 *version.Version
+)
+
+func init() {
+	VaultVersion110 = version.Must(version.NewSemver(consts.VaultVersion110))
+	VaultVersion111 = version.Must(version.NewSemver(consts.VaultVersion111))
```
benashz, line 51: We don't need `VaultVersion190` ?

vault/resource_aws_secret_backend.go
```diff
@@ -99,31 +100,31 @@ func awsSecretBackendResource() *schema.Resource {
...
 	}
 
 	// Mount Migration is only available for versions >= 1.10
-	client, e := provider.GetClient(diff, meta)
-	if e != nil {
-		return e
-	}
-
-	remountEnabled, _, err := semver.GreaterThanOrEqual(ctx, client, consts.VaultVersion10)
+	minVersion, err := version.NewVersion(consts.VaultVersion110)
```
benashz, line 118: Could we use `VaultVersion110` instead?

vault/resource_kubernetes_secret_backend.go
```diff
@@ -22,7 +22,7 @@ const (
 
 func kubernetesSecretBackendResource() *schema.Resource {
 	resource := &schema.Resource{
-		CreateContext: MinVersionCheckWrapper(kubernetesSecretBackendCreateUpdate, consts.VaultVersion11),
+		CreateContext: MinVersionCheckWrapper(kubernetesSecretBackendCreateUpdate, consts.VaultVersion111),
```
benashz, line 25: Looks like `MinVersionCheckWrapper()` should probably be updated to take a `version.Version` instead of a version string?

vault/provider.go
vinay-gopalan: Added

vault/resource_consul_secret_backend.go
vinay-gopalan: This happened due to Goland's formatting for sorting imports, not sure what the right order is in your files but Goland reformats these for me upon saves

vault/resource_consul_secret_backend_role.go
vinay-gopalan: See comment above regarding Goland's formatting

vault/provider.go
```diff
@@ -869,17 +881,22 @@ func ReadContextWrapper(f schema.ReadContextFunc) schema.ReadContextFunc {
 // wrapped schema.CreateContextFunc.
 func MinVersionCheckWrapper(f schema.CreateContextFunc, minVersion string) schema.CreateContextFunc {
 	return func(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-		client, e := provider.GetClient(d, meta)
-		if e != nil {
-			return diag.FromErr(e)
+		versionMap := map[string]*version.Version{
```
benashz, line 884: I think if you change signature of the outer function to take a `version.Version` then you can probably drop the map altogether.

vinay-gopalan: I tried doing that, but unfortunately at the time of passing the version variables to the `CreateContext` function of a resource, the `VaultVersionXXX` version variables do not initialize, and are always being passed to this function as `nil`

vault/resource_aws_secret_backend.go
vinay-gopalan: Added a `mustAddMountMigrationSchema` util that adds a boolean to disable remounts. An explicit test to disable remounts has been demonstrated in the Consul Secret backend tests

vault/resource_consul_secret_backend.go
vinay-gopalan: Fixed

vault/resource_consul_secret_backend_role.go
vinay-gopalan: Fixed

vault/resource_aws_secret_backend.go
```diff
@@ -96,39 +96,32 @@ func awsSecretBackendResource() *schema.Resource {
...
 		},
-	}
+	})
 }
 
-func mountMigrationCustomizeDiff(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
-	if !diff.HasChange("path") {
-		return nil
-	}
+func getMountMigrationDiff(field string) schema.CustomizeDiffFunc {
```
benashz, line 102:
```suggestion
func getMountCustomizeDiffFunc(field string) schema.CustomizeDiffFunc {
```

vault/resource_consul_secret_backend_test.go
```diff
@@ -173,6 +173,100 @@ func TestConsulSecretBackend_Bootstrap(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "disable_remount", "true"),
+					getUUIDForMount(path, &newUUID),
+					testCheckMountUUIDs(oldUUID, newUUID, false),
+				),
+			},
+		},
+	})
+}
+
+func getUUIDForMount(path string, uuid *string) resource.TestCheckFunc {
```
benashz, line 240:
```suggestion
func testGetUUIDForMount(path string, uuid *string) resource.TestCheckFunc {
```

vault/resource_consul_secret_backend_test.go
```diff
@@ -173,6 +173,100 @@ func TestConsulSecretBackend_Bootstrap(t *testing.T) {
...
 }
 
+func TestConsulSecretBackend_remount(t *testing.T) {
+	path := acctest.RandomWithPrefix("tf-test-consul")
+	updatedPath := acctest.RandomWithPrefix("tf-test-consul-updated")
+	token := "026a0c16-87cd-4c2d-b3f3-fb539f592b7e"
+
+	resourceName := "vault_consul_secret_backend.test"
+
+	var oldUUID string
```
benashz, line 183: I prefer to use a struct for this type of thing, and then pass the an instance ref around. This is okay for this case however.

benashz (approved):
A few minor nits/suggestions to address; then :+1:

vault/resource_consul_secret_backend_test.go
vinay-gopalan: Updated to use a `testMountStore` object. Thanks!

vault/resource_ad_secret_backend.go
```diff
@@ -217,10 +218,15 @@ func adSecretBackendResource() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
-		Schema: fields,
+		CustomizeDiff: mountMigrationCustomizeDiff_FieldBackend,
+		Schema:        fields,
 	}
 }
 
+func mountMigrationCustomizeDiff_FieldBackend(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) error {
```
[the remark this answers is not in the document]
vinay-gopalan: Fixed as per feedback. Thanks!

vault/resource_ad_secret_backend_test.go
vinay-gopalan: Was able to set up a test for `consul_secret_backend` resource which compares UUIDs for the mounts to verify that the mount was appropriately remounted/recreated.

vault/provider.go
vinay-gopalan: Updated after our discussion. Thanks!

PR #1596, opened 2022-09-06
Add redis elasticache database plugin support

maxcoulombe:

```release-note
feat: Add support for the upcoming Redis-ElastiCache Database Plugin
```

```
/usr/local/go/bin/go tool test2json -t /tmp/GoLand/___1TestAccDatabaseSecretBackendConnection_redisElastiCache_in_github_com_hashicorp_terraform_provider_vault_vault.test -test.v -test.paniconexit0 -test.run ^\QTestAccDatabaseSecretBackendConnection_redisElastiCache\E$
2022/09/06 17:34:50 [INFO] Using Vault token with the following policies: root
=== RUN   TestAccDatabaseSecretBackendConnection_redisElastiCache
--- PASS: TestAccDatabaseSecretBackendConnection_redisElastiCache (0.86s)
PASS

```

Config example:
```
provider "vault" {
  token   = "root"
}

resource "vault_database_secret_backend_connection" "redis_db" {
  backend       = "database"
  name          = "redis"
  allowed_roles = ["*"]

  redis_elasticache {
    url = "master.vault-plugin-elasticache-test.6ylpiw.use1.cache.amazonaws.com:6379"
  }
}

resource "vault_database_secret_backend_static_role" "redis_role" {
  name            = "redis"
  backend         = "database"
  username        = "vault-test"
  rotation_period = 300
  db_name         = vault_database_secret_backend_connection.redis_db.name
}
```


vinay-gopalan (commented):
Looking good! I think you also need a `setRedisElastiCacheConnectionData` function that reads the data from the TF config. This method can then be called from `getDatabaseAPIDataForEngine` as done here: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_database_secret_backend_connection.go#L833

codegen/generate.go
```diff
@@ -223,14 +223,15 @@ func docFilePath(tfTp tfType, endpoint string) (string, error) {
...
-//  endpoint: /transform/encode/{role_name}
-//  normalized: transform_encode
+//	endpoint: /transform/decode/{role_name}
+//	normalized: transform_decode
+//
+//	endpoint: /transform/encode/{role_name}
+//	normalized: transform_encode
 func normalizeDocEndpoint(endpoint string) string {
 	endpoint = stripCurlyBraces(endpoint)
 	endpoint = strings.TrimRight(endpoint, "name")
```
maxcoulombe, line 237: Not sure if it's legit? This was modified by `make fmt`. I'm using go v1.19.1.

vault/resource_database_secret_backend_connection.go
```diff
@@ -831,10 +869,14 @@ func getDatabaseAPIDataForEngine(engine *dbEngine, idx int, d *schema.ResourceDa
...
 	case dbEngineElasticSearch:
 		setElasticsearchDatabaseConnectionData(d, prefix, data)
+	case dbEngineRedisElastiCache:
+		setRedisElastiCacheDatabaseConnectionData(d, prefix, data)
 	case dbEngineSnowflake:
 		setDatabaseConnectionDataWithUserPass(d, prefix, data)
 	case dbEngineRedshift:
 		setDatabaseConnectionDataWithDisableEscaping(d, prefix, data)
+	default:
+		return nil, fmt.Errorf("unrecognized DB engine: %v", engine)
```
maxcoulombe, line 879: I added a guard since engines are specific enum values and should never silently pass through. Switch cases should [almost always contain a default case](https://stackoverflow.com/questions/4649423/should-switch-statements-always-contain-a-default-clause).

maxcoulombe:
> Looking good! I think you also need a `setRedisElastiCacheConnectionData` function that reads the data from the TF config. This method can then be called from `getDatabaseAPIDataForEngine` as done here: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_database_secret_backend_connection.go#L833

Thanks for catching that. Out of curiosity whats the functional purpose of adding that? It seems like everything was working fine including showing the diff during updates and populating the statefile so I'm wondering if it's a "dead" piece of code and if not what use case can I use to validate it's working as intended?

vault/resource_database_secret_backend_connection.go
vinay-gopalan: Nice! That's a good call, thanks for adding that

vinay-gopalan:
> > Looking good! I think you also need a `setRedisElastiCacheConnectionData` function that reads the data from the TF config. This method can then be called from `getDatabaseAPIDataForEngine` as done here: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_database_secret_backend_connection.go#L833
> 
> Thanks for catching that. Out of curiosity whats the functional purpose of adding that? It seems like everything was working fine including showing the diff during updates and populating the statefile so I'm wondering what use case can I use to validate it's working as intended?

Good question! I believe in the context of your test, the SDK was mainly only checking if the respective fields are populated with the particular values within the resource diff and state. However, since we never actually read those values from the resource diff and write them to Vault, an Import test would show that the test would fail. Import tests read the resource from Vault and verify the values are the what we expect them to be, thereby providing a more complete loop on the test. Here's an example of an import test for a DB engine: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_database_secret_backend_connection_test.go#L247-L252

vault/resource_database_secret_backend_connection.go
```diff
@@ -556,6 +561,39 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
...
+					"url": {
+						Type:        schema.TypeString,
+						Required:    true,
+						Description: "The configuration endpoint for the ElastiCache cluster to connect to.",
+					},
+					"username": {
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "The AWS access key id to use to talk to ElastiCache. If omitted the credentials chain provider is used instead.",
+						Sensitive:   true,
```
benashz, line 579: Please fold long lines at 100 columns

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -842,6 +842,40 @@ func TestAccDatabaseSecretBackendConnection_snowflake(t *testing.T) {
...
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "allowed_roles.0", "*"),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "verify_connection", "true"),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "redis_elasticache.0.url", url),
+				),
+			},
+			{
+				ResourceName:            testDefaultDatabaseSecretBackendResource,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"verify_connection", "couchbase.0.password"},
```
benashz, line 873: Shouldn't that be:  `redis_elasticache.0.password` ?

website/docs/r/database_secret_backend_connection.md
```diff
@@ -92,6 +92,9 @@ The following arguments are supported:
 
 * `influxdb` - (Optional) A nested block containing configuration options for InfluxDB connections.
 
+* `redis_elasticache` - (Optional) A nested block containing configuration options for Redis ElastiCache connections.
+
+
```
benashz, line 97: is this extra new line required?

benashz (commented):
Looking good, just have a few issues/nits/suggestions

vinay-gopalan (approved):
LGTM after updating the import to ignore redis password and adding the docs!

vault/resource_database_secret_backend_connection_test.go
maxcoulombe: Whoops thanks for catching that.

benashz (approved):
LGTM!

PR #1600, opened 2022-09-08
Accept data source values in the token field for Consul secrets backend

### LINKED ISSUE #1595: Resource vault_consul_secret_backend thinks token is unspecified (but exists from data source).
### Terraform Version
Terraform v1.2.8
on darwin_arm64
+ provider registry.terraform.io/hashicorp/consul v2.15.1
+ provider registry.terraform.io/hashicorp/vault v3.8.2

### Affected Resource(s)
- vault_consul_secret_backend

```hcl
resource "consul_acl_token" "vault" {
  description = "Token for Vault secrets engine."
  policies    = ["global-management"]
}

data "consul_acl_token_secret_id" "vault" {
  accessor_id = consul_acl_token.vault.id
}

resource "vault_consul_secret_backend" "consul" {
  path        = "consul"
  description = "Access Consul tokens"

  address                   = "127.0.0.1:8500"
  scheme                    = "http"
  token                     = data.consul_acl_token_secret_id.vault.secret_id
  bootstrap                 = false
  default_lease_ttl_seconds = "3600"
  max_lease_ttl_seconds     = "36000"
}
```

### Expected Behavior
The token for the vault secrets backend is being retrieved via a data source that is "known after apply". the expected behaviour is for the code to apply successfully. 

### Actual Behavior
When the configuration is attempting to plan or apply the following message is returned.
```
│ Error: field 'bootstrap' must be set to true when 'token' is unspecified
```

### Steps to Reproduce
using the code provided
1. `terraform apply`

### Important Factoids
Consul's ACL system has been bootstrapped previously and a valid ACL token is available in the environment variables.

robmonte:
With the recent update to the Consul secret backend, we force users to set the field `bootstrap` to true when they do not provide a token. This was mistakenly triggering when the token is set to a data source, as the provider initially sees it as an empty value.

This bug fix determines that the empty value from a data source is not the known final value, and so allows the provider to continue. 

```release-note
BUG FIXES:
* Correctly allow data source values for Consul tokens
```

```
ok      github.com/hashicorp/terraform-provider-vault/vault     5.025s
```

Closes #1595

### DIFF
--- a/vault/resource_consul_secret_backend.go
+++ b/vault/resource_consul_secret_backend.go
@@ -296,13 +296,17 @@ func consulSecretsBackendCustomizeDiff(_ context.Context, diff *schema.ResourceD
 	newToken := diff.Get("token").(string)
 	newBootstrap := diff.Get("bootstrap").(bool)
 
-	// If the user sets bootstrap to false but doesn't provide a token, disallow it.
-	if newToken == "" && !newBootstrap {
+	// Disallow the following:
+	//   1. Bootstrap is false, the token field is empty, and we know this is the final value of token.
+	if !newBootstrap && newToken == "" && diff.NewValueKnown("token") {
 		return fmt.Errorf("field 'bootstrap' must be set to true when 'token' is unspecified")
 	}
 
-	// If the user sets bootstrap to true and also provides a token, disallow it.
-	if newToken != "" && newBootstrap {
+	// Disallow the following:
+	//   1. Bootstrap is true and the token field is set to something.
+	//   2. Bootstrap is true and the token field is empty, but we don't know the final value of token.
+	if newBootstrap && newToken != "" ||
+		(newBootstrap && newToken == "" && !diff.NewValueKnown("token")) {
 		return fmt.Errorf("field 'bootstrap' must be set to false when 'token' is specified")
 	}
 
--- a/vault/resource_consul_secret_backend.go
+++ b/vault/resource_consul_secret_backend.go
@@ -294,20 +294,21 @@ func consulSecretBackendConfigPath(backend string) string {
 
 func consulSecretsBackendCustomizeDiff(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
 	newToken := diff.Get("token").(string)
-	newBootstrap := diff.Get("bootstrap").(bool)
-
-	// Disallow the following:
-	//   1. Bootstrap is false, the token field is empty, and we know this is the final value of token.
-	if !newBootstrap && newToken == "" && diff.NewValueKnown("token") {
-		return fmt.Errorf("field 'bootstrap' must be set to true when 'token' is unspecified")
-	}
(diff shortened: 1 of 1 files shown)


vault/resource_consul_secret_backend.go
```diff
@@ -296,13 +296,17 @@ func consulSecretsBackendCustomizeDiff(_ context.Context, diff *schema.ResourceD
...
 
-	// If the user sets bootstrap to true and also provides a token, disallow it.
-	if newToken != "" && newBootstrap {
+	// Disallow the following:
+	//   1. Bootstrap is true and the token field is set to something.
+	//   2. Bootstrap is true and the token field is empty, but we don't know the final value of token.
+	if newBootstrap && newToken != "" ||
+		(newBootstrap && newToken == "" && !diff.NewValueKnown("token")) {
 		return fmt.Errorf("field 'bootstrap' must be set to false when 'token' is specified")
 	}
```
robmonte, line 311: This if statement can actually be simplified to:
```go
if newBootstrap &&
		(newToken != "" || newToken == "" && !diff.NewValueKnown("token")) {
		return fmt.Errorf("field 'bootstrap' must be set to false when 'token' is specified")
	}
```

but I actually find the first version to be easier to read logic-wise. Open to changing it though.

vault/resource_consul_secret_backend.go
```diff
@@ -296,13 +296,17 @@ func consulSecretsBackendCustomizeDiff(_ context.Context, diff *schema.ResourceD
 	newToken := diff.Get("token").(string)
 	newBootstrap := diff.Get("bootstrap").(bool)
 
-	// If the user sets bootstrap to false but doesn't provide a token, disallow it.
-	if newToken == "" && !newBootstrap {
+	// Disallow the following:
+	//   1. Bootstrap is false, the token field is empty, and we know this is the final value of token.
+	if !newBootstrap && newToken == "" && diff.NewValueKnown("token") {
```
benashz, line 301: I like to group related booleans:

```suggestion
	if !newBootstrap && (newToken == "" && diff.NewValueKnown("token")) {
```

vault/resource_consul_secret_backend.go
```diff
@@ -296,13 +296,17 @@ func consulSecretsBackendCustomizeDiff(_ context.Context, diff *schema.ResourceD
 	newToken := diff.Get("token").(string)
 	newBootstrap := diff.Get("bootstrap").(bool)
 
-	// If the user sets bootstrap to false but doesn't provide a token, disallow it.
-	if newToken == "" && !newBootstrap {
+	// Disallow the following:
```
benashz, line 299: I think the flow would be a little bit cleaner with a single if/else statement like:

```
if diff.Get("bootstrap").(bool) {
   .... do the validation 
} else {
   .... do the other validation
}
```

robmonte: Please let me know if my latest commit is what you were envisioning.

benashz (approved):
LGTM!

PR #1618, opened 2022-09-27
Rework first-class AWS login support

benashz:


internal/consts/consts.go
```diff
@@ -79,14 +79,22 @@ const (
...
 	FieldAuthLoginAzure           = "auth_login_azure"
-	FieldIdentity                 = "identity"
-	FieldSignature                = "signature"
-	FieldPKCS7                    = "pkcs7"
-	FieldNonce                    = "nonce"
 	FieldIAMHttpRequestMethod     = "iam_http_request_method"
 	FieldIAMHttpRequestURL        = "iam_http_request_url"
 	FieldIAMHttpRequestBody       = "iam_http_request_body"
 	FieldIAMHttpRequestHeaders    = "iam_http_request_headers"
+	FieldAWSAccessKeyID           = "aws_access_key_id"
```
vinay-gopalan, line 86: nit: I wonder if we should now rename the pre-existing constants defined here to be more generic. Like `FieldAccessKey` and `FieldSecretKey` https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/consts/consts.go#L48

vinay-gopalan (approved):
LGTM

internal/consts/consts.go
benashz: Funny you should say that, I did just that for the Login MFA work.

vinay-gopalan: Nice!

PR #1620, opened 2022-09-28
Add support for login MFA resources

### LINKED ISSUE #1431: Add support for Login MFA
Add support for the new (1.10) [Login MFA](https://www.vaultproject.io/docs/auth/login-mfa#login-mfa) feature.

benashz:
This PR adds support for the Login MFA methods: `duo`, `okta`, `pingid`, `totp`.

It also adds support for configuring Login MFA enforcement.

Closes #1431


internal/identity/mfa/mfa.go
```diff
@@ -0,0 +1,617 @@
...
+			return diag.FromErr(fmt.Errorf("resource ID is empty"))
+		}
+
+		path, err := config.GetRequestPathWithID(id)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		// login MFA does not support partial updates unfortunately,
+		// so we update() becomes very similar to create.
```
vinay-gopalan, line 504: nit
```suggestion
		// so update() becomes very similar to create.
```

vault/resource_identity_mfa_duo_test.go
```diff
@@ -0,0 +1,86 @@
...
+	resourceName := mfa.ResourceNameDuo + ".test"
+
+	checksCommon := []resource.TestCheckFunc{
+		resource.TestCheckResourceAttrSet(resourceName, consts.FieldUUID),
+		resource.TestCheckResourceAttrSet(resourceName, consts.FieldMethodID),
+		resource.TestCheckResourceAttr(resourceName, consts.FieldType, mfa.MethodTypeDuo),
+		resource.TestCheckResourceAttr(resourceName, consts.FieldNamespaceID, "root"),
+	}
+
+	res, err := mfa.GetDuoSchemaResource()
```
vinay-gopalan, line 26: It's not fully clear to me why we are doing this paired with the switch case. It looks like we're getting the full Duo resource (along with all context functions), mostly to add empty string checks for `name` and `mount_accessor`, and that too _if_ the Duo schema has the Name and Mount Acc. field on it. But since we're using `getMethodSchemaResource`, don't we always know that the resource schema will have those fields on it?

It seems like we can achieve the same behavior by simply appending these checks without getting the Duo schema resource and checking a known case with a switch case block:
```
resource.TestCheckResourceAttr(resourceName, consts.FieldName, "")
resource.TestCheckResourceAttr(resourceName, consts.FieldMountAccessor, "")
```

vinay-gopalan: I also see that this is behavior specific to Okta and Duo and not the other MFA resources in this PR. Maybe we're using this to add a guard for when the schema for Duo and Okta changes? 🤔 Let me know if I'm missing something in my understanding

vinay-gopalan (commented):
Looking good! Had a couple of comments/questions

vault/resource_identity_mfa_duo_test.go
benashz: Fixed in 75bc3995bb942b958c52cd71e3b23c1eed9a9d8e

internal/identity/mfa/mfa.go
benashz: Fixed in 75bc3995bb942b958c52cd71e3b23c1eed9a9d8e

vinay-gopalan (approved):
Looks great!

PR #1621, opened 2022-10-03
docs: document the `disable_remount` field

vinay-gopalan:
Adds documentation to Auth/Secret backends for the new `disable_remount` field added to disable mount migration.


benashz (approved):
LGTM!

PR #1622, opened 2022-10-05
Fix erroneous persistent diff in the vault_token resource.

benashz:
The vault_token resource would produce a persistent diff for the `renewable` and `policies` fields. This PR ensures a valid diff is produced upon plan.

Other fixes:

- factor out more constants
- clean up extraneous bits in consts.go


vinay-gopalan (approved):
LGTM!

PR #1648, opened 2022-10-27
Fix GCP auth with service account credentials

magat:
We are trying to use the provider with SA credentials, but the authentication does not succeed because the required scope is not present in the JWT request: "https://www.googleapis.com/auth/cloud-platform"

This scope is mandatory according to the documentation https://developer.hashicorp.com/vault/docs/auth/gcp#gcp-credentials.

The JWT sign request fails with the following error:
```
{"error":"invalid_scope","error_description":"Invalid OAuth scope or ID token audience provided."}
```

I was able to reproduce this in the `TestAuthLoginGCP_Login` test, but only after setting the initialized:true setting on the `AuthLoginGCP.AuthLoginCommon` object. I guess the test is not ran when this value is not positioned.

This fixes the JWT token generation, but the provider's call to Vault is rejected because the `credentials` and `service_account` parameters are not expected by the endpoint. I added them to the `copyParamsExcluding` call and changed the expected parameters in the test.

I can now run this test with valid credentials (see below).

### Community Note

```release-note
Fix GCP authentication backend when using Service Account credentials
```

```
$ GOOGLE_APPLICATION_CREDENTIALS="/tmp/creds.json"
TF_ACC_GOOGLE_SERVICE_ACCOUNT="xxx@yyyy.iam.gserviceaccount.com" go test -run TestAuthLoginGCP_Login
PASS
ok      github.com/hashicorp/terraform-provider-vault/internal/provider 0.491s
...
```

### DIFF
--- a/internal/provider/auth_gcp.go
+++ b/internal/provider/auth_gcp.go
@@ -99,6 +99,8 @@ func (l *AuthLoginGCP) Login(client *api.Client) (*api.Secret, error) {
 		consts.FieldNamespace,
 		consts.FieldMount,
 		consts.FieldJWT,
+		consts.FieldCredentials,
+		consts.FieldServiceAccount,
 	)
 	if err != nil {
 		return nil, err
@@ -123,7 +125,7 @@ func (l *AuthLoginGCP) getJWT(ctx context.Context) (string, error) {
 
 	if v, ok := l.params[consts.FieldCredentials]; ok && v.(string) != "" {
 		// get the token from IAM
-		creds, err := getGCPOauthCredentials(ctx, v.(string))
+		creds, err := getGCPOauthCredentials(ctx, v.(string), "https://www.googleapis.com/auth/cloud-platform")
 		if err != nil {
 			return "", fmt.Errorf(
 				"JSON credentials are not valid, err=%w", err)
--- a/internal/provider/auth_gcp.go
+++ b/internal/provider/auth_gcp.go
@@ -21,6 +21,8 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 )
 
+const defaultGCPScope = "https://www.googleapis.com/auth/cloud-platform"
+
 // GetGCPLoginSchema for the gcp authentication engine.
 func GetGCPLoginSchema(authField string) *schema.Schema {
 	return getLoginSchema(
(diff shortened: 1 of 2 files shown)


internal/provider/auth_gcp.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 )
 
+const defaultGCPScope = "https://www.googleapis.com/auth/cloud-platform"
```
benashz, line 24: I wonder if it would make sense to add a new scheme field for configuring the scopes in the `provider` resource block. It could default to `[]string{defaultGCPScope}`?

internal/provider/auth_gcp.go
```diff
@@ -99,6 +101,8 @@ func (l *AuthLoginGCP) Login(client *api.Client) (*api.Secret, error) {
 		consts.FieldNamespace,
 		consts.FieldMount,
 		consts.FieldJWT,
+		consts.FieldCredentials,
```
benashz, line 104: Good catch!

internal/provider/auth_gcp.go
```diff
@@ -123,7 +127,7 @@ func (l *AuthLoginGCP) getJWT(ctx context.Context) (string, error) {
 
 	if v, ok := l.params[consts.FieldCredentials]; ok && v.(string) != "" {
 		// get the token from IAM
-		creds, err := getGCPOauthCredentials(ctx, v.(string))
+		creds, err := getGCPOauthCredentials(ctx, v.(string), defaultGCPScope)
```
benashz, line 130: Getting the scopes from the configured params might make better sense here, since it would allow the user to set whatever they'd like. See the related comment above.

benashz (commented):
Looking good.  Just a suggestion for making scopes a configurable.

internal/provider/auth_gcp.go
magat: I am not sure I understand the need behind configurable scopes. Since Vault requires the default scope to be set, what value would another scope bring here ? 
In other words, if the token was created with an additional scope, would Vault/the provider use this scope ?

You know this part way better than I do, so I'll follow your advice, but I'd like to understand :)

benashz: Good point. I think we can keep it in sync with the way that `vault login method=gcp` works today. If we ever need to support configuring "other" scopes then we have pretty easy path forward.

benashz (approved):
LGTM 👍 

gilberthdez:
I know it just got merged... but any idea when it will be released?(I was just facing this same problem 😓 )

benashz:
> I know it just got merged... but any idea when it will be released?(I was just facing this same problem 😓 )

Sorry to hear you are having the same issue. We haven't set a date for v3.10.1 just yet. It should probably be mid next week at the earliest.

gilberthdez:
thanks for the reply!, yeah next week is totally fine 👍 

also thanks to @magat for the report and the fix! 🙌

benashz:
Please pardon the delay in getting this fix out. It will go out in tomorrow's [v3.11.0](https://github.com/hashicorp/terraform-provider-vault/milestone/25) release.

PR #1651, opened 2022-10-27
Update the TFC tests to be less revealing

benashz:


tvoran (approved):
:shipit: 

Though I also wonder if these would be good to use `SkipTestEnvSet` with, so they're explicitly opt-out instead of skipped automatically.

vinay-gopalan (approved):

benashz:
> Though I also wonder if these would be good to use `SkipTestEnvSet` with, so they're explicitly opt-out instead of skipped automatically.

Definitely! We can do that when we add support in CI.

PR #1653, opened 2022-10-28
Temporarily skip azure secrets tests on vault-1.12

benashz:
Drop CheckTestVaultVersion() in favour of provider.IsAPISupported()


vinay-gopalan (approved):
Looks great!

PR #1659, opened 2022-11-02
Redis - Add TFVP Support

Zlaticanin:

```release-note
feat: Add TFVP support for the Redis Database Plugin
```

```
$ make testacc TESTARGS='-v -run=TestAccDatabaseSecretBackendConnection_redis'
2022/11/01 16:01:26 [INFO] Using Vault token with the following policies: root
=== RUN   TestAccDatabaseSecretBackendConnection_redis
--- PASS: TestAccDatabaseSecretBackendConnection_redis (1.15s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     (cached)
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 IMPROVEMENTS:
 * Add support for externally managed Group Member IDs to Vault Identity Group: ([#1630](https://github.com/hashicorp/terraform-provider-vault/pull/1630))
 * Support configuring vault version handling: ([#1646](https://github.com/hashicorp/terraform-provider-vault/pull/1646))
+* Add Redis database secrets engine support: ([#1659](https://github.com/hashicorp/terraform-provider-vault/pull/1659))
 
 BUGS:
 * Ensure that namespaced github auth mounts are destroyed: ([#1637](https://github.com/hashicorp/terraform-provider-vault/pull/1637))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 IMPROVEMENTS:
 * Add support for externally managed Group Member IDs to Vault Identity Group: ([#1630](https://github.com/hashicorp/terraform-provider-vault/pull/1630))
 * Support configuring vault version handling: ([#1646](https://github.com/hashicorp/terraform-provider-vault/pull/1646))
-* Add Redis database secrets engine support: ([#1659](https://github.com/hashicorp/terraform-provider-vault/pull/1659))
 
 BUGS:
 * Ensure that namespaced github auth mounts are destroyed: ([#1637](https://github.com/hashicorp/terraform-provider-vault/pull/1637))
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -612,7 +612,8 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
 						Type:     schema.TypeString,
 						Optional: true,
 						Description: "The contents of a PEM-encoded CA cert file " +
-							"to use to verify the Redis server's identity",
+							"to use to verify the Redis server's identity.",
+						Default: false,
 					},
 				},
 			},
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -610,7 +610,8 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
 					"ca_cert": {
 						Type:        schema.TypeString,
 						Optional:    true,
-						Description: "The path to a PEM-encoded CA cert file to use to verify the Redis server's identity",
+						Description: "The contents of a PEM-encoded CA cert file to use to verify the Redis server's identity",
+						Default:     "",
 					},
 				},
 			},
--- a/vault/resource_database_secret_backend_connection.go
+++ b/vault/resource_database_secret_backend_connection.go
@@ -602,16 +602,17 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
 						Default:     false,
 					},
 					"insecure_tls": {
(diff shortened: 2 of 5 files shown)


vault/resource_database_secret_backend_connection.go
```diff
@@ -561,6 +566,57 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
...
+						Description: "Specifies whether to use TLS when connecting to Redis.",
+						Default:     false,
+					},
+					"insecure_tls": {
+						Type:        schema.TypeBool,
+						Optional:    true,
+						Description: "Specifies whether to skip verification of the server certificate when using TLS.",
+						Default:     false,
+					},
+					"ca_cert": {
```
benashz, line 610: We have a few unrelated PRs out where the user would like to be able to pass this in as a string, wonder if it would be worth supporting both forms? Reading a file from a Terraform file is not always supported in a pipeline, and requiring it can make the deployment a little more cumbersome.

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -842,6 +842,45 @@ func TestAccDatabaseSecretBackendConnection_snowflake(t *testing.T) {
...
+	name := acctest.RandomWithPrefix("db")
+
+	resource.Test(t, resource.TestCase{
+		Providers:    testProviders,
+		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy: testAccDatabaseSecretBackendConnectionCheckDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDatabaseSecretBackendConnectionConfig_redis(name, backend, host, username, password),
+				Check: testComposeCheckFuncCommonDatabaseSecretBackend(name, backend, pluginName,
```
benashz, line 864: I wonder if we should include any of the other fields here, e.g. `insecure_tls` ?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 IMPROVEMENTS:
 * Add support for externally managed Group Member IDs to Vault Identity Group: ([#1630](https://github.com/hashicorp/terraform-provider-vault/pull/1630))
 * Support configuring vault version handling: ([#1646](https://github.com/hashicorp/terraform-provider-vault/pull/1646))
+* Add Redis database secrets engine support: ([#1659](https://github.com/hashicorp/terraform-provider-vault/pull/1659))
```
benashz, line 5: No need to update the changelog yet. We will generate it from the PR description when we cutting the 3.11.0 release.

benashz (commented):
Looking good! I had a few minor suggestions/comments.

vault/resource_database_secret_backend_connection.go
Zlaticanin: I will change the description. This is actually doing what you are referring to. We are not passing the path to the actual file, but the contents. I will update the description to say that. Thank you for the suggestion 👍

vault/resource_database_secret_backend_connection_test.go
Zlaticanin: I added `insecure_tls` and `tls` fields. Tested both with `true` and `false` - they pass.

vault/resource_database_secret_backend_connection.go
```diff
@@ -561,6 +566,58 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
...
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "The contents of a PEM-encoded CA cert file to use to verify the Redis server's identity",
+						Default:     "",
```
benashz, line 614: nit: probably don't need a default here.

vault/resource_database_secret_backend_connection.go
```diff
@@ -561,6 +566,58 @@ func getDatabaseSchema(typ schema.ValueType) schemaMap {
...
+					"insecure_tls": {
+						Type:        schema.TypeBool,
+						Optional:    true,
+						Description: "Specifies whether to skip verification of the server certificate when using TLS.",
+						Default:     false,
+					},
+					"ca_cert": {
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "The contents of a PEM-encoded CA cert file to use to verify the Redis server's identity",
```
benashz, line 613: nit: mind wrapping lines at 100 columns?

website/docs/r/database_secret_backend_connection.md
```diff
@@ -167,6 +169,24 @@ Exactly one of the nested blocks of configuration options must be supplied.
...
+
+* `port` - (Required) The default port to connect to if no port is specified as
+  part of the host.
+
+* `tls` - (Optional) Whether to use TLS when connecting to Redis.
+
+* `insecure_tls` - (Optional) Whether to skip verification of the server
+  certificate when using TLS.
+
+* `ca_cert` - (Optional) The path to a PEM-encoded CA cert file to use to verify the Redis server's identity.
```
benashz, line 188: Here it mentions `path`, probably want to update it to be better aligned with the field's description.

website/docs/r/database_secrets_mount.md
```diff
@@ -397,6 +400,24 @@ Supported list of database secrets engines that can be configured:
...
+
+* `port` - (Optional) The default port to connect to if no port is specified as
+  part of the host.
+
+* `tls` - (Optional) Whether to use TLS when connecting to Redis.
+
+* `insecure_tls` - (Optional) Whether to skip verification of the server
+  certificate when using TLS.
+
+* `ca_cert` - (Optional) The path to a PEM-encoded CA cert file to use to verify the Redis server's identity.
```
benashz, line 419: Same comment as in `database_secret_backend_connection.md`

benashz (approved):
A few minor nits still to address, otherwise looks great!

vinay-gopalan (approved):
Looks great!

PR #1160, opened 2021-08-27
Add Basic Constraints attribute to vault_pki_secret_backend_intermediate_cert_request

tomwerneruk:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
Updates `vault_pki_secret_backend_intermediate_cert_request` resource to include the `add_basic_constraints` argument. This adds extensions required when signing intermediate CSRs by Microsoft AD CS Root Authorities.
```

PR #1661, opened 2022-11-03
Update PR "Add Basic Constraints attribute to vault_pki_secret_backend_intermediate_cert_request"

robmonte:
This is a new branch to update the community PR https://github.com/hashicorp/terraform-provider-vault/pull/1160 to resolve merge conflicts and add test coverage. 

```
❯ make testacc TESTARGS='-run=^TestPki -count=1'
TF_ACC=1 go test -run=^TestPki -count=1 -timeout 30m ./...
ok      github.com/hashicorp/terraform-provider-vault/vault     58.566s
```

```release-note
Improvement: `vault_pki_secret_backend_intermediate_cert_request` resource to include the `add_basic_constraints` argument. This adds extensions required when signing intermediate CSRs by Microsoft AD CS Root Authorities.
```

Relates #1160

### DIFF
--- a/vault/resource_pki_secret_backend_intermediate_cert_request.go
+++ b/vault/resource_pki_secret_backend_intermediate_cert_request.go
@@ -185,7 +185,7 @@ func pkiSecretBackendIntermediateCertRequestResource() *schema.Resource {
 			},
 			"add_basic_constraints": {
 				Type:        schema.TypeBool,
-				Description: "Whether to add a Basic Constraints extension with CA: true",
+				Description: "Set 'CA: true' in basic constraints to the CSR to allow signing with ADCS.",
 				ForceNew:    true,
 				Default:     false,
 				Optional:    true,
--- a/vault/resource_pki_secret_backend_intermediate_cert_request.go
+++ b/vault/resource_pki_secret_backend_intermediate_cert_request.go
@@ -183,6 +183,13 @@ func pkiSecretBackendIntermediateCertRequestResource() *schema.Resource {
 				ForceNew:      true,
 				ConflictsWith: []string{"managed_key_name"},
 			},
+			"add_basic_constraints": {
+				Type:        schema.TypeBool,
+				Description: "Whether to add a Basic Constraints extension with CA: true",
+				ForceNew:    true,
+				Default:     false,
+				Optional:    true,
+			},
 		},
 	}
 }
@@ -223,19 +230,20 @@ func pkiSecretBackendIntermediateCertRequestCreate(ctx context.Context, d *schem
 	}
 
 	data := map[string]interface{}{
-		"common_name":          d.Get("common_name").(string),
-		"format":               d.Get("format").(string),
-		"private_key_format":   d.Get("private_key_format").(string),
-		"exclude_cn_from_sans": d.Get("exclude_cn_from_sans").(bool),
-		"ou":                   d.Get("ou").(string),
-		"organization":         d.Get("organization").(string),
-		"country":              d.Get("country").(string),
-		"locality":             d.Get("locality").(string),
-		"province":             d.Get("province").(string),
-		"street_address":       d.Get("street_address").(string),
-		"postal_code":          d.Get("postal_code").(string),
-		"managed_key_name":     d.Get("managed_key_name").(string),
-		"managed_key_id":       d.Get("managed_key_id").(string),
+		"common_name":           d.Get("common_name").(string),
+		"format":                d.Get("format").(string),
+		"private_key_format":    d.Get("private_key_format").(string),
+		"exclude_cn_from_sans":  d.Get("exclude_cn_from_sans").(bool),
+		"ou":                    d.Get("ou").(string),
+		"organization":          d.Get("organization").(string),
+		"country":               d.Get("country").(string),
+		"locality":              d.Get("locality").(string),
+		"province":              d.Get("province").(string),
+		"street_address":        d.Get("street_address").(string),
+		"postal_code":           d.Get("postal_code").(string),
+		"managed_key_name":      d.Get("managed_key_name").(string),
+		"managed_key_id":        d.Get("managed_key_id").(string),
+		"add_basic_constraints": d.Get("add_basic_constraints").(bool),
 	}
 
 	if intermediateType != "kms" {
--- a/vault/resource_pki_secret_backend_intermediate_cert_request.go
+++ b/vault/resource_pki_secret_backend_intermediate_cert_request.go
@@ -184,11 +184,12 @@ func pkiSecretBackendIntermediateCertRequestResource() *schema.Resource {
 				ConflictsWith: []string{"managed_key_name"},
 			},
 			"add_basic_constraints": {
-				Type:        schema.TypeBool,
-				Description: "Set 'CA: true' in basic constraints to the CSR to allow signing with ADCS.",
-				ForceNew:    true,
-				Default:     false,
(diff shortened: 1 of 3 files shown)


vault/resource_pki_secret_backend_intermediate_cert_request.go
```diff
@@ -183,6 +183,13 @@ func pkiSecretBackendIntermediateCertRequestResource() *schema.Resource {
 				ForceNew:      true,
 				ConflictsWith: []string{"managed_key_name"},
 			},
+			"add_basic_constraints": {
+				Type:        schema.TypeBool,
+				Description: "Whether to add a Basic Constraints extension with CA: true",
```
benashz, line 188: nit: the description seems a bit of. I think something like `Set 'CA: true' in the 'Basic Constraints' extension.` is a bit more concise. Also it seem that this is only used as a work-around for AD Certificate Services. May want to include that as well.

vault/resource_pki_secret_backend_intermediate_cert_request.go
```diff
@@ -223,19 +230,20 @@ func pkiSecretBackendIntermediateCertRequestCreate(ctx context.Context, d *schem
...
+		"ou":                    d.Get("ou").(string),
+		"organization":          d.Get("organization").(string),
+		"country":               d.Get("country").(string),
+		"locality":              d.Get("locality").(string),
+		"province":              d.Get("province").(string),
+		"street_address":        d.Get("street_address").(string),
+		"postal_code":           d.Get("postal_code").(string),
+		"managed_key_name":      d.Get("managed_key_name").(string),
+		"managed_key_id":        d.Get("managed_key_id").(string),
+		"add_basic_constraints": d.Get("add_basic_constraints").(bool),
```
benashz, line 246: nit: there is no real need to do the type assert on a value to map[string]interface{}, but since all the other values are like that, I suppose it is fine.

vault/resource_pki_secret_backend_intermediate_cert_request_test.go
```diff
@@ -16,20 +16,26 @@ func TestPkiSecretBackendIntermediateCertRequest_basic(t *testing.T) {
...
-					resource.TestCheckResourceAttr(resourceName, "common_name", "test.my.domain"),
-					resource.TestCheckResourceAttr(resourceName, "uri_sans.#", "1"),
-					resource.TestCheckResourceAttr(resourceName, "uri_sans.0", "spiffe://test.my.domain"),
-				),
+				Config: testPkiSecretBackendIntermediateCertRequestConfig_basic(path, false),
+				Check:  resource.ComposeTestCheckFunc(testCheckFunc...),
+			},
+			{
+				Config: testPkiSecretBackendIntermediateCertRequestConfig_basic(path, true),
+				Check:  resource.ComposeTestCheckFunc(testCheckFunc...),
```
benashz, line 38: It doesn't look like we are checking the value of `add_basic_constraints`

website/docs/r/pki_secret_backend_intermediate_cert_request.html.md
```diff
@@ -82,6 +82,7 @@ The following arguments are supported:
 * `managed_key_id` - (Optional) The ID of the previously configured managed key. This field is
   required if `type` is `kms` and it conflicts with `managed_key_name`
 
+* `add_basic_constraints` - (Optional) Add CA basic constraints to the CSR to allow signing with ADCS
```
benashz, line 85: nit: may want to harmonize this with the schem field's Description.

benashz (commented):
Looking good, only a few minor nits+suggestions.

vault/resource_pki_secret_backend_intermediate_cert_request.go
robmonte: Agreed, and I think the description included on the docs page describes it a lot better.

vault/resource_pki_secret_backend_intermediate_cert_request.go
```diff
@@ -183,6 +183,13 @@ func pkiSecretBackendIntermediateCertRequestResource() *schema.Resource {
 				ForceNew:      true,
 				ConflictsWith: []string{"managed_key_name"},
 			},
+			"add_basic_constraints": {
+				Type:        schema.TypeBool,
+				Description: "Set 'CA: true' in basic constraints to the CSR to allow signing with ADCS.",
```
benashz, line 188: nit: may want to unpack what `ADCS`, also may want mention that is a really specific configuration, similar to what the API docs have.

vault/resource_pki_secret_backend_intermediate_cert_request_test.go
```diff
@@ -16,20 +16,28 @@ func TestPkiSecretBackendIntermediateCertRequest_basic(t *testing.T) {
...
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(resourceName, "backend", path),
-					resource.TestCheckResourceAttr(resourceName, "type", "internal"),
-					resource.TestCheckResourceAttr(resourceName, "common_name", "test.my.domain"),
-					resource.TestCheckResourceAttr(resourceName, "uri_sans.#", "1"),
-					resource.TestCheckResourceAttr(resourceName, "uri_sans.0", "spiffe://test.my.domain"),
-				),
+				Config: testPkiSecretBackendIntermediateCertRequestConfig_basic(path, false),
+				Check: resource.ComposeTestCheckFunc(append(testCheckFunc,
+					resource.TestCheckResourceAttr(resourceName, "add_basic_constraints", "false"))...),
```
benashz, line 35: 👍

benashz (approved):
Looks good! I would just recommend improving the docs a bit, see my related comments.

vault/resource_pki_secret_backend_intermediate_cert_request.go
robmonte: Actually, looking back at the community PR again, the two recent comments mention wanting this for Google CAS and not AD CS. Maybe we shouldn't name a specific one?

In that case, would you have a specific way you want this stated?

benashz: > Actually, looking back at the community PR again, the two recent comments mention wanting this for Google CAS and not AD CS. Maybe we shouldn't name a specific one?

Ideally our docs would closely align with https://developer.hashicorp.com/vault/api-docs/secret/pki#add_basic_constraints. Especially the "workaround" bit.

> 
> In that case, would you have a specific way you want this stated?

https://developer.hashicorp.com/vault/api-docs/secret/pki#add_basic_constraints seems to suggest that the field provides a "workaround" for ADCS. Perhaps we need to update the API docs as well?

website/docs/r/pki_secret_backend_intermediate_cert_request.html.md
```diff
@@ -82,6 +82,7 @@ The following arguments are supported:
 * `managed_key_id` - (Optional) The ID of the previously configured managed key. This field is
   required if `type` is `kms` and it conflicts with `managed_key_name`
 
+* `add_basic_constraints` - (Optional) Adds a Basic Constraints extension with 'CA: true'. Only needed as a workaround in some compatibility scenarios with Active Directory Certificate Services.
```
benashz, line 85: :+1:  - if you could wrap at 100 columns that would be great.

benashz (approved):
just one minor nit then :+1:

benashz (approved):

PR #1239, opened 2021-11-29
feat: kv-v2 generic secret metadata

vfoucault:
Added support for kv-v2 secret's metadata:

* cas_required

* custom_metadata

* delete_version_after

* max_versions

Thought `cas_required` can be set (via the http api `options` block), none of the above settings could be set with the regular resources.

It might become handy to read/write the secret metadata, to set per secret these settings.

usage example:

_resource_ 
```hcl
resource "vault_generic_secret" "foo" {
  path          = "secrets/foo"
  data_json = jsonencode({"blah": "diblah"})
}

resource "vault_generic_secret_metadata" "foo" {
  path = vault_generic_secret.foo.path
  custom_metadata = {
    blah = "diblah"
    foo  = "bar"
  }
  cas_required         = true
  delete_version_after = "730h"
  max_versions         = 17
}
```
_data source_

```hcl
data "vault_generic_secret_metadata" "foo" {
  path = "secrets/foo"
}

output "foo_metadata" {
  value = vault_generic_secret_metadata.foo
}

output "foo_custom_metadata" "foo" {
  value = vault_generic_secret_metadata.foo.custom_metadata
}
```

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Relates to #491 to #1156

Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md):

```release-note
* **New Resource** `vault_generic_secret_metadata`: Configure Vault's KVv2 secret metadata
* **New Data Source** `vault_generic_secret_metadata`: read Vault's KVv2 secret metadata
```

Output from acceptance testing:

```
$ ==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./...) -v -run=TestResourceGenericSecretMetadata -run=genericSecretMetadataDataSource -timeout 20m
?       github.com/hashicorp/terraform-provider-vault   [no test files]
?       github.com/hashicorp/terraform-provider-vault/cmd/coverage      [no test files]
?       github.com/hashicorp/terraform-provider-vault/cmd/generate      [no test files]
testing: warning: no tests to run
PASS
ok      github.com/hashicorp/terraform-provider-vault/codegen   0.359s [no tests to run]
[...]

PR #1687, opened 2022-12-02
Add support for a KV V2 Secret Metadata resource

vinay-gopalan:
Adds support for adding custom metadata to a KV V2 secret resource.

```
$ make testacc TESTARGS='-run=TestAccKVSecretV2'
=== RUN   TestAccKVSecretV2
--- PASS: TestAccKVSecretV2 (2.35s)
PASS
...
```


vinay-gopalan:
> @vinay-gopalan Seems like this is the same as #1239

@joao-oneill-unbabel the code in this PR slightly differs from the one you mentioned, in that this resource is compatible with the latest dedicated KV V2 Secret engine support that was added to the TFVP in v3.7.0 as opposed to the legacy resource `vault_generic_secret` 😄

website/docs/r/kv_secret_metadata_v2.html.md
```diff
@@ -0,0 +1,99 @@
...
+  data_json                  = jsonencode(
+  {
+    zip       = "zap",
+    foo       = "bar",
+  }
+  )
+}
+
+resource "vault_kv_secret_metadata_v2" "example" {
+  mount                = vault_mount.kvv2.path
```
tvoran, line 39: Similar question here:

```suggestion
  mount                = vault_mount.example.path
```

website/docs/r/kv_secret_v2.html.md
```diff
@@ -69,6 +69,11 @@ The following arguments are supported:
 * `data_json` - (Required) JSON-encoded string that will be
   written as the secret data at the given path.
 
+* `custom_metadata` - (Optional) A nested block containing configuration options for the
```
benashz, line 72: Isn't this `map[string]string` ? Not sure what is meant by "nested block".

website/docs/r/kv_secret_backend_v2.html.md
```diff
@@ -65,7 +65,8 @@ No additional attributes are exported by this resource.
 
 ## Import
 
-The KV-V2 secret backend can be imported using the `path`, e.g.
+The KV-V2 secret backend can be imported using its unique ID,
+the `${mount}/config`, e.g.
 
 ```
 $ terraform import vault_kv_secret_backend_v2.config kvv2/config
```
benashz, line 72: nit: may want to rename the resource `example`, the repeating `config` makes this example a little confusing.

vault/resource_kv_secret_v2.go
```diff
@@ -100,6 +102,43 @@ func kvSecretV2Resource(name string) *schema.Resource {
 				Default:     false,
 				Description: "If set to true, permanently deletes all versions for the specified key.",
 			},
+
+			consts.FieldCustomMetadata: {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Custom metadata to be set for the secret",
```
benashz, line 109:
```suggestion
				Description: "Custom metadata to be set for the secret.",
```

vault/resource_kv_secret_v2.go
```diff
@@ -100,6 +102,43 @@ func kvSecretV2Resource(name string) *schema.Resource {
...
+							Optional: true,
+							Description: "If set, specifies the length of time before " +
+								"a version is deleted",
+						},
+						consts.FieldData: {
+							Type:     schema.TypeMap,
+							Optional: true,
+							Computed: true,
+							Description: "A map of arbitrary string to string valued " +
+								"user-provided metadata meant to describe the secret",
```
benashz, line 136:
```suggestion
								"user-provided metadata meant to describe the secret.",
```

vault/resource_kv_secret_v2.go
```diff
@@ -100,6 +102,43 @@ func kvSecretV2Resource(name string) *schema.Resource {
...
+							Optional: true,
+							Computed: true,
+							Description: "If true, all keys will require the cas " +
+								"parameter to be set on all write requests.",
+						},
+						consts.FieldDeleteVersionAfter: {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Description: "If set, specifies the length of time before " +
+								"a version is deleted",
```
benashz, line 129:
```suggestion
								"a version is deleted.",
```

vault/resource_kv_secret_v2.go
```diff
@@ -184,13 +261,72 @@ func kvSecretV2Read(_ context.Context, d *schema.ResourceData, meta interface{})
...
 			}
 		}
+
 	}
 
 	return nil
 }
 
+func readKVV2Metadata(d *schema.ResourceData, client *api.Client) (map[string]interface{}, error) {
+	mount := d.Get(consts.FieldMount).(string)
```
benashz, line 286: probably want to bail out early if `d.Id()` is `""` ? An alternative to that, would be to change the signature to take the `mount` and `name` instead of `d`.

vault/resource_kv_secret_v2.go
```diff
@@ -184,13 +261,72 @@ func kvSecretV2Read(_ context.Context, d *schema.ResourceData, meta interface{})
...
 				}
+
+				// Read & Set custom metadata
+				if _, ok := v[consts.FieldCustomMetadata]; ok {
+					cm, err := readKVV2Metadata(d, client)
+					if err != nil {
+						return diag.FromErr(err)
+					}
+
+					customMetadata := []interface{}{cm}
```
benashz, line 272: nit: variable is only used once, you might want to inline the value in the call to `d.Set`, below.

vault/resource_kv_secret_v2.go
```diff
@@ -100,6 +102,43 @@ func kvSecretV2Resource(name string) *schema.Resource {
...
+						consts.FieldDeleteVersionAfter: {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Description: "If set, specifies the length of time before " +
+								"a version is deleted",
+						},
+						consts.FieldData: {
+							Type:     schema.TypeMap,
+							Optional: true,
+							Computed: true,
```
benashz, line 134: Should this ever be a computed value? Presumably the default is empty.

vault/resource_kv_secret_v2.go
```diff
@@ -108,6 +147,33 @@ func getKVV2Path(mount, name, prefix string) string {
...
 }
 
+func getKVV2MetadataPath(mount, name string) string {
+	return fmt.Sprintf("%s/metadata/%s", mount, name)
+}
+
+func getCustomMetadata(d *schema.ResourceData) map[string]interface{} {
+	data := map[string]interface{}{}
+
+	metadataFields := []string{
```
benashz, line 157: nit: variable name is a bit verbose, relative to its function. May want to rename to `fields`

vault/resource_kv_secret_v2.go
```diff
@@ -108,6 +147,33 @@ func getKVV2Path(mount, name, prefix string) string {
...
+		consts.FieldMaxVersions,
+		consts.FieldCASRequired,
+		consts.FieldDeleteVersionAfter,
+		consts.FieldData,
+	}
+	fieldPrefix := fmt.Sprintf("%s.0", consts.FieldCustomMetadata)
+	for _, k := range metadataFields {
+		fieldKey := fmt.Sprintf("%s.%s", fieldPrefix, k)
+		vaultStateKey := k
```
benashz, line 166: nit: you might be able to drop this intermediate variable in favor of assigning directly to `k` on line `168` below.

vault/resource_kv_secret_v2.go
```diff
@@ -184,13 +261,72 @@ func kvSecretV2Read(_ context.Context, d *schema.ResourceData, meta interface{})
...
+	if err != nil {
+		return nil, err
+	}
+
+	if resp == nil {
+		log.Printf("[DEBUG] no metadata found for secret")
+		return nil, nil
+	}
+
+	metadataFields := map[string]string{
```
benashz, line 302: This map may be useful as a global that can be used to replace the duplicate remapping logic in `getCustomMetadata()`, and any where else you are accessing these fields.

vault/resource_kv_secret_v2_test.go
```diff
@@ -55,17 +60,23 @@ func testKVSecretV2Config(mount, name string) string {
...
+  name                = "%s"
+  delete_all_versions = true
+  data_json = jsonencode(
+    {
+      zip  = "zap",
+      foo  = "bar",
+      flag = false
+    }
   )
+  custom_metadata {
```
benashz, line 73: nit: might be worth preserving the original test config as well i.e. testing when without and without the `custom_metadata` being set.

benashz (commented):
Looking good. Have some initial feedback for you.

vault/resource_kv_secret_v2.go
vinay-gopalan: I wanted to make sure we are able to import the custom metadata and set it to the TF state for the KV secret, specifically in the cases that the secret metadata was created outside of TF and via Vault. That was my thought process around setting this field to also be possibly computed. Is that overkill/unnecessary?

vault/resource_kv_secret_v2.go
vinay-gopalan: Ah good call, thanks for the suggestion! Updated the code and removed the duplication

vault/resource_kv_secret_v2_test.go
vinay-gopalan: Good call, added an extra test step

vault/resource_kv_secret_v2.go
vinay-gopalan: I removed the `Computed` from the individual nested fields, but still set `Computed: true` for the parent `custom_metadata` field

website/docs/r/kv_secret_v2.html.md
vinay-gopalan: Added docs to clarify the configuration structure in e5eef1f

vault/resource_kv_secret_v2.go
```diff
@@ -5,14 +5,24 @@ import (
...
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/vault/api"
 
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 	"github.com/hashicorp/terraform-provider-vault/internal/provider"
 )
 
+var metadataFields = map[string]string{
```
benashz, line 19: Probably want to include `KV` in the name, since the `vault` package is quite large.

vault/resource_kv_secret_v2.go
```diff
@@ -184,13 +254,62 @@ func kvSecretV2Read(_ context.Context, d *schema.ResourceData, meta interface{})
...
+		if val, ok := resp.Data[vaultKey]; ok {
+			// the delete_version_after field is written to
+			// Vault as an integer but is returned as a string
+			// of the format "3h12m10s"
+			if vaultKey == consts.FieldDeleteVersionAfter {
+				t, err := time.ParseDuration(val.(string))
+				if err != nil {
+					return nil, fmt.Errorf("error parsing duration, err=%s", err)
+				}
+				data[tfKey] = t.Seconds()
```
benashz, line 303: nit: you can probably set `val = t.Seconds()` here, then you can drop the `else{}`. Makes things a bit easier to read.

website/docs/r/kv_secret_v2.html.md
```diff
@@ -69,13 +76,30 @@ The following arguments are supported:
...
+
+* `cas_required` - (Optional) If true, all keys will require the cas
+  parameter to be set on all write requests.
+
+* `delete_version_after` - (Optional) If set, specifies the length of time before
+  a version is deleted. Accepts duration in integer seconds.
+
+* `data` - (Optional) A map of arbitrary string to string valued user-provided
+  metadata meant to describe the secret.
+
```
benashz, line 102: nit: this description is a bit hard to read. I'd suggest simplifying it:

```suggestion
* `data` - (Optional) A string to string map describing the secret.
```

benashz (approved):
A few minor nits to address, then :+1:

vinay-gopalan:
1 acceptance test suite is failing due to GH request rate limiting; proceeding to merge since the work done here does not affect the failing tests

PR #1693, opened 2022-12-07
Ensure both name and mount accessor match before returning

mbillow:
### Description

Currently, this logic is only checking the name of the target entity alias against the list of configured aliases returned by the LookupEntity function. Functionally, this doesn't matter to Terraform because all it cares about is if there is a collision which is represented by anything being returned from this function. 

When a collision is detected though, what is returned from this logic is shown to the user. Because we are only checking the name and not mount accessor, we could potentially return the wrong alias if an entity has the same alias name on multiple auth backends.

#### Example

Error from Terraform
``` 
╷
│ Error: entity alias "demo--qa/default" already exists for mount accessor "auth_kubernetes_bfdaf8e8", id="922e1079-6566-939b-7d68-bb154773d078"
│ 
│   with vault_identity_entity_alias.service_account_entity_alias["demo--qa.default"],
│   on service_accounts.tf line 5, in resource "vault_identity_entity_alias" "service_account_entity_alias":
│    5: resource "vault_identity_entity_alias" "service_account_entity_alias" {
│ 
│ In the case where this error occurred during the creation of more than one alias, it may be necessary to assign a unique alias name to each of affected resources and then rerun the apply. After a successful apply the desired original alias names can then be reassigned
╵
```

Note that it returns the alias ID `922e1079-6566-939b-7d68-bb154773d078`. Which when you look that up in Vault:
```
~
➜ vault read identity/entity-alias/id/922e1079-6566-939b-7d68-bb154773d078
Key                          Value
---                          -----
canonical_id                 0d34a3c1-8818-825d-7d27-87af262634c4
creation_time                2022-11-15T17:19:21.514285809Z
custom_metadata              <nil>
id                           922e1079-6566-939b-7d68-bb154773d078
last_update_time             2022-11-15T17:19:21.514285809Z
local                        false
merged_from_canonical_ids    <nil>
metadata                     <nil>
mount_accessor               auth_kubernetes_e76323c8
mount_path                   auth/k8s-cluster2/
mount_type                   kubernetes
name                         demo--qa/default
namespace_id                 root
```

That is an alias attached to `auth_kubernetes_e76323c8` not `auth_kubernetes_bfdaf8e8`.

### Community Note

```release-note
 - Fixes bug where wrong entity alias ID could be returned during duplicate alias preflight check.
```

### DIFF
--- a/internal/identity/entity/entity.go
+++ b/internal/identity/entity/entity.go
@@ -151,7 +151,7 @@ func LookupEntityAlias(client *api.Client, params *FindAliasParams) (*Alias, err
 				return nil, err
 			}
 
-			if a.Name == params.Name {
+			if a.Name == params.Name && a.MountAccessor == params.MountAccessor {
 				return &a, nil
 			}
 		}


benashz:
Thanks for the submission @mbillow. Would you mind updating/extending the unit tests to test your proposed change? You should find them in `internal/identity/entity/entity_test.go`.

Ben

benashz:
We are going to take this PR over. Thanks again for your submission @mbillow!

benashz:
Moved to #1708

PR #1710, opened 2022-12-22
Add support for importing the PKI CRL config

benashz:
- Adapt tests to support older vault versions
- Make some API data and schema set operations simpler
- Fix broken build

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```


vault/resource_pki_secret_backend_crl_config_test.go
```diff
@@ -7,125 +7,149 @@ import (
...
+			checks = baseChecks
+		}
+		return resource.ComposeAggregateTestCheckFunc(checks...)(state)
+	}
+}
 
+func TestPkiSecretBackendCrlConfig_lte_111(t *testing.T) {
+	setupCRLConfigTest(t, func() {
+		testutil.TestAccPreCheck(t)
+		SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion112)
```
vinay-gopalan, line 57: Should this be `SkipIfAPIVersionGTE`? Looks like we want this test suite to run for Vault versions 1.11 and lower, is that correct?

vinay-gopalan (approved):
PR is looking great! Nice style precedent that we should probably use going forward for setting up tests wherever we can 😄 

Had one potential fix for one of the test pre-checks, but looks good to me otherwise

vault/resource_pki_secret_backend_crl_config_test.go
benashz: Good catch! Should be fixed in 56451ca115b82780d3cfc84a89650025317fff55

PR #1712, opened 2022-12-30
Add resource for new GCP impersonated account endpoints

mdgreenfield:
### Community Note

Relates To: https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/129

```release-note
Adds `vault_gcp_secret_impersonated_account` resource
```

Note, included the `-v` with `go test`.
```
$ TESTARGS="--run GCPSecretImpersonatedAccount" make testacc
TF_ACC=1 go test --run GCPSecretImpersonatedAccount -v -timeout 30m ./...
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
2022/12/29 17:00:18 [INFO] Using Vault token with the following policies: root
=== RUN   TestGCPSecretImpersonatedAccount
--- PASS: TestGCPSecretImpersonatedAccount (1.62s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     2.105s
```

### DIFF
--- a/vault/resource_gcp_secret_impersonated_account.go
+++ b/vault/resource_gcp_secret_impersonated_account.go
@@ -1,12 +1,15 @@
 package vault
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 	"github.com/hashicorp/terraform-provider-vault/internal/provider"
 )
 
@@ -17,17 +20,16 @@ var (
 
 func gcpSecretImpersonatedAccountResource() *schema.Resource {
 	return &schema.Resource{
-		Create: gcpSecretImpersonatedAccountCreate,
-		Read:   gcpSecretImpersonatedAccountRead,
-		Update: gcpSecretImpersonatedAccountUpdate,
-		Delete: gcpSecretImpersonatedAccountDelete,
-		Exists: gcpSecretImpersonatedAccountExists,
+		CreateContext: gcpSecretImpersonatedAccountCreate,
+		ReadContext:   ReadContextWrapper(gcpSecretImpersonatedAccountRead),
+		UpdateContext: gcpSecretImpersonatedAccountUpdate,
+		DeleteContext: gcpSecretImpersonatedAccountDelete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			StateContext: schema.ImportStatePassthroughContext,
 		},
 
 		Schema: map[string]*schema.Schema{
-			"backend": {
+			consts.FieldBackend: {
 				Type:        schema.TypeString,
 				Required:    true,
 				Description: "Path where the GCP secrets engine is mounted.",
@@ -66,13 +68,13 @@ func gcpSecretImpersonatedAccountResource() *schema.Resource {
 	}
 }
 
-func gcpSecretImpersonatedAccountCreate(d *schema.ResourceData, meta interface{}) error {
+func gcpSecretImpersonatedAccountCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, err := provider.GetClient(d, meta)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
-	backend := d.Get("backend").(string)
+	backend := d.Get(consts.FieldBackend).(string)
 	impersonatedAccount := d.Get("impersonated_account").(string)
 
 	path := gcpSecretImpersonatedAccountPath(backend, impersonatedAccount)
@@ -81,39 +83,39 @@ func gcpSecretImpersonatedAccountCreate(d *schema.ResourceData, meta interface{}
 
 	data := map[string]interface{}{}
 	gcpSecretImpersonatedAccountUpdateFields(d, data)
-	d.SetId(path)
 	_, err = client.Logical().Write(path, data)
 	if err != nil {
 		d.SetId("")
-		return fmt.Errorf("error writing GCP Secrets backend impersonated account %q: %s", path, err)
+		return diag.Errorf("error writing GCP Secrets backend impersonated account %q: %s", path, err)
 	}
+	d.SetId(path)
 	log.Printf("[DEBUG] Wrote GCP Secrets backend impersonated account %q", path)
 
-	return gcpSecretImpersonatedAccountRead(d, meta)
+	return gcpSecretImpersonatedAccountRead(ctx, d, meta)
 }
 
-func gcpSecretImpersonatedAccountRead(d *schema.ResourceData, meta interface{}) error {
+func gcpSecretImpersonatedAccountRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, err := provider.GetClient(d, meta)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	path := d.Id()
 
 	backend, err := gcpSecretImpersonatedAccountBackendFromPath(path)
 	if err != nil {
-		return fmt.Errorf("invalid path %q for GCP secrets backend impersonated account: %s", path, err)
+		return diag.Errorf("invalid path %q for GCP secrets backend impersonated account: %s", path, err)
 	}
 
 	impersonatedAccount, err := gcpSecretImpersonatedAccountNameFromPath(path)
 	if err != nil {
-		return fmt.Errorf("invalid path %q for GCP Secrets backend impersonated account: %s", path, err)
+		return diag.Errorf("invalid path %q for GCP Secrets backend impersonated account: %s", path, err)
 	}
 
 	log.Printf("[DEBUG] Reading GCP Secrets backend impersonated account %q", path)
 	resp, err := client.Logical().Read(path)
 	if err != nil {
-		return fmt.Errorf("error reading GCP Secrets backend impersonated account %q: %s", path, err)
+		return diag.Errorf("error reading GCP Secrets backend impersonated account %q: %s", path, err)
 	}
 
 	log.Printf("[DEBUG] Read GCP Secrets backend impersonated account %q", path)
@@ -123,18 +125,18 @@ func gcpSecretImpersonatedAccountRead(d *schema.ResourceData, meta interface{}) 
 		return nil
 	}
 
-	if err := d.Set("backend", backend); err != nil {
-		return err
+	if err := d.Set(consts.FieldBackend, backend); err != nil {
+		return diag.FromErr(err)
 	}
 	if err := d.Set("impersonated_account", impersonatedAccount); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	for _, k := range []string{"token_scopes", "service_account_email", "service_account_project"} {
 		v, ok := resp.Data[k]
 		if ok {
 			if err := d.Set(k, v); err != nil {
-				return fmt.Errorf("error reading %s for GCP Secrets backend impersonated account %q: %q", k, path, err)
+				return diag.Errorf("error reading %s for GCP Secrets backend impersonated account %q: %q", k, path, err)
 			}
 		}
 	}
@@ -142,10 +144,10 @@ func gcpSecretImpersonatedAccountRead(d *schema.ResourceData, meta interface{}) 
 	return nil
 }
 
-func gcpSecretImpersonatedAccountUpdate(d *schema.ResourceData, meta interface{}) error {
+func gcpSecretImpersonatedAccountUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, err := provider.GetClient(d, meta)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	path := d.Id()
@@ -157,17 +159,17 @@ func gcpSecretImpersonatedAccountUpdate(d *schema.ResourceData, meta interface{}
 
 	_, err = client.Logical().Write(path, data)
 	if err != nil {
-		return fmt.Errorf("error updating GCP Secrets backend impersonated account %q: %s", path, err)
+		return diag.Errorf("error updating GCP Secrets backend impersonated account %q: %s", path, err)
 	}
 	log.Printf("[DEBUG] Updated GCP Secrets backend impersonated account %q", path)
 
-	return gcpSecretImpersonatedAccountRead(d, meta)
+	return gcpSecretImpersonatedAccountRead(ctx, d, meta)
 }
 
-func gcpSecretImpersonatedAccountDelete(d *schema.ResourceData, meta interface{}) error {
+func gcpSecretImpersonatedAccountDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, err := provider.GetClient(d, meta)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	path := d.Id()
@@ -175,7 +177,7 @@ func gcpSecretImpersonatedAccountDelete(d *schema.ResourceData, meta interface{}
 	log.Printf("[DEBUG] Deleting GCP secrets backend impersonated account %q", path)
 	_, err = client.Logical().Delete(path)
 	if err != nil {
-		return fmt.Errorf("error deleting GCP secrets backend impersonated account %q", path)
+		return diag.Errorf("error deleting GCP secrets backend impersonated account %q", path)
 	}
 	log.Printf("[DEBUG] Deleted GCP secrets backend impersonated account %q", path)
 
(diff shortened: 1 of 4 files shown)


robmonte (commented):
I just have 2 nits at the moment regarding the test cases. 

I haven't had a whole lot of time to fully dig into the implementation yet but my first pass over it looks good, and it matches up well with the static account resource as expected.

One question; I wasn't successful with running the test cases. I noticed the TF test plan at the bottom of the file doesn't create a backend resource, which could explain why I get `no handler for route`. But it looks as though you didn't have this problem, was there some other step you were performing to get the tests to function?

mdgreenfield:
@robmonte, thanks for looking over the code. I'll take care of the two changes you highlighted in the morning.

Regarding the `no handler for route` error you're seeing. I suspect that is because the impersonated account support in the GCP secrets plugin has not yet been tagged and the Vault dependency hasn't been updated. For my testing, I manually built the plugin and enabled it prior to running my tests.

website/docs/r/gcp_secret_impersonated_account.html.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `backend` - (Required, Forces new resource) Path where the GCP Secrets Engine is mounted
+
+* `impersonated_account` - (Required, Forces new resource) Name of the Impersonated Account to create
+
+* `service_account_email` - (Required, Forces new resource) Email of the GCP service account to manage.
```
robmonte, line 45: Nit/Question: Would saying "Email of the GCP service account to impersonate" be more descriptive? This is more of a question on my part as I'm not a consistent user of GCP, so maybe "manage" is the standard term here and is actually better.

robmonte (commented):
@mdgreenfield Sorry for the delay, I have been out of PTO. I was using an updated Vault branch as the feature worked when running your provider branch manually, as only the test case failed. I found one more issue that I didn't notice at first, plus the original thing I commented:

[This line](https://github.com/hashicorp/terraform-provider-vault/pull/1712/files#diff-e0a54876c5900d3f4a32326b871bf66f7ee56452a14d3fa8d5f8053516a03ae9R185) seems to be mistakenly commenting out the expected value and you set "gcp" manually which I assume was an accidental leftover from manual testing you might've run. Additionally, what I meant about the missing resource backend was that there's no "vault_gcp_secret_backend" defined in the test plan, which is referenced as the backend [on the same line](https://github.com/hashicorp/terraform-provider-vault/pull/1712/files#diff-e0a54876c5900d3f4a32326b871bf66f7ee56452a14d3fa8d5f8053516a03ae9R185) when uncommented.

I am including suggested fixes in the comments of this review for those things above for reference, along with some other callouts I had. As for functionality, I did not find any issues nor have comments on implementation details.

Please let me know if you have any questions or feedback!

vault/resource_gcp_secret_impersonated_account.go
```diff
@@ -0,0 +1,235 @@
...
+func gcpSecretImpersonatedAccountResource() *schema.Resource {
+	return &schema.Resource{
+		Create: gcpSecretImpersonatedAccountCreate,
+		Read:   gcpSecretImpersonatedAccountRead,
+		Update: gcpSecretImpersonatedAccountUpdate,
+		Delete: gcpSecretImpersonatedAccountDelete,
+		Exists: gcpSecretImpersonatedAccountExists,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
```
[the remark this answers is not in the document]
vinay-gopalan: Thanks for the question @robmonte! Yes, the comments above regarding the deprecation of `Create`, `Read` etc. are accurate, and we prefer using the `Context` functions over the deprecated operations. 

As for the wrapper methods, here's the documentation for these functions:

- `ReadContextWrapper` provides common read operations to the wrapped `schema.ReadContextFunc`. This wrapper method is used on all `ReadContext` methods to set some common schema fields to the TF state (for eg., each resource has a `namespace` field that needs to be read from Vault and set to the TF state). [Here is an example](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kubernetes_secret_backend.go#L26) of it being used in the codebase.

- `MountCreateContextWrapper` performs a minimum version requirement check prior to the wrapped `schema.CreateContextFunc.` This wrapper method should be wrapped on the `CreateContext` method if there is a minimum version check required for this feature. Since this is a `Vault v1.13.0` feature, I'm guessing we should add a minimum version check with Vault 1.13. This would look [something like this](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kubernetes_secret_backend.go#L25), and would require us to add a `provider.VaultVersion113` to the codebase as [done here](https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/provider/meta.go#L36)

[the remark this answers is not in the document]
vinay-gopalan: `MustAddMountMigrationSchema` shouldn't be needed in this case, since it deals with adding specific fields to the schema in case the resource supports mount migration (only resources that actually mount a secret backend need these schema fields, so we should be good in this case)

vault/resource_gcp_secret_impersonated_account.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	log.Printf("[DEBUG] Read GCP Secrets backend impersonated account %q", path)
+	if resp == nil {
+		log.Printf("[WARN] GCP Secrets backend impersonated account %q not found, removing from state", path)
+		d.SetId("")
+		return nil
+	}
+
+	if err := d.Set("backend", backend); err != nil {
```
vinay-gopalan, line 126: similar comment around constant usage :)

vault/resource_gcp_secret_impersonated_account.go
```diff
@@ -0,0 +1,235 @@
...
+	if v, ok := d.GetOk("service_account_email"); ok {
+		data["service_account_email"] = v.(string)
+	}
+
+	if v, ok := d.GetOk("token_scopes"); ok {
+		data["token_scopes"] = v.(*schema.Set).List()
+	}
+}
+
+func gcpSecretImpersonatedAccountExists(d *schema.ResourceData, meta interface{}) (bool, error) {
```
vinay-gopalan, line 195: related to comment above, this can be removed

vault/resource_gcp_secret_impersonated_account_test.go
```diff
@@ -0,0 +1,191 @@
...
+	noBindings := testGCPSecretImpersonatedAccount_accessToken(backend, impersonatedAccount, credentials, serviceAccountEmail)
+
+	resourceName := "vault_gcp_secret_impersonated_account.test"
+	resource.Test(t, resource.TestCase{
+		Providers:    testProviders,
+		PreCheck:     func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy: testGCPSecretImpersonatedAccountDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: noBindings,
```
vinay-gopalan, line 43: We can inline this
```suggestion
				Config: testGCPSecretImpersonatedAccount_accessToken(backend, impersonatedAccount, credentials, serviceAccountEmail),
```

vault/resource_gcp_secret_impersonated_account.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func gcpSecretImpersonatedAccountUpdateFields(d *schema.ResourceData, data map[string]interface{}) {
+	if v, ok := d.GetOk("service_account_email"); ok {
+		data["service_account_email"] = v.(string)
+	}
+
+	if v, ok := d.GetOk("token_scopes"); ok {
+		data["token_scopes"] = v.(*schema.Set).List()
+	}
```
vinay-gopalan, line 192: We should also read `service_account_project` from the terraform config; currently that field is never being written to Vault

vault/resource_gcp_secret_impersonated_account_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testGCPSecretImpersonatedAccount_attrs(backend, impersonatedAccount),
```
vinay-gopalan, line 45: This is an older style of writing tests in the codebase. For new resource additions, we prefer to have a test structure  that is something [along the lines of this](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kv_secret_v2_test.go#L24-L64)

As shown in the test above, the complete loop of testing contains an initial test step, an update test step, and finally an import test step. The first two test cases confirm that values are obtained from the TF config as expected (`.tf` file). The Import test step verifies that the expected values were written to Vault, and the values that are read from Vault are what we had initially obtained from the TF config. 

In this case, the checks you have from L46-L51 are accurate, but we can remove L45 entirely, since this check adds a lot of repeated code to the codebase and we are trying to move away from this style for new resources.
```suggestion
```

vault/resource_gcp_secret_impersonated_account_test.go
```diff
@@ -0,0 +1,191 @@
...
+				ResourceName:            "vault_gcp_secret_impersonated_account.test",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{},
+			},
+		},
+	})
+}
+
+func testGCPSecretImpersonatedAccount_attrs(backend, impersonatedAccount string) resource.TestCheckFunc {
```
vinay-gopalan, line 64: as described in a comment above, we can remove this function as it is no longer needed in our testing loop

vault/resource_gcp_secret_impersonated_account.go
```diff
@@ -0,0 +1,235 @@
...
+	gcpSecretImpersonatedAccountNameFromPathRegex    = regexp.MustCompile("^.+/impersonated-account/(.+)$")
+)
+
+func gcpSecretImpersonatedAccountResource() *schema.Resource {
+	return &schema.Resource{
+		Create: gcpSecretImpersonatedAccountCreate,
+		Read:   gcpSecretImpersonatedAccountRead,
+		Update: gcpSecretImpersonatedAccountUpdate,
+		Delete: gcpSecretImpersonatedAccountDelete,
+		Exists: gcpSecretImpersonatedAccountExists,
```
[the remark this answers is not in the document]
vinay-gopalan:
```suggestion

```

vault/resource_gcp_secret_impersonated_account_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+	"github.com/hashicorp/terraform-provider-vault/testutil"
+	"golang.org/x/oauth2/google"
+)
+
+// This test requires that you pass credentials for a user or service account having the IAM rights
+// listed at https://www.vaultproject.io/docs/secrets/gcp/index.html for the project you are testing
+// on. The credentials must also allow setting IAM permissions on the project being tested.
+func TestGCPSecretImpersonatedAccount(t *testing.T) {
```
vinay-gopalan, line 22: If this is a Vault 1.13 feature, this test will fail until Vault 1.13 is released and we can run tests against it in CI; in order to not break CI, we can add this line at the start of the test function to skip this test as it is meant to be run on an upcoming version of Vault.

```suggestion
func TestGCPSecretImpersonatedAccount(t *testing.T) {
    testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)
```

vinay-gopalan (commented):
Looking good! Thank you for working on this feature 😄 
Left some comments, happy to answer any further questions!

mdgreenfield:
No problem @robmonte, I was also on leave but I'll look at getting these comments addressed.

website/docs/r/gcp_secret_impersonated_account.html.md
mdgreenfield: Switching from "manage" to "impersonate" makes sense to me.

vault/resource_gcp_secret_impersonated_account.go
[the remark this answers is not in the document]
mdgreenfield: Nice. I did not know this.

vault/resource_gcp_secret_impersonated_account_test.go
mdgreenfield: Neat

mdgreenfield:
> [This line](https://github.com/hashicorp/terraform-provider-vault/pull/1712/files#diff-e0a54876c5900d3f4a32326b871bf66f7ee56452a14d3fa8d5f8053516a03ae9R185) seems to be mistakenly commenting out the expected value and you set "gcp" manually which I assume was an accidental leftover from manual testing you might've run. Additionally, what I meant about the missing resource backend was that there's no "vault_gcp_secret_backend" defined in the test plan, which is referenced as the backend [on the same line](https://github.com/hashicorp/terraform-provider-vault/pull/1712/files#diff-e0a54876c5900d3f4a32326b871bf66f7ee56452a14d3fa8d5f8053516a03ae9R185) when uncommented.

Yeah, that was left over from some manual testing. I added `testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)` per @vinay-gopalan's note. And after building Vault locally with the updated `vault-plugin-secrets-gcp` dependency, everything  looks much better now.

However, I am seeing this error which I wasn't seeing before:

```
=== RUN   TestGCPSecretImpersonatedAccount
    resource_gcp_secret_impersonated_account_test.go:34: Step 2/3 error: Error running apply: exit status 1

        Error: error updating GCP Secrets backend impersonated account "tf-test-gcp-3920713680288013531/impersonated-account/tf-test-1070837978685140673": Error making API request.

        URL: PUT http://127.0.0.1:8200/v1/tf-test-gcp-3920713680288013531/impersonated-account/tf-test-1070837978685140673
        Code: 404. Errors:

        * no handler for route "tf-test-gcp-3920713680288013531/impersonated-account/tf-test-1070837978685140673". route entry not found.

          with vault_gcp_secret_impersonated_account.test,
          on terraform_plugin_test.tf line 2, in resource "vault_gcp_secret_impersonated_account" "test":
           2: resource "vault_gcp_secret_impersonated_account" "test" {

    testing_new.go:87: Error running post-test destroy, there may be dangling resources: exit status 1

        Error: error deleting GCP secrets backend impersonated account "tf-test-gcp-3920713680288013531/impersonated-account/tf-test-1070837978685140673"
```

And in the Vault server audit logs I can clearly see where it looks like the impersonated account "tf-test-gcp-3920713680288013531/impersonated-account/tf-test-1070837978685140673" is being deleted before the "update" `resource.TestStep` runs. Does something jump out at you that I'm doing wrong? I suspect there might be something in the test configuration that I'm missing.

mdgreenfield:
Ah, I see my issue. The update Terraform block does not include the backend and so the mount is removed.

mdgreenfield:
One last error that I'll need to debug tomorrow.

```
--- FAIL: TestGCPSecretImpersonatedAccount (2.46s)
    resource_gcp_secret_impersonated_account_test.go:34: Step 2/3 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

          # vault_gcp_secret_impersonated_account.test will be updated in-place
          ~ resource "vault_gcp_secret_impersonated_account" "test" {
                id                      = "tf-test-gcp-5732402503560457735/impersonated-account/tf-test-1742435248193293619"
              ~ token_scopes            = [
                  + "https://www.googleapis.com/auth/cloud-platform.read-only",
                    # (1 unchanged element hidden)
                ]
                # (4 unchanged attributes hidden)
            }

```

mdgreenfield:
The issue above is fixed in https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/165

robmonte:
Great bug find! Pretty neat the provider helped uncover it 😄 Will review that PR as soon as I can so we can close out both of these!

robmonte (approved):
The tests pass for me now with the new changes! I'm getting ready to merge, but would like final confirmation that you also see tests passing on your end with the bugfix merged @mdgreenfield 

Thanks again!

mdgreenfield:
Ran tests with `-v` and confirmed the test is running and passing - 

```text
TF_ACC=1 go test -v --run GCPSecretImpersonatedAccount -timeout 30m ./...
PASS
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/decode    1.323s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/datasources/transform/encode    0.702s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/resources/transform/alphabet    0.883s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/resources/transform/role        1.531s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/resources/transform/template    0.558s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-vault/generated/resources/transform/transformation      1.171s [no tests to run]
PASS
PASS
PASS
PASS
2023/02/02 08:20:38 [INFO] Using Vault token with the following policies: root
=== RUN   TestGCPSecretImpersonatedAccount
--- PASS: TestGCPSecretImpersonatedAccount (3.41s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     5.297s
```

PR #1713, opened 2023-01-03
Set ForceNew on the path field of namespaces

kschoche:
With Terraform Vault provider, the current behaviour is that changing "path" argument of "vault_namespace" resource creates a new namespace but does not destroy the old one, nor update the existing namespace.


vault/resource_namespace_test.go
```diff
@@ -45,6 +45,15 @@ func TestAccNamespace(t *testing.T) {
...
 		CheckDestroy: testNamespaceDestroy(namespacePath),
 		Steps: []resource.TestStep{
+			{
+				Config: testNamespaceConfig(namespacePath),
+				Check:  testNamespaceCheckAttrs(),
+			},
+			{
+				Config: testNamespaceConfig(namespacePath + "-foo"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("vault_namespace.test", consts.FieldPath, namespacePath+"-foo")),
```
benashz, line 55: I wonder if we should also check to that the old namespace was properly deleted in Vault? You can do that in another `TestStep`? I think `testNamespaceDestroy()` might work here.

benashz (commented):
Looking great! Just had a comment about checking that the old namespace was properly deleted.

vault/resource_namespace_test.go
kschoche: Great idea, I was actually thinking about adding that, and so will do so!

kschoche: I added this using `testNamespaceDestroy()` as an additional `Check` as part of the second step and it looks like it works to check that the original namespace with that path was destroyed. I wasnt sure what to use for the `Config` portion of an additional `TestStep` yet, but if you have ideas I'd love to chat offline!

benashz (approved):
LGTM!

PR #1737, opened 2023-01-24
Fix issue where removing optional fields in database secrets backend connection resource did not reset the fields to their default values

maxcoulombe:
### Community Note

### Overview

The username & disable_escaping fields are not managed correctly causing unresolvable drift or bad updates if a previously defined value is removed from the stanza.

### To reproduce:
* Create a database backend connection resource with a username field and disable_escaping set to true in the DB type's stanza, example for postgre:
```
provider "vault" {
  token   = "my-token"
}

resource "vault_mount" "my-db" {
  path = "postgres"
  type = "database"
}

resource "vault_database_secret_backend_connection" "my-connection" {
  backend = vault_mount.my-db.path
  name    = "postgres"

  postgresql {
    connection_url     = "postgres://my-username:my-password@localhost:5432"
    username             = "my-username"
    disable_escaping = true
  }
}
```
* terraform apply
* Remove the username and disable_escaping fields from the stanza
```
resource "vault_database_secret_backend_connection" "my-connection" {
  (...)
  postgresql {
    connection_url       = "postgres://my-username:my-password@localhost:5432"
    #username             = "my-username"
    #disable_escaping = true
  }
}
```
* terraform apply
* even if the apply is successful, subsequent `terraform plan` will show unexpected drift:
```

  # vault_database_secret_backend_connection.postgresql_role will be updated in-place
  ~ resource "vault_database_secret_backend_connection" "postgresql_role" {
        id                       = "postgres/config/postgres"
        name                     = "postgres"
        # (6 unchanged attributes hidden)

      ~ postgresql {
          - username                = "my-username" -> null
          - disable_escaping    = true -> null
            # (5 unchanged attributes hidden)
        }
    }
```

### New behaviour:
* Following a successful `terraform apply`, subsequent `terraform plan` shows everything is up-to-date indicating that the statefile was updated correctly and the artificial drift is gone. The disable_escaping field also reverts back to its default `false` value in Vault.
```
No changes. Your infrastructure matches the configuration.
```

This is backward compatible. Users currently experiencing this issue will have to execute a successful apply once after upgrading their provider to a version with the fix, then all subsequent plans will no longer show the unwanted drift.

### Changelog:
```release-note
BUGS: Fix issue where removing optional fields in database secrets backend connection resource with postgresql did not reset fields to the default values
```

```
$ make testacc TESTARGS='-run=TestAccDatabaseSecretBackendConnection_postgresql'
2023/01/24 09:39:05 [INFO] Using Vault token with the following policies: root
=== RUN   TestAccDatabaseSecretBackendConnection_postgresql
--- PASS: TestAccDatabaseSecretBackendConnection_postgresql (0.66s)
PASS
```

### DIFF
--- a/vault/resource_database_secret_backend_connection_test.go
+++ b/vault/resource_database_secret_backend_connection_test.go
@@ -1481,13 +1481,13 @@ resource "vault_database_secret_backend_connection" "test" {
   root_rotation_statements = ["FOOBAR"]
 
   postgresql {
-	  connection_url          = "%s"
+      connection_url          = "%s"
       max_open_connections    = "%s"
       max_idle_connections    = "%s"
       max_connection_lifetime = "%s"
       username                = "%s"
       password                = "%s"
-	  username_template       = "%s"
+      username_template       = "%s"
       disable_escaping        = true
   }
 }
--- a/vault/resource_database_secret_backend_connection_test.go
+++ b/vault/resource_database_secret_backend_connection_test.go
@@ -760,11 +760,6 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.username_template", ""),
 				),
 			},
-			{
-				Config:             testAccDatabaseSecretBackendConnectionConfig_postgresql_reset_optional_values(name, backend, parsedURL),
-				PlanOnly:           true,
-				ExpectNonEmptyPlan: false,
-			},
 		},
 	})
 }
--- a/vault/resource_database_secret_backend_connection_test.go
+++ b/vault/resource_database_secret_backend_connection_test.go
@@ -716,6 +716,7 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
 	}
 
 	username := parsedURL.User.Username()
+	password, _ := parsedURL.User.Password()
 	backend := acctest.RandomWithPrefix("tf-test-db")
 	pluginName := dbEnginePostgres.DefaultPluginName()
 	name := acctest.RandomWithPrefix("db")
@@ -726,7 +727,7 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
 		CheckDestroy: testAccDatabaseSecretBackendConnectionCheckDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccDatabaseSecretBackendConnectionConfig_postgresql(name, backend, userTempl, parsedURL),
+				Config: testAccDatabaseSecretBackendConnectionConfig_postgresql(name, backend, userTempl, username, password, parsedURL),
 				Check: testComposeCheckFuncCommonDatabaseSecretBackend(name, backend, pluginName,
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "allowed_roles.#", "2"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "allowed_roles.0", "dev"),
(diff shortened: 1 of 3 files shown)


vault/resource_database_secret_backend_connection_test.go
```diff
@@ -739,6 +740,7 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.max_idle_connections", "0"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.max_connection_lifetime", "0"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.username", username),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.password", password),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.disable_escaping", "true"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.username_template", userTempl),
 				),
```
maxcoulombe, line 746: I wasn't sure how to test the exact problematic scenario automatically and if it is worth an explicit test or is too niche. I did not see tests that validate the plan has no changes to apply following other operations.

vinay-gopalan: Hmm, I think test parameters such as `PlanOnly` or `ExpectNonEmptyPlan` might help you achieve this test structure and confirm that a plan was successfully validated/invalidated either with/without a drift. Here is an example of this being used in the codebase — [example](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_token_test.go#L203-L212)

maxcoulombe: Got it to work, was looking for a expectEmptyPlan option but the framework uses the double negative of `ExpectNonEmptyPlan = false` to achieve this.

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -735,14 +739,32 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
...
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.password", ""),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.disable_escaping", "false"),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.username_template", ""),
+				),
+			},
+			{
+				Config:             testAccDatabaseSecretBackendConnectionConfig_postgresql_reset_optional_values(name, backend, parsedURL),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: false,
+			},
```
maxcoulombe, line 767: Wondering if there's (or should there be?) a more built-in way to check this by default? My understanding is that for 99.9% of cases after a successful `terraform apply` a subsequent plan should show no changes, otherwise it's an unintended drift bug.

vinay-gopalan: Just confirmed that the Terraform SDK by default will run a `terraform plan` after each test step to ensure that the the updates take place with no drift. Adding a test step that causes the plan to be non-empty results in an error from the SDK after the test step, something like: 
```
=== RUN   TestAccDatabaseSecretBackendConnection_postgresql
    resource_database_secret_backend_connection_test.go:771: Step 2/3 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # vault_database_secret_backend_connection.test will be updated in-place
          ~ resource "vault_database_secret_backend_connection" "test" {
                id                       = "tf-test-db-3935684109521394435/config/db-4017655576009088209"
                name                     = "db-4017655576009088209"
                # (6 unchanged attributes hidden)
        
              ~ postgresql {
                  - disable_escaping        = true -> null
                  - username                = "postgres" -> null
                    # (4 unchanged attributes hidden)
                }
            }
```

vinay-gopalan (approved):
Looks good, and thanks for adding a fix for this!

PR #1746, opened 2023-02-07
Fix qr_size config for identity mfa TOTP

davidadeleon:
This PR fixes an issue where `qr_size` would be set to zero when attempting to create this resource using the provider. A user would also be unable to specify a `qr_size` in the tf code, and would instead need to log into Vault and manually configure this option. The test has also been modified accordingly to verify defaults as well as manual config setting is executed correctly against Vault.

### Community Note

```release-note
Fix qr_size config for identity mfa totp
```

Output from testing:

```
$ go test -timeout 30s -run ^TestIdentityMFATOTP$ github.com/hashicorp/terraform-provider-vault/vault -v
2023/02/07 13:18:18 [INFO] Using Vault token with the following policies: root
=== RUN   TestIdentityMFATOTP
=== PAUSE TestIdentityMFATOTP
=== CONT  TestIdentityMFATOTP
--- PASS: TestIdentityMFATOTP (3.67s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/vault	4.236s
```

### DIFF
--- a/internal/identity/mfa/totp.go
+++ b/internal/identity/mfa/totp.go
@@ -45,7 +45,8 @@ var (
 		},
 		consts.FieldQRSize: {
 			Type:        schema.TypeInt,
-			Computed:    true,
+			Optional:    true,
+			Default:     200,
 			Description: `The pixel size of the generated square QR code.`,
 		},
 		consts.FieldAlgorithm: {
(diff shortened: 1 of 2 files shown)


benashz:
I think this is actually an issue vault. It should be setting the default value for this field. We may want to open an issue for that. Other than that, this fix looks good!

benashz (approved):
LGTM!

benashz:
We found some other related issues, so we will open new PR to address those.

benashz:
Fixed with #1749

Issue #1780, opened 2023-03-01
kv_secret_v2 corrupted state


rjhornsby:
### Terraform Version
v1.3.9

### Affected Resource(s)
- vault_kv_secret_v2

### Terraform Configuration Files
```hcl
terraform {

  required_version = ">= 1.0.0"

  required_providers {
    vault = {
      source = "hashicorp/vault"
      version = ">= 3.13.0"
    }
  }
}

resource "vault_kv_secret_v2" "packer-s3-bucket" {
  mount               = "client/my/msp/secret"
  name                = "packer/aws/0123456789/vmimport"
  delete_all_versions = true
  data_json           = jsonencode({
    foo = "bar"
  })
}
```

### Expected Behavior
Terraform needs to record the state properly.

### Actual Behavior

While the provider creates/writes the correct entry into vault itself (the mount exists, permissions are good, etc), it does not record the TF state correctly and you end up with a corrupted state file.

The actual state file contains:

```
...
"mount": "client",
"name": "msp/secret/data/packer/aws/0123456789/vmimport",
"namespace": null,
"options": null,
"path": "client/my/msp/secret/data/packer/aws/0123456789/vmimport"
...
```

Note how the `path` is correct, and matches what is expected from the TF code. However, the `mount` is stored as the first path component. The second path component, "`my`" here, is dropped, and the remaining path is stored in `name` with the improper inclusion of `data` in the value.

The state file should match the TF code, like so:

```
...
"mount": "client/my/msp/secret",
"name": "packer/aws/0123456789/vmimport",
"namespace": null,
"options": null,
"path": "client/my/msp/secret/data/packer/aws/0123456789/vmimport"
...
```

The result of this corrupted state is that:
a) a plan against this state shows a delta because TF recognizes when it reads back the state that "`client`" does not match the expected `mount` value.
b) an apply against this state fails because of (a) and because TF tries to use these corrupted values to construct a path to query vault for the infrastructure state. This re-constructed path does not exist in vault, so vault reports back 404.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform plan`

### Important Factoids

#1186 appears to be a similar but separate bug in terms of how vault paths are handled and assumptions made about how a vault path should/can be broken down.

[...]

PR #1781, opened 2023-03-01
Fix KV incorrect metadata path for prefixed mounts

### LINKED ISSUE #1770: vault_kv_secret_v2 incorrect metadata path
Since upgrading to provider version 3.13.0, our terraform pipeline can no longer update secrets in our kvv2 secret engine. Permissions did not change and a provider downgrade fixed the issue for now.

### Terraform Version
provider registry.terraform.io/hashicorp/vault v3.13.0

### Affected Resource(s)
- vault_kv_secret_v2

```hcl
resource "vault_kv_secret_v2" "credentials" {
  for_each            = toset(var.our-list-of-data)
  mount               = "our-prefix/our-secret-engine"
  name                = each.key
  cas                 = 1
  delete_all_versions = true
  data_json           = jsonencode(our-data[each.key])
}

```

### Error
```
Error: Error making API request.
│ 
│ URL: GET https://redacted.url/v1/our-prefix/metadata/data/our-secret-name
│ Code: 403. Errors:
│ 
│ * 1 error occurred:
│ 	* permission denied
│ 
```

### Expected Behavior
Terraform should be able to update the resource as with provider version 3.12.0

### Actual Behavior
The above error message appears. Note that the path does not contain the name of our secret engine but "metadata/data", so the API call seems to have been constructed incorrectly.

### Steps to Reproduce
1. Create a secret engine with a prefix in its path (for example "prefix/kvv2" instead of just "kvv2")
2. Create a token with permissions to manage secrets in this engine
3. Use the token to create a vault_kv_secret_v2 in this engine

### Important Factoids
The secret engine must contain a "/" in its path, see example above

### References
May have been introduced by https://github.com/hashicorp/terraform-provider-vault/pull/1722

vinay-gopalan:
In v3.13.0, we aimed to incorporate a fix for the construction of the metadata path for KV V2 secrets (#1722). Unfortunately, this introduced a regression for customers that had prefixed mounts (mounts that were prefixed with a string and contained a `/`, for example `cloud-metadata/kvv2-mount`).  This PR includes a fix that correctly parses mount and secret names from the ID path using regexes.

A test has been modified to also test for prefixed mounts. This test fails without the addition of the fixes in this PR.

Fixes #1770 #1780 

```
$ make testacc TESTARGS='-v -run TestAccKVSecretV2'
=== RUN   TestAccKVSecretV2
--- PASS: TestAccKVSecretV2 (5.46s)
PASS

```

### DIFF
--- a/vault/resource_kv_secret_v2.go
+++ b/vault/resource_kv_secret_v2.go
@@ -8,7 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"strings"
+	"regexp"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -19,12 +19,17 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/internal/provider"
 )
 
-var kvMetadataFields = map[string]string{
-	consts.FieldMaxVersions:        consts.FieldMaxVersions,
-	consts.FieldCASRequired:        consts.FieldCASRequired,
-	consts.FieldDeleteVersionAfter: consts.FieldDeleteVersionAfter,
-	consts.FieldCustomMetadata:     consts.FieldData,
-}
+var (
+	kvV2SecretMountFromPathRegex = regexp.MustCompile("^(.+)/data/.+$")
+	kvV2SecretNameFromPathRegex  = regexp.MustCompile("^.+/data/(.+)$")
+
+	kvMetadataFields = map[string]string{
+		consts.FieldMaxVersions:        consts.FieldMaxVersions,
+		consts.FieldCASRequired:        consts.FieldCASRequired,
+		consts.FieldDeleteVersionAfter: consts.FieldDeleteVersionAfter,
+		consts.FieldCustomMetadata:     consts.FieldData,
+	}
+)
 
 func kvSecretV2Resource(name string) *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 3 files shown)


vault/resource_kv_secret_v2.go
```diff
@@ -364,3 +366,25 @@ func kvSecretV2Delete(_ context.Context, d *schema.ResourceData, meta interface{
 
 	return nil
 }
+
+func getKVV2SecretNameFromPath(path string) (string, error) {
```
raymonstah, line 370: Would it be useful to add a table test for these functions?

raymonstah (approved):

fairclothjm (approved):

vault/resource_kv_secret_v2.go
vinay-gopalan: added table tests in latest commit; good call!

mouellet:
When can we expect a release with this fix? Without it, risk of ending up with a corrupted state is pretty high!

vinay-gopalan:
Apologies for the delay. We had to move some things around in the release to incorporate some features that needed to be added after Vault 1.13 was released. The fix for this issue has been released as part of TFVP [v3.14.0](https://github.com/hashicorp/terraform-provider-vault/releases/tag/v3.14.0). We appreciate your patience, and please let us know if you see this issue again! 🙏🏼

PR #1795, opened 2023-03-14
Enable Vault 1.13 in CI

vinay-gopalan:
This PR upgrades the build matrix so we are also running tests against Vault 1.13.


fairclothjm (approved):
👍

PR #1804, opened 2023-03-21
Vault 13682/add backend resource and test

Zlaticanin:
1/2 

This PR is for the backend config resource that mounts the secrets engine and writes the config.

In the other PR (linked later) role resource will be written that sets up all the necessary parameters to manage the lifecycle of the role.

```
2023/03/21 14:38:05 [INFO] Using Vault token with the following policies: root
=== RUN   TestAccMongoDBAtlasSecretBackend_basic
--- PASS: TestAccMongoDBAtlasSecretBackend_basic (1.38s)
PASS

```


internal/consts/consts.go
```diff
@@ -244,26 +246,27 @@ const (
...
-	MountTypeRadius     = "radius"
-	MountTypeOCI        = "oci"
-	MountTypeOIDC       = "oidc"
-	MountTypeJWT        = "jwt"
-	MountTypeAzure      = "azure"
-	MountTypeGitHub     = "github"
-	MountTypeAD         = "ad"
-	MountTypeConsul     = "consul"
-	MountTypeTerraform  = "terraform"
+	MountTypeDatabase     = "database"
```
vinay-gopalan, line 249: I would avoid this formatting :)

testutil/testutil.go
```diff
@@ -186,6 +186,11 @@ func GetTestRMQCreds(t *testing.T) (string, string, string) {
 	return v[0], v[1], v[2]
 }
 
+func GetTestMDBACreds(t *testing.T) (string, string) {
+	v := SkipTestEnvUnset(t, "PRIVATE_KEY", "PUBLIC_KEY")
```
vinay-gopalan, line 190: Could we have a prefix to the env variables, something like `MONGODB_ATLAS_PRIVATE_KEY`? Although it's longer I'd rather opt on the side of being over explicit 😅

vinay-gopalan (commented):
Looking great! I can do another pass when the docs are up 😄

internal/consts/consts.go
Zlaticanin: I am pretty sure that this part was me resolving the conflicts after the PR was up. Didn't add this. Is there a way to check?

testutil/testutil.go
Zlaticanin: Makes sense. Will do
Thank you!

internal/consts/consts.go
Zlaticanin: Hmm.. Not entirely sure what you mean here :D 
The one that I added was `MountTypeMongoDBAtlas = "mongodbatlas" `. Not sure why it formatted all of the mount types?

Zlaticanin:
> Looking great! I can do another pass when the docs are up 😄

Are there any doc PRs that I can look at and use as reference?

website/docs/r/mongodbatlas_secret_backend.html.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault#namespace).
+  *Available only for Vault Enterprise*.
```
Zlaticanin, line 37: Not sure if this is needed, although I saw it listed in most of the docs?

vault/resource_mongodbatlas_secret_backend.go
```diff
@@ -0,0 +1,132 @@
...
+	"golang.org/x/net/context"
+)
+
+var mongodbAtlasAPIFields = []string{
+	consts.FieldPublicKey,
+	consts.FieldPrivateKey,
+}
+
+func mongodbAtlasSecretBackendResource() *schema.Resource {
+	return provider.MustAddMountMigrationSchema(&schema.Resource{
```
vinay-gopalan, line 23: So after having a closer look at the code, I believe we can remove this. The `vault_mount` resource is responsible for remounting/unmounting, and if there is a path change it will always perform mount migration.
```suggestion
	&schema.Resource{
```

vault/resource_mongodbatlas_secret_backend.go
```diff
@@ -0,0 +1,132 @@
...
+	if e != nil {
+		return diag.FromErr(e)
+	}
+
+	path := d.Id()
+	if err := d.Set(consts.FieldPath, path); err != nil {
+		return diag.FromErr(err)
+	}
+
+	backend := strings.Trim(path, "/config")
```
vinay-gopalan, line 99: Did some testing and realized this should be `TrimSuffix` instead 😅 
```suggestion
	backend := strings.TrimSuffix(path, "/config")
```

vault/resource_mongodbatlas_secret_backend_test.go
```diff
@@ -0,0 +1,71 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, fmt.Sprintf("%s/config", mount)),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPrivateKey, updatedPrivateKey),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPublicKey, updatedPublicKey),
+				),
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{consts.FieldDisableRemount, consts.FieldPrivateKey},
```
vinay-gopalan, line 52: Can remove this as per comment above; this field no longer exists in the schema
```suggestion
				ImportStateVerifyIgnore: []string{consts.FieldPrivateKey},
```

internal/consts/consts.go
vinay-gopalan: Ah my bad! I misread this. No issues here!

vinay-gopalan (commented):
Looking great! Ran some tests and had some comments; almost there 😄

website/docs/r/mongodbatlas_secret_backend.html.md
vinay-gopalan: Yup! This is to be added to all docs. The `UpdateSchemaResource` wrapper you added in `provider.go` around the resource is what adds this schema field to all resources in the provider :)

Zlaticanin: Ah, that makes sense. Thank you!

vault/resource_mongodbatlas_secret_backend.go
```diff
@@ -0,0 +1,126 @@
...
+		CustomizeDiff: getMountCustomizeDiffFunc(consts.FieldPath),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			consts.FieldBackend: {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "Path where MongoDB Atlas secret backend will be mounted",
```
vinay-gopalan, line 32:
```suggestion
				Description:  "Path where MongoDB Atlas secret backend is mounted",
```

website/docs/r/mongodbatlas_secret_backend.html.md
```diff
@@ -0,0 +1,55 @@
+---
+layout: "vault"
+page_title: "Vault: vault_mongodbatlas_secret_backend resource"
+sidebar_current: "docs-vault-resource-mongodbatlas-secret-backend"
+description: |-
+Creates an MongoDB Atlas secret backend for Vault.
```
vinay-gopalan, line 6: nit: 
```suggestion
Creates a MongoDB Atlas secret backend for Vault.
```

website/docs/r/mongodbatlas_secret_backend.html.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+# vault\_mongodbatlas\_secret\_backend
+
+Creates an MongoDB Atlas Secret Backend for Vault.
```
vinay-gopalan, line 11: nit:
```suggestion
Creates a MongoDB Atlas Secret Backend for Vault.
```

vinay-gopalan (commented):
Looking great! I think we just need to include the docs in the `vault.erb` file so that the docs show up in the sidebar. Here's [an example](https://github.com/hashicorp/terraform-provider-vault/blob/main/website/vault.erb#L384-L386) of how this was done for the KMIP Secret Backend 😄

vinay-gopalan (approved):
LGTM! Thanks for working on this 🙏🏼

PR #1816, opened 2023-04-10
add role resource + test

Zlaticanin:
2/2 Second part of adding TFVP support for MongoDB Atlas secrets engine.


vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	if v, ok := d.GetOk("project_roles"); ok {
+		data["project_roles"] = v
+	}
+	if v, ok := d.GetOk("ttl"); ok {
+		data["ttl"] = v
+	}
+	if v, ok := d.GetOk("max_ttl"); ok {
+		data["max_ttl"] = v
+	}
```
vinay-gopalan, line 140: We can make a list of API fields and loop over them to reduce some of the repeated bits 😄  
Example of this [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kubernetes_secret_backend_role.go#L144-L160)

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+	if v, ok := resp.Data["ttl"]; ok {
+		if err := d.Set("ttl", v); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting state key 'ttl': %s", err))
+		}
+	}
+	if v, ok := resp.Data["max_ttl"]; ok {
+		if err := d.Set("max_ttl", v); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting state key 'max_ttl': %s", err))
+		}
+	}
```
vinay-gopalan, line 218: Similar comment as above. We can make a list of API fields and loop over them as done [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kubernetes_secret_backend_role.go#L200-L218)

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func mongodbAtlasSecretRoleRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := provider.GetClient(d, meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	path := d.Id()
+	if err := d.Set(consts.FieldPath, path); err != nil {
```
vinay-gopalan, line 158: We will also need to set the `backend` field to state here. We can probably do that by extracting the value using a regex. An example of constructing such a regex can be found [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_azure_auth_backend_role.go#L20), and the usage of this regex to obtain the backend can be found [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_azure_auth_backend_role.go#L336-L345) 😄

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the role",
+			},
+			"organization_id": {
+				Type:        schema.TypeString,
+				Required:    false,
```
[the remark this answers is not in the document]
vinay-gopalan: These are probably failing in CI build right now because currently the fields are neither Required, Optional or Computed. Those errors should go away when we update these to be `Optional: true` instead of `Required: false`

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+				Description:  "Path where MongoDB Atlas secret backend is mounted",
+				ValidateFunc: provider.ValidateNoLeadingTrailingSlashes,
+			},
+			consts.FieldPath: {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "Path where MongoDB Atlas backend is mounted",
+				ValidateFunc: provider.ValidateNoLeadingTrailingSlashes,
+			},
+			"name": {
```
vinay-gopalan, line 40: We have a constant for `name` in `consts.go`. If we don't have constants for any of the fields in this file, we should add them to `consts.go` and update the code to use constants here and in the tests. Thanks in advance for extending the constants! 🙏🏼

vinay-gopalan (commented):
Looking great so far! Had some initial comments; thanks for working on this! 😄

vault/resource_mongodbatlas_secret_role.go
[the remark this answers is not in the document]
Zlaticanin: Makes sense, fixed now. Thank you!

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+			},
+			"roles": {
+				Type:        schema.TypeList,
+				Required:    true,
+				Description: "List of roles that the API Key needs to have",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"id_addresses": {
```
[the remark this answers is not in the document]
Zlaticanin: Ah my bad 😆 Don't know how I did that, they are not even close on the keyboard lol

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,238 @@
...
+				Type:        schema.TypeList,
+				Required:    false,
+				Description: "Roles assigned when an org API key is assigned to a project API key",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"ttl": {
+				Type:        schema.TypeString,
+				Optional:    true,
```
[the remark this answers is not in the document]
Zlaticanin: Got it! Thank you!

vault/resource_mongodbatlas_secret_role.go
Zlaticanin: That looks a lot cleaner and nicer! Thank you!

vault/resource_mongodbatlas_secret_role.go
Zlaticanin: Got it! Did find a field for `ttl`, the rest are added in `consts.go` and are being used in both files now.

vault/resource_mongodbatlas_secret_role.go
Zlaticanin: Got it! Thank you for the examples, very helpful! 🙏

vault/resource_mongodbatlas_secret_role_test.go
```diff
@@ -0,0 +1,97 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldTTL, "30"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMaxTTL, "60"),
+				),
+			},
+			{
+				Config: testAccMongoDBAtlasSecretBackendRole_basic(mount, privateKey, publicKey, "tf-test-role-updated", "7cf5a45a9ccf6400e60981b7", "5cf5a45a9ccf6400e60981b6", "ORG_READ_ONLY", "192.168.1.5, 192.168.1.6", "192.168.1.3/35", "GROUP_READ_ONLY", "30min", "1hr"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldBackend, mount),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldName, "tf-test-role-updated"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldOrganizationID, "7cf5a45a9ccf6400e60981b7"),
```
Zlaticanin, line 52: Currently using the same values for these fields. Got them from our docs, it needs to be updated. 
@vinay-gopalan I was wondering if we should test maybe test different fields in these configs? Kinda similarly to [this](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kubernetes_secret_backend_role_test.go#L18)?

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,215 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			consts.FieldBackend: {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "Path where MongoDB Atlas secret backend is mounted",
+				ValidateFunc: provider.ValidateNoLeadingTrailingSlashes,
+			},
+			consts.FieldPath: {
```
vinay-gopalan, line 37: We don't need a `path` field. The ID for the resource will store the path :)

vault/resource_mongodbatlas_secret_role_test.go
```diff
@@ -0,0 +1,107 @@
...
+	privateKey, publicKey := testutil.GetTestMDBACreds(t)
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		PreCheck:          func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy:      testCheckMountDestroyed(resourceType, consts.MountTypeMongoDBAtlas, consts.FieldPath),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccMongoDBAtlasSecretBackendRole_basic(mount, privateKey, publicKey, "tf-test-role",
+					"7cf5a45a9ccf6400e60981b7", "5cf5a45a9ccf6400e60981b6", "ORG_MEMBER",
```
vinay-gopalan, line 29: nit: should we set these using variables?

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,215 @@
...
+		consts.FieldTTL,
+		consts.FieldMaxTTL,
+	}
+	for _, k := range fields {
+		if d.HasChange(k) {
+			data[k] = d.Get(k)
+		}
+	}
+
+	if _, err := client.Logical().Write(path, data); err != nil {
```
vinay-gopalan, line 134: nit: to match up with the read with context below
```suggestion
	if _, err := client.Logical().WriteWithContext(ctx, path, data); err != nil {
```

vinay-gopalan (commented):
Looking great! Could we also include the docs in this PR?

website/docs/r/mongodbatlas_secret_role.html.md
```diff
@@ -0,0 +1,85 @@
...
+written in cleartext to state and plan files generated by Terraform, and
+will appear in the console output when Terraform runs. Protect these
+artifacts accordingly. See
+[the main provider documentation](../index.html)
+for more details.
+
+## Example Usage
+
+```hcl
+resource "vault_mongodbatlas_secret_backend" "config" {
```
vinay-gopalan, line 23: This would also need a `vault_mount` to be set up right?

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,209 @@
...
+	"log"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+)
+
+var mongodbAtlasSecretBackendFromPathRegex = regexp.MustCompile("^(.+)/roles/.+$")
```
vinay-gopalan, line 18:
```suggestion
var mongodbAtlasSecretMountFromPathRegex = regexp.MustCompile("^(.+)/roles/.+$")
```

vault/resource_mongodbatlas_secret_role.go
```diff
@@ -0,0 +1,209 @@
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Roles assigned when an org API key is assigned to a project API key",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			consts.FieldTTL: {
+				Type:        schema.TypeString,
+				Computed:    true,
```
vinay-gopalan, line 87: I tested things out and looks like these are configurable, so we can update them to be `Optional`
```suggestion
				Optional:    true,
```

vinay-gopalan (commented):
Looking great! Had some comments, a few of which have been scoped under this [PR](https://github.com/hashicorp/terraform-provider-vault/pull/1819). Let me know what you think!

website/docs/r/mongodbatlas_secret_role.html.md
```diff
@@ -0,0 +1,91 @@
...
+
+No additional attributes are exported by this resource.
+
+## Import
+
+The MongoDB Atlas secret role can be imported using the full path to the role
+of the form: `<mount_path>/roles/<role_name>` e.g.
+
+```
+$ terraform import vault_mongodbatlas_secret_role.example mongodbatlas/roles/example-role
```
vinay-gopalan, line 90:
```suggestion
$ terraform import vault_mongodbatlas_secret_role.role mongodbatlas/roles/example-role
```

vinay-gopalan (approved):
LGTM! Thanks for working on this 🙏🏼

Zlaticanin:
@vinay-gopalan Thank you so much for the reviews and help! 🙏

PR #1833, opened 2023-05-01
ldap dynamic credential data source

raymonstah:
Adds a data source for LDAP dynamic credentials.
Also renames the dynamic credentials resource to use `mount` instead of `path`, as discussed in a previous PR.


vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	if err := d.Set(consts.FieldDistinguishedNames, response.distinguishedNames); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set(consts.FieldPassword, response.password); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set(consts.FieldUsername, response.username); err != nil {
+		return diag.FromErr(err)
+	}
```
Zlaticanin, line 111: Can we use a loop here? Something similar that was done [here](https://github.com/hashicorp/terraform-provider-vault/blob/283770a5a1be807848999e02aab7635ae47fdeca/vault/resource_ldap_secret_backend.go#L207-L228)

website/docs/d/ldap_dynamic_role_credentials.html.md
```diff
@@ -0,0 +1,92 @@
+---
+layout: "vault"
+page_title: "Vault: vault_ldap_dynamic_credentials data source"
+sidebar_current: "docs-vault-datasource-ldap-dynamic_credentials"
+description: |-
+Reads dynamic role credentials from an LDAP secret backend in Vault
```
Zlaticanin, line 6:
```suggestion
   Reads dynamic role credentials from an LDAP secret backend in Vault
```

vault/data_source_ldap_dynamic_role_credentials.go
raymonstah: although technically possible by converting the response into a `map[string]any{}`, i'm a bit hesitant to do it here because we'd be giving up type safety after parsing & validating the secret.

if others feel like it's worth the tradeoff, i'm more than happy to change it! i know the duplication isn't pretty.

fairclothjm (approved):

vault/data_source_ldap_dynamic_role_credentials.go
Zlaticanin: Makes sense. I think safety is more important

Zlaticanin (approved):
LGTM! 👍

vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	if secret == nil {
+		return diag.FromErr(fmt.Errorf("no role found at %q", fullPath))
+	}
+
+	response, err := parseLDAPDynamicCredSecret(secret)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(response.username)
```
vinay-gopalan, line 93: We should set the ID to the `response.leaseID` here since we have that unique and important lease info readily available here. Since we didn't have that info available for the static roles, I wanted to check in on how we are setting the ID for that data source on [this line ](https://github.com/hashicorp/terraform-provider-vault/blob/VAULT-8099/secrets/ldap/vault/data_source_ldap_static_role_credentials.go#L99)

@raymonstah @fairclothjm do we want to set the ID to a more unique combination of values? I realized that it could be possible that two IDs would clash if the credentials are read for the same username (or is that not a concern?). Wanted to get your thoughts on if there was potentially a more unique choice for the ID in case of static role credentials

vinay-gopalan, line 93:
```suggestion
	d.SetId(response.leaseID)
```

vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+		leaseID = leaseIDRaw.(string)
+	}
+
+	if leaseDurationRaw, ok := secret.Data[consts.FieldLeaseDuration]; ok {
+		leaseDuration = leaseDurationRaw.(string)
+	}
+
+	if leaseRenewableRaw, ok := secret.Data[consts.FieldLeaseRenewable]; ok {
+		leaseRenewable = leaseRenewableRaw.(bool)
+	}
```
vinay-gopalan, line 141: These values are part of the `secret` structure and aren't nested under `secret.Data`. We can update this to be
```suggestion
	leaseID = secret.LeaseID
	leaseDuration = secret.LeaseDuration
	leaseRenewable = secret.Renewable
```

vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+				ForceNew:    true,
+				Description: "Name of the role.",
+			},
+			consts.FieldLeaseID: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Lease identifier assigned by Vault.",
+			},
+			consts.FieldLeaseDuration: {
+				Type:        schema.TypeString,
```
vinay-gopalan, line 39: Suggest managing this as a `TypeInt`. Will make our lives a whole lot easier in the long run 😅

vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	leaseRenewable     bool
+	distinguishedNames []string
+	password           string
+	username           string
+}
+
+func parseLDAPDynamicCredSecret(secret *api.Secret) (lDAPDynamicCredResponse, error) {
+	var (
+		leaseID            string
+		leaseDuration      string
```
vinay-gopalan, line 127:
```suggestion
		leaseDuration      int
```

vault/data_source_ldap_dynamic_role_credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	if err := d.Set(consts.FieldUsername, response.username); err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
+}
+
+type lDAPDynamicCredResponse struct {
+	leaseID            string
+	leaseDuration      string
```
vinay-gopalan, line 117:
```suggestion
	leaseDuration      int
```

website/docs/d/ldap_dynamic_role_credentials.html.md
```diff
@@ -0,0 +1,92 @@
...
+dn: cn={{.Username}},ou=users,dc=example,dc=org
+changetype: delete
+EOT
+  rollback_ldif = <<EOT
+dn: cn={{.Username}},ou=users,dc=example,dc=org
+changetype: delete
+EOT
+  default_ttl   = "60s"
+  max_ttl       = "60s"
+}
```
vinay-gopalan, line 51:
```suggestion
}

```

vinay-gopalan (commented):
Looking great! Had a few comments around handling the lease data and setting a unique ID, but should be ready to get in after those! 🙏🏼

vault/data_source_ldap_dynamic_role_credentials.go
raymonstah: ah yeah, using the `leaseID` seems preferable; i can update that.

as for the static credentials, I'm not sure if there is something unique we could use. I'm also not sure if this is a concern. could you explain the significance of setting the ID?

vinay-gopalan: The IDs are how Terraform keeps track of all the terraform resource/datasources. These IDs should be unique, and you can [see examples](https://github.com/hashicorp/terraform-provider-vault/blob/VAULT-8099/secrets/ldap/vault/resource_ldap_secret_backend.go#L209-L212) of how removing these IDs from the TF state can cause the terraform resource to be destroyed and cleaned up.

Causing two resources/data sources to have the same ID can cause either to be overwrited or a mismanagement of the TF state for that object, so we should avoid any ID clashes for resources and data sources. Let me know if you have any other questions!

Issue #602, opened 2019-11-12
vault provider does not use namespace creating child token


matteomazza91:
### Terraform Version
Terraform v0.12.6
+ provider.vault v2.6.0

### Terraform Configuration Files
main.tf

```hcl
provider "vault" {
  version = "~> 2.6"
  address = "https://MY_VAULT_HOSTNAME/"
  namespace = "MY_NAMESPACE"
  alias   = "my-alias"

  auth_login {
    path = "auth/ldap/login/${var.login_username}"
    parameters = {
      password = var.login_password
    }
  }
}

module "my_test_module" {
  providers = {
    vault = "vault.my-alias"
  }
  source = "./modules/vault/"
}
output "vault_test" {
  value = module.my_test_module.vault_test
}
```

./modules/vault/main.tf

```hcl
data "vault_generic_secret" "vault_secret_test" {
  path = "some/path"
}

output "vault_test" {
  value = data.vault_generic_secret.vault_secret_test.data
}
```

### Debug Output
```
2019-11-12T12:47:48.990Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: -----------------------------------------------------
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: 2019/11/12 12:47:49 [DEBUG] Vault API Request Details:
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: ---[ REQUEST ]---------------------------------------
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: POST /v1/auth/token/create HTTP/1.1
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: Host: MY_VAULT_HOSTNAME
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: User-Agent: Go-http-client/1.1
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: Content-Length: 128
2019-11-12T12:47:49.398Z [DEBUG] plugin.terraform-provider-vault_v2.6.0_x4: X-Vault-Namespace: 
[...]

PR #1840, opened 2023-05-04
Revert #1830

benashz:
This reverts commit 5d5882e9ca12d2ca6fa628bb62e9af00aaf64ae6.

PR #1841, opened 2023-05-04
Add back support for deriving the provider namespace from the Vault token's

benashz:
This PR reverts the revert PR #1840, and adds back support for deriving the provider namespace from the Vault token.

It supports the following configurations:
- namespace set in the token and not set on the provider: use the token's namespace for provisioning
- namespace set on the provider: use that for provisioning
- namespace set on an `auth_login*` and not set on the provider: use the `auth_login*` namespace for provisioning (essentially point 1 above)

Other fixes:
- [x] log a warning if the token TTL is below a certain value e.g 10m
- [x] document Token TTL best practices

Closes https://github.com/hashicorp/terraform-provider-vault/issues/602


benashz:
Going to add some tests.

internal/provider/meta.go
```diff
@@ -216,40 +217,93 @@ func NewProviderMeta(d *schema.ResourceData) (interface{}, error) {
...
+		// a child token is always created in the namespace of the parent token.
+		token, err = createChildToken(d, client, tokenNamespace)
 		if err != nil {
 			return nil, err
 		}
+
+		client.SetToken(token)
+	}
+
+	if namespace == "" && tokenNamespace != "" {
```
ncabatoff, line 287: How do I specify that I want provisioning to take place in the root namespace if my token is using a different namespace?

benashz: You would need to set the namespace here: 
https://registry.terraform.io/providers/hashicorp/vault/latest/docs#namespace

ncabatoff: The root namespace is named `""`, how do you distinguish that from namespace not set?

benashz: I think that correctly handling the default namespace is out of scope for this PR. We can fix that in the future, since it could introduce more complications for the user.

fairclothjm (approved):

vinay-gopalan (approved):

PR #1842, opened 2023-05-04
remove deprecated function oktaAuthBackendExists

jsab:
Without this, terraform always tries to create `vault_okta_auth_backend` resources that use the `namespace` attribute (and fails because they already exist) because it searches in the provider namespace for auths instead of in the appropriate namespace.

### Community Note

Relates OR Closes #0000

```release-note
Fixed resource vault_okta_auth_backend not honoring namespace attribute correctly
```

```

...
```

### DIFF
--- a/vault/resource_okta_auth_backend.go
+++ b/vault/resource_okta_auth_backend.go
@@ -28,7 +28,6 @@ func oktaAuthBackendResource() *schema.Resource {
 		Delete: oktaAuthBackendDelete,
 		Read:   ReadWrapper(oktaAuthBackendRead),
 		Update: oktaAuthBackendUpdate,
-		Exists: oktaAuthBackendExists,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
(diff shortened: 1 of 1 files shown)


vault/resource_okta_auth_backend.go
```diff
@@ -291,7 +291,12 @@ func oktaAuthBackendDelete(d *schema.ResourceData, meta interface{}) error {
 }
 
 func oktaAuthBackendExists(d *schema.ResourceData, meta interface{}) (bool, error) {
```
benashz, line 293: The `*Exists()` functions are deprecated so we should be able to remove this function along with the reference here: https://github.com/hashicorp/terraform-provider-vault/blob/9a7fdcfc52fe0e242df10907bdf8e61111adabc6/vault/resource_okta_auth_backend.go#LL31C1-L31C33

benashz (commented):
Thanks for the PR @jsab . Please see my comments about dropping the `oktaAuthBackendExists` function all together.

vault/resource_okta_auth_backend.go
jsab: @benashz I have adjusted the PR and tested the provider locally confirming that this also fixes the issue.

benashz (approved):
LGTM!

Thank you for contributing to HashiCorp!

jsab:
@benashz When can we expect the next release?

PR #1844, opened 2023-05-04
Add multi-issuer support to generate root cert resource

vinay-gopalan:
Main changes:
- Adds the multi-issuer API fields to the Root Cert resource and adds a completely new test.
- Updates docs for the aforementioned resource.

Additional changes:
- Updates some of the other PKI resources to use non-deprecated context methods from the SDK. The following resources are updated:
    - `pki_secret_backend_config_ca`
    - `pki_secret_backend_cert`
    - `pki_secret_backend_intermediate_set_signed`
    - `pki_secret_backend_root_sign_intermediate`
    - `pki_secret_backend_sign`


vinay-gopalan:
Looking at failing tests.

raymonstah (approved):
looks like there's a test failure for `TestPkiSecretBackendConfigCA_basic`, but other than that looks 👍🏼

vinay-gopalan:
After spending a decent amount of time trying to fix the tests for the Config CA resource, I decided to keep those separate for now and we can decide to maybe add them in later on. The behavior of the computed fields is very inconsistent between different Vault versions. Since the computed fields are more of a nice to have than a necessity for the feature set, I'm going to refine the scope of this PR in the interest of time to focus on the Root Cert resource updates — which are the main bits for Multi-Issuer support

PR #1876, opened 2023-05-26
Fix DB engine password overwrite

vinay-gopalan:
Adds a fix to only update the password if it is a new resource or if there is a change. This fix helps the provider maintain state even when an external rotate root command has been initiated for the DB connection's password


vault/resource_database_secret_backend_connection_test.go
```diff
@@ -767,6 +767,54 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
...
+				Check: testComposeCheckFuncCommonDatabaseSecretBackend(name, backend, pluginName,
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.username", username),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "postgresql.0.password", password),
+				),
+			},
+			{
+				PreConfig: func() {
+					client := testProvider.Meta().(*provider.ProviderMeta).GetClient()
+					rotateRootPath := fmt.Sprintf("%s/rotate-root/%s", backend, name)
+					resp, err := client.Logical().Write(rotateRootPath, nil)
```
vinay-gopalan, line 800: The rotate root does not seem to be working as expected. At present this test passes even when it should fail since no root rotation is actually taking place. Looking into this a bit more

vault/resource_database_secret_backend_connection.go
```diff
@@ -1510,8 +1511,14 @@ func setDatabaseConnectionDataWithUserPass(d *schema.ResourceData, prefix string
...
 	data["username"] = d.Get(prefix + "username")
 
-	if v, ok := d.GetOk(prefix + "password"); ok {
-		data["password"] = v.(string)
+	// Vault does not return the password in the API. If the root credentials have been rotated, sending
+	// the old password in the update request would break the connection config. Thus we only send it,
+	// if it actually changed to still support updating it for non-rotated cases.
+	passwordKey := prefix + consts.FieldPassword
+	if v, ok := d.GetOk(passwordKey); ok {
+		if d.IsNewResource() || d.HasChange(passwordKey) {
```
vinay-gopalan, line 1519: The first `d.IsNewResource` case might be redundant. Will double check and remove if it is so

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -1492,6 +1540,30 @@ resource "vault_database_secret_backend_connection" "test" {
...
+resource "vault_mount" "db" {
+  path = "%s"
+  type = "database"
+}
+
+resource "vault_database_secret_backend_connection" "test" {
+  backend = vault_mount.db.path
+  name = "%s"
+  allowed_roles = ["dev", "prod"]
+  root_rotation_statements = []
```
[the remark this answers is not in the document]
vinay-gopalan: Ah interesting! Thanks for catching that and whoops 😅

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -767,6 +767,54 @@ func TestAccDatabaseSecretBackendConnection_postgresql(t *testing.T) {
...
 
+// This test makes sure that the DB connection resource is still
+// operational even when an external rotate root call is made to update
+// its credentials
+// Prerequisites: run the Postgres container found in docker-compose.yaml
+func TestAccDatabaseSecretBackendConnection_externalRotateRoot(t *testing.T) {
+	MaybeSkipDBTests(t, dbEnginePostgres)
+
+	username := "postgres"
+	password := "secret"
```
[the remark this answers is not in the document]
vinay-gopalan: Wouldn't that cause the Postgres container in CI to have it's credentials rotated? Thereby causing our other tests that use that same container to fail?

My thought was to have this test be a local only test where we will have to set up a Docker container locally — with the knowledge that the container's password will get rotated out after and it won't be usable after this test. The idea would be the container is only set up for that test and then destroyed once it's no longer usable by other tests

Curious to hear thoughts from folks! I don't think we have any rotate root tests in CI currently so this might be new areas that might need to be documented.

raymonstah (approved):

vault/resource_database_secret_backend_connection_test.go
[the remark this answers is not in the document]
raymonstah: Yeah I have it noted as a comment [here](https://github.com/hashicorp/terraform-provider-vault/pull/1876/files#diff-4fa185ba74c22b153981dd218fee9f1b4b962fe279ab789137724f251815fb05R776).

This would have to be revisited if we introduced new tests that also relied on the same PG container.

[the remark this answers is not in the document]
raymonstah: I'll see if additional reruns causes any flakiness, and if so, I'll skip this test in CI.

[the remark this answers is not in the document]
raymonstah: Alternatively we can introduce https://github.com/ory/dockertest, which it looks like we're already using in Vault.

docker-compose.yaml
```diff
@@ -56,3 +56,11 @@ services:
       - LDAP_ADMIN_PASSWORD=adminpassword
       - LDAP_USERS=alice,bob,foo
       - LDAP_PASSWORDS=password1,password2,password3
+
+  postgres:
```
fairclothjm, line 60: Do we really need an entirely new container for this?

Can we just reuse the existing mysql container for this and the existing rotate root test [TestAccDatabaseSecretBackendConnectionTemplatedUpdateExcludePassword_mysql](https://github.com/hashicorp/terraform-provider-vault/blob/76aec3409c93fc78f4b5d8d2b0d055af454a377b/vault/resource_database_secret_backend_connection_test.go#LL582C6-L582C80)? If we do end up reusing that test can we just assert that the user and password are not changed in the TF state after the rotate root operation?

raymonstah: That seems reasonable to use mysql to test this change. I'm fine either way, but leaning towards a new unit test to make it obvious what we're testing for. @vinay-gopalan thoughts? I know you had originally wrote the test for Postgres.

fairclothjm: I don't want this comment to block progress. So feel free to disregard if you see fit. However, it seems to me that the `TestAccDatabaseSecretBackendConnectionTemplatedUpdateExcludePassword_mysql` test is intended to test this exact scenario but it fails to assert that the password is unchanged in the state after the rotate root operation.

fairclothjm: But, presumably, TestAccDatabaseSecretBackendConnectionTemplatedUpdateExcludePassword_mysql would have failed that assertion before this change so maybe I am wrong there.

vault/resource_database_secret_backend_connection_test.go
```diff
@@ -661,7 +661,9 @@ func TestAccDatabaseSecretBackendConnectionTemplatedUpdateExcludePassword_mysql(
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "allowed_roles.1", "prod"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "verify_connection", "true"),
 					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "mysql.0.connection_url", testConnURL),
-					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "mysql.0.max_connection_lifetime", "10"),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "mysql.0.max_connection_lifetime", "15"),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "mysql.0.username", secondaryRootUsername),
+					resource.TestCheckResourceAttr(testDefaultDatabaseSecretBackendResource, "mysql.0.password", secondaryRootPassword),
```
fairclothjm, line 666: Should we add a step that asserts that a subsequent apply with a new password will result in a change?

fairclothjm (approved):
LGTM! Just a suggestion on an additional test step.

vault/resource_database_secret_backend_connection_test.go
fairclothjm: Sorry, you are correct that it will fail. I have had ldap on my mind too much lately 😅

PR #1877, opened 2023-05-30
Add AWS Static Roles support

kpcraig:
This PR adds support for the new (ideally, in vault 1.14) AWS static roles feature.


vault/data_source_aws_static_credentials_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+	"testing"
+)
+
+func TestAccDataSourceAWSStaticCredentials(t *testing.T) {
+	//mountPath := acctest.RandomWithPrefix("tf-test-aws-static")
+	//accessKey, secretKey := testutil.GetTestAWSCreds(t) // from the environment, causes test to skip if unset
+
+	resource.Test(t, resource.TestCase{
+		IsUnitTest: false,
```
vinay-gopalan, line 15: We generally add in our acceptance test check in the PreChecks below, similar to what is done [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/data_source_ldap_static_role_credentials_test.go#L26)

Doesn't mean this won't work or is wrong, just wanted to call out the pattern in the codebase :)

vinay-gopalan (commented):
Looking great!

vault/data_source_aws_static_credentials.go
```diff
@@ -0,0 +1,79 @@
...
+	log.Printf("[DEBUG] Read %q from Vault", fullPath)
+	if secret == nil {
+		return diag.FromErr(fmt.Errorf("no role found at %q", fullPath))
+	}
+
+	d.SetId(fullPath)
+
+	// Unfortunately we called the secret fields in vault different things between the dynamic and static credentials;
+	// so we're left with the choice here of keeping them common among terraform data sources, or keeping them
+	// the same between vault and terraform.
```
fairclothjm, line 70: Out of scope for this PR, but is it worthwhile changing the naming for static creds in Vault? Would that improve things here?

kpcraig: it could - i'm not entirely sure what our style guidance is here regarding what should match (i suppose everything everywhere _should_ match, ideally)

vault/data_source_aws_static_credentials_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/testutil"
+	"testing"
+)
+
+func TestAccDataSourceAWSStaticCredentials(t *testing.T) {
+	_, _ = testutil.GetTestAWSCreds(t)
```
kpcraig, line 13: Related to this, which looks a little weird, I imagine:

I found if i included `access_key` and `secret_key` in the backend specification for the test, it would break vault's ability to get information from IAM ("Invalid security token"). I'm not sure whether this is a bug with my local vault, a possible issue in the static roles implementation in vault, or some other weird local artifact.

If I run solely with access credentials set via env var, things work.

kpcraig:
I've pulled this out of draft to indicate it's ready for deeper review

vault/data_source_aws_static_credentials.go
fairclothjm: I think at the very least we should try make sure the naming within Vault is consistent between dynamic and static credentials. If that isn't possible then we should do best effort to make it as straightforward as possible.

vinay-gopalan: Yeah I think Vault is very inconsistent about the access key fields for AWS. I have seen `access_key`, `access_key_id` and `aws_access_key_id` 😓 

I think it's best to match the API field in Vault for less confusion. We have precedents for that with other AWS resources. So I would suggest making the schema field `FieldAccessKeyID` (you might have to add a constant for that)

vault/data_source_aws_static_credentials_test.go
vinay-gopalan: Ah yes, I think this has something to do with how the AWS Secrets Engine always needs the STS token passed as an environment variable. 

Going to cc @calvn since they recently confirmed this behavior in the AWS Secrets Engine when the team was working on the static roles implementation for Vault. It doesn't look like token info can be passed to the AWS Config Root endpoint, is that correct?

vault/data_source_aws_static_credentials.go
kpcraig: Updating this, thanks to Calvin and JM we got the field name changed merged into vault.

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+	"github.com/hashicorp/terraform-provider-vault/util"
+)
+
+func awsSecretBackendStaticRoleResource() *schema.Resource {
+	fields := map[string]*schema.Schema{
+		consts.FieldMount: {
```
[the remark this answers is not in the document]
fairclothjm: Yes, we would probably use ExactlyOneOf although I haven't tested it. I think for now we can revert to using backend until we get a better pattern decided on--which we don't want to wait on here.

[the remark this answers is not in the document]
kpcraig: That sounds good.

In the meantime, I did some testing and it looks like we can do this, although when it came to still allowing both to be unset, I was having some trouble, with both `ExactlyOneOf` and some `Default` + `Computed` issues.

vault/data_source_aws_static_credentials_test.go
```diff
@@ -0,0 +1,55 @@
...
+	username := testutil.SkipTestEnvUnset(t, "AWS_STATIC_USER")[0]
+	mount := acctest.RandomWithPrefix("tf-aws-static")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			//SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion114)
```
fairclothjm, line 20: Should we enable this PreCheck?

vault/resource_aws_secret_backend_static_role_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestAccAWSSecretBackendStaticRole(t *testing.T) {
+	mount := acctest.RandomWithPrefix("tf-aws-static")
+	_, _ = testutil.GetTestAWSCreds(t)
+	username := testutil.SkipTestEnvUnset(t, "AWS_STATIC_USER")[0]
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			//SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion114)
```
fairclothjm, line 19: Same comment, should we enable this?

vault/resource_aws_secret_backend_static_role_test.go
```diff
@@ -0,0 +1,50 @@
...
+		},
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
```
fairclothjm, line 22: Can we add an import step? Example: https://github.com/hashicorp/terraform-provider-vault/blob/abd3bce786b4a069360495265ab4ae03a25c1051/vault/resource_aws_secret_backend_test.go#L42

vault/data_source_aws_static_credentials_test.go
fairclothjm: I looked back at the investigation and this is what I found: There is not a way to provide the session token to the AWS secrets engine to the config other than the aws env var. This is likely intentional since sts creds are not supposed to be long lived and therefore not intended to be used as creds you provide to the root config endpoint (outside of testing purposes). The typical recommendation is to provide IAM user creds and then rotate root creds.

That being said @kpcraig can you check to see if you have any env variables set in your shell session where you vault server and/or TF tests are running? Maybe these are causing issues?

vault/data_source_aws_static_credentials_test.go
kpcraig: yeah, I missed re-including it after local tests

vault/data_source_aws_static_credentials_test.go
kpcraig: As a followup, it seems like the minimum version that works is having env-vars set in the vault environment, although setting actively incorrect settings in a terraform file will break things.

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -94,12 +96,21 @@ func readAWSStaticRoleResource(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	rolePath := d.Id()
-	log.Printf("[DEBUG] Reading %q", rolePath)
+	path := d.Id()
+	log.Printf("[DEBUG] Reading %q", path)
+	pathPieces := strings.Split(path, "/")
+	if len(pathPieces) < 3 || pathPieces[len(pathPieces)-2] != "static-roles" {
```
fairclothjm, line 102: I think this will fail for mount paths with arbitrary slashes, e.g. `foo/bar/static-roles/baz` where `foo/bar` is the mount path.

We could use a regex like https://github.com/hashicorp/terraform-provider-vault/blob/5e8979d3b444a1aa44dcc1af0f15213f6e48997e/vault/resource_kubernetes_secret_backend_role.go#L241

vault/data_source_aws_static_credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+}
+
+func awsStaticCredentialsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := provider.GetClient(d, meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	mount := d.Get(consts.FieldBackend).(string)
```
vinay-gopalan, line 55: nit, for consistency
```suggestion
	backend := d.Get(consts.FieldBackend).(string)
```

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -0,0 +1,147 @@
...
+func createUpdateAWSStaticRoleResource(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := provider.GetClient(d, meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	mount := d.Get(consts.FieldBackend).(string)
+	role := d.Get(consts.FieldName).(string)
+	rolePath := fmt.Sprintf("%s/static-roles/%s", mount, role)
+	log.Printf("[DEBUG] Creating AWAS static role at %q", rolePath)
```
vinay-gopalan, line 76:
```suggestion
	log.Printf("[DEBUG] Creating AWS static role at %q", rolePath)
```

vault/resource_aws_secret_backend_static_role.go
vinay-gopalan: +1 one using regexes to get this info! We can never assume what kind of paths the customers will be using and I have made this mistake in the past before 😅

vault/resource_aws_secret_backend_static_role_test.go
```diff
@@ -0,0 +1,55 @@
...
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion114)
+		},
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAWSStaticReourceConfig(mount, a, s, username),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("vault_aws_secret_backend_static_role.role", "name", "test"),
```
vinay-gopalan, line 27: suggest moving `"vault_aws_secret_backend_static_role.role"` out to a constant like `resourceName`, and updating the field names to use constants 😄 
```suggestion
					resource.TestCheckResourceAttr("vault_aws_secret_backend_static_role.role", consts.FieldName, "test"),
```

vinay-gopalan (commented):
Looking great!

vault/resource_aws_secret_backend_static_role.go
fairclothjm: The role name can also contain slashes

kpcraig: a quick question, what should "a/static-roles/static-roles/b" parse to?

fairclothjm: Given a regex like `regexp.MustCompile("^(.+)/static-roles/.+$")`, `a/static-roles/static-roles/b` would parse to:

mount name: `a/static-roles`
role name: `b`

I think that is sufficient. We can't stop a user from completely shooting themselves in the foot.

kpcraig: I added a regex (`(.+)/static-roles/(.+)`) 

One could probably argue for a narrowing of the dot or adding $/^ marks.

vault/data_source_aws_static_credentials_test.go
```diff
@@ -0,0 +1,59 @@
...
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAWSStaticDataSourceConfig(mount, username),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resourceName, consts.FieldAccessKey),
+					resource.TestCheckResourceAttrSet(resourceName, consts.FieldSecretKey),
+				),
+			},
+			//testutil.GetImportTestStep("vault_aws_static_access_credentials.creds", false, nil),
```
vinay-gopalan, line 34: Can be removed, data sources are essentially imports in themselves so they don't need this step 😄

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -0,0 +1,170 @@
...
+
+func readAWSStaticRoleResource(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, err := provider.GetClient(d, meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	path := d.Id()
+	log.Printf("[DEBUG] Reading %q", path)
+	backend, _, err := parseAWSStaticRolePath(path)
```
vinay-gopalan, line 110: We should also capture and set the `name` here

vault/resource_aws_secret_backend_static_role.go
fairclothjm: Are there any edge cases we would hit if we don't add the beginning and end of line anchors?

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -0,0 +1,171 @@
...
+			ForceNew:    true,
+		},
+		consts.FieldRotationPeriod: {
+			Type:        schema.TypeInt,
+			Required:    true,
+			Description: "How often Vault should rotate the password of the user entry.",
+		},
+	}
+	return &schema.Resource{
+		CreateContext: createUpdateAWSStaticRoleResource,
```
vinay-gopalan, line 58: We should add the `MountCreateContextWrapper` here, and ensure that this resource is only enabled on Vault versions >= 1.14. An example on how this is used can be found [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_ldap_secret_backend.go#L124)

vault/resource_aws_secret_backend_static_role.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	path := d.Id()
+	log.Printf("[DEBUG] Reading %q", path)
+	backend, role, err := parseAWSStaticRolePath(path)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("invalid id %q; %w", path, err))
+	}
+
+	err = d.Set(consts.FieldBackend, backend)
```
vinay-gopalan, line 115: We should also return an error here if it is non-nil

vault/resource_aws_secret_backend_static_role_test.go
```diff
@@ -0,0 +1,123 @@
...
+		PreCheck: func() {
+			testutil.TestAccPreCheck(t)
+			SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion114)
+		},
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAWSStaticReourceConfig(mount, username),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", "test"),
```
vinay-gopalan, line 29: we can use the constants here as well
```suggestion
					resource.TestCheckResourceAttr(resourceName, consts.FieldName, "test"),
```

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,71 @@
...
+When using the outputs of this data source to authenticate with the [Terraform Provider for AWS](https://registry.terraform.io/providers/hashicorp/aws/latest/docs) or
+the [Terraform Provider for AWS Cloud Control](https://registry.terraform.io/providers/hashicorp/awscc/latest/docs),
+the credentials leased from Vault cannnot be renewed.
+Ensure that the lease is long enough for Terraform to complete.
+
+## Example Usage
+
+```hcl
+resource "vault_aws_secret_backend" "aws" {
+	path = "%s"
```
vinay-gopalan, line 31: We should update the `%s` values in examples with test strings

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,71 @@
...
+  access_key = data.vault_aws_access_credentials.creds.access_key
+  secret_key = data.vault_aws_access_credentials.creds.secret_key
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `backend` - (Required) The path to the LDAP secret backend to
```
vinay-gopalan, line 59: The `namespace` parameter is also supported here, so we can add a block for that similar to the static role docs

website/docs/r/aws_secret_backend_static_role.md
```diff
@@ -0,0 +1,64 @@
...
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault#namespace).
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Optional) The unique path this backend should be mounted at. Must
+  not begin or end with a `/`. Defaults to `aws`
+
+* `username` - (Required) The username of the existing LDAP entry to manage password rotation for.
```
vinay-gopalan, line 49:
```suggestion
* `username` - (Required) The username of the existing AWS IAM to manage password rotation for.
```

vinay-gopalan (commented):
Looking great! A couple more comments

website/docs/r/aws_secret_backend_static_role.md
kpcraig: 🤦🏽‍♀️  I even checked to make sure I didn't miss any

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
...
+the [Terraform Provider for AWS Cloud Control](https://registry.terraform.io/providers/hashicorp/awscc/latest/docs),
+the credentials leased from Vault cannnot be renewed.
+Ensure that the lease is long enough for Terraform to complete.
+
+## Example Usage
+
+```hcl
+resource "vault_aws_secret_backend" "aws" {
+	path = "my-aws"
+	description = "Obtain AWS credentials."
```
fairclothjm, line 32: nit: The standard is 2 spaces for hcl files. I think we should try to be consistent in our docs.

vault/data_source_aws_static_credentials_test.go
fairclothjm: Should we add a comment explaining why we are calling GetTestAWSCreds but not using the return values?

fairclothjm: Also, I don't think I quite follow what the problem is. I see in resource_aws_secret_backend_role_test.go we are passing the values to the config: https://github.com/hashicorp/terraform-provider-vault/blob/716f90065eda7d13a11b87decf0ae53f259556b5/vault/resource_aws_secret_backend_role_test.go#L35. What is the difference here?

vault/resource_aws_secret_backend_static_role.go
kpcraig: added!

kpcraig:
I also want to check: this should still be merged against `vault-next`?

website/docs/d/aws_static_credentials.md
kpcraig: shoutout to the part of the test code that ended up as 3 space tabs somehow

kpcraig: spaced!

kpcraig:
Branch updated after @vinay-gopalan's merge of `main` into `vault-next`

vault/data_source_aws_static_credentials_test.go
kpcraig: I have found that in this case anyway, setting the backend values at all breaks the test, even if they are correct, and match those that the vault instance knows.

kpcraig: So, after some investigation, it looks like in some cases the Access/Secret pair is insufficient, and the method I was using to obtain credentials required a session token for access. Using my personal AWS account, I was able to get it working, so I've re-added the access/secret fields for the aws backend in the tests.

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
+---
+layout: "vault"
+page_title: "Vault: vault_aws_static_credentials data source"
+sidebar_current: "docs-vault-datasource-aws-static_credentials"
```
vinay-gopalan, line 4: looks like we're using a hyphen in the erb so we should match up here
```suggestion
sidebar_current: "docs-vault-datasource-aws-static-credentials"
```

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
...
+resource "vault_aws_secret_backend_static_role" "role" {
+  backend = vault_aws_secret_backend.aws.path
+  name = "test"
+  username = "my-test-user"
+  rotation_period = "3600"
+}
+
+
+# generally, these blocks would be in a different module
+data "vault_ldap_static_credentials" "creds" {
```
vinay-gopalan, line 44:
```suggestion
data "vault_aws_access_credentials.creds" "creds" {
```

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
...
+  backend   = vault_aws_secret_backend.aws.path
+  role_name = vault_ldap_secret_backend_static_role.role.role_name
```
vinay-gopalan, line 46:
```suggestion
  name = vault_aws_secret_backend_static_role.role.name
```

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault#namespace).
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Required) The path to the LDAP secret backend to
```
vinay-gopalan, line 64:
```suggestion
* `backend` - (Required) The path to the AWS secret backend to
```

website/docs/d/aws_static_credentials.md
```diff
@@ -0,0 +1,76 @@
...
+read credentials from, with no leading or trailing `/`s.
+
+* `role_name` - (Required) The name of the LDAP secret backend static role to read
```
vinay-gopalan, line 67:
```suggestion
* `name` - (Required) The name of the AWS secret backend static role to read
```

website/vault.erb
```diff
@@ -201,6 +205,10 @@
                             <a href="/docs/providers/vault/r/aws_secret_backend_role.html">vault_aws_secret_backend_role</a>
                         </li>
 
+                        <li<%= sidebar_current("docs-vault-resource-aws-auth-backend-role") %>>
```
vinay-gopalan, line 208:
```suggestion
                        <li<%= sidebar_current("docs-vault-resource-aws-secret-backend-static-role") %>>
```

vinay-gopalan (approved):
LGTM! Thanks for all the work on this 🙏🏼

PR #1880, opened 2023-05-31
Add support to Import Issuers to a mount in PKI engine

vinay-gopalan:
Adds the `imported_issuers` and `imported_keys` parameters to resources in order to allow issuers to be imported between mounts. This is needed to complete the full PKI multi-issuer workflow.


internal/consts/consts.go
```diff
@@ -284,6 +284,18 @@ const (
 	FieldIssuerID                      = "issuer_id"
 	FieldKeyRef                        = "key_ref"
 	FieldPemBundle                     = "pem_bundle"
+	FieldCAChain                       = "ca_chain"
```
vinay-gopalan, line 287: This small switch must have happened during all the merge conflict resolutions 😅 Happy to move it back if folks really feel the need for it but left it as is for now

vault/resource_pki_secret_backend_intermediate_set_signed.go
```diff
@@ -43,22 +61,42 @@ func pkiSecretBackendIntermediateSetSignedCreate(ctx context.Context, d *schema.
...
 	if err != nil {
 		return diag.Errorf("error creating intermediate set-signed on PKI secret backend %q: %s", backend, err)
 	}
 	log.Printf("[DEBUG] Created intermediate set-signed on PKI secret backend %q", backend)
 
 	d.SetId(path)
+
+	computedIssuerFields := []string{consts.FieldImportedIssuers, consts.FieldImportedKeys}
+
+	for _, k := range computedIssuerFields {
```
vinay-gopalan, line 83: While this is a version specific field added in 1.11, since Vault returns no response on versions 1.10 and below, this was easier to handle in this way. We only set the multi-issuer fields to state if they are returned in the response, otherwise we set the computed values to nil to resolve any potential TF drift. Happy to talk more about this if folks would like

fairclothjm (approved):

fairclothjm (approved):

PR #1884, opened 2023-06-02
Move core provider factory functionality to the provider package.

benashz:
Currently, it is not possible to test `meta.NewProviderMeta()` since it depends the `vault.Provider()`, which creates a cyclical dependency. It also makes sense that the provider factory would be in the lower level 'provider' package.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -43,23 +43,25 @@ type Description struct {
 
 type ResourceRegistry map[string]*Description
 
+type ResourcesMap map[string]*schema.Resource
+
 func NewProvider(
 	dataRegistry ResourceRegistry,
 	resourceRegistry ResourceRegistry,
-	resourceMap ...map[string]*schema.Resource,
+	extraResourcesMaps ...ResourcesMap,
 ) *schema.Provider {
 	dataSourcesMap, err := parse(dataRegistry)
 	if err != nil {
 		panic(err)
 	}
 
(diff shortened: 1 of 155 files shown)


internal/provider/provider.go
```diff
@@ -0,0 +1,295 @@
...
+		panic(err)
+	}
+
+	resourcesMap, err := parse(resourceRegistry)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, m := range resourceMap {
+		MustAddSchemaResource(m, resourcesMap, nil)
```
vinay-gopalan, line 62: should this be
```suggestion
		MustAddSchemaResource(m, resourceMap, nil)
```

benashz: Variable names a bit confusing, I think what we have there is correct. It comes from:
https://github.com/hashicorp/terraform-provider-vault/pull/1884/files#diff-7e69de666c9006fa0a0fb8fe4e4f8c418acce7c555f69f8087d88d16b434f2c1L59

vinay-gopalan: Ah you're right, thanks for clarifying

vinay-gopalan (approved):
LGTM!

internal/provider/provider.go
benashz: Make the variable names a little less confusing: f663882da8043ae07f30111e8b206349b1e2b592

vinay-gopalan: Nice!

PR #1887, opened 2023-06-06
Add support for retrying kv-v2 secret data writes

benashz:
Sometimes the kv secrets engine might be performing a migration from v1 to v2, this can cause issues writing secret data to those mounts. Adding a retry should mitigate the effects of the migration.

Extends the work done under #1579  to `vault_kv_secret_v2`

### DIFF
--- a/vault/resource_generic_secret.go
+++ b/vault/resource_generic_secret.go
@@ -7,12 +7,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"net/http"
-	"time"
 
-	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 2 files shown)


vinay-gopalan (commented):
Looks great! Should we also add this to the [KV V1 resource](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kv_secret.go#L71-L73)?

benashz:
> Looks great! Should we also add this to the [KV V1 resource](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kv_secret.go#L71-L73)?

Think it is only needed on a kv-v2 secrets.

vinay-gopalan (approved):

PR #1896, opened 2023-06-08
Update resource name for vault_ldap_secret_library

shisheng10:
When trying to use the resource `vault_ldap_secret_library` as shown in [Example Usage](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/ldap_secret_backend_library_set#example-usage), the following error is shown during `terraform plan`.

`The provider hashicorp/vault does not support resource type "vault_ldap_secret_library".`

Reference #1832, the correct resource name should be `vault_ldap_secret_backend_library_set`.


fairclothjm (approved):
Thanks!

PR #1911, opened 2023-06-20
Add Upgrade guide for PKI Multi-Issuer Feature

vinay-gopalan:
This PR adds some documentation to existing resources that can potentially be affected when upgrading to the newly enabled multi-issuer feature. Furthermore, this PR also adds an upgrade guide describing the following:

- best practices when attempting to upgrade the Provider from a non multi-issuer enabled version
- importing multi-issuer info from Vault into the TF state
- recommendations on which Vault versions to use with the multi-issuer feature


Zlaticanin (approved):
LGTM! 👍

PR #1927, opened 2023-07-07
Run tests directly on GHA runner

raymonstah:
Runs tests and acceptance tests outside of a container to see if it's any faster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,8 +148,8 @@ jobs:
         env:
           VAULT_TOKEN: "root"
           VAULT_ADDR: "http://localhost:8200"
-          # NATIVE is true if not using jobs.<job_id>.container
-          NATIVE: "true"
+          # RUNS_IN_CONTAINER is false if not using jobs.<job_id>.container
+          RUNS_IN_CONTAINER: "false"
           MYSQL_URL: "root:mysql@tcp(mysql:3306)/"
           MYSQL_CONNECTION_URL: "{{username}}:{{password}}@tcp(mysql:3306)/"
           MYSQL_CONNECTION_USERNAME: "root"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -140,7 +147,9 @@ jobs:
       - name: Acceptance Tests
         env:
           VAULT_TOKEN: "root"
-          VAULT_ADDR: "http://vault:8200"
+          VAULT_ADDR: "http://localhost:8200"
+          # NATIVE is true if not using jobs.<job_id>.container
+          NATIVE: "true"
```
fairclothjm, line 152: Want to make sure I am understanding.

1. Does this env variable act as a switch for cases where TFVP is making direct calls to the DB?
2. Does the comment mean that we are setting NATIVE to "true" because we are not running in a container?

fairclothjm (approved):
LGTM! just a few questions for clarification

.github/workflows/build.yml
raymonstah: Yes that's right! It could also be named `NOT_DOCKERIZED` but I was trying to come up with a name that doesn't use negation.

fairclothjm: I think making the name more descriptive might be better. Could we flip it? `RUNS_IN_CONTIANER` and set it to false?

fairclothjm:
I am curious if this has been rebased on top of the other test improvement changes? And does that cause any noticeable improvement with this change?

raymonstah:
> I am curious if this has been rebased on top of the other test improvement changes? And does that cause any noticeable improvement with this change?

Looks like this does speed up the acceptance tests by a bit! Worth merging in my opinion.

fairclothjm:
I think as a result of this merge we are seeing some errors and warning on the build page: https://github.com/hashicorp/terraform-provider-vault/actions/runs/5536134545

Screenshot 2023-07-12 at 4 36 53 PM

PR #1928, opened 2023-07-07
Add new auth_login_token_file method

benashz:
The new method is targeted for use in the context of a Terraform Cloud run. It's use outside of that context are not recommended, since the method requires storing vault tokens on disk.


fairclothjm (approved):
LGTM, Thanks!

vinay-gopalan (approved):
LGTM!

PR #1955, opened 2023-07-25
Retry Write on kv-v2 config

benashz:
TFVP fails to create a KV secret backed when Vault is upgrading from kv-v1  to kv-v2 storage.  The upgrade process is done async, so the provider should be able to retry the original request whenever the error is encountered.

This PR adds support for retrying create/update operations on the `vault_kv_secret_backend_v2` resource.

Example error:

"Error: error writing config data to kv_2/config, err=Error making API request.\n\nNamespace: foo\nURL: PUT https://localhost:8200/v1/kv_2/config\nCode: 400. Errors:\n\n* Waiting for the primary to upgrade from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service when the primary is finished."


util/util.go
```diff
@@ -405,3 +406,63 @@ func Remount(d *schema.ResourceData, client *api.Client, mountField string, isAu
...
+		MaxTries:    60,
+		Delay:       time.Millisecond * 500,
+		StatusCodes: []int{http.StatusBadRequest},
+	}
+}
+
+// RetryWrite attempts to retry a Logical.Write() to Vault for the
+// RetryRequestOpts. Primary useful for handling some of Vault's eventually
+// consistent APIs.
+func RetryWrite(client *api.Client, path string, data map[string]interface{}, req *RetryRequestOpts) (*api.Secret, error) {
```
vinay-gopalan, line 437: Nice! Are we now also good to remove `writeSecreteDataWithRetry` from [this line](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kv_secret_v2.go#L394) since it would no longer be used?

vinay-gopalan (approved):
LGTM, had one question

util/util.go
benashz: Ah, good catch! FIxed with a4ced5baf23b58c45d2547ba9b71610b2e78a140

Issue #1964, opened 2023-07-31
`vault_gcp_auth_backend` resource is missing the `accessor` attribute


arcud:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

v1.5.4

### Affected Resource(s)

- vault_gcp_auth_backend

### Terraform Configuration Files
```hcl
resource "vault_identity_entity" "someentity" {
  name = "somename"
  policies = ["somepolicy"]
}

data "vault_auth_backend" "myauthbackend" {
  path = vault_gcp_auth_backend.default.path
}

resource "vault_identity_entity_alias" "someentity" {

  name           = "somename"
  mount_accessor = data.vault_auth_backend.myauthbackend.accessor
  canonical_id   = vault_identity_entity.someentity.id
}

```

### Expected Behavior

`vault_gcp_auth_backend` does not have the `accessor` attribute.

### Actual Behavior

In order to get the `accessor` attribute, I need to get it from a data source, which seems unnecessary. Can we please add the `accessor` attribute to the `vault_gcp_auth_backend` resource?

Issue #882, opened 2020-10-21
tune is not supported for vault_gcp_auth_backend resource


ghost:
Hi there,

When I'm enabling GCP auth method. I want to tune token_ttl however it do not support tune parameter 

Code 
```
resource "vault_gcp_auth_backend" "gcp_auth_backend" {
  credentials  = "JUST TEST"
  path = "gcp"

  tune {
    max_lease_ttl      = "1h"
    default_lease_ttl  = "1h"
  }
}
```
Error
```
$ terraform apply 

Error: Unsupported block type

  on main.tf line 5, in resource "vault_gcp_auth_backend" "gcp_auth_backend":
   5:   tune {

Blocks of type "tune" are not expected here.
```

I was expecting to see something seemlier  as resource `vault_jwt_auth_backend` which supports tune argument

papovyr:
What will be best option here to tune auth backend ?

amhn:
You could import this backend as a generic backend and use tune there.

```
resource "vault_auth_backend" "gcp_auth_backend" {
  path = "gcp"

  tune {
    max_lease_ttl      = "1h"
    default_lease_ttl  = "1h"
  }
}
```
After `terraform import vault_auth_backend.gcp_auth_backend gcp` you should be able to configure it.

This is only a workaround because with both vault_auth_backend and vault_gcp_auth_backend an initial apply without import will fail.

papovyr:
@amhn thanks for response
If i use 
```
resource "vault_auth_backend" "gcp_auth_backend" {
  path = "gcp"

  tune {
    max_lease_ttl      = "1h"
    default_lease_ttl  = "1h"
  }
}
```
How do i configure GCP credentials ? 

If i use generic endpoint I need to match format of JSON for GCP service account JSON which terraform doesn't support 

Will be best to support tune on `vault_gcp_auth_backend` resource

amhn:
Sorry, my answer was not complete. You would use both backend types for the same path, apply one and import it as the other.
```
resource "vault_gcp_auth_backend" "gcp_auth_backend" {
  credentials  = "JUST TEST"
  path = "gcp"
}

resource "vault_auth_backend" "gcp_auth_backend" {
  path = "gcp"

  tune {
    max_lease_ttl      = "1h"
    default_lease_ttl  = "1h"
  }
}
```
[...]

PR #1980, opened 2023-08-15
Add accessor computed attribute for gcp auth backend

jdelnano:
### Description

This PR updates the schema and read functionality for the `vault_gcp_auth_backend` resource in order to define, set, and expose the `accessor` attribute.

Closes https://github.com/hashicorp/terraform-provider-vault/issues/1964 since this will expose the `accessor` attribute for the `vault_gcp_auth_backend` resource.

### Checklist

```

...
```

### Community Note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## Unreleased
-FEATURES:
+IMPROVEMENTS:
 * Add accessor attribute for `vault_gcp_auth_backend` resource: ([#1980](https://github.com/hashicorp/terraform-provider-vault/pull/1980))
 
 BUGS:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## Unreleased
+FEATURES:
+* Add accessor attribute for `vault_gcp_auth_backend` resource: ([#1980](https://github.com/hashicorp/terraform-provider-vault/pull/1980))
 
 BUGS:
 * Fixes a panic that can occur when Vault [lookup-self](https://developer.hashicorp.com/vault/api-docs/auth/token#lookup-a-token-self) API returns nil token info ([#1978](https://github.com/hashicorp/terraform-provider-vault/pull/1978))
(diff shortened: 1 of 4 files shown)


vault/resource_gcp_auth_backend_test.go
```diff
@@ -46,6 +48,7 @@ func TestGCPAuthBackend_basic(t *testing.T) {
 			{
 				Config: testGCPAuthBackendConfig_update(path, gcpJSONCredentials),
 				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckAuthMountExists("vault_gcp_auth_backend.test", &resAuthFirst),
```
austingebauer, line 51: I wonder if moving this to `testutil.go` would make sense instead of keeping it in the github auth resource test file?

austingebauer (approved):
Looks good to me! Thanks @jdelnano 👍

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 ## Unreleased
+FEATURES:
```
austingebauer, line 2: Would you also mind updating the documentation to include the accessor as a attribute? Similar to something like https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/github_auth_backend#attributes-reference. Thanks!

austingebauer, line 2: I think this is more of an improvement than a feature. Suggest to change this to `IMPROVEMENTS:`.

jdelnano: Ah, my bad -- will update now

jdelnano: Makes sense, will update.

vault/resource_gcp_auth_backend_test.go
jdelnano: Hey Austin,

I like your suggestion, but would you be open to capturing that work in a separate Github issue or addressing it as part of https://github.com/hashicorp/terraform-provider-vault/issues/882 (which I'd be more than happy to pick up and was planning on doing so!). The reason I suggest that is, should we pursue this, I believe the other helper functions linked [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_auth_backend_test.go#L274-L333) that also assert against the status of an auth mount) should also be relocated to `./testutil/testutil.go`.

Are you ok if I refactor the auth mount test logic in `./testutil/testutil.go` as part of https://github.com/hashicorp/terraform-provider-vault/issues/882 and we go ahead and merge this as is? Let me know, and thank you so much for the quick review!

austingebauer: Thanks for the detailed reply, @jdelnano! That sounds reasonable to me 👍

PR #2013, opened 2023-09-13
add local var for aws secret engine

zmadell523:
### Description
Vault core supports marking mounts as [local](https://developer.hashicorp.com/vault/api-docs/system/mounts#local) in order to keep a mount from being replicated to performance replicas. This implementation mimics what has been done in other backend mounts such as the [consul mount](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_consul_secret_backend.go).

Relates OR Closes #0000

### Checklist

### Community Note

### DIFF
--- a/vault/resource_aws_secret_backend.go
+++ b/vault/resource_aws_secret_backend.go
@@ -268,6 +268,7 @@ func awsSecretBackendRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("description", mount.Description)
 	d.Set("default_lease_ttl_seconds", mount.Config.DefaultLeaseTTL)
 	d.Set("max_lease_ttl_seconds", mount.Config.MaxLeaseTTL)
+	d.Set("local", mount.Local)
 
 	return nil
 }
--- a/vault/resource_aws_secret_backend.go
+++ b/vault/resource_aws_secret_backend.go
@@ -98,6 +98,13 @@ func awsSecretBackendResource() *schema.Resource {
 				Computed:    true,
 				Description: "Template describing how dynamic usernames are generated.",
 			},
+			"local": {
+				Type:        schema.TypeBool,
+				ForceNew:    true,
+				Optional:    true,
+				Default:     false,
+				Description: "Specifies if the secret backend is local only",
+			},
 		},
 	})
 }
@@ -143,12 +150,14 @@ func awsSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
 	iamEndpoint := d.Get("iam_endpoint").(string)
 	stsEndpoint := d.Get("sts_endpoint").(string)
 	usernameTemplate := d.Get("username_template").(string)
+	local := d.Get("local").(bool)
 
 	d.Partial(true)
 	log.Printf("[DEBUG] Mounting AWS backend at %q", path)
 	err := client.Sys().Mount(path, &api.MountInput{
 		Type:        consts.MountTypeAWS,
 		Description: description,
+		Local:       local,
 		Config: api.MountConfigInput{
 			DefaultLeaseTTL: fmt.Sprintf("%ds", defaultTTL),
 			MaxLeaseTTL:     fmt.Sprintf("%ds", maxTTL),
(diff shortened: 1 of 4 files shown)


vault/resource_aws_secret_backend.go
```diff
@@ -98,6 +98,13 @@ func awsSecretBackendResource() *schema.Resource {
 				Computed:    true,
 				Description: "Template describing how dynamic usernames are generated.",
 			},
+			"local": {
+				Type:        schema.TypeBool,
+				ForceNew:    true,
+				Optional:    true,
+				Default:     false,
```
vinay-gopalan, line 105: Suggest making this `Computed` instead, which lets TF know that this state can be read and set to state from Vault. Ideally we like to let Vault handle all the default values it can :)

vault/resource_aws_secret_backend.go
```diff
@@ -143,12 +150,14 @@ func awsSecretBackendCreate(d *schema.ResourceData, meta interface{}) error {
...
 	stsEndpoint := d.Get("sts_endpoint").(string)
 	usernameTemplate := d.Get("username_template").(string)
+	local := d.Get("local").(bool)
 
 	d.Partial(true)
 	log.Printf("[DEBUG] Mounting AWS backend at %q", path)
 	err := client.Sys().Mount(path, &api.MountInput{
 		Type:        consts.MountTypeAWS,
 		Description: description,
+		Local:       local,
```
vinay-gopalan, line 160: Looking great! We should also set this value to state if we read from Vault. We can add the following in the read method as done on [these lines here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_aws_secret_backend.go#L258-L261)

```
d.Set("local", mount.Local)
```

vinay-gopalan (commented):
Thanks for working on this! Looking good, left a couple comments. I'm not sure if this parameter can also be passed in when tuning a mount, but if it can, we would also need to add that in the `awsSecretBackendUpdate` method as well :)

zmadell523:
> Thanks for working on this! Looking good, left a couple comments. I'm not sure if this parameter can also be passed in when tuning a mount, but if it can, we would also need to add that in the `awsSecretBackendUpdate` method as well :)

Vault core requires a mount to be rebuilt to change the value of `Local` so I don't think it is exposed to tune.

vault/resource_aws_secret_backend.go
zmadell523: Are Computed and Default mutually exclusive? I was trying to make sure that it would default to false, since this would be a net new flag, and changing to true requires a recreate on the resource

zmadell523: It sounds like `Computed` will effectively pull this value from Vault's default (which is false) if the optional value is not provided.

vinay-gopalan: Never mind, since this is a `ForceNew` and we don't want to recreate existing TF pipelines, it's fine to leave this as is

vault/resource_aws_secret_backend.go
```diff
@@ -98,6 +98,13 @@ func awsSecretBackendResource() *schema.Resource {
 				Computed:    true,
 				Description: "Template describing how dynamic usernames are generated.",
 			},
+			"local": {
```
vinay-gopalan, line 101: Could we also add a check in the test file [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_aws_secret_backend_test.go#L17) that ensures the fields default works as expected? We can add the following check:

```
resource.TestCheckResourceAttr(resourceName, "local", "false"),
```

vinay-gopalan (commented):
Hey @zmadell523, thanks for making the changes! The PR is looking great, could we also add the documentation for the newly added parameter over [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/website/docs/r/aws_secret_backend.html.md?plain=1#L30)? Thanks! 🙏🏼

vinay-gopalan (approved):

PR #2052, opened 2023-10-12
Fix failing data oidc client creds test

benashz:
Fixes failing tests introduced in #2048


vinay-gopalan (approved):
LGTM, were we also thinking of marking the `client_type` field as ForceNew?

benashz:
> LGTM, were we also thinking of marking the `client_type` field as ForceNew?

That probably makes sense since it is immutable on the Vault side. Should probably covered in another PR.

PR #1997, opened 2023-09-01
Add tune block support to gcp auth backend and relocate auth mount test functions

jdelnano:
### Description
This PR closes https://github.com/hashicorp/terraform-provider-vault/issues/882 by adding `tune` block support, adding the arguments shown [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/auth_mount.go#L26-L78), such as `default_lease_ttl`, etc. In order to accomplish this, the specific changes in this PR are:

- Update `./vault/resource_gcp_auth_backend.go` (using `vault/resource_github_auth_backend_test.go` as a reference) to:
  - Add the `tune` field to the resource schema
  - Update the resource's `Read` and `Update` methods to properly apply `tune` changes to the GCP auth mount
- Using `./vault/resource_github_auth_backend_test.go` as a reference, update `./vault/resource_gcp_auth_backend_test.go` to incorporate assertions against `tune` block settings (along with small cosmetic changes) 
- Based on [discussion](https://github.com/hashicorp/terraform-provider-vault/pull/1980#discussion_r1297794666) during the review of #1980, relocate auth mount test utility functions to `./testutil/testutil.go`
- Update the following test files to use the newly re-homed auth mount test utility functions:
  - `./vault/resource_gcp_auth_backend_test.go`
  - `./vault/resource_github_auth_backend_test.go`
  - `./vault/resource_auth_backend_test.go`

### Checklist
- [x] Added [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md) entry (only for user-facing changes)
- [x] Acceptance tests where run against all supported Vault Versions

### Output from acceptance testing:
I ran the following tests, which are affected/updated by these changes, against Vault `1.14.1` (non-enterprise), running as a Docker container on my local machine. All commands shown below had passing tests except one test, which failed since it involved namespaces and I do not have a Vault enterprise server.

Results file for GithubAuthBackend tests:  [github_auth_backend_acc_output.txt](https://github.com/hashicorp/terraform-provider-vault/files/12491882/github_auth_backend_acc_output.txt)
```
$ make testacc TESTARGS='-run=TestAccGithubAuthBackend*'
```

Results file for the GcpAuthBackend* tests:  [gcp_auth_backend_acc_output.txt](https://github.com/hashicorp/terraform-provider-vault/files/12491892/gcp_auth_backend_acc_output.txt)
```
$ make testacc TESTARGS='-run=TestGCPAuthBackend*'
```

[...]

PR #2110, opened 2023-12-13
Fixes failing tests for `make test` run in build step

vinay-gopalan:
This PR adds missing `testutil.SkipTestAcc(t)` to a few tests that were updated as a part of #1997. The build step was previously failing since these tests were not being skipped when running `make test` (these should be skipped and only run on `make testacc`). This fixes the build

### DIFF
--- a/vault/resource_gcp_auth_backend_test.go
+++ b/vault/resource_gcp_auth_backend_test.go
@@ -32,7 +32,7 @@ const gcpJSONCredentials string = `
 `
 
 func TestGCPAuthBackend_basic(t *testing.T) {
-	testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)
+	testutil.SkipTestAcc(t)
 
 	var resAuthFirst api.AuthMount
 	path := resource.PrefixedUniqueId("gcp-basic-")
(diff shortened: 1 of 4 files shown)


vault/resource_gcp_auth_backend_test.go
```diff
@@ -32,7 +32,7 @@ const gcpJSONCredentials string = `
 `
 
 func TestGCPAuthBackend_basic(t *testing.T) {
-	testutil.SkipTestEnvSet(t, testutil.EnvVarSkipVaultNext)
```
vinay-gopalan, line 35: This test no longer needs this skip step since all features being tested have been released. Since I was in the code I went ahead and removed it

vault/resource_github_auth_backend_test.go
```diff
@@ -90,9 +91,7 @@ func TestAccGithubAuthBackend_ns(t *testing.T) {
 			{
 				Config: testAccGithubAuthBackendConfig_ns(ns, path, testGHOrg),
 				Check: resource.ComposeTestCheckFunc(
-					testutil.TestAccCheckAuthMountExists(resourceName,
-						&resAuth,
-						testProvider.Meta().(*provider.ProviderMeta).MustGetClient()),
+					githubAuthMountExistsHelperNS(resourceName, &resAuth),
```
vinay-gopalan, line 94: This test was not getting the correct namespaced client; a wrapper method that fetches the Vault client in the correct namespace has been added, and the test now passes in CI

fairclothjm (approved):

PR #2146, opened 2024-02-16
Fix unhandled error in d/azure_access_creds

benashz:
We were not handling the error from `cloudConfigFromName()` when reading azure access credentials.

Refactor some of the code, and add tests for `getAzureCloudConfigFromName()` which was renamed from `cloudConfigFromName()`


vinay-gopalan (approved):
Good catch, thanks for also including a test for the invalid case!

PR #2151, opened 2024-02-21
Fix documentation and remove erroneous credential parameters

kpcraig:
followup to: https://github.com/hashicorp/terraform-provider-vault/pull/2128


fairclothjm (approved):
Thanks!

Issue #2061, opened 2023-10-20
[Bug]: 


mackill2:
### Terraform Core Version

1.3.3.

### Terraform Vault Provider Version

3.21.0

### Vault Server Version

1.13.5+ent.shm

### Affected Resource(s)

vault_pki_secret_backend_issuer

  update for this resource errase parameters like name or  issuing_certificates\

 one CI/CD 
  # module.vault-pki-issuers-hard["pki-test-1"].module.intermediate["pki-test-intermediate"].vault_pki_secret_backend_issuer.intermediate["v.0.0.3"] will be updated in-place
  ~ resource "vault_pki_secret_backend_issuer" "intermediate" {
        id                        = "pki-int-1/issuer/2242fd5c-9214-068c-910f-f57632235de6"
      ~ issuing_certificates      = [
          + "http://pki-st.pse.pl/v1/pki-int-1/issuer/v.0.0.3/pem",
        ]
        # (10 unchanged attributes hidden)

Second one PIPELINE ( without changing resource parametrs )

  # module.vault-pki-issuers-hard["pki-test-1"].module.intermediate["pki-test-intermediate"].vault_pki_secret_backend_issuer.intermediate["v.0.0.3"] will be updated in-place
  ~ resource "vault_pki_secret_backend_issuer" "intermediate" {
        id                        = "pki-int-1/issuer/2242fd5c-9214-068c-910f-f57632235de6"
      + issuer_name               = "v.0.0.3"
        # (10 unchanged attributes hidden)
    }

### Expected Behavior

Don't errase resource parameters

### Actual Behavior

Errase  updated resource parameters

### Terraform Configuration Files

data

pki:
  pki-test-1:
    path: "pki-1"
    description: "Testowe PKI CA Main"
    default_lease_ttl_seconds: "31536000"
    max_lease_ttl_seconds: "157680000"
    issuing_certificates:
      - "http://127.0.0.1:8200/v1/pki/ca"
    crl_distribution_points:
      - "http://127.0.0.1:8200/v1/pki/crl"
    crl_build_disable: false
    crl_expiry: "72h"
    root_cert_issuers:
      - issuer_name: "v.0.0.2"
        managed: true
        default: false
        type: "internal"
        common_name: "Testowe PKI CA Main"
        ttl: "365d"
        format: "pem"
        private_key_format: "der"
        key_type: "rsa"
        key_bits: 4096  
        exclude_cn_from_sans: true        
[...]

PR #2186, opened 2024-03-12
Fix PKI Issuer overwrites on updates

### LINKED ISSUE #1990: vault_pki_secret_backend_issuer overwrites existing unchanged attributes when applying unrelated attribute changes
### Description

When individual attributes change on a vault_pki_secret_backend_issuer resource only changed attributes are sent in the API call to update a issuer, however the call is made with PUT (synonymous with POST). 

Per the documentation, this overwrites values and this is the observed behavior.   It should be a PATCH.  

Observed in 3.19, `crl_distribution_points` is overwritten when setting an `issuer_name.`   The inverse occurs on a subsequent plan & apply.

```

> vault read pki/issuer/:issuer_ref

  # truncated output
  crl_distribution_points           [http://example.com/crl.crl]
  enable_aia_url_templating         false
  issuer_id                         :issuer_ref
  issuer_name                       n/a

> terraform plan


  ~ resource "vault_pki_secret_backend_issuer" "pki_issuer" {
        id                        = "pki/issuer/:issuer_ref"
      + issuer_name               = "my_issuer_name"
        # (11 unchanged attributes hidden)
    }

> terraform apply
...
> vault read pki/issuer/:issuer_ref

  crl_distribution_points           []
  issuer_id                         :issuer_ref
  issuer_name                       my_issuer_name
  issuing_certificates              []
```

### References

 https://developer.hashicorp.com/vault/api-docs/secret/pki#update-issuer
https://github.com/hashicorp/terraform-provider-vault/blob/91a56a478b810d5444c7acc1a0cbeef70cc27e1f/vault/resource_pki_secret_backend_issuer.go#L178

None

vinay-gopalan:
### Description
Previously, the resource was not employing a JSON Merge Patch when making update requests to the Vault server. This was causing all data to be overwritten for the issuer. This fix follows the Vault documentation and makes a `PATCH` request to the issuer API when sending update info. [Vault docs](https://developer.hashicorp.com/vault/api-docs/secret/pki#update-issuer)

Relates OR Closes #1990 #2061 

### Checklist

```
$ make testacc TESTARGS='-run=TestAccPKISecretBackendIssuer_basic'
=== RUN   TestAccPKISecretBackendIssuer_basic
    resource_pki_secret_backend_issuer_test.go:29: Vault server version "1.17.0-beta1+ent"
--- PASS: TestAccPKISecretBackendIssuer_basic (3.82s)
PASS
```

### DIFF
--- a/vault/resource_pki_secret_backend_issuer.go
+++ b/vault/resource_pki_secret_backend_issuer.go
@@ -175,9 +175,16 @@ func pkiSecretBackendIssuerUpdate(ctx context.Context, d *schema.ResourceData, m
 
 	// only write to Vault if a patch is required
 	if patchRequired {
-		_, err := client.Logical().WriteWithContext(ctx, path, data)
-		if err != nil {
-			return diag.Errorf("error writing data to %q, err=%s", path, err)
+		if d.IsNewResource() {
+			_, err := client.Logical().WriteWithContext(ctx, path, data)
+			if err != nil {
+				return diag.Errorf("error writing issuer data to %q, err=%s", path, err)
+			}
+		} else {
+			_, err := client.Logical().JSONMergePatch(ctx, path, data)
+			if err != nil {
+				return diag.Errorf("error updating issuer data at %q, err=%s", path, err)
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


vault/resource_pki_secret_backend_issuer.go
```diff
@@ -175,9 +175,16 @@ func pkiSecretBackendIssuerUpdate(ctx context.Context, d *schema.ResourceData, m
...
-			return diag.Errorf("error writing data to %q, err=%s", path, err)
+		if d.IsNewResource() {
+			_, err := client.Logical().WriteWithContext(ctx, path, data)
+			if err != nil {
+				return diag.Errorf("error writing issuer data to %q, err=%s", path, err)
+			}
+		} else {
+			_, err := client.Logical().JSONMergePatch(ctx, path, data)
+			if err != nil {
+				return diag.Errorf("error updating issuer data at %q, err=%s", path, err)
```
benashz, line 186: Would be great if we could some how extend the test coverage to include these sorts of error conditions.

benashz (approved):
LGTM!

vault/resource_pki_secret_backend_issuer.go
vinay-gopalan: Good call, added an extra test step to validate the error conditions. Thanks!

PR #2235, opened 2024-05-08
Add missing parameters to `vault_mount`

vinay-gopalan:
### Description
Adds parameters to the resource that were missing, including newly 1.16 added fields.

### Checklist

```

...
```

### DIFF
--- a/vault/resource_mount.go
+++ b/vault/resource_mount.go
@@ -455,9 +455,11 @@ func readMount(d *schema.ResourceData, meta interface{}, excludeType bool) error
 	if err := d.Set(consts.FieldAllowedResponseHeaders, mount.Config.AllowedResponseHeaders); err != nil {
 		return err
 	}
-	if err := d.Set(consts.FieldDelegatedAuthAccessors, mount.Config.DelegatedAuthAccessors); err != nil {
-		return err
-	}
+
+	// @TODO add this back in when Vault 1.16.3 is released
+	//if err := d.Set(consts.FieldDelegatedAuthAccessors, mount.Config.DelegatedAuthAccessors); err != nil {
+	//	return err
+	//}
 	if err := d.Set(consts.FieldListingVisibility, mount.Config.ListingVisibility); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


fairclothjm (approved):
LGTM

vault/resource_mount.go
```diff
@@ -236,8 +305,32 @@ func updateMount(d *schema.ResourceData, meta interface{}, excludeType bool) err
 		path = newPath
 	}
 
-	if d.HasChange("allowed_managed_keys") {
-		config.AllowedManagedKeys = expandStringSlice(d.Get("allowed_managed_keys").(*schema.Set).List())
+	if d.HasChange(consts.FieldAllowedManagedKeys) {
```
benashz, line 308: Out of curiosity, is there a reason we are conditionally setting these new parameters, rather than always setting them?

benashz (commented):
I had one question about this change, other than that it looks pretty straightforward. 

It does look like the tests are not passing e.g: https://github.com/hashicorp/terraform-provider-vault/actions/runs/8994620751/job/24708374987?pr=2235#step:9:1766 

Will you also be doing the docs update under this PR?

vault/resource_mount.go
vinay-gopalan: It's mostly because they all have varied minimum version requirements. We can't always set them since some fields are only enabled on certain Vault versions.

The idea is that all these params will only be added in if the user provides them / makes a change to them. The expectation is that if a user is providing us a field to use, then they are aware of their Vault Server and its API features/limitations. We can also individually add version checks for all the older versions to the parameters,I figured this would be easier and less "If this version but less than that version do this, else that"

vinay-gopalan:
> I had one question about this change, other than that it looks pretty straightforward.
> 
> It does look like the tests are not passing e.g: https://github.com/hashicorp/terraform-provider-vault/actions/runs/8994620751/job/24708374987?pr=2235#step:9:1766
> 
> Will you also be doing the docs update under this PR?

Good catch! Docs slipped my mind and I will add them in. And yes! The tests failures are expected, looks like the TFVP helped uncover a bug in a newly added Vault 1.16.x field. There is a [fix PR up on the Vault repo](https://github.com/hashicorp/vault/pull/26876). For now, I'll add a `TODO` note to fix that test, and we can update that test once the fix is merged into Vault. Thanks for the thoroughness!

fairclothjm (approved):
LGTM! Thanks for doing this!

PR #2246, opened 2024-05-24
Add PKI EST configuration support

stevendpclark:
### Description

This PR adds new data source and resource types for the [PKI EST configuration API](https://developer.hashicorp.com/vault/api-docs/secret/pki#set-est-configuration). The PR is built on top of https://github.com/hashicorp/terraform-provider-vault/pull/2235 as EST support requires the new mount fields.

This also requires Vault 1.16.3+ent at a minimum as the following fixes are required for this to work correctly

-  https://github.com/hashicorp/vault/pull/26876
-  https://github.com/hashicorp/vault-enterprise/pull/5785

<details>

<summary>Sample TF script used for testing (along with the added tests)</summary>

```hcl
variable "pki_int_path" {
  type    = string
  default = "pki-int"
}

variable "est_pass" {
  type    = string
  default = "estpass"
}

resource "vault_mount" "pki_root" {
  path        = "pki-root"
  type        = "pki"
  description = "PKI root mount"

  max_lease_ttl_seconds = "31536000"
}

resource "vault_pki_secret_backend_key" "root_key" {
  backend  = vault_mount.pki_root.path
  type     = "internal"
  key_name = "root-ca-key"
  key_type = "rsa"
  key_bits = 4096
}

resource "vault_pki_secret_backend_root_cert" "root_cert" {
  backend              = vault_mount.pki_root.path
  type                 = "existing"
  key_ref              = vault_pki_secret_backend_key.root_key.key_name
  common_name          = "Root CA"
  ttl                  = "364d"
  format               = "pem"
  exclude_cn_from_sans = true
  ou                   = "My OU"
  organization         = "My organization"
}

resource "vault_auth_backend" "cert" {
  type = "cert"

  tune {
    max_lease_ttl = "10s"
  }
}

  type = "userpass"

  tune {
    max_lease_ttl = "10s"
  }
}

resource "vault_policy" "est_policy" {
  name   = "est-policy"
  policy = <<EOT
path "${var.pki_int_path}/est/*" {
  capabilities = ["read", "update", "create"]
}

path "${var.pki_int_path}/roles/est-clients/*" {
  capabilities = ["read", "update", "create"]
}
EOT
}

resource "vault_cert_auth_backend_role" "cert" {
  backend = vault_auth_backend.cert.path

  name           = "est-ca"
  display_name   = "EST Client CA"
  certificate    = vault_pki_secret_backend_root_cert.root_cert.certificate
  token_policies = [vault_policy.est_policy.name]
  token_type     = "batch"
}

resource "vault_generic_endpoint" "est_user" {
  path                 = format("auth/%s/users/estuser", vault_auth_backend.userpass.path)
  ignore_absent_fields = true

  data_json = <<EOT
{
  "policies": ["${vault_policy.est_policy.name}"],
  "password": "${var.est_pass}"
}
EOT
}

resource "vault_mount" "int" {
  path        = var.pki_int_path
  type        = "pki"
  description = "PKI EST mount"

  delegated_auth_accessors = [vault_auth_backend.userpass.accessor, vault_auth_backend.cert.accessor]
  allowed_response_headers = ["Content-Transfer-Encoding", "Content-Length", "WWW-Authenticate"]
  max_lease_ttl_seconds    = "31536000"
}

resource "vault_pki_secret_backend_key" "int_key" {
  backend  = vault_mount.int.path
  type     = "internal"
  key_name = "int-ca-key"
  key_type = "ec"
}

resource "vault_pki_secret_backend_intermediate_cert_request" "int_est_csr" {
  backend = vault_mount.int.path

  type    = "existing"
  key_ref = vault_pki_secret_backend_key.int_key.key_name

  common_name = "Intermediary EST Authority"
}

resource "vault_pki_secret_backend_root_sign_intermediate" "root_sign_est_intermediary" {
  backend = vault_mount.pki_root.path

  common_name = vault_pki_secret_backend_intermediate_cert_request.int_est_csr.common_name
  csr         = vault_pki_secret_backend_intermediate_cert_request.int_est_csr.csr
  ttl         = "31d"
}

resource "vault_pki_secret_backend_intermediate_set_signed" "int_est" {
  backend = vault_mount.int.path

  certificate = vault_pki_secret_backend_root_sign_intermediate.root_sign_est_intermediary.certificate
}

resource "vault_pki_secret_backend_role" "est_clients" {
  backend = vault_mount.int.path

  name             = "est-clients"
  max_ttl          = "30h"
  allow_ip_sans    = true
  allow_subdomains = true
  allow_any_name   = true
  require_cn       = false
  no_store         = true
}

resource "vault_pki_secret_backend_config_est" "est_config" {
  backend = vault_mount.int.path

  enabled             = true
  default_mount       = true
  default_path_policy = format("role:%s", vault_pki_secret_backend_role.est_clients.name)
  label_to_path_policy = {
    "test-label" : format("role:%s", vault_pki_secret_backend_role.est_clients.name),
    "sign-all" : "sign-verbatim"
  }
  authenticators {
    cert = {
      accessor  = vault_auth_backend.cert.accessor
      cert_role = vault_cert_auth_backend_role.cert.name
    }
    userpass = {
      accessor = vault_auth_backend.userpass.accessor
    }
  }
  enable_sentinel_parsing = true
  audit_fields = ["csr", "common_name", "alt_names", "ip_sans", "uri_sans", "other_sans",
    "signature_bits", "exclude_cn_from_sans", "ou", "organization", "country",
    "locality", "province", "street_address", "postal_code", "serial_number",
  "use_pss", "key_type", "key_bits", "add_basic_constraints"]
}
```

</details>

### Checklist

```

...
```

### Community Note

### DIFF
--- a/vault/data_source_pki_secret_backend_config_est.go
+++ b/vault/data_source_pki_secret_backend_config_est.go
@@ -27,22 +27,22 @@ func pkiSecretBackendConfigEstDataSource() *schema.Resource {
 			consts.FieldEnabled: {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Is the EST feature enabled",
+				Description: "Specifies whether EST is enabled",
 			},
 			consts.FieldDefaultMount: {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Is this mount providing the cluster's default EST mount",
+				Description: "If set, this mount is registered as the default `.well-known/est` URL path. Only a single mount can enable this across a Vault cluster",
 			},
 			consts.FieldDefaultPathPolicy: {
 				Type:        schema.TypeString,
 				Computed:    true,
-				Description: "The behavior of the default_mount when enabled",
+				Description: "Required to be set if default_mount is enabled. Specifies the behavior for requests using the default EST label. Can be sign-verbatim or a role given by role:<role_name>",
 			},
 			consts.FieldLabelToPathPolicy: {
 				Type:        schema.TypeMap,
 				Computed:    true,
-				Description: "A pairing of EST label to the configured EST behavior for it",
+				Description: "A pairing of an EST label with the redirected behavior for requests hitting that role. The path policy can be sign-verbatim or a role given by role:<role_name>. Labels must be unique across Vault cluster, and will register .well-known/est/<label> URL paths",
 			},
 			consts.FieldAuthenticators: {
 				Type:        schema.TypeList,
@@ -66,7 +66,7 @@ func pkiSecretBackendConfigEstDataSource() *schema.Resource {
 			consts.FieldEnableSentinelParsing: {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Are fields from the provided CSR parsed out for Sentinel policies",
+				Description: "If set, parse out fields from the provided CSR making them available for Sentinel policies",
 			},
 			consts.FieldAuditFields: {
 				Type:        schema.TypeList,
--- a/vault/resource_pki_secret_backend_config_est.go
+++ b/vault/resource_pki_secret_backend_config_est.go
@@ -30,28 +30,28 @@ func pkiSecretBackendConfigEstResource() *schema.Resource {
 			consts.FieldBackend: {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "The PKI secret backend the resource belongs to.",
+				Description: "The PKI secret backend the resource belongs to",
 				ForceNew:    true,
 			},
 			consts.FieldEnabled: {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "Is the EST feature enabled",
+				Description: "Specifies whether EST is enabled",
 			},
 			consts.FieldDefaultMount: {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "Is this the cluster's default EST mount",
+				Description: "If set, this mount will register the default `.well-known/est` URL path. Only a single mount can enable this across a Vault cluster",
 			},
 			consts.FieldDefaultPathPolicy: {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "The behavior of the default_mount when enabled",
+				Description: "Required to be set if default_mount is enabled. Specifies the behavior for requests using the default EST label. Can be sign-verbatim or a role given by role:<role_name>",
 			},
 			consts.FieldLabelToPathPolicy: {
 				Type:        schema.TypeMap,
 				Optional:    true,
-				Description: "A pairing of EST label to the configured EST behavior for it",
+				Description: "Configures a pairing of an EST label with the redirected behavior for requests hitting that role. The path policy can be sign-verbatim or a role given by role:<role_name>. Labels must be unique across Vault cluster, and will register .well-known/est/<label> URL paths",
 			},
 			consts.FieldAuthenticators: {
 				Type:        schema.TypeList,
@@ -75,7 +75,7 @@ func pkiSecretBackendConfigEstResource() *schema.Resource {
 			consts.FieldEnableSentinelParsing: {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "Enable parsing of fields from the provided CSR for Sentinel policies",
+				Description: "If set, parse out fields from the provided CSR making them available for Sentinel policies",
 			},
 			consts.FieldAuditFields: {
 				Type:        schema.TypeList,
--- a/vault/resource_pki_secret_backend_config_est_test.go
+++ b/vault/resource_pki_secret_backend_config_est_test.go
@@ -177,8 +177,7 @@ resource "vault_pki_secret_backend_config_est" "test" {
 }
 
 data "vault_pki_secret_backend_config_est" "test" {
-  depends_on = [vault_pki_secret_backend_config_est.test]
-  backend = vault_mount.test.path
+  backend = vault_pki_secret_backend_config_est.test.backend	
 }
 `, pkiPath)
 }
@@ -196,8 +195,7 @@ resource "vault_pki_secret_backend_config_est" "test" {
 }
 
 data "vault_pki_secret_backend_config_est" "test" {
-  depends_on = [vault_pki_secret_backend_config_est.test]
-  backend = vault_mount.test.path
+  backend = vault_pki_secret_backend_config_est.test.backend	
 }
 `, path)
 }
--- a/website/docs/d/pki_secret_backend_config_est.html.md
+++ b/website/docs/d/pki_secret_backend_config_est.html.md
@@ -46,25 +46,25 @@ The following arguments are supported:
  
 ## Attributes Reference
 
-* `authenticators` - Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators))
+* `authenticators` - Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators)).
  
-* `default_mount` - Is this mount providing the cluster's default EST mount
+* `default_mount` -  If set, this mount is registered as the default `.well-known/est` URL path. Only a single mount can enable this across a Vault cluster.
  
-* `default_path_policy` - The behavior of the default_mount when enabled
+* `default_path_policy` - Required to be set if default_mount is enabled. Specifies the behavior for requests using the default EST label. Can be sign-verbatim or a role given by role:<role_name>.
  
-* `enable_sentinel_parsing` - Are fields from the provided CSR parsed out for Sentinel policies
+* `enable_sentinel_parsing` - If set, parse out fields from the provided CSR making them available for Sentinel policies.
+
+* `enabled` - Specifies whether EST is enabled.
  
-* `enabled` - Is the EST feature enabled
+* `label_to_path_policy` - A pairing of an EST label with the redirected behavior for requests hitting that role. The path policy can be sign-verbatim or a role given by role:<role_name>. Labels must be unique across Vault cluster, and will register .well-known/est/<label> URL paths.
  
-* `label_to_path_policy` - A pairing of EST label to the configured EST behavior for it
+* `audit_fields` - Fields parsed from the CSR that appear in the audit and can be used by sentinel policies.
  
-* `audit_fields` - Fields parsed from the CSR that appear in the audit and can be used by sentinel policies
- 
-* `last_updated` - A read-only timestamp representing the last time the configuration was updated
+* `last_updated` - A read-only timestamp representing the last time the configuration was updated.
 
 <a id="nestedatt--authenticators"></a>
 ### Nested Schema for `authenticators`
 
-* `cert` - "The accessor and cert_role properties for cert auth backends"
+* `cert` - "The accessor and cert_role properties for cert auth backends".
  
-* `userpass` - "The accessor property for user pass auth backends"
+* `userpass` - "The accessor property for user pass auth backends".
--- a/website/docs/r/pki_secret_backend_config_est.html.md
+++ b/website/docs/r/pki_secret_backend_config_est.html.md
@@ -14,9 +14,9 @@ Allows setting the EST configuration on a PKI Secret Backend
 
 ```hcl
 resource "vault_mount" "pki" {
-	path        = "pki-root"
-	type        = "pki"
-    description = "PKI secret engine mount"
+  path        = "pki-root"
+  type        = "pki"
+  description = "PKI secret engine mount"
 }
 
 resource "vault_pki_secret_backend_role" "est_role" {
@@ -73,30 +73,30 @@ The following arguments are supported:
 * `backend` - (Required) The path to the PKI secret backend to
   read the EST configuration from, with no leading or trailing `/`s.
 
-* `authenticators` - (Optional) Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators))
+* `authenticators` - (Optional) Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators)).
 
-* `default_mount` - Is this mount providing the cluster's default EST mount
+* `default_mount` - (Optional) If set, this mount will register the default `.well-known/est` URL path. Only a single mount can enable this across a Vault cluster.
 
-* `default_path_policy` - (Optional) The behavior of the default_mount when enabled
+* `default_path_policy` - (Optional) Required to be set if default_mount is enabled. Specifies the behavior for requests using the default EST label. Can be sign-verbatim or a role given by role:<role_name>.
 
-* `enable_sentinel_parsing` - (Optional) Are fields from the provided CSR parsed out for Sentinel policies
+* `enable_sentinel_parsing` - (Optional) If set, parse out fields from the provided CSR making them available for Sentinel policies.
 
-* `enabled` - (Required) Is the EST feature enabled
+* `enabled` - (Optional) Specifies whether EST is enabled.
 
-* `label_to_path_policy` - (Optional) A pairing of EST label to the configured EST behavior for it
+* `label_to_path_policy` - (Optional) Configures a pairing of an EST label with the redirected behavior for requests hitting that role. The path policy can be sign-verbatim or a role given by role:<role_name>. Labels must be unique across Vault cluster, and will register .well-known/est/<label> URL paths.
 
-* `audit_fields` - (Optional) Fields parsed from the CSR that appear in the audit and can be used by sentinel policies
+* `audit_fields` - (Optional) Fields parsed from the CSR that appear in the audit and can be used by sentinel policies.
 
 <a id="nestedatt--authenticators"></a>
 ### Nested Schema for `authenticators`
 
-* `cert` - "The accessor (required) and cert_role (optional) properties for cert auth backends"
+* `cert` - "The accessor (required) and cert_role (optional) properties for cert auth backends".
 
-* `userpass` - "The accessor (required) property for user pass auth backends"
+* `userpass` - "The accessor (required) property for user pass auth backends".
 
 ## Attributes Reference
 
-* `last_updated` - A read-only timestamp representing the last time the configuration was updated
+* `last_updated` - A read-only timestamp representing the last time the configuration was updated.
 
 ## Import
 


vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,181 @@
...
+	d.SetId(path)
+
+	return pkiSecretBackendConfigEstRead(ctx, d, meta)
+}
+
+func pkiSecretBackendConfigEstRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return readPKISecretBackendConfigEst(ctx, d, meta)
+}
+
+func pkiSecretBackendConfigEstDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
vinay-gopalan, line 151: Looks like there is no `DELETE` method on this endpoint (makes sense, config endpoints in Vault usually don't). In that case we can just `return nil` for this method, akin to what is done for the [KVV2 Config resource](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_kv_secret_backend_v2.go#L135-L137)

We let users control the updates on the configs (which includes resets, a form of update in this case) by updating their TF config file. This keeps it in line w/ how a user would expect to perform the reset in Vault (a `vault write pki/config/est` as opposed to `vault delete pki/config/est`).

This also enables the user to keep track of their reset EST config in the TF state. This method is called upon a `terraform destroy`, so TF would also nuke the state of the EST config resource when it gets reset (unless there's an actual removal of the resource in Vault, TF practice would be to manage the lifecycle for as long as possible 😄 ).

vault/data_source_pki_secret_backend_config_est_test.go
```diff
@@ -0,0 +1,49 @@
...
+			testutil.TestAccPreCheck(t)
+			testutil.TestEntPreCheck(t)
+			SkipIfAPIVersionLT(t, testProvider.Meta(), provider.VaultVersion116)
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testPKISecretEmptyEstConfigDataSource(backend),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(dataName, consts.FieldBackend, backend),
+					resource.TestCheckResourceAttrSet(dataName, consts.FieldEnabled),
```
vinay-gopalan, line 29: nit: can we validate that we get the expected defaults from Vault for these fields?
```suggestion
					resource.TestCheckResourceAttr(dataName, consts.FieldEnabled, "false"),
```

vault/data_source_pki_secret_backend_config_est_test.go
```diff
@@ -0,0 +1,49 @@
...
+					resource.TestCheckResourceAttrSet(dataName, consts.FieldDefaultMount),
+					resource.TestCheckResourceAttrSet(dataName, consts.FieldEnableSentinelParsing),
```
vinay-gopalan, line 31: Looks like Audit Fields also has a list of defaults. Is it worth checking them here?

vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,181 @@
...
+func pkiSecretBackendConfigEstResource() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Manages Vault PKI EST configuration",
+		CreateContext: provider.MountCreateContextWrapper(pkiSecretBackendConfigEstWrite, provider.VaultVersion116),
+		UpdateContext: pkiSecretBackendConfigEstWrite,
+		ReadContext:   pkiSecretBackendConfigEstRead,
+		DeleteContext: pkiSecretBackendConfigEstDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: func(_ context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				id := d.Id()
```
vinay-gopalan, line 29: My understanding is the main reason for adding this bit is that both the data source and the resource are sharing the same `readPKISecretBackendConfigEst` method. While I see the benefits of having a common util, I think splitting them out a little might be beneficial for us in the long run. In my experience, it's always better to try and avoid writing custom importers as much as possible. They sometimes behave unexpectedly in edge cases where users input weird things 😅 

Once things get garbled in the TF state for users in production, it's hard for us to rollback these changes (especially with a field like `backend`, which is a `ForceNew` and would attempt to destroy and recreate)

vinay-gopalan: My suggestion would be:
- For the data source, verify the Vault server version requirements. We also can always expect that the `backend` field will be provided (`d.GetOk` is allowed)
- For the resource, we don't need to verify the Vault server version requirements again (already taken care of by `MountCreateContextWrapper`). For the read method, we should extract the `backend` parameter from the `d.Id()` instead. I would suggest stripping the end `/config/est` instead of splitting the string on a `/`. We can never expect how many slashes users might put into their input values, so this can lead to incorrect TF states

Typically for Read methods in resources, we try not to have any `d.Get` commands in order to make the import work seamlessly. We should try to construct our IDs and resources in a way where all relevant info for the TF state is available through the read method.

Let me know if this makes sense! Happy to chat about this further if you had other thoughts :)

vinay-gopalan (commented):
Looking great! Thanks for working on this, had a few comments/suggestions :)

vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,181 @@
...
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func pkiSecretBackendConfigEstWrite(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if err := verifyPkiEstFeatureSupported(meta); err != nil {
```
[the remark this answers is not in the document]
stevendpclark: This was added to verify that we were talking to an Enterprise version of Vault, but at the same time I also added the version. 

To my knowledge there isn't a way within `provider.MountCreateContextWrapper` to make it check for a version and the enterprise version?

vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,181 @@
...
+		Schema: map[string]*schema.Schema{
+			consts.FieldBackend: {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The PKI secret backend the resource belongs to.",
+				ForceNew:    true,
+			},
+			consts.FieldEnabled: {
+				Type:        schema.TypeBool,
+				Required:    true,
```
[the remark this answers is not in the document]
stevendpclark: Fair, I was thinking more along the lines that it didn't really make sense from the provider's point of view to leave this as optional but it makes sense to follow the API itself.

vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,181 @@
...
+		consts.FieldEnabled:               d.Get(consts.FieldEnabled).(bool),
+		consts.FieldDefaultMount:          d.Get(consts.FieldDefaultMount).(bool),
+		consts.FieldDefaultPathPolicy:     d.Get(consts.FieldDefaultPathPolicy).(string),
+		consts.FieldLabelToPathPolicy:     d.Get(consts.FieldLabelToPathPolicy).(map[string]interface{}),
+		consts.FieldAuthenticators:        authenticator,
+		consts.FieldEnableSentinelParsing: d.Get(consts.FieldEnableSentinelParsing).(bool),
+		consts.FieldAuditFields:           d.Get(consts.FieldAuditFields).([]interface{}),
+	}
+
+	log.Printf("[DEBUG] Updating EST config on PKI secret backend %q:\n%v", backend, data)
```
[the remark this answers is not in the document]
stevendpclark: Nope none of the fields are carry sensitive info.

vault/data_source_pki_secret_backend_config_est_test.go
stevendpclark: This and the similar comment about AuditFields below, do we want the TFVP tests to start failing if Vault changes the default values? Seems a little brittle which is why these are a minimal set based checks but the full suite is validated within the resource check when the test actually sets all the values. 

If you feel strongly I have no issue adding them in, or maybe add a comment in this saying this is more thoroughly tested within the `TestAccPKISecretBackendConfigEst_basic` test?

vault/resource_pki_secret_backend_config_est.go
stevendpclark: I think I understand have a look at https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/8183fe41f06467d6c25bb1f4b6484bbfeff8bf2e, I'm assuming we still need to keep the requirements check due to the enterprise requirement.

vault/data_source_pki_secret_backend_config_est_test.go
stevendpclark: See https://github.com/hashicorp/terraform-provider-vault/pull/2246#discussion_r1616052909

vault/resource_pki_secret_backend_config_est.go
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/8183fe41f06467d6c25bb1f4b6484bbfeff8bf2e

vault/resource_pki_secret_backend_config_est.go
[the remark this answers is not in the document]
stevendpclark: Marked as Optional in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/8183fe41f06467d6c25bb1f4b6484bbfeff8bf2e

vault/data_source_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,165 @@
...
+				Description: "Is the EST feature enabled",
+			},
+			consts.FieldDefaultMount: {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Is this mount providing the cluster's default EST mount",
+			},
+			consts.FieldDefaultPathPolicy: {
+				Type:        schema.TypeString,
+				Computed:    true,
```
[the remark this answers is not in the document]
stevendpclark: No worries, but that is the reason on [FieldAuditFields](https://github.com/hashicorp/terraform-provider-vault/pull/2246/files#diff-545444cb362211ba72a83bfa1a1cf804f0ad1cfe553907bdcedc12eaba30ab27R83) within the resource Computed is set, we get back a default set of values from Vault if not set.

stevendpclark:
Updated PR to the latest version of main as https://github.com/hashicorp/terraform-provider-vault/pull/2235 was merged, along with disabling a test to workaround an unreleased bug (will be in 1.16.3) in Vault Enterprise https://github.com/hashicorp/vault-enterprise/pull/5785

vault/resource_pki_secret_backend_config_est.go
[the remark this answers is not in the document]
vinay-gopalan: Ah, I see what you're saying here, @stevendpclark! I realize that ideally `provider.MountCreateContextWrapper` should also perform enterprise checks if needed. Since it may be out of scope for this PR, I'm fine moving ahead with the manual checks you have implemented here. I can add in an update in a follow-up PR to add the enterprise checks :)

vault/data_source_pki_secret_backend_config_est_test.go
vinay-gopalan: Definitely a nit, don't feel strongly about it! I'm happy with the comment you have added

vault/resource_pki_secret_backend_config_est_test.go
```diff
@@ -0,0 +1,203 @@
...
+  enable_sentinel_parsing = true
+  audit_fields = ["csr", "common_name", "alt_names", "ip_sans", "uri_sans", "other_sans",
+                  "signature_bits", "exclude_cn_from_sans", "ou", "organization", "country", 
+                  "locality", "province", "street_address", "postal_code", "serial_number",
+                  "use_pss", "key_type", "key_bits", "add_basic_constraints"]
+}
+
+data "vault_pki_secret_backend_config_est" "test" {
+  depends_on = [vault_pki_secret_backend_config_est.test]
+  backend = vault_mount.test.path
```
vinay-gopalan, line 181: nit: since the `backend` field for the config resource is also pointing to the `vault_mount` path, we can have a cleaner dependency graph and remove `depends_on` from the line above:
```suggestion
  backend = vault_pki_secret_backend_config_est.test.backend
```

vault/resource_pki_secret_backend_config_est_test.go
```diff
@@ -0,0 +1,203 @@
...
+	type        = "pki"
+    description = "PKI secret engine mount"
+}
+
+resource "vault_pki_secret_backend_config_est" "test" {
+  backend = vault_mount.test.path
+}
+
+data "vault_pki_secret_backend_config_est" "test" {
+  depends_on = [vault_pki_secret_backend_config_est.test]
```
vinay-gopalan, line 199: nit: same as above on dependency graph

vault/resource_pki_secret_backend_config_est.go
```diff
@@ -0,0 +1,178 @@
...
+		return diag.Errorf("error updating EST config for PKI secret backend %q: %s", backend, err)
+	}
+	log.Printf("[DEBUG] Updated EST config on PKI secret backend %q", backend)
+
+	d.SetId(path)
+
+	return pkiSecretBackendConfigEstRead(ctx, d, meta)
+}
+
+func pkiSecretBackendConfigEstRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
vinay-gopalan, line 149: Thanks for separating the read methods!

website/docs/r/pki_secret_backend_config_est.html.md
```diff
@@ -0,0 +1,109 @@
...
+
+Allows setting the EST configuration on a PKI Secret Backend
+
+## Example Usage
+
+```hcl
+resource "vault_mount" "pki" {
+	path        = "pki-root"
+	type        = "pki"
+    description = "PKI secret engine mount"
```
vinay-gopalan, line 19: nit: indentation
```suggestion
	description = "PKI secret engine mount"
```

website/docs/r/pki_secret_backend_config_est.html.md
```diff
@@ -0,0 +1,109 @@
...
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault/index.html#namespace).
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Required) The path to the PKI secret backend to
+  read the EST configuration from, with no leading or trailing `/`s.
+
+* `authenticators` - (Optional) Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators))
+
+* `default_mount` - Is this mount providing the cluster's default EST mount
```
vinay-gopalan, line 78: nit: Ideally we try to match up with Vault's documentation for the field as much as possible
```suggestion
* `default_mount` - (Optional) If set, this mount will register the default `.well-known/est` URL path. Only a single mount can enable this across a Vault cluster
```

website/docs/r/pki_secret_backend_config_est.html.md
```diff
@@ -0,0 +1,109 @@
...
+
+* `default_path_policy` - (Optional) The behavior of the default_mount when enabled
+
+* `enable_sentinel_parsing` - (Optional) Are fields from the provided CSR parsed out for Sentinel policies
```
vinay-gopalan, line 82: nit: in line with above
```suggestion
* `enable_sentinel_parsing` - (Optional) If set,  parse out fields from the provided CSR making them available for Sentinel policies.
```

website/docs/r/pki_secret_backend_config_est.html.md
```diff
@@ -0,0 +1,109 @@
...
+
+* `enabled` - (Required) Is the EST feature enabled
```
vinay-gopalan, line 84:
```suggestion
* `enabled` - (Required)  Specifies whether EST is enabled.
```

website/docs/d/pki_secret_backend_config_est.html.md
```diff
@@ -0,0 +1,70 @@
...
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Required) The path to the PKI secret backend to
+  read the EST configuration from, with no leading or trailing `/`s.
+ 
+## Attributes Reference
+
+* `authenticators` - Lists the mount accessors EST should delegate authentication requests towards (see [below for nested schema](#nestedatt--authenticators))
+ 
+* `default_mount` - Is this mount providing the cluster's default EST mount
```
vinay-gopalan, line 51: nit: similar comments as resource fields

vinay-gopalan (approved):
PR looks great! Had some comments/suggestions on a few documentation lines, but nothing blocking. Thanks for working on this!

vault/resource_pki_secret_backend_config_est_test.go
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/5f5961a4a31806fc4509223e1234830c7a669590

website/docs/r/pki_secret_backend_config_est.html.md
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/5f5961a4a31806fc4509223e1234830c7a669590

I actually fixed the lines above to match up with the indentation across the rest of the example.

website/docs/r/pki_secret_backend_config_est.html.md
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/5f5961a4a31806fc4509223e1234830c7a669590

website/docs/r/pki_secret_backend_config_est.html.md
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/5f5961a4a31806fc4509223e1234830c7a669590

website/docs/d/pki_secret_backend_config_est.html.md
stevendpclark: Fixed in https://github.com/hashicorp/terraform-provider-vault/pull/2246/commits/5f5961a4a31806fc4509223e1234830c7a669590

vinay-gopalan (approved):

PR #2256, opened 2024-06-04
Add new WIF fields for GCP Auth (Vault Enterprise only)

vinay-gopalan:
### Description
Adds the following new fields to the GCP Auth Backend resource to enable the WIF workflow:

- `identity_token_audience`
- `identity_token_ttl`
- `identity_token_key`
- `service_account_email`

### Checklist

```
$ make testacc TESTARGS='-run=TestGCPAuthBackend_WIF'
=== RUN   TestGCPAuthBackend_WIF
--- PASS: TestGCPAuthBackend_WIF (1.96s)
PASS
```


vault/resource_gcp_auth_backend.go
```diff
@@ -165,37 +190,46 @@ func gcpAuthBackendConfigPath(path string) string {
...
 	log.Printf("[DEBUG] Enabled gcp auth backend %q", path)
 
 	d.SetId(path)
 
-	return gcpAuthBackendUpdate(d, meta)
+	return gcpAuthBackendUpdate(ctx, d, meta)
 }
 
-func gcpAuthBackendUpdate(d *schema.ResourceData, meta interface{}) error {
+func gcpAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
fairclothjm, line 229: maybe out of scope for this PR, but should we pass the context to the Tune call on line 271? I am not sure if we have a helper for that already. We could leave a TODO comment on L271 if we don't want to address it in this PR.

vault/resource_gcp_secret_backend.go
```diff
@@ -131,7 +131,7 @@ func gcpSecretBackendCreate(ctx context.Context, d *schema.ResourceData, meta in
 
 	d.Partial(true)
 	log.Printf("[DEBUG] Mounting GCP backend at %q", path)
-	useAPIVer117Ent := provider.IsAPISupported(meta, provider.VaultVersion117Ent)
+	useAPIVer117Ent := provider.IsAPISupported(meta, provider.VaultVersion117) && provider.IsEnterpriseSupported(meta)
```
fairclothjm, line 134: Was the previous check not sufficient to check for ent support?

vinay-gopalan: @fairclothjm upon testing with different versions I realized that since `IsAPISupported` only checks if a semantic version is greater than the other, the previous check was passing for non-enterprise versions since `v1.17.0` > `v1.16.3+ent`. It looks like the safest bet is to check the enterprise metadata separately from the version comparison

vault/resource_gcp_auth_backend.go
vinay-gopalan: Good call, added a TODO note

Zlaticanin:
Should be add another usage example [here](https://github.com/hashicorp/terraform-provider-vault/blob/8d50bb694555cc113cdbd5941376898b06b0a54e/website/docs/r/gcp_auth_backend.html.md?plain=1#L13) for WIF? Similar how we've done for [others](https://github.com/hashicorp/terraform-provider-vault/blob/0aa919ec5f1b791d3938eee5121eaf27c0341491/website/docs/r/azure_secret_backend.html.md#example-usage-vault-19-and-above)

vinay-gopalan:
> Should be add another usage example [here](https://github.com/hashicorp/terraform-provider-vault/blob/8d50bb694555cc113cdbd5941376898b06b0a54e/website/docs/r/gcp_auth_backend.html.md?plain=1#L13) for WIF? Similar how we've done for [others](https://github.com/hashicorp/terraform-provider-vault/blob/0aa919ec5f1b791d3938eee5121eaf27c0341491/website/docs/r/azure_secret_backend.html.md#example-usage-vault-19-and-above)

Ah thanks for the reminder! Added examples for both Auth and Secrets

vault/resource_gcp_secret_backend.go
fairclothjm: Thanks that makes sense.  After this PR lands are we planning to update the other WIF TFVP resources to make the same check?

fairclothjm: Oh, looks like you may have covered those in this PR already with recent commits.

fairclothjm (approved):

Zlaticanin (approved):
LGTM! 🥳

PR #2270, opened 2024-06-12
Add support for missing `max_retries` param in `vault_aws_auth_backend_client` resource

### LINKED ISSUE #2074: [Enhancement]: Add max_retries to vault_aws_auth_backend_client resource
### Description

The [vault_aws_auth_backend_client](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/aws_auth_backend_client) does not have the `max_retries` argument to configure retry behavior of the AWS SDK as outlined in the [API docs](https://developer.hashicorp.com/vault/api-docs/auth/aws#configure-client). 

This should be added to the resource to align with the AWS client config API `/auth/aws/config/client`

vault_aws_auth_backend_client

```terraform
resource "vault_auth_backend" "example" {
  type = "aws"
}

resource "vault_aws_auth_backend_client" "example" {
  backend     = vault_auth_backend.example.path
  access_key  = "INSERT_AWS_ACCESS_KEY"
  secret_key  = "INSERT_AWS_SECRET_KEY"
  max_retries = 0
}
```

### References

https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/aws_auth_backend_client
https://developer.hashicorp.com/vault/api-docs/auth/aws#configure-client

Yes

Zlaticanin:
### Description
Add support for the missing `max_retries` parameter.

Relates OR Closes #2074 

### Checklist

```
$ make testacc TESTARGS='-run=TestAccAWSAuthBackendClient_basic'
=== RUN   TestAccAWSAuthBackendClient_basic
=== PAUSE TestAccAWSAuthBackendClient_basic
=== CONT  TestAccAWSAuthBackendClient_basic
--- PASS: TestAccAWSAuthBackendClient_basic (1.28s)
PASS
...
```


HamzaShili65 (commented):
LGTM.

robmonte (approved):
LGTM, I made a small suggestion if you like the idea and can test it

vault/resource_aws_auth_backend_client_test.go
```diff
@@ -240,13 +241,20 @@ func testAccAWSAuthBackendClientCheck_attrs(backend string) resource.TestCheckFu
...
-			if resp.Data[apiAttr] != instanceState.Attributes[stateAttr] {
-				return fmt.Errorf("expected %s (%s) of %q to be %q, got %q", apiAttr, stateAttr, endpoint, instanceState.Attributes[stateAttr], resp.Data[apiAttr])
+			if apiAttr == consts.FieldMaxRetries {
+				if resp.Data[apiAttr].(json.Number).String() != instanceState.Attributes[stateAttr] {
+					return fmt.Errorf("expected %s (%s) of %q to be %q, got %q", apiAttr, stateAttr, endpoint, instanceState.Attributes[stateAttr], resp.Data[apiAttr].(json.Number).String())
+				}
+			} else {
+				if resp.Data[apiAttr] != instanceState.Attributes[stateAttr] {
+					return fmt.Errorf("expected %s (%s) of %q to be %q, got %q", apiAttr, stateAttr, endpoint, instanceState.Attributes[stateAttr], resp.Data[apiAttr])
+				}
```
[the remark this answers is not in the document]
Zlaticanin: Thank you! Yes, that looks cleaner and more concise. I tested it and it works as expected, I will go ahead and make the change. Thank you again 🙏

vault/resource_aws_auth_backend_client_test.go
```diff
@@ -240,13 +241,15 @@ func testAccAWSAuthBackendClientCheck_attrs(backend string) resource.TestCheckFu
...
 			consts.FieldSTSRegion:              consts.FieldSTSRegion,
 			consts.FieldIAMServerIDHeaderValue: consts.FieldIAMServerIDHeaderValue,
+			consts.FieldMaxRetries:             consts.FieldMaxRetries,
 		}
 		for stateAttr, apiAttr := range attrs {
-			if resp.Data[apiAttr] == nil && instanceState.Attributes[stateAttr] == "" {
-				continue
+			respApiAttr := resp.Data[apiAttr]
+			if apiAttr == consts.FieldMaxRetries {
+				respApiAttr = resp.Data[apiAttr].(json.Number).String()
```
vinay-gopalan, line 249: Can we add a nil check here? I recently ran into a bit of code like this, and Vault's response behavior had changed for the test to return `nil` for that value which caused a panic on lines like this. 

We should add a nil check for `resp.Data[appiAttr]` that returns an error if the value is nil and was expected. If it is non nil we should be able to convert it to a json.Number

Zlaticanin: I am sorry, I merged it right before you put up this comment. I opened up a new PR to address this: https://github.com/hashicorp/terraform-provider-vault/pull/2275

robmonte: Good catch, do you think both `resp.Data != nil` and `resp.Data[apiAttr] != nil` would need to be checked?

vinay-gopalan: @Zlaticanin no problem at all, thanks for putting up the follow up PR! 🙏🏼 

@robmonte, I think Vault's server returning `nil` for `resp.Data` in this case would be very odd behavior, I think panicking on that might be better? I think just the `resp.Data[apiAttr] != nil` check should suffice

PR #2327, opened 2024-09-17
feat: data_source vault_transit_secret_backend_key

### LINKED ISSUE #2326: [New Data Source]: vault_transit_secret_backend_key
### Description

I would like to read information about a transit key (see [docs](https://developer.hashicorp.com/vault/api-docs/secret/transit#read-key)) using terraform.

```terraform
data "vault_transit_secret_backend_key" "example" {
  backend = vault_mount.transit.path
  name    = "example-key"
}
```

### References

https://developer.hashicorp.com/vault/api-docs/secret/transit#read-key

No

simonostendorf:
### Description

Add `vault_transit_secret_backend_key` data source.

Closes #2326

### Checklist

```
$ make testacc TESTARGS='-run=TestAccDataSourceTransitSecretKey'

everything ok
```

### Community Note


vault/data_source_transit_secret_backend_key.go
```diff
@@ -0,0 +1,176 @@
...
+				ForceNew:    true,
+				Description: "Full path where transit backend is mounted.",
+			},
+			consts.FieldName: {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the key.",
+			},
+			"type": {
```
stevendpclark, line 34: There are a few keys missing from this

- auto_rotate_period
- imported_key (it's listed as imported below)
- latest_version
- min_available_version

There are also a few optional keys depending on options set in the response returned see [formatKeyPolicy](https://github.com/hashicorp/vault/blob/main/builtin/logical/transit/path_keys.go#L335)

- key_size is returned if set on creation
- imported_key_allow_rotation is returned if imported_key is true
- backup_info and restore_info are returned if the key was backed up/restored
- kdf, kdf_mode, convergent_encryption and possibly convergent_encryption_version are returned if derived is true

stevendpclark (changes requested):
A great start to this @simonostendorf, thanks!

I've added a comment about keys that are missing and we are also missing the associated documentation updates for the new data source.

vault/data_source_transit_secret_backend_key.go
simonostendorf: > There are a few keys missing from this

Sorry I used an old vault release without some of these keys and after testing it with the new version I forgot to add them.

simonostendorf:
> A great start to this @simonostendorf, thanks!
> 
> I've added a comment about keys that are missing and we are also missing the associated documentation updates for the new data source.

Thanks for your review, after testing with the latest Vault version I noticed that there are no more key details in the api response. But that was the reason I wanted this data source. I can try to finish this PR, but I can not say how long it will take, because it can no longer fulfil my goal.

PR #2328, opened 2024-09-18
removed outdated cas optoion from example

ksemele:
Just an example code edited.


ksemele:
It was missed when the docs were cleared here: https://github.com/hashicorp/vault/issues/18724

vinay-gopalan (approved):
Thanks!

PR #2368, opened 2024-11-19
Update CODEOWNERS

lursu:
### Description
Updated to the broader Vault team so that ecosystem doesn't become a bottleneck.


fairclothjm (approved):
Thanks!

PR #2375, opened 2024-12-02
Add Fallback parameters matching Vault PR

kpcraig:
This is the TFVP parity PR for https://github.com/hashicorp/vault/pull/29051


vault/resource_aws_secret_backend.go
```diff
@@ -99,6 +100,23 @@ func awsSecretBackendResource() *schema.Resource {
 				Optional:    true,
 				Description: "Specifies a custom HTTP STS endpoint to use.",
 			},
+			consts.FieldSTSRegion: {
```
kpcraig, line 103: This field predates the linked PR but wasn't included in TFVP yet, so i added it while I was here.

(See: https://github.com/hashicorp/vault/pull/29051/files#diff-11906daf04096f16c364abff4761df97c3b879915152cee7f14a03e1f74dc32eL51)

Zlaticanin: Nice! 
Do we know in what release this was added? Was it semi-recent? Asking because I was wondering if we should gate this param with the vault version as well.

Zlaticanin (approved):
LGTM! 👍

vault/resource_aws_secret_backend.go
kpcraig: I thought it was older, but apparently it was added in october - i'll gate it behind the 1.19 flag.

vault/resource_aws_secret_backend_test.go
```diff
@@ -76,6 +76,33 @@ func TestAccAWSSecretBackend_basic(t *testing.T) {
...
+		CheckDestroy:      testCheckMountDestroyed(resourceType, consts.MountTypeAWS, consts.FieldPath),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSecretBackendConfig_fallback(path, accessKey, secretKey),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, path),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDescription, "test description"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldSTSEndpoint, "https://sts.us-west-1.amazonaws.com"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldSTSRegion, "us-west-1"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldSTSFallbackRegions+".0", "us-east-2"),
```
vinay-gopalan, line 96: nit: we can also check lengths here for extra validation
```suggestion
					resource.TestCheckResourceAttr(resourceName, consts.FieldSTSFallbackRegions+".#", "2"),
```

vault/resource_aws_secret_backend_test.go
```diff
@@ -76,6 +76,33 @@ func TestAccAWSSecretBackend_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSSecretBackend_fallback(t *testing.T) {
```
vinay-gopalan, line 79: Nice! Could we also add an update step and an import step to complete out the full testing loop?

vinay-gopalan (commented):
Looking great so far! Had a comment on adding a couple more test steps. Could we also update the docs and include the new fields? Thanks!

vault/resource_aws_secret_backend_test.go
kpcraig: Added an import test

vault/resource_aws_secret_backend_test.go
kpcraig: Added an import and update test.

website/docs/r/aws_secret_backend.html.md
```diff
@@ -83,6 +83,12 @@ for credentials issued by this backend.
 
 * `sts_endpoint` - (Optional) Specifies a custom HTTP STS endpoint to use.
 
+* `sts_region` - (Optional) Specifies the region of the STS endpoint. Should be included if `sts_endpoint` is supplied.
```
vinay-gopalan, line 86: Should we also add the version constraints for the new params?

```suggestion
* `sts_region` - (Optional) Specifies the region of the STS endpoint. Should be included if `sts_endpoint` is supplied. Requires Vault 1.19+.
```

vinay-gopalan (approved):
LGTM, had a non-blocking comment on adding version constraints to the docs

vinay-gopalan (approved):

PR #2412, opened 2025-02-24
Add support for disabled_validations to the PKI CMPv2 config resource

victorr:
### Description

Add support for disabled_validations to the PKI CMPv2 config resource.

### Checklist

```

...
```

### Community Note

### DIFF
--- a/vault/data_source_pki_secret_backend_config_cmpv2.go
+++ b/vault/data_source_pki_secret_backend_config_cmpv2.go
@@ -63,6 +63,15 @@ func pkiSecretBackendConfigCMPV2DataSource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			consts.FieldDisabledValidations: {
+				Type:        schema.TypeList,
+				Required:    false,
+				Optional:    true,
+				Description: "A comma-separated list of validations not to perform on CMPv2 messages.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 			consts.FieldLastUpdated: {
 				Type:        schema.TypeString,
 				Computed:    true,
@@ -108,6 +117,7 @@ func readCMPV2Config(ctx context.Context, d *schema.ResourceData, client *api.Cl
 		consts.FieldDefaultPathPolicy,
 		consts.FieldEnableSentinelParsing,
 		consts.FieldAuditFields,
+		consts.FieldDisabledValidations,
 		consts.FieldLastUpdated,
 	}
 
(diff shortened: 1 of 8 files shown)


vault/data_source_pki_secret_backend_config_cmpv2.go
```diff
@@ -63,6 +63,15 @@ func pkiSecretBackendConfigCMPV2DataSource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			consts.FieldDisabledValidations: {
```
kitography, line 66: Every other field seems to have a "computed" line - why not this one?  How is this different than AuditFields in that sense?  I'm also not really sure how optional works here?  Most of these don't have that listed.

website/docs/r/pki_secret_backend_config_cmpv2.html.md
```diff
@@ -72,8 +73,10 @@ The following arguments are supported:
 * `enable_sentinel_parsing` - (Optional) If set, parse out fields from the provided CSR making them available for Sentinel policies.
 
 * `enabled` - (Optional) Specifies whether CMPv2 is enabled.
-
+  
 * `audit_fields` - (Optional) Fields parsed from the CSR that appear in the audit and can be used by sentinel policies.
+  
+* `disabled_validations` - (Optional) A comma-separated list of validations not to perform on CMPv2 messages.
```
kitography, line 79: Might be nice to mention what the (current) supported options are.

kitography (commented):
Looks good - I'd love another set of eyes since I haven't added one of these fields before (and I don't know, eg. how CHANGELOG works).  Added a couple confused comments.

vault/data_source_pki_secret_backend_config_cmpv2.go
stevendpclark: The computed argument is mainly used when a value is returned from the server when no initial value from Terraform itself was provided. In this case since I believe the value returned will be an empty list it isn't required.

Optional only means the field does not need to be specified within the resource block

website/docs/r/pki_secret_backend_config_cmpv2.html.md
stevendpclark: I believe the norm isn't to specify a list here as it can fall out of sync with the server side of things, especially across revisions. 

Its the same reasoning the TFVP team have shied away from enforcing the valid values on the input schema as it causes end-user friction that we need to update and publish a new TFVP version every time something on the server changes

kitography (approved):
Looks great!

PR #2414, opened 2025-02-25
Add Automated Root Rotation Parameters to AWS Auth/Secrets and DB Secrets

vinay-gopalan:
### Description
Adds new automated root rotation parameters that will be available as part of Vault 1.19 Enterprise to the following resources:
- `vault_aws_secret_backend`
- `vault_database_secret_backend_connection`
- `vault_database_secrets_mount`
- `vault_aws_auth_backend_client`

### Checklist

```
$ make testacc TESTARGS='-run=TestAccAWSSecretBackend_automatedRotation'
=== RUN   TestAccAWSSecretBackend_automatedRotation
    resource_aws_secret_backend_test.go:167: Vault server version "1.20.0-beta1+ent"
--- PASS: TestAccAWSSecretBackend_automatedRotation (1.87s)
PASS
```


vault/resource_aws_auth_backend_client.go
```diff
@@ -5,6 +5,7 @@ package vault
 
 import (
 	"context"
+	automatedrotationutil "github.com/hashicorp/terraform-provider-vault/internal/rotation"
```
kpcraig, line 8: maybe it doesn't apply to tfvp, but _occasionally_ some linter somewhere will complain to me about this sort of import order.

kpcraig (dismissed):

robmonte (approved):
Looks good, only addition I can think of is a test case that can actually check if a rotation job is registered and that `disable_automated_rotation` removes the job, and similarly the same result when zeroing out all fields. Don't think we have the support for that level of checking yet though, so something to consider for the future.

kpcraig (approved):

vinay-gopalan:
Will follow up with a docs PR

PR #2429, opened 2025-03-12
Update pki_secret_backend_crl_config to be more resilent to unknown response fields

stevendpclark:
### Description

Do not pass in the PKI CRL configuration response map directly into `util.SetResourceData` as this will fail if an unknown field is returned from Vault that isn't listed in the resource's schema. We encountered this within the Vault 1.19 release where the resource broke when `max_crl_entries` was added. 

```
│ Error: error setting resource data for key "max_crl_entries", err=Invalid address to set: []string{"max_crl_entries"}
│
│   with vault_pki_secret_backend_crl_config.crl_config,
│   on main.tf line 8, in resource "vault_pki_secret_backend_crl_config" "crl_config":
│    8: resource "vault_pki_secret_backend_crl_config" "crl_config" {

```

Also this includes a slight refactoring to update the methods to the non-deprecated values along with removing duplication by adding a common method that builds up the list of fields based on Vault version.

### Checklist
- [X] Acceptance tests where run against all supported Vault Versions

```

...
```

### Community Note

### DIFF
--- a/vault/resource_pki_secret_backend_crl_config.go
+++ b/vault/resource_pki_secret_backend_crl_config.go
@@ -4,10 +4,11 @@
 package vault
 
 import (
-	"fmt"
+	"context"
 	"log"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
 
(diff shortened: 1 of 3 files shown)


vault/resource_pki_secret_backend_crl_config.go
```diff
@@ -183,35 +155,59 @@ func pkiSecretBackendCrlConfigRead(d *schema.ResourceData, meta interface{}) err
...
+	if _, ok := d.GetOk(consts.FieldBackend); !ok {
 		// ensure that the backend is set on import
-		if err := d.Set("backend", strings.TrimRight(path, crlConfigPathBase)); err != nil {
-			return err
+		if err := d.Set(consts.FieldBackend, strings.TrimRight(path, crlConfigPathBase)); err != nil {
+			return diag.Errorf("failed setting field %s: %v", consts.FieldBackend, err)
 		}
 	}
 
-	if err := util.SetResourceData(d, config.Data); err != nil {
```
stevendpclark, line 201: This is the main fix here, passing in config.Data into util.SetResourceData was causing issues when the Data map contained an unknown field

sgmiller (approved):

PR #2430, opened 2025-03-12
add docs for automated root rotation fields to secrets/auth engines

vinay-gopalan:
Documentation for Automated Rotation fields

### DIFF
--- a/website/docs/r/aws_auth_backend_client.html.md
+++ b/website/docs/r/aws_auth_backend_client.html.md
@@ -36,6 +36,8 @@ resource "vault_aws_auth_backend_client" "example" {
   identity_token_audience = "<TOKEN_AUDIENCE>"
   identity_token_ttl      = "<TOKEN_TTL>"
   role_arn                = "<AWS_ROLE_ARN>"
+  rotation_schedule       = "0 * * * SAT"
+  rotation_window         = 3600
 }
 ```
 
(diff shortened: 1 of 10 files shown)


website/docs/r/aws_auth_backend_client.html.md
```diff
@@ -103,6 +107,18 @@ The following arguments are supported:
 * `max_retries` - (Optional) Number of max retries the client should use for recoverable errors. 
     The default `-1` falls back to the AWS SDK's default behavior.
 
+* `rotation_period` - (Optional) The amount of time in seconds Vault should wait before rotating the root credential.
+    A zero value tells Vault not to rotate the root credential. The minimum rotation period is 10 seconds. Requires Vault Enterprise 1.19+.
+
+* `rotation_schedule` - (Optional) The schedule, in [cron-style time format](https://en.wikipedia.org/wiki/Cron),
```
kpcraig, line 113: Do we want something here about mutual exclusivity of rotation_window vs. period?

kpcraig (approved):

kpcraig:
left a comment, but especially if we need this for release, I don't think it's blocking.

website/docs/r/aws_auth_backend_client.html.md
vinay-gopalan: Good question! Decided to leave that out since strictly speaking, TFVP does not enforce any mutual exclusivity on the fields. It is standard practice to let Vault enforce its mutual exclusivity. It would be technically possible to set both via TFVP, but Vault would return an error so the flow would fail and let the user know

PR #2445, opened 2025-04-01
Add `subscribe_event_types` field to policy document data source

Zlaticanin:
### Description

This PR adds missing `subscribe_event_types` field to `data_source_policy_document` data source.
https://registry.terraform.io/providers/hashicorp/vault/latest/docs/data-sources/policy_document 

### Checklist
- [x] Added [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md) entry (only for user-facing changes)
- [ ] Acceptance tests where run against all supported Vault Versions

### Output from acceptance testing:

```
$ go test -v ./vault -run=TestDataSourcePolicyDocument

=== RUN   TestDataSourcePolicyDocument
--- PASS: TestDataSourcePolicyDocument (0.93s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     1.661s

...
```

PR #2466, opened 2025-05-13
Revert change to allow for [] capabilities

Zlaticanin:
Fixing a regression I introduced with this [PR](https://github.com/hashicorp/terraform-provider-vault/pull/2445). My understanding by looking at the docs, both [TFVP](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/data-sources/policy_document#argument-reference) and [Vault](https://developer.hashicorp.com/vault/docs/concepts/policies#capabilities), was that this param is required and shouldn't be empty. Looking at the Vault code more closely now, we do allow for [], and I misunderstood the docs. Reverting this change and added a test case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+BUGS:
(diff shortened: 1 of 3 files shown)


fairclothjm (approved):
LGTM

PR #2476, opened 2025-05-26
VAULT-35631: support RLQ fields "group_by" and "secondary_rate"

bosouza:
### Description

This PR adds supports to the new `group_by` and `secondary_rate` fields to the `vault_quota_rate_limit` resource, added to the API in https://github.com/hashicorp/vault/pull/30447. These new collective and entity-based rate limits are an enterprise-only functionality, but the default option `group_by: ip` is allowed on community edition and maintains the standard, backwards-compatible behavior of rate limit quotas.

As described in code comments I had a lot of trouble with the default value semantics for the fields, as leaving `group_by` unset is allowed and defaults to `ip`, and leaving `secondary_rate` unset makes it default to 0 on `ip` and `none` modes, and `rate` on the entity-based modes. There are 2 approaches I considered:

1. **Optional + set state if in config**: this is the approach that's already being used by the inheritable field, it is configured as an `Optional` field but **not** `Computed`, and we only call the `Set` function to commit its value to the tf state if it was already in the config/state, which means that if the user doesn't specify anything in the TF config then the value won't be written to the state.

  Pros:
   * when the practitioner updates the `inheritable` field from some value to unset in the TF config the result if for the TF state to contain a zero value for the setting. Still not perfect, but the SDK v2 can't distinguish between zero value and unset, so that's the best we can hope for

   Cons:
   * doesn't work with TF import. That's actually a bug we have, even if `inheritable` is set in the backend the field will be left unset in the resulting imported configuration
   * the default value of fields like `inheritable` can't be referenced by the TF config, it's only visible when set.
2. **Optional + computed**: mark the field as both optional and computed, so regardless if the `group_by` and `secondary_rate` fields are set in the TF config they will show up

Pros:
   * TF import works properly
   * fields can always be depended upon to be a source of truth, even having the default values assigned by the external API

Cons:
   * when the practitioner changes a field from a value to unset the terraform state is maintained. As seen in the tests, that results in a bug where removing `secondary_rate` doesn't bring it back to the default state expected if the RLQ had been originally created with the field unset.

Overall it seems like the approach 2 is the best. I considered migrating `inheritable` to this same approach, fixing the bug with terraform import, but that could be considered a breaking change so I'm not sure if it's a viable approach.

Jira: [VAULT-35631](https://hashicorp.atlassian.net/browse/VAULT-35631)

### Checklist

```
$ make testacc TESTARGS='-run=TestQuotaRateLimitGroupBy'
TF_ACC=1 go test -run=TestQuotaRateLimitGroupBy -timeout 30m ./...
ok      github.com/hashicorp/terraform-provider-vault/codegen   0.014s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/framework/validators     0.016s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/identity/entity  0.034s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/provider 0.020s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/ephemeral  0.014s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/sys        0.013s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/testutil  0.005s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util      0.006s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util/mountutil    0.003s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/vault     0.014s
...
```

### Community Note

[VAULT-35631]: https://hashicorp.atlassian.net/browse/VAULT-35631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/vault/resource_quota_rate_limit.go
+++ b/vault/resource_quota_rate_limit.go
@@ -48,6 +48,12 @@ func quotaRateLimitResource() *schema.Resource {
 				Description:  "The maximum number of requests at any given second to be allowed by the quota rule. The rate must be positive.",
 				ValidateFunc: validation.FloatAtLeast(0.0),
 			},
+			"secondary_rate": {
+				Type:         schema.TypeFloat,
+				Optional:     false,
+				Description:  `Only available when using the "entity_then_ip" or "entity_then_none" group_by modes. This is the rate limit applied to the requests that fall under the "ip" or "none" groupings, while the authenticated requests that contain an entity ID are subject to the "rate" field instead. Defaults to the same value as "rate".`,
+				ValidateFunc: validation.FloatAtLeast(0.0),
+			},
 			"interval": {
 				Type:         schema.TypeInt,
 				Optional:     true,
@@ -70,6 +76,18 @@ func quotaRateLimitResource() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Description: "If set to true on a quota where path is set to a namespace, the same quota will be cumulatively applied to all child namespace. The inheritable parameter cannot be set to true if the path does not specify a namespace. Only the quotas associated with the root namespace are inheritable by default.",
+			},
+			"group_by": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Description: `Attribute by which to group requests by. Valid group_by modes are: 1) "ip" that groups` +
+					` requests by their source IP address (group_by defaults to ip if unset); 2) "none" that groups all` +
+					` requests that match the rate limit quota rule together; 3) "entity_then_ip" that groups requests` +
+					` by their entity ID for authenticated requests that carry one, or by their IP for unauthenticated` +
+					` requests (or requests whose authentication is not connected to an entity); and 4) "entity_then_none"` +
+					` which also groups requests by their entity ID when available, but the rest is all grouped together` +
+					` (i.e. unauthenticated or with authentication not connected to an entity).`,
+				ValidateFunc: validation.StringInSlice([]string{"ip", "none", "entity_then_ip", "entity_then_none"}, false),
 			},
 		},
 	}
@@ -109,6 +127,20 @@ func quotaRateLimitCreate(d *schema.ResourceData, meta interface{}) error {
 		if v, ok := d.GetOk(consts.FieldRole); ok {
 			data[consts.FieldRole] = v
 		}
+	}
+
+	if v, ok := d.GetOk("group_by"); ok {
+		if !provider.IsAPISupported(meta, provider.VaultVersion120) {
+			return fmt.Errorf("group_by is only supported in Vault 1.20 and later")
+		}
+		data["group_by"] = v
+	}
+
+	if v, ok := d.GetOk("secondary_rate"); ok {
+		if !provider.IsAPISupported(meta, provider.VaultVersion120) {
+			return fmt.Errorf("secondary_rate is only supported in Vault 1.20 and later")
+		}
+		data["secondary_rate"] = v
 	}
 
 	_, err := client.Logical().Write(path, data)
@@ -151,6 +183,10 @@ func quotaRateLimitRead(d *schema.ResourceData, meta interface{}) error {
 		if _, ok := d.GetOkExists("inheritable"); ok {
 			fields = append(fields, "inheritable")
 		}
+	}
+
+	if provider.IsAPISupported(meta, provider.VaultVersion120) {
+		fields = append(fields, "group_by", "secondary_rate")
 	}
 
 	for _, k := range fields {
--- a/vault/resource_quota_rate_limit.go
+++ b/vault/resource_quota_rate_limit.go
@@ -52,7 +52,6 @@ func quotaRateLimitResource() *schema.Resource {
 			"secondary_rate": {
 				Type:         schema.TypeFloat,
 				Optional:     true,
-				Computed:     true,
 				Description:  `Only available when using the "entity_then_ip" or "entity_then_none" group_by modes. This is the rate limit applied to the requests that fall under the "ip" or "none" groupings, while the authenticated requests that contain an entity ID are subject to the "rate" field instead. Defaults to the same value as "rate".`,
 				ValidateFunc: validation.FloatAtLeast(0.0),
 			},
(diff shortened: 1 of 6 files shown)


vault/resource_quota_rate_limit.go
```diff
@@ -153,6 +185,10 @@ func quotaRateLimitRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if provider.IsAPISupported(meta, provider.VaultVersion120) {
```
bosouza, line 188: following the existing pattern but it's unclear what's the point of this: If the server supports the new fields it's going to return them, if it doesn't it won't, there's no point in checking the version

bosouza: also, should we verify if it's an ent Vault?

vault/resource_quota_rate_limit.go
```diff
@@ -71,10 +79,38 @@ func quotaRateLimitResource() *schema.Resource {
 				Optional:    true,
 				Description: "If set to true on a quota where path is set to a namespace, the same quota will be cumulatively applied to all child namespace. The inheritable parameter cannot be set to true if the path does not specify a namespace. Only the quotas associated with the root namespace are inheritable by default.",
 			},
+			"group_by": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				Description: `Attribute by which to group requests by. Valid group_by modes are: 1) "ip" that groups` +
```
mpalmi:
```suggestion
				Description: `Attribute by which to group requests. Valid group_by modes are: 1) "ip" that groups` +
```

vault/resource_quota_rate_limit.go
```diff
@@ -188,21 +248,38 @@ func quotaRateLimitUpdate(d *schema.ResourceData, meta interface{}) error {
...
+
+	if v, ok := d.GetOk("secondary_rate"); ok {
+		if !provider.IsAPISupported(meta, provider.VaultVersion120) {
+			return fmt.Errorf("secondary_rate is only supported in Vault Enterprise 1.20 and later")
+		}
+		data["secondary_rate"] = v
+	}
+
 	if provider.IsAPISupported(meta, provider.VaultVersion115) {
+		// we should probably fail if the field is set on an unsupported version instead of ignoring it, but changin
```
mpalmi:
```suggestion
		// we should probably fail if the field is set on an unsupported version instead of ignoring it, but changing
```

vault/resource_quota_rate_limit.go
```diff
@@ -188,21 +248,38 @@ func quotaRateLimitUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		// that would be a breaking change
 		if v, ok := d.GetOkExists("inheritable"); ok {
 			data["inheritable"] = v.(bool)
 		}
 	}
 
 	if provider.IsAPISupported(meta, provider.VaultVersion112) {
+		// we should probably fail if the field is set on an unsupported version instead of ignoring it, but changin
```
mpalmi:
```suggestion
		// we should probably fail if the field is set on an unsupported version instead of ignoring it, but changing
```

vault/resource_quota_rate_limit.go
```diff
@@ -188,21 +248,38 @@ func quotaRateLimitUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		// that would be a breaking change
 		if v, ok := d.GetOk(consts.FieldRole); ok {
 			data[consts.FieldRole] = v
 		}
 	}
 
 	_, err := client.Logical().Write(path, data)
 	if err != nil {
-		d.SetId("")
```
mpalmi, line 205: Is this intentional? It looks like we're still doing this for other error cases, so why is this unique?

website/docs/r/quota_rate_limit.md
```diff
@@ -55,6 +55,18 @@ The following arguments are supported:
 
 * `inheritable` - (Optional) If set to `true` on a quota where path is set to a namespace, the same quota will be cumulatively applied to all child namespace. The inheritable parameter cannot be set to `true` if the path does not specify a namespace. Only the quotas associated with the root namespace are inheritable by default. Requires Vault 1.15+.
 
+* `group_by` - (Optional) Attribute by which to group requests by. Valid `group_by` modes are: 1) `ip` that groups
```
mpalmi:
```suggestion
* `group_by` - (Optional) Attribute by which to group requests. Valid `group_by` modes are: 1) `ip` that groups 
```

website/docs/r/quota_rate_limit.md
```diff
@@ -55,6 +55,18 @@ The following arguments are supported:
 
 * `inheritable` - (Optional) If set to `true` on a quota where path is set to a namespace, the same quota will be cumulatively applied to all child namespace. The inheritable parameter cannot be set to `true` if the path does not specify a namespace. Only the quotas associated with the root namespace are inheritable by default. Requires Vault 1.15+.
 
+* `group_by` - (Optional) Attribute by which to group requests by. Valid `group_by` modes are: 1) `ip` that groups 
+  requests by their source IP address (**`group_by` defaults to `ip` if unset, which is the only supported mode in
```
mpalmi: Putting myself in a user's shoes, I find this description is a little confusing.

We're not really describing what "grouping" requests does. I suggest that we explicitly call that out. Maybe something to the effect of "defines the group by which a rate limit applies" and then diving into the examples.

mpalmi: Same goes for the schema description earlier.

vault/resource_quota_rate_limit.go
```diff
@@ -48,6 +49,13 @@ func quotaRateLimitResource() *schema.Resource {
 				Description:  "The maximum number of requests at any given second to be allowed by the quota rule. The rate must be positive.",
 				ValidateFunc: validation.FloatAtLeast(0.0),
 			},
+			"secondary_rate": {
+				Type:         schema.TypeFloat,
```
mpalmi, line 53: I was going to suggest using a nillable field to provide more control over the unset default case you described, but it looks like `typeObject` (perhaps intentionally) isn't an exported `ValueType`. I think the trade-offs and conclusion you described make sense. Thanks for spelling it out so clearly.

mpalmi (commented):
Couple of comments, but mostly nits and cosmetic/clarity stuff. Feel free to cherry-pick which ones you'd like to address!

vault/resource_quota_rate_limit.go
bosouza: this was a bug, we shouldn't delete the tf state of the resource on a failed update

PR #2480, opened 2025-05-27
Add managed key parameters for SSH CA config

rculpepper:
### Description

Relates OR Closes #0000

### Checklist

```

...
```

### Community Note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,16 @@
 ## Unreleased
 
 FEATURES:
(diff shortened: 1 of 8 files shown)


victorr (commented):
The changes look good to me. 

I see that we do have unit tests for `vault_managed_keys`, so I suggest that one or more tests be added for this new functionality.

Hopefully that is not impractical?

rculpepper:
> The changes look good to me.
> 
> I see that we do have unit tests for `vault_managed_keys`, so I suggest that one or more tests be added for this new functionality.
> 
> Hopefully that is not impractical?

Unfortunately, I don't think that's doable. We'd need to be able to create a managed key to use, and I don't think we can do that currently

stevendpclark (approved):

PR #2488, opened 2025-06-05
Add support for Transit CMAC endpoint

rculpepper:
### Description

Relates OR Closes #0000

### Checklist

```

...
```

### Community Note

### DIFF
--- a/vault/data_source_transit_cmac.go
+++ b/vault/data_source_transit_cmac.go
@@ -1,0 +1,130 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package vault
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+)
+
+func transitCMACDataSource() *schema.Resource {
+	return &schema.Resource{
+		Read: provider.ReadWrapper(transitCMACDataSourceRead),
+
+		Schema: map[string]*schema.Schema{
+			consts.FieldPath: {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The Transit secret backend the key belongs to.",
+			},
+			consts.FieldName: {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the CMAC key to use.",
+			},
+			consts.FieldKeyVersion: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The version of the key to use",
+			},
+			consts.FieldInput: {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Specifies the base64 encoded input data. One of input or batch_input must be supplied.",
+			},
+			consts.FieldMACLength: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "Specifies the MAC length to use (POST body parameter). The mac_length cannot be larger than the cipher's block size.",
+			},
+			consts.FieldURLMACLength: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "Specifies the MAC length to use (URL parameter). If provided, this value overrides mac_length. The url_mac_length cannot be larger than the cipher's block size.",
+			},
+			consts.FieldBatchInput: {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Specifies a list of items for processing. When this parameter is set, any supplied 'input' or 'context' parameters will be ignored. Responses are returned in the 'batch_results' array component of the 'data' element of the response. Any batch output will preserve the order of the batch input. If the input data value of an item is invalid, the corresponding item in the 'batch_results' will have the key 'error' with a value describing the error.",
+				Elem:        &schema.Schema{Type: schema.TypeMap},
+			},
+			consts.FieldCMAC: {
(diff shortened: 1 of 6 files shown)


vault/data_source_transit_cmac.go
```diff
@@ -0,0 +1,130 @@
...
+	keyName := d.Get(consts.FieldName).(string)
+
+	payload := map[string]interface{}{}
+
+	signAPIStringFields := []string{
+		consts.FieldKeyVersion,
+		consts.FieldInput,
+		consts.FieldSignatureContext,
+		consts.FieldMACLength,
+		consts.FieldURLMACLength,
```
victorr: `url_mac_length` should not be part of the payload, should it?

vault/data_source_transit_cmac.go
```diff
@@ -0,0 +1,130 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the CMAC key to use.",
+			},
+			consts.FieldKeyVersion: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The version of the key to use",
+			},
+			consts.FieldInput: {
```
victorr, line 35: Could you use `AtLeastOneOf` with `input` and `batch_input`? I have not used it myself, but it should work here I think?

website/docs/d/transit_cmac.html.md
```diff
@@ -0,0 +1,81 @@
...
+  `input` or `batch_input` must be supplied.
+
+* `mac_length` - (Optional) Specifies the MAC length to use (POST body parameter).
+  The `mac_length` cannot be larger than the cipher's block size.
+
+* `url_mac_length` - (Optional) Specifies the MAC length to use (URL parameter).
+  If provided, this value overrides `mac_length`. The `url_mac_length` cannot
+  be larger than the cipher's block size.
+
+* `reference` - (Optional)
```
victorr, line 50: `reference` is not in the datasource schema, should it be?

vault/data_source_transit_cmac_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+`
+
+var cmacBatchInputBlocks = `
+data "vault_transit_cmac" "test" {
+    path        = vault_mount.test.path
+    name        = vault_transit_secret_backend_key.test.name
+	batch_input = [
+		{
+		  input = "adba32=="
```
victorr, line 41: Please consider adding a test with batch input that uses `reference`, and maybe `mac_length` (the Transit API docs example seems to indicate that that is possible).

website/docs/d/transit_cmac.html.md
```diff
@@ -0,0 +1,81 @@
+---
+layout: "vault"
+page_title: "Vault: vault_transit_cmac data source"
```
victorr, line 3: I believe this needs to be added to `website/vault.erb`.

website/docs/d/transit_cmac.html.md
rculpepper: You can only provide `reference` within `batch_input`, so it's not its own field in the schema

vault/batch_input_util.go
```diff
@@ -0,0 +1,86 @@
...
+var intFields = []string{
+	consts.FieldMACLength,
+	consts.FieldKeyVersion,
+}
+
+// When batch_input is provided as a map, all of the fields get parsed as strings,
+// which results in an error if non-string parameters are included, because Vault
+// expects a different type. convertBatchInput converts these values to their correct
+// types to avoid this error
+func convertBatchInput(batchInput interface{}) ([]map[string]interface{}, error) {
```
stevendpclark: Could we pass in the intFields as a param to this function to make it a little more generic and usable in other situations?

vault/data_source_transit_sign.go
```diff
@@ -122,7 +129,6 @@ func transitSignDataSourceRead(d *schema.ResourceData, meta interface{}) error {
 		consts.FieldSignatureAlgorithm,
 		consts.FieldMarshalingAlgorithm,
 		consts.FieldSaltLength,
-		consts.FieldBatchInput,
```
stevendpclark, line 125: Should we get rid of `consts.FieldPrehashed` from this list?

vault/data_source_transit_verify.go
```diff
@@ -126,20 +130,27 @@ func transitVerifyDataSourceRead(d *schema.ResourceData, meta interface{}) error
...
 		consts.FieldHashAlgorithm,
 		consts.FieldInput,
 		consts.FieldSignature,
 		consts.FieldHMAC,
 		consts.FieldCMAC,
 		consts.FieldReference,
-		consts.FieldBatchInput,
 		consts.FieldContext,
 		consts.FieldSignatureContext,
 		consts.FieldPrehashed,
```
stevendpclark: Same here `consts.FieldPrehashed` probably should be removed?

vault/resource_transit_secret_backend_key.go
```diff
@@ -340,7 +340,7 @@ func transitSecretBackendKeyRead(d *schema.ResourceData, meta interface{}) error
 
 	ikeys := secret.Data["keys"]
 	keys := []interface{}{}
-	if ikeys != nil || secret.Data["type"] != "hmac" { // hmac type does not return keys
+	if ikeys != nil {
```
stevendpclark, line 343: hmac now returns keys?

rculpepper: No, but that extra check isn't necessary. The nil check is enough, since that'll be `nil` for any type that doesn't return keys. The type check also causes issues with CMAC keys, because they don't return keys either

victorr (approved):
LGTM!

PR #2523, opened 2025-07-10
add `jwks_pairs` argument to vault_jwt_auth_backend resource

thyton:
### Description

Relates OR Closes #0000

### Checklist

```

...
```

### Community Note

## PCI review checklist


vault/resource_jwt_auth_backend.go
```diff
@@ -110,6 +110,14 @@ func jwtAuthBackendResource() *schema.Resource {
 				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URL. If not set, system certificates are used.",
 			},
 
+			"jwks_pairs": {
```
Zlaticanin, line 113: When did we add this field? Was it in recent Vault versions, or? Wondering if we should gate it with a version check or not.

vinay-gopalan: Good call, officially TFVP v5 supports Vault versions >= 1.15.x, so we can assume that customers should at least be on that minimum Vault version when using this feature. Doing a little digging in the docs, looks like this was added in Vault 1.16, so a version check might be a good addition!

thyton:
Thank you @Zlaticanin @vinay-gopalan for the feedback! Let me know if anything else is missing

vault/resource_jwt_auth_backend.go
```diff
@@ -373,15 +392,19 @@ func jwtAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 	configuration := map[string]interface{}{}
 	for _, configOption := range matchingJwtMountConfigOptions {
 		if _, ok := d.GetOkExists(configOption); ok || d.HasChange(configOption) {
-			configuration[configOption] = d.Get(configOption)
-
-			if configOption == "provider_config" {
+			switch configOption {
+			case "jwks_pairs":
+				if useAPIVer116 {
+					configuration[configOption] = d.Get(configOption)
```
vinay-gopalan, line 398: nit: for visibility, should we log to the user that we are ignoring the field due to the version requirements?

vinay-gopalan (dismissed):
LGTM! Had 1 nit around a log when ignoring the field, but should be good otherwise. Thanks for adding this in 🙏🏼

vault/resource_jwt_auth_backend.go
maithyton: fixed in [07a538a](https://github.com/hashicorp/terraform-provider-vault/pull/2523/commits/07a538a5ee822f75e1350431fe179596b8d4da24)

Zlaticanin (approved):
Great job! 👏

Issue #2572, opened 2025-08-19
[Bug]: `Provider GetResourceIdentitySchemas Not Implemented` Error


auttley:
### Terraform Core Version

1.12.2

### Terraform Vault Provider Version

5.2.0

### Vault Server Version

N/A

### Affected Resource(s)

N/A - All, provider fails on a `terraform validate`.

### Expected Behavior

Terraform runs without issue.

### Actual Behavior

Error on completion of initialization due to provider bug.

### Relevant Error/Panic Output Snippet

```shell
╷
│ Error: Failed to load plugin schemas
│ 
│ Error while loading schemas for plugin components: Failed to obtain
│ provider schema: Could not load the schema for provider
│ registry.terraform.io/hashicorp/vault: failed to retrieve schema from
│ provider "registry.terraform.io/hashicorp/vault": Provider
│ GetResourceIdentitySchemas Not Implemented: A GetResourceIdentitySchemas
│ call was received by the provider, however the provider does not implement
│ the call. Either upgrade the provider to a version that implements resource
│ identity support or this is a bug in Terraform that should be reported to
│ the Terraform maintainers...
╵
```

### Terraform Configuration Files

```
terraform {
  required_providers {
    vault = {
      source  = "hashicorp/vault"
      version = "= 5.2.0"
    }
  }
  required_version = ">= 1.12"
}
```

### Steps to Reproduce

* `terraform init`
* `terraform validate`

### Debug Output

```
Initializing the backend...
Initializing provider plugins...
- Finding hashicorp/vault versions matching "5.2.0"...
- Installing hashicorp/vault v5.2.0...
- Installed hashicorp/vault v5.2.0 (signed by HashiCorp)
Terraform has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that Terraform can guarantee to make the same selections by default when
you run "terraform init" in the future.

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.
```

```
╷
│ Error: Failed to load plugin schemas
│ 
│ Error while loading schemas for plugin components: Failed to obtain provider schema: Could not load
[...]

PR #2577, opened 2025-08-19
CI: extend test matrix to include terraform versions

benashz:
Run acceptance tests against a matrix of Terraform and Vault versions. This should help to prevent issues like: #2572


fairclothjm (dismissed):
LGTM

Does this matrix increase the build time significantly? If yes, maybe we could just build the latest Vault version with the previous TF version?

benashz:
> LGTM
> 
> Does this matrix increase the build time significantly? If yes, maybe we could just build the latest Vault version with the previous TF version?

Thanks. No change to build times.

PR #2605, opened 2025-09-17
Allow Unsetting String and Array Values to Auth Mount

kitography:
### Description
This changes the implementation to the tuneable auth parameters to use JSON rather than a struct.  This way which parameters were actually changed / set is transmitted to vault.  This means that instead of ignoring nil/zero/empty values, those values explictly exist, and fields can therefore be unset (that is set to nil/zero/empty value).

This is an updated implementation of this PR, to avoid changes that wouldn't be backwards compatible: https://github.com/hashicorp/terraform-provider-vault/pull/2532

There is a vault change that enables the same sort of "unset" behavior for these fields from the command line, but that change is independent (that is, neither of these changes are reliant on the other to function):

ENT: https://github.com/hashicorp/vault-enterprise/pull/9383
CE: https://github.com/hashicorp/vault/pull/31555

### Checklist

```
=== RUN   TestResourceMount_AllowedResponseHeaders_EmptySlice
--- PASS: TestResourceMount_AllowedResponseHeaders_EmptySlice (3.81s)
PASS

```

### Community Note

## PCI review checklist

-- It's sufficient to just revert this PR

-- No change to security controls, these values were already accessible /unsettable through the API

### DIFF
--- a/vault/resource_mount.go
+++ b/vault/resource_mount.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


vault/resource_mount.go
```diff
@@ -287,23 +288,22 @@ func updateMount(ctx context.Context, d *schema.ResourceData, meta interface{},
 		return err
 	}
 
-	config := api.MountConfigInput{
-		DefaultLeaseTTL: fmt.Sprintf("%ds", d.Get(consts.FieldDefaultLeaseTTL)),
-		MaxLeaseTTL:     fmt.Sprintf("%ds", d.Get(consts.FieldMaxLeaseTTL)),
-		Options:         mountOptions(d),
+	mapConfig := map[string]interface{}{
```
fairclothjm, line 294: Should we leave a comment explaining why we aren't using `api.MountConfigInput` here?

vault/resource_mount.go
```diff
@@ -360,7 +360,7 @@ func updateMount(ctx context.Context, d *schema.ResourceData, meta interface{},
 	// TODO: remove this work-around once VAULT-5521 is fixed
 	var tries int
 	for {
-		if err := client.Sys().TuneMountWithContext(ctx, path, config); err != nil {
+		if err := tuneMountWithMap(client, ctx, path, mapConfig); err != nil {
```
fairclothjm: Should we leave a comment explaining why are making a raw HTTP request instead of using the Vault client?

fairclothjm (dismissed):
Great work! I left some suggestions on adding some clarifying comments.

vault/resource_mount.go
kitography: Great suggestion!  Added (sorry, please re-review)

vault/resource_mount.go
kitography: Great suggestion, Added!  (Sorry, please re-review)

vault/resource_mount.go
```diff
@@ -511,3 +517,16 @@ func mountOptions(d *schema.ResourceData) map[string]string {
 	}
 	return options
 }
+
+func tuneMountWithMap(c *api.Client, ctx context.Context, path string, config map[string]interface{}) error {
```
benashz, line 521: nit: typically the context.Context is the first argument, would you mind updating this?

kitography: sure

fairclothjm (approved):
LGTM

benashz (approved):

PR #2623, opened 2025-10-08
secrets/aws: Add max_retries field to AWS Secret Backend.

Balaji2198:
### Description

This PR adds support for the [`max_retries` field in the AWS Secret engine](https://developer.hashicorp.com/vault/api-docs/auth/aws#max_retries).

Relates OR Closes #0000

### Checklist

```
$ make testacc TESTARGS='-run=TestAccAWSSecretBackend_wif'
TF_ACC=1 go test -run=TestAccAWSSecretBackend_wif -timeout 30m ./...
ok      github.com/hashicorp/terraform-provider-vault/codegen   1.005s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/framework/validators     1.318s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/identity/entity  1.725s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/provider 2.401s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/ephemeral  2.395s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/sys        1.988s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/testutil  3.127s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util      2.444s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util/mountutil    2.876s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/vault     2.801s
...
```

### Community Note

## PCI review checklist

### DIFF
--- a/vault/resource_aws_secret_backend.go
+++ b/vault/resource_aws_secret_backend.go
@@ -151,6 +151,12 @@ func awsSecretBackendResource() *schema.Resource {
 				Computed:    true,
 				Description: "The TTL of generated identity tokens in seconds.",
 			},
+			consts.FieldMaxRetries: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "Number of max retries the client should use for recoverable errors.",
+			},
 		},
 	}, false)
 
@@ -262,6 +268,9 @@ func awsSecretBackendCreate(ctx context.Context, d *schema.ResourceData, meta in
 		if v, ok := d.GetOk(consts.FieldIdentityTokenTTL); ok && v != 0 {
 			data[consts.FieldIdentityTokenTTL] = v.(int)
 		}
+		if v, ok := d.GetOk(consts.FieldMaxRetries); ok && v != 0 {
+			data[consts.FieldMaxRetries] = v.(int)
+		}
 	}
 
 	if region != "" {
@@ -367,6 +376,9 @@ func awsSecretBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
 			if err := d.Set(consts.FieldIdentityTokenTTL, resp.Data[consts.FieldIdentityTokenTTL]); err != nil {
 				return diag.Errorf("error reading %s for AWS Secret Backend %q: %q", consts.FieldIdentityTokenTTL, path, err)
 			}
+			if err := d.Set(consts.FieldMaxRetries, resp.Data[consts.FieldMaxRetries]); err != nil {
+				return diag.Errorf("error reading %s for AWS Secret Backend %q: %q", consts.FieldMaxRetries, path, err)
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


vault/resource_aws_secret_backend.go
```diff
@@ -370,6 +380,10 @@ func awsSecretBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
 		}
 	}
 
+	if err := d.Set(consts.FieldMaxRetries, resp.Data[consts.FieldMaxRetries]); err != nil {
+		return diag.Errorf("error reading %s for AWS Secret Backend %q: %q", consts.FieldMaxRetries, path, err)
+	}
+
```
aahel: This should be done in if resp != nil block. 
Also check for presence of field in resp before setting state.

vault/resource_aws_secret_backend.go
```diff
@@ -452,6 +466,11 @@ func awsSecretBackendUpdate(ctx context.Context, d *schema.ResourceData, meta in
 			}
 		}
 
+		maxRetries := d.Get(consts.FieldMaxRetries).(int)
+		if maxRetries != 0 {
+			data[consts.FieldMaxRetries] = maxRetries
+		}
+
```
aahel: Have you tested if maxRetries 0 is not allowed ?
I think it would be better to use GetOk to check presence first

vault/resource_aws_secret_backend.go
```diff
@@ -264,6 +270,10 @@ func awsSecretBackendCreate(ctx context.Context, d *schema.ResourceData, meta in
 		}
 	}
 
+	if v, ok := d.GetOk(consts.FieldMaxRetries); ok && v != 0 {
+		data[consts.FieldMaxRetries] = v.(int)
+	}
```
aahel: Have you tested if maxRetries 0 is not allowed ?

vault/resource_aws_secret_backend_test.go
```diff
@@ -57,6 +58,7 @@ func TestAccAWSSecretBackend_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, consts.FieldIAMEndpoint, "https://iam.amazonaws.com"),
 					resource.TestCheckResourceAttr(resourceName, consts.FieldSTSEndpoint, "https://sts.us-west-1.amazonaws.com"),
 					resource.TestCheckResourceAttr(resourceName, consts.FieldLocal, "false"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMaxRetries, "5"),
 				),
```
aahel, line 60: do we have a test which verifies the case when max retries is not given in tf config ?

vault/resource_aws_secret_backend.go
Balaji2198: `GetOk` by default ignores the value 0, so even if we set `max_retries` = 0 it'll be set as default value of -1.

vault/resource_aws_secret_backend.go
```diff
@@ -452,6 +468,10 @@ func awsSecretBackendUpdate(ctx context.Context, d *schema.ResourceData, meta in
 			}
 		}
 
+		if v, ok := d.GetOk(consts.FieldMaxRetries); ok {
+			data[consts.FieldMaxRetries] = v.(int)
```
aahel: If we remove the max_retries from tf config terraform wont detect a diff and hence not get updated.
since its a computed field.

Here its better to keep it optional with a default value (as define in vault api docs) and remove Computed

Also instead of GetOK check we should do
```
data[consts.FieldMaxRetries] = d.Get(consts.FieldMaxRetries)
```

this will set scehma default if field is not there in config

aahel (changes requested):

PR #2627, opened 2025-10-10
Add ttl and max_ttl to gcp secret backend

vijayavelsekar:
### Description

Adding ttl and max_ttl attributes to [vault_gcp_secret_backend](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/gcp_secret_backend) resource associated with the [/gcp/config](https://developer.hashicorp.com/vault/api-docs/secret/gcp#write-config) endpoint

### Checklist

### Community Note

## PCI review checklist


vault/resource_gcp_secret_backend.go
```diff
@@ -239,6 +253,8 @@ func gcpSecretBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
 			consts.FieldIdentityTokenAudience,
 			consts.FieldIdentityTokenTTL,
 			consts.FieldServiceAccountEmail,
+			consts.FieldTTL,
```
siyer-corp: Do these need to be in the ver 1.17 check? It seems to be available since pre 1.15

vault/resource_gcp_secret_backend_test.go
```diff
@@ -249,6 +258,8 @@ EOF
   description = "test description"
   default_lease_ttl_seconds = 1800
   max_lease_ttl_seconds = 43200
+  ttl = 3600
```
siyer-corp, line 268: Might be a good idea to have some negative tests, also check the behavior when the field is not present/removed.

vault/resource_gcp_secret_backend.go
vijayavelsekar: You are right, but gcp secret backend is available from 1.16 and we only had these two fields at that time. I'll make the changes accordingly.

vijayavelsekar: Done

vault/resource_gcp_secret_backend_test.go
vijayavelsekar: Done. Find below screenshot for both positive and negative cases.
In testGCPSecretBackend_initialConfig, I have removed the ttl and max_ttl.
In testGCPSecretBackend_updateConfig, I have added the ttl and max_ttl.
Also added remove ttl scenario on update config operation.

Screenshot 2025-10-15 at 6 07 06 PM

image

vault/resource_gcp_secret_backend.go
```diff
@@ -225,25 +239,36 @@ func gcpSecretBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
...
-	if useAPIVer117Ent {
-		resp, err := client.Logical().ReadWithContext(ctx, gcpSecretBackendConfigPath(path))
-		if err != nil {
-			return diag.FromErr(err)
-		}
-		if resp == nil {
-			return diag.FromErr(fmt.Errorf("GCP backend config %q not found", path))
-		}
+	fields := []string{}
+	if provider.IsAPISupported(meta, provider.VaultVersion116) {
```
siyer-corp: Do we need this check? Both fields are present in older versions of vault as well.

vijayavelsekar: Done.

fairclothjm (approved):
LGTM!

PR #2629, opened 2025-10-14
add retry fields to azure backend config

aahel:
### Description

This pr adds support for the following missing fields in `vault_azure_auth_backend_config`. Api [doc](https://developer.hashicorp.com/vault/api-docs/auth/azure)

- max_retries

- retry_delay

- max_retry_delay

 Closes [VAULT-39759](https://hashicorp.atlassian.net/browse/VAULT-39759)

### Checklist

Screenshot 2025-10-14 at 12 55 23 PM

...

### Community Note

## PCI review checklist

[VAULT-39759]: https://hashicorp.atlassian.net/browse/VAULT-39759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vault/resource_azure_auth_backend_config_test.go
```diff
@@ -97,6 +102,67 @@ func TestAccAzureAuthBackend_wif(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldBackend, backend),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMaxRetries, "10"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldRetryDelay, "5s"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMaxRetryDelay, "3m0s"),
+				),
+			},
+			{
+				// Test with zero max_retries (should be allowed)
+				Config: testAccAzureAuthBackendConfig_retryFields(backend, 5, "1s", "30s"),
```
Balaji2198: The comment refers to testing with zero `max_retries`, but the config sets it to 5. Could we align the comment with the test input or update the input to match the comment?

vault/resource_azure_auth_backend_config.go
```diff
@@ -19,6 +21,22 @@ import (
 
 var azureAuthBackendConfigFromPathRegex = regexp.MustCompile("^auth/(.+)/config$")
 
+// durationDiffSuppressFunc compares duration strings semantically to suppress
+// diffs when durations are equivalent (e.g., "120s" vs "2m0s")
+func durationDiffSuppressFunc(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	// Parse both values as durations
+	oldDuration, err1 := time.ParseDuration(oldValue)
+	newDuration, err2 := time.ParseDuration(newValue)
```
benashz: I wonder if having a single err var would be a better approach here, you could join the errors using `error.Join()`; then log the error at least.

vault/resource_azure_auth_backend_config.go
```diff
@@ -19,6 +21,22 @@ import (
...
+
+	// If both parse successfully, compare the actual durations
+	if err1 == nil && err2 == nil {
```
benashz: You can use the single "joined" err here instead.

vault/resource_azure_auth_backend_config.go
```diff
@@ -19,6 +21,22 @@ import (
...
+		return oldDuration == newDuration
+	}
+
+	// If either fails to parse, fall back to string comparison
+	return oldValue == newValue
```
benashz: I wonder if we should move this comparison to come before any parsing. That would save a few calls if they are equal. I would suggest using https://cs.opensource.google/go/go/+/go1.25.3:src/strings/strings.go;l=1194

vault/resource_azure_auth_backend_config.go
```diff
@@ -19,6 +21,22 @@ import (
 
 var azureAuthBackendConfigFromPathRegex = regexp.MustCompile("^auth/(.+)/config$")
 
+// durationDiffSuppressFunc compares duration strings semantically to suppress
+// diffs when durations are equivalent (e.g., "120s" vs "2m0s")
+func durationDiffSuppressFunc(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
benashz: Would you mind adding an explicit test for this function?

vault/resource_azure_auth_backend_config.go
```diff
@@ -80,6 +98,26 @@ func azureAuthBackendConfigResource() *schema.Resource {
...
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "4s",
+				Description:      "The initial delay between retries for Azure API requests. Defaults to '4s'.",
+				DiffSuppressFunc: durationDiffSuppressFunc,
+			},
+			consts.FieldMaxRetryDelay: {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "60s",
```
benashz: For the provider we like to use seconds (schema.TypeInt) rather than having to deal with Go's duration parsing. Is there a particular reason we cannot use that approach here?

vault/resource_azure_auth_backend_config.go
```diff
@@ -80,6 +98,26 @@ func azureAuthBackendConfigResource() *schema.Resource {
...
+			consts.FieldMaxRetries: {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     3,
+				Description: "Maximum number of retries for Azure API requests. Defaults to 3.",
+			},
+			consts.FieldRetryDelay: {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "4s",
```
benashz: For the provider we like to use seconds (schema.TypeInt) rather than having to deal with Go's duration parsing. Is there a particular reason we cannot use that approach here?

vault/resource_azure_auth_backend_config.go
```diff
@@ -80,6 +98,26 @@ func azureAuthBackendConfigResource() *schema.Resource {
...
+				Description:      "The initial delay between retries for Azure API requests. Defaults to '4s'.",
+				DiffSuppressFunc: durationDiffSuppressFunc,
```
benashz: If we go with an seconds int we can do away with the durationDiffSupressFunc()

vault/resource_azure_auth_backend_config.go
```diff
@@ -19,6 +21,22 @@ import (
...
+// durationDiffSuppressFunc compares duration strings semantically to suppress
+// diffs when durations are equivalent (e.g., "120s" vs "2m0s")
+func durationDiffSuppressFunc(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	// Parse both values as durations
+	oldDuration, err1 := time.ParseDuration(oldValue)
+	newDuration, err2 := time.ParseDuration(newValue)
+
+	// If both parse successfully, compare the actual durations
+	if err1 == nil && err2 == nil {
+		return oldDuration == newDuration
```
benashz: I wonder if we should be comparing Seconds() rather than the entire duration?

vault/resource_azure_auth_backend_config.go
aahel: the api and cli use go's duration strings. i think it would be good to follow same pattern here

vault/resource_azure_auth_backend_config.go
aahel: I think comparing durations would be more precise (converting to seconds might have float64 precision errors)

vault/resource_azure_auth_backend_config.go
aahel: this was done to keep inputs similar to our api which supports go's duration strings

vault/resource_azure_auth_backend_config.go
aahel: the api and cli use go's duration strings. i think it would be good to follow same pattern here

vault/resource_azure_auth_backend_config.go
aahel: done

vault/resource_azure_auth_backend_config.go
aahel: done

vault/resource_azure_auth_backend_config.go
aahel: done

vault/resource_azure_auth_backend_config.go
aahel: done

vault/resource_azure_auth_backend_config.go
benashz: That's a bit of a trap that we fell into before. TFVP is not meant to be CLI compatible; there is no need for that since it has access to the Vault API.

vault/resource_azure_auth_backend_config.go
benashz: Yeah, let's avoid that. Switching to duration as integer type makes thing simple and is standard for this project.

vault/resource_azure_auth_backend_config.go
```diff
@@ -190,6 +216,26 @@ func azureAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+			return diag.Errorf("failed to parse retry_delay from API response: %v (value: %v)", err, v)
+		}
+		if err := d.Set(consts.FieldRetryDelay, int(time.Duration(ns).Seconds())); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	if v, ok := secret.Data[consts.FieldMaxRetryDelay]; ok {
+		ns, err := parseutil.ParseInt(v)
+		if err != nil {
+			return diag.Errorf("failed to parse max_retry_delay from API response: %v (value: %v)", err, v)
```
benashz: Please pass the field name here rather than duplicating it.

vault/resource_azure_auth_backend_config.go
```diff
@@ -190,6 +216,26 @@ func azureAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
 		}
 	}
 
+	// Handle retry delay fields - convert nanoseconds from API to seconds
+	if v, ok := secret.Data[consts.FieldRetryDelay]; ok {
+		ns, err := parseutil.ParseInt(v)
+		if err != nil {
+			return diag.Errorf("failed to parse retry_delay from API response: %v (value: %v)", err, v)
```
benashz: Please pass the field name here rather than duplicating it.

vault/resource_azure_auth_backend_config.go
```diff
@@ -190,6 +216,26 @@ func azureAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+		if err := d.Set(consts.FieldRetryDelay, int(time.Duration(ns).Seconds())); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	if v, ok := secret.Data[consts.FieldMaxRetryDelay]; ok {
+		ns, err := parseutil.ParseInt(v)
+		if err != nil {
+			return diag.Errorf("failed to parse max_retry_delay from API response: %v (value: %v)", err, v)
+		}
+		if err := d.Set(consts.FieldMaxRetryDelay, int(time.Duration(ns).Seconds())); err != nil {
```
benashz: Can any of these delay values be passed as signed integers or zero? I wonder what happens in that case. Is the a better way to prevent any of those from occurring?

vault/resource_azure_auth_backend_config.go
```diff
@@ -115,6 +135,11 @@ func azureAuthBackendWrite(ctx context.Context, d *schema.ResourceData, meta int
 		consts.FieldEnvironment:  environment,
 	}
 
+	// Always send retry fields (using schema defaults when not specified)
+	data[consts.FieldMaxRetries] = d.Get(consts.FieldMaxRetries)
```
benashz, line 139: Why do we always send the default? Could you expand on this a bit more in the comment above?

vault/resource_azure_auth_backend_config_test.go
```diff
@@ -25,8 +25,13 @@ func TestAccAzureAuthBackendConfig_import(t *testing.T) {
 		CheckDestroy:             testAccCheckAzureAuthBackendConfigDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccAzureAuthBackendConfig_basic(backend),
-				Check:  testAccAzureAuthBackendConfigCheck_attrs(backend),
+				Config: testAccAzureAuthBackendConfig_retryFields(backend, 7, 8, 90),
+				Check: resource.ComposeTestCheckFunc(
+					testAccAzureAuthBackendConfigCheck_attrs(backend),
+					resource.TestCheckResourceAttr("vault_azure_auth_backend_config.config", consts.FieldMaxRetries, "7"),
```
benashz: can you split up the setting of each field in separate steps as well?

vault/resource_azure_auth_backend_config_test.go
```diff
@@ -97,6 +102,67 @@ func TestAccAzureAuthBackend_wif(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactories(context.Background(), t),
+		PreCheck:                 func() { testutil.TestAccPreCheck(t) },
+		CheckDestroy:             testCheckMountDestroyed(resourceType, consts.MountTypeAzure, consts.FieldBackend),
+		Steps: []resource.TestStep{
+			{
+				// Test with all retry fields set
+				Config: testAccAzureAuthBackendConfig_retryFields(backend, 5, 10, 120),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldBackend, backend),
```
benashz, line 147: again, it would a good idea to configure each field individually to see what the behavior would be. Like a user first configures max_retries, then ...

vault/resource_azure_auth_backend_config_test.go
```diff
@@ -97,6 +102,67 @@ func TestAccAzureAuthBackend_wif(t *testing.T) {
 	})
 }
 
+func TestAccAzureAuthBackendConfig_retryFields(t *testing.T) {
```
benashz: Ideally these test steps could be included in TestAccAzureAuthBackendConfig_basic above. I don't think they warrant creating an entirely new test case.

benashz (commented):
Looking good. I had some questions/suggestions for your consideration.

vault/resource_azure_auth_backend_config.go
aahel: Only passing the field if its set using GetOk does not work in one case. If the user passes 0 for max_retries in config GetOk returns false(even though its set in config). Hence the field is not passed to vault and it stores default(3). this creates perpetual diffs during tf apply and  also prevents setting it to 0 which is allowed by the api.
```
if v, ok := d.GetOk(consts.FieldMaxRetries); ok {
		data[consts.FieldMaxRetries] = v
	}
```

The current impl passes whatever is set or passes schema default(3) if field is not set. So this works for all values including 0 (int's default).

vault/resource_azure_auth_backend_config.go
aahel: both 0 and -ve gets saved as is in vault. So in this vaut would return the same and we our state gets set ( even if they are -ve or 0).Have validated this works with -ve/0 values as well. The cli and api allow this.

Do we add a validation to prevent adding -ve values in write method  ?  This should ideally be vaullt api's responsibility.

vault/resource_azure_auth_backend_config_test.go
aahel: done

vault/resource_azure_auth_backend_config_test.go
aahel: done

vault/resource_azure_auth_backend_config_test.go
aahel: done

vault/resource_azure_auth_backend_config.go
```diff
@@ -190,6 +216,23 @@ func azureAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+		consts.FieldRetryDelay,
+		consts.FieldMaxRetryDelay,
+	}
+	for _, field := range retryDelayFields {
+		if v, ok := secret.Data[field]; ok {
+			ns, err := parseutil.ParseInt(v)
+			if err != nil {
+				return diag.Errorf("failed to parse %s from API response: %v (value: %v)", field, err, v)
+			}
+			if err := d.Set(field, int(time.Duration(ns).Seconds())); err != nil {
```
benashz, line 230: This will probably break when we address the issue in the azure auth plugin e.g. no longer returning nanos; would it make sense to add a mitigation now? I'd be okay punting that change to future work.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ FEATURES:
 
 * New parameters for `vault_terraform_cloud_secret_role` to support multi-team tokens, by @drewmullen ([#2498](https://github.com/hashicorp/terraform-provider-vault/pull/2498))
 
+* Add retry configuration fields (`max_retries`, `retry_delay`, `max_retry_delay`) to `vault_azure_auth_backend_config` resource for Azure API request resilience
```
benashz: I think we need to add the PR link like the other log lines.

vault/resource_azure_auth_backend_config.go
aahel: could you let me know what mitigation you had in mind in the provider code ? 
since its just an int we wont know its nanosecond or second. we could have some assumption that upto 3600 is second and beyond that is nanosecond but that would have high risk of bugs

benashz (approved):

PR #2635, opened 2025-10-21
Add support for Azure Static Secrets

Zlaticanin:
This PR adds support for Azure Static Secrets. It introduces a new `vault_azure_secret_backend_static_role` resource that defines a static Azure role managed by the Azure Secrets Engine. The resource uses the new Terraform Plugin Framework schema. Vault manages the associated Azure AD application’s credentials automatically based on the role definition.

The PR also adds an ephemeral resource `vault_azure_static_credentials`. This resource acts as a data source for reading or generating Azure static credentials without persisting them in Terraform state. 

Both resources include acceptance tests. The ephemeral tests follow the same pattern as KVv2 and Database ephemeral resources, using the Echo provider to verify that the credentials are correctly generated and returned at runtime.

Setup:
```
  subscription_id = var.subscription_id
  tenant_id       = var.tenant_id
  client_secret   = var.client_secret
  client_id       = var.client_id
}

resource "vault_azure_secret_backend_static_role" "azurerole" {
  backend = vault_azure_secret_backend.azure.path
  role    = "azurerole"

  application_object_id = var.application_object_id
  ttl     = 63079000
  metadata = {
      environment = "test"
      team        = "eco"
  }
}

ephemeral "vault_azure_static_credentials" "azurerole" {
  mount_id = vault_azure_secret_backend_static_role.azurerole.id
  backend = vault_azure_secret_backend.azure.path
  role    = vault_azure_secret_backend_static_role.azurerole.role

  metadata = {
    foo = "bar"
    bob = "baz"
  }
}
```


internal/provider/fwprovider/provider.go
digivava, line 1: Can you include a PR description? Not only is it helpful for reviewers, but since this is a public repo it's helpful for people who are looking at the commit history to understand why a certain PR was made, since they can't see the Jira tickets, especially if this PR ends up causing a regression or something. (God forbid!)

digivava, line 1: Can you add a changelog entry? (Or is there a process to add that later?)

website/docs/r/azure_secret_backend_static_role.html.md
```diff
@@ -0,0 +1,63 @@
+---
+layout: "vault"
+page_title: "Vault: vault_azure_secret_backend_static_role resource"
+sidebar_current: "docs-vault-resource-azure-secret-backend-static-role"
+description: |-
+  Creates an azure secret backend static role for Vault.
```
digivava: nit: I assume we want capital "Azure" here

internal/vault/secrets/azure/azure_static_role_resource.go
digivava, line 1: Was there a decision made to put these files in their own unique folder in `internal/vault/secrets` instead of just inside of `vault` like the rest of the resources? 

I notice that the similar `vault/resource_ldap_secret_backend_static_role.go` is saved alongside `vault/resource_ldap_secret_backend.go`, and there's a `vault/resource_azure_secret_backend_role.go` already there too.

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+func (r *AzureSecretsStaticRoleResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_azure_secret_backend_static_role"
+}
+
+func (r *AzureSecretsStaticRoleResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldBackend: schema.StringAttribute{
+				MarkdownDescription: "The path where the Azure secrets backend is mounted.",
+				Optional:            true,
```
digivava: In the argument reference in the doc you have this field as "(Required)", so I think either this or the doc needs to be adjusted.

I see there's "Computed: true" on there, so maybe you want to add to the doc (and the MarkdownDescription) what it will be computed to by default if the value is not provided by the user.

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldApplicationObjectID: schema.StringAttribute{
+				MarkdownDescription: "Application object ID for an existing service principal that is managed by the static role.",
+				Required:            true,
+			},
+			consts.FieldTTL: schema.Int64Attribute{
```
digivava, line 89: The argument reference on the doc page says "Defaults to 2 years." Are you supposed to set the "Default" here, like you have on the Backend field?

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+				Optional:            true,
+				Sensitive:           true,
+			},
+			consts.FieldExpiration: schema.StringAttribute{
+				MarkdownDescription: "A future expiration time for the imported credential, in RFC3339 format.",
+				Optional:            true,
+			},
+			consts.FieldSkipImportRotation: schema.BoolAttribute{
+				MarkdownDescription: "If true, skip rotation of the client secret on import.",
+				Optional:            true,
```
digivava, line 114: I notice in the doc you have "Valid only during creation". Is there a way to restrict that somewhere? (Feel free to point it out if you're already doing it elsewhere and I didn't notice.)

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+	case float64:
+		return int64(v), nil
+	case string:
+		if d, err := time.ParseDuration(v); err == nil {
+			return int64(d / time.Second), nil
+		}
+		n, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			return 0, fmt.Errorf("unsupported TTL string: %q", v)
+		}
```
digivava, line 359: I assume we also want to convert to absolute value of the number, so we don't get a negative number as a TTL.

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+
+// normalizeTTL converts Vault TTL (string or number) into seconds as int64
+// Since Vault may return either a string or number for TTL, we need to
+// normalize it, so we have a consistent type.
+func normalizeTTL(ttl any) (int64, error) {
+	if ttl == nil {
+		return 0, nil
+	}
+
+	switch v := ttl.(type) {
```
digivava, line 345: What if ttl is an int? Currently this method will just return an error.

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func makeID(backend, role string) string {
+	return fmt.Sprintf("%s/%s/%s", backend, staticRolesAffix, role)
+}
+
+// normalizeTTL converts Vault TTL (string or number) into seconds as int64
+// Since Vault may return either a string or number for TTL, we need to
+// normalize it, so we have a consistent type.
+func normalizeTTL(ttl any) (int64, error) {
```
digivava, line 340: Let's add a unit test for this function.

internal/vault/secrets/azure/azure_static_role_resource.go
```diff
@@ -0,0 +1,361 @@
...
+	if _, err := cli.Logical().WriteWithContext(ctx, path, vaultRequest); err != nil {
+		resp.Diagnostics.AddError(errutil.VaultUpdateErr(err))
+		return
+	}
+
+	data.ID = types.StringValue(makeID(backend, role))
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func buildVaultRequestFromModel(ctx context.Context, data *AzureStaticRoleModel, includeSkipImport bool) (map[string]any, diag.Diagnostics) {
```
digivava, line 247: Can we add a unit test for this function?

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,165 @@
...
+
+	Backend  types.String `tfsdk:"backend"`
+	Role     types.String `tfsdk:"role"`
+	Metadata types.Map    `tfsdk:"metadata"`
+
+	ClientID             types.String `tfsdk:"client_id"`
+	ClientSecret         types.String `tfsdk:"client_secret"`
+	SecretID             types.String `tfsdk:"secret_id"`
+	Expiration           types.String `tfsdk:"expiration"`
+	CombinedRoleMetadata types.Map    `tfsdk:"combined_role_metadata"`
```
digivava: Is this one supposed to also be documented in the "attributes reference" on the doc page?

digivava (changes requested):

internal/provider/fwprovider/provider.go
Zlaticanin: The changelog gets generated later when we do the release for all changes that are being released.

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: Yes, we are now using framework schema for our resources, so now we have a new location where we create them. There aren't many examples of this new process yet, only a couple.

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: No we don't have to since this will be returned by Vault, and we have a default set there.

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: Yes, we are doing it in Vault. So the `skip_import_rotation` field only has an affect on Create, and is ignored on updates. 
https://github.com/hashicorp/vault-plugin-secrets-azure-enterprise/blob/main/path_static_roles.go#L244

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: Good catch. I'll make this required here as well to stay consistent with others.

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: I pulled this code out in a function, so I can reuse it for both Create and Update functions. I do have the test written for the workflow here: https://github.com/hashicorp/terraform-provider-vault/blob/8b2bb5d84a63503fe08f0170410389832e4849fa/internal/vault/secrets/azure/azure_static_role_resource_test.go#L26-L120

It tests both the initial config and then makes sure we can update it as well. Let me know if you had some other testing in mind here

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
Zlaticanin: Are you referring to `CombinedRoleMetadata` here? If so, then we don't want to include it in the "attributes reference" on the docs because we are not returning this on read. 
I had to add this `CombinedRoleMetadata` because of how ephemerals work. Since we allow `metadata` to be set both on role creation and on the credentials endpoint, and ephemerals don't support plan modifiers or state handling, we had to have this separate field for it.

This is the error I was getting when trying use same field for both
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Provider produced invalid ephemeral resource instance
│ 
│   with ephemeral.vault_azure_static_access_credentials.azurerole,
│ The provider for vault_azure_static_access_credentials produced an inconsistent result: .metadata: planned value cty.MapVal(map[string]cty.Value{"environment":cty.StringVal("test"),
│ "milena":cty.StringVal("zlaticanin"), "team":cty.StringVal("ecoddooo")}) does not match config value cty.MapVal(map[string]cty.Value{"bob":cty.StringVal("baz"), "foo":cty.StringVal("bar")}).
```

Because of how ephemerals work, Vinay and I decided that this would be the best approach.

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: I don't think we have to do this as we don't allow setting negative numbers in Vault, it would error out.
```
Error writing data to azure/static-roles/my-role: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/azure/static-roles/my-role
Code: 400. Errors:

* error converting input for field "ttl": cannot provide negative value '-31536000'
```

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: added here: https://github.com/hashicorp/terraform-provider-vault/pull/2635/commits/042d6ed8d5abdf23a18bc94a3ca4873985fba687

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: Not sure how to do this since with this new process of how we write this resources, my [role resource is package](https://github.com/hashicorp/terraform-provider-vault/blob/8b2bb5d84a63503fe08f0170410389832e4849fa/internal/vault/secrets/azure/azure_static_role_resource.go#L4) `azure` and [my test resource](https://github.com/hashicorp/terraform-provider-vault/blob/8b2bb5d84a63503fe08f0170410389832e4849fa/internal/vault/secrets/azure/azure_static_role_resource_test.go#L4) has to have `_test` extension in the package name to be able to run. Since these are not in the same package, we can't reference the helper. 
Trying to find a way on how to do this, but there aren't many examples 😬

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
```diff
@@ -0,0 +1,59 @@
...
+  backend = vault_azure_secret_backend.azure.path
+  role    = vault_azure_secret_backend_static_role.example.role
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `backend` - (Required) Path to the mounted Azure Secrets Engine where the static role resides.
```
vinay-gopalan, line 55: we should add the parameters that are available to all Ephemeral resources by default: `namespace` and `mount_id`. Example docs for `mount_id` [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/website/docs/ephemeral-resources/kv_secret_v2.html.md?plain=1#L62)

internal/vault/secrets/azure/azure_static_role_resource.go
vinay-gopalan: Might be worth also setting it to be `Computed` so the default can be set by Vault

Zlaticanin: done here: https://github.com/hashicorp/terraform-provider-vault/pull/2635/commits/c51c4daf72e49b5d7fb3dc908c215ce2715d0fcc

website/docs/r/azure_secret_backend_static_role.html.md
```diff
@@ -0,0 +1,63 @@
...
+    owner = "vault"
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `role` - (Required) Name of the static role to create or manage.
```
vinay-gopalan, line 57: should document the `namespace` parameter

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
Zlaticanin: Good catch! Thank you.

Added here: https://github.com/hashicorp/terraform-provider-vault/pull/2635/commits/c51c4daf72e49b5d7fb3dc908c215ce2715d0fcc

copied `mount_id` from the example you sent above, and copied the `namespace` docs from [azure access credential resource](https://github.com/hashicorp/terraform-provider-vault/blob/main/website/docs/d/azure_access_credentials.html.md?plain=1#L67-L70)

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
```diff
@@ -0,0 +1,68 @@
...
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault/index.html#namespace).
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Required) Path to the mounted Azure Secrets Engine where the static role resides.
+
+* `role` - (Required) The name of the static role to generate or read credentials for.
+
+* `metadata` - (Optional) A map of key-value pairs to associate with the static role and include in the credential response.
+
+## Attributes Reference
```
vinay-gopalan, line 62: should `combined_role_metadata` be here?

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,166 @@
...
+			for k, v := range inMeta {
+				kvPairs = append(kvPairs, fmt.Sprintf("%s=%s", k, v))
+			}
+			readData = map[string][]string{"metadata": kvPairs}
+		}
+	}
+
+	var sec *api.Secret
+	if readData != nil {
+		sec, err = c.Logical().ReadWithData(path, readData)
```
vinay-gopalan: nit: using ctx methods
```suggestion
		sec, err = c.Logical().ReadWithDataWithContext(ctx, path, readData)
```

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
vinay-gopalan: Yes, to be more precise, this is not just for ephemeral resources, but TF resources in general. In this case, we have a scenario where the data input to Vault (`metadata`) and the response returned from Vault will always have a constant drift, as the data returned by the API is input + extra fields

To avoid a constant state drift in TF, generally we tend to split these fields to isolate the behavior. `metadata` is provided to Vault, while `combined_role_metadata` is the read-only value returned by the Vault API

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
Zlaticanin: Hmm.. I am not sure? It's a little bit confusing to me since in Vault, if we set `metadata` on creds endpoint, we only see those fields on that first response. Every next read will not return those fields, as they are not persisted to storage (and this is intended behavior, not a bug like we initially thought). 
I wasn't sure if I should include it since I couldn't really even see that field set when I was doing local testing. What do you think? @vinay-gopalan

website/docs/r/azure_secret_backend_static_role.html.md
Zlaticanin: done herE: https://github.com/hashicorp/terraform-provider-vault/pull/2635/commits/bd59ff7ad527cc62832c41b0dae73b5fce050532

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
vinay-gopalan: > I wasn't sure if I should include it since I couldn't really even see that field set when I was doing local testing

Want to make sure I understand what you're saying here. Does this mean you were sending metadata to the creds read endpoint but were not seeing input + role metadata being returned? The idea here is that if the user wants to use the complete metadata, then they can always use the `combined_role_metadata` field

In the case where you don't send `metadata` to Vault, `combined_role_metadata` should still be set, and would equal to your role metadata, IIUC.

vinay-gopalan: The docs should tell the user how and when to use `combined_role_metadata`. Currently, the docs do not mention how to make use of the complete metadata returned on a creds read by Vault

internal/vault/secrets/azure/azure_static_role_resource.go
digivava: Yeah this is one of the tradeoffs of having test code in a separate package like you're doing, which is typically done if you want to have more "black-box"-style "external" tests. But you can have multiple test files in the same directory, even if they're in different packages.

I asked my wise mentor from the Boundary team for advice here, and he mentioned it being quite common to have a directory like:
```
azure
  azure.go (package: azure)
  azure_test.go (package: azure)
  azure_ext_test.go (package: azure_test)
```
Unit tests for things like internal functions can go in azure_test.go, and if you need more black-box-style testing, it can go in azure_ext_test.go.

website/docs/ephemeral-resources/azure_secrets_static_credentials.html.md
Zlaticanin: So, I updated our docs slightly to differentiate between the input and output fields.
But to be honest, I am a little confused on how we can print this after our apply? We can't use outputs here, and after I apply the ephemeral resource 
```
ephemeral "vault_azure_static_credentials" "azurerole" {
  mount_id = vault_azure_secret_backend_static_role.azurerole.id
  backend = vault_azure_secret_backend.azure.path
  role    = vault_azure_secret_backend_static_role.azurerole.role

  metadata = {
    foo = "bar-updated"
    bob = "baz"
  }
}
```
```
vault_azure_secret_backend.azure: Creating...
vault_azure_secret_backend.azure: Creation complete after 0s [id=azure]
vault_azure_secret_backend_static_role.azurerole: Creating...
vault_azure_secret_backend_static_role.azurerole: Creation complete after 1s [id=azure/static-roles/azurerole]
ephemeral.vault_azure_static_credentials.azurerole: Opening...
ephemeral.vault_azure_static_credentials.azurerole: Opening complete after 0s
ephemeral.vault_azure_static_credentials.azurerole: Closing...
ephemeral.vault_azure_static_credentials.azurerole: Closing complete after 0s
```

it doesn't print out the response, like we do in Vault. So how can we see this `merged_metadata`?

internal/vault/secrets/azure/azure_static_role_resource.go
Zlaticanin: TIL! Thank you. Just created a new test file with the same package name: https://github.com/hashicorp/terraform-provider-vault/pull/2635/commits/06b75a407dfc06e17e29f6637bbf9184511850f1

vinay-gopalan (approved):

digivava (approved):
Looking good!

internal/vault/secrets/azure/azure_helpers_test.go
```diff
@@ -0,0 +1,42 @@
...
+// Test_normalizeTTL tests the normalizeTTL function with various input types.
+// The function should correctly parse and convert the input to an int64 representing seconds.
+func Test_normalizeTTL(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   any
+		want    int64
+		wantErr bool
+	}{
+		{"nil", nil, 0, false},
```
benashz, line 17: In the future it would be good to avoid using positional values.

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,172 @@
...
+	RequestMetadata types.Map    `tfsdk:"request_metadata"`
+
+	ClientID     types.String `tfsdk:"client_id"`
+	ClientSecret types.String `tfsdk:"client_secret"`
+	SecretID     types.String `tfsdk:"secret_id"`
+	Expiration   types.String `tfsdk:"expiration"`
+	Metadata     types.Map    `tfsdk:"metadata"`
+}
+
+type AzureStaticCredsAPIModel struct {
```
benashz, line 48: It would be great if we could document the public members of this package.

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package ephemeralsecrets
```
benashz, line 4: usually the package name matches it's parent directory.

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,172 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral"
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/base"
```
benashz, line 15: typically we group local module imports separately from all other imports.

internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/client"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/errutil"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/model"
+	"github.com/hashicorp/vault/api"
+)
+
+const fieldRequestMetadata = "request_metadata"
```
benashz, line 22: I believe we normally store all field names in the `consts` package.

PR #2640, opened 2025-10-24
Upgrade go version and acceptance test matrix

vinay-gopalan:
### Description
- Includes new Vault 1.21 Ent in the testing pipeline
- upgrades go version and Vault dependencies

## PCI review checklist


.github/workflows/build.yml
```diff
@@ -154,10 +154,9 @@ jobs:
       matrix:
         image:
         - "vault-enterprise:1.16.25-ent"
-        - "vault-enterprise:1.17.18-ent"
-        - "vault-enterprise:1.18.14-ent"
```
fairclothjm, line 158: Are we dropping support for these? If yes, should we note that in the docs/upgrade guide? We note our current support here: https://registry.terraform.io/providers/hashicorp/vault/latest/docs/guides/version_5_upgrade#what-vault-server-versions-are-supported-in-version-5x

fairclothjm (commented):
Overall LGTM. Left a question about supported versions.

.github/workflows/build.yml
vinay-gopalan: Good question. The goal is not to drop support, but just not testing against those versions. My thought in this regard was that since we do test the LTS version of Vault 1.16, we should be good to not test against Vault versions that are no longer actively maintained

I am not strong on this change though and can revert it if you think it could cause more issues down the line due to lack of test coverage!

vinay-gopalan: It's a good callout that we will probably have to update documentation to specify which versions we test against, so the overall change might not be worth it. I can remove this and we can revisit the contract for the TFVP going forward

vinay-gopalan: Opted to keep the 1.17.x and 1.18.x versions in the build matrix for this release. Will revisit in the following release

sapleu-hc (approved):

robmonte (approved):

PR #2644, opened 2025-10-31
Added fields related to namespace used to create a role in kubernetes auth method

tejashwiniingalagi:
### Description

added fields bound_service_account_names, bound_service_account_namespaces to create a role for kubernetes auth method in vault. Handled below conditions
1. Show an error if both are empty
2. If only one of the above set then set the other to empty.

Relates OR Closes #0000

### Checklist

Screenshot 2025-11-06 at 4 14 27 PM

```





<!--- Please keep this note for the community --->

### Community Note

<!--- Thank you for keeping this note for the community --->


## PCI review checklist

<!-- heimdall_github_prtemplate:grc-pci_dss-2024-01-05 -->

### CODE REVIEW

File: vault/resource_kubernetes_auth_backend_role.go [hunk 1]
```diff
@@ -93,8 +116,36 @@ func kubernetesAuthBackendRoleUpdateFields(d *schema.ResourceData, data map[stri
...
+		if boundServiceAccountNamespaceSelector, ok := d.GetOk("bound_service_account_namespace_selector"); ok {
+			data["bound_service_account_namespace_selector"] = boundServiceAccountNamespaceSelector.(string)
+		}
+	} else {
+		// always send the current value from state on update so the backend preserves or clears it
+		if v := d.Get("bound_service_account_namespace_selector"); v != nil {
+			data["bound_service_account_namespace_selector"] = v.(string)
+		} else {
+			data["bound_service_account_namespace_selector"] = ""
+		}
```
aahel: v != nil check is not needed here since Get will by default return "".  So this can be simplified to

```
 data["bound_service_account_namespace_selector"] = d.Get("bound_service_account_namespace_selector").(string)
```

aahel (changes requested):

vault/resource_kubernetes_auth_backend_role.go
tejashwiniingalagi: done

vault/data_source_kubernetes_auth_backend_role.go
```diff
@@ -44,6 +44,12 @@ func kubernetesAuthBackendRoleDataSource() *schema.Resource {
 			Computed:    true,
 			Description: "List of namespaces allowed to access this role. If set to \"*\" all namespaces are allowed, both this and bound_service_account_names can not be set to \"*\".",
 		},
+		"bound_service_account_namespace_selector": {
```
benashz, line 47: Would you mind pulling this field out to consts.go? Feel free refactor the other other fields in a subsequent PR.

vault/resource_kubernetes_auth_backend_role.go
```diff
@@ -79,6 +85,23 @@ func kubernetesAuthBackendRoleResource() *schema.Resource {
...
 
 		Schema: fields,
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, m interface{}) error {
+			// At least one of bound_service_account_namespaces or bound_service_account_namespace_selector must be provided
+			namespaces, hasNamespaces := d.GetOk("bound_service_account_namespaces")
+			selector, hasSelector := d.GetOk("bound_service_account_namespace_selector")
+
+			// Check if namespaces is empty set
+			namespacesEmpty := !hasNamespaces || namespaces.(*schema.Set).Len() == 0
```
benashz: Is there reason we need to check the len of namespaces here? I believe `d.GetOk()` does this for you.

vault/resource_kubernetes_auth_backend_role.go
```diff
@@ -79,6 +85,23 @@ func kubernetesAuthBackendRoleResource() *schema.Resource {
 		},
 
 		Schema: fields,
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, m interface{}) error {
```
benashz: This function makes some assumptions about the Vault API and may in fact duplicate Vault's input validation. I wonder if we could let Vault take care of the input validation rather than having the provider do it? The error would be bubbled up to the user in this case.

vault/resource_kubernetes_auth_backend_role.go
```diff
@@ -93,8 +116,25 @@ func kubernetesAuthBackendRoleUpdateFields(d *schema.ResourceData, data map[stri
...
 	}
 
-	if boundServiceAccountNamespaces, ok := d.GetOk("bound_service_account_namespaces"); ok {
-		data["bound_service_account_namespaces"] = boundServiceAccountNamespaces.(*schema.Set).List()
+	if create {
+		if boundServiceAccountNamespaces, ok := d.GetOk("bound_service_account_namespaces"); ok {
+			data["bound_service_account_namespaces"] = boundServiceAccountNamespaces.(*schema.Set).List()
+		}
+	} else {
+		// always send the current value from state on update so the backend preserves or clears it
```
benashz: I don't understand this bit. Why do we differentiate between create and update here?

vault/resource_kubernetes_auth_backend_role.go
```diff
@@ -93,8 +116,25 @@ func kubernetesAuthBackendRoleUpdateFields(d *schema.ResourceData, data map[stri
...
+		data["bound_service_account_namespace_selector"] = d.Get("bound_service_account_namespace_selector").(string)
+		// "bound_service_account_namespace_selector"] = d.Get("bound_service_account_namespace_selector").(string)
+	}
+
+	// Always set bound_service_account_namespace_selector to ensure proper clearing on updates
+	if create {
```
benashz: Same as my comment above

nit: having duplicate conditionals on `create` make the code harder to understand, I wonder if we could merge this conditional block with the one above.

vault/data_source_kubernetes_auth_backend_role.go
tejashwiniingalagi: done. Moved all the fields to consts.go

vault/resource_kubernetes_auth_backend_role.go
tejashwiniingalagi: removed this code. Not doing the validations now

vault/resource_kubernetes_auth_backend_role.go
tejashwiniingalagi: done

vault/resource_kubernetes_auth_backend_role.go
tejashwiniingalagi: Update was setting the old value. However I have removed this code and tested with only one block. Its working as expected

vault/resource_kubernetes_auth_backend_role.go
tejashwiniingalagi: done

PR #2654, opened 2025-11-11
added vault_azure_access_credentials ephemeral resource

aahel:
### Description
This pr adds vault_azure_access_credentials ephemeral resource. This reads ephemeral dynamic credentials from an Azure secret backend in Vault.

These credentials are not stored in Terraform state. 

### Checklist

Screenshot 2025-11-13 at 2 21 35 PM

### Community Note

## PCI review checklist

### DIFF
--- a/website/docs/ephemeral-resources/azure_access_credentials.html.md
+++ b/website/docs/ephemeral-resources/azure_access_credentials.html.md
@@ -61,7 +61,7 @@ The following arguments are supported:
 
 * `validate_creds` - (Optional) Whether generated credentials should be validated before being returned.
 
-* `num_sequential_successes` - (Optional) If 'validate_creds' is true, the number of sequential successes required to validate generated credentials. Defaults to 8.
+* `num_sequential_successes` - (Optional) If 'validate_creds' is true, the number of sequential successes required to validate generated credentials. Defaults to 4.
 
 * `num_seconds_between_tests` - (Optional) If 'validate_creds' is true, the number of seconds to wait between each test of generated credentials. Defaults to 1.
 


internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	c, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	log.Printf("[DEBUG] Sleep to prevent azure rate limits.This happens when open is called multiple times quickly during plan,apply,refresh")
+	time.Sleep(5 * time.Second)
```
fairclothjm: Is there a better way to handle this? Here we always wait 5 seconds, even if the API is healthy. This slows down every plan, apply, and refresh.

Instead, could we sleep and retry (with exponential backoff) only if we receive a rate-limit error? Does Vault return a reliable error in this case?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+FEATURES:
```
Zlaticanin, line 15: If our process remained the same, we don't need a changelog entry here. We add it once we do a release.

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: Vault returns a generic 500 status code.  We would have to do a `strings.Contains(errMsg, "concurrent requests being made")` on the error message 
Screenshot 2025-11-14 at 8 07 04 AM

CHANGELOG.md
aahel: removed my changelog entry. Will someone else add the changelog before 5.5.0 ?

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: have implemented exp retry with back off based on error message and status code(500)

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+// https://learn.microsoft.com/en-us/graph/sdks/national-clouds
+const (
+	azurePublicCloudEnvName = "AZUREPUBLICCLOUD"
+	azureChinaCloudEnvName  = "AZURECHINACLOUD"
+	azureUSGovCloudEnvName  = "AZUREUSGOVERNMENTCLOUD"
+
+	// Default values for credential validation
+	defaultNumSecondsBetweenTests   = 1
+	defaultMaxCredValidationSeconds = 300
+	defaultNumSequentialSuccesses   = 8
```
aahel: @Zlaticanin 
i have kept all the default values same as the corresponding data source. But i have seen it tends to fail a lot if we require 8 consecutive successes within 300 seconds.
 Should i reduce `defaultNumSequentialSuccesses`  to 4 ? or let it be 8 to be consistent with the data source

CHANGELOG.md
aahel: @fairclothjm can you confirm this as well. Asking since I had added Changelog for some of my previous pr's as well

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+	ClientID     string `json:"client_id" mapstructure:"client_id"`
+	ClientSecret string `json:"client_secret" mapstructure:"client_secret"`
+}
+
+// Schema defines this resource's schema which is the data that is available in
+// the resource's configuration, plan, and state
+func (r *AzureAccessCredentialsEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"backend": schema.StringAttribute{
```
Zlaticanin: It would be good to use consts here, like we did for `consts.FieldLeaseDuration`. If there are any that we don't already have, we can add them here: https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/consts/consts.go

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+				Optional:            true,
+			},
+			"num_seconds_between_tests": schema.Int64Attribute{
+				MarkdownDescription: "If 'validate_creds' is true, the number of seconds to wait between each test of generated credentials.",
+				Optional:            true,
+			},
+			"max_cred_validation_seconds": schema.Int64Attribute{
+				MarkdownDescription: "If 'validate_creds' is true, the number of seconds after which to give up validating credentials.",
+				Optional:            true,
+			},
```
Zlaticanin: Small nit, should we add the defaults here directly maybe? Like we have it in the [old](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/data_source_azure_access_credentials.go#L63-L79) data source?

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+	ClientID       types.String `tfsdk:"client_id"`
+	ClientSecret   types.String `tfsdk:"client_secret"`
+	LeaseID        types.String `tfsdk:"lease_id"`
+	LeaseDuration  types.Int64  `tfsdk:"lease_duration"`
+	LeaseStartTime types.String `tfsdk:"lease_start_time"`
+	LeaseRenewable types.Bool   `tfsdk:"lease_renewable"`
+}
+
+// AzureAccessCredentialsAPIModel describes the Vault API data model.
+type AzureAccessCredentialsAPIModel struct {
```
Zlaticanin: I don't see this being used anywhere? Should we use this instead of the manual `secret.Data` casts?

internal/vault/secrets/ephemeral/azure_access_credentials.go
Zlaticanin: Do we see the same types of failures with the old data source as well? 🤔

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+		resp.Diagnostics.Append(resp.Result.Set(ctx, &data)...)
+		return
+	}
+
+	// Credential validation logic
+	configPath := backend + "/config"
+	var config *api.Secret
+	getConfigData := func() (map[string]interface{}, error) {
+		if config == nil {
+			c, err := c.Logical().ReadWithContext(ctx, configPath)
```
Zlaticanin: This seems like a bug? We are using the same variable name (`c`), for both the client side and the returned secret. This seems like it would give us a compile error? We should probably rename the returned secret: https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/data_source_azure_access_credentials.go#L180

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+		})
+		if err != nil {
+			log.Printf("[WARN] Failed to marshal private data: %s", err)
+		} else {
+			resp.Private.SetKey(ctx, "lease_data", privateData)
+		}
+	}
+
+	// If we're not supposed to validate creds, we're done
+	if !data.ValidateCreds.ValueBool() {
```
Zlaticanin, line 281: What happens if `validate_creds` is unset/null here?

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -0,0 +1,508 @@
...
+	bo.MaxElapsedTime = 5 * time.Minute
+
+	err = backoff.RetryNotify(
+		func() error {
+			var readErr error
+			secret, readErr = c.Logical().ReadWithContext(ctx, credsPath)
+			if readErr != nil {
+				errMsg := readErr.Error()
+				// Check if this is a rate limit error from Azure
+				if strings.Contains(errMsg, "concurrent requests being made") {
```
Zlaticanin: Is this the only error Azure returns? It seems like if Azure returns anything other than this string, this would get skipped, no?

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: ephemeral resource schema does not support default

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: Removed this struct. we are not casting secret.Data anywhere

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: It's not a bug. here c inside is shadowing the outside the client c in the scope of the if block. After this point we are not using c as client inside the if block

aahel: but shadowing does create confusion. renamed it to `configSecret`

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: ValueBool() returns false if its unset or null

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: this is the only relevant error we get when we do multiple requests simultaneosly. vault api returns a generic 500 status code. So we have to check this message.

https://learn.microsoft.com/en-us/graph/errors (check 409 status code here)
There is another possible error(429) but i have not encountered that. We could have checked codes like 409 and 429 but vault api is returning generic 500 status code.

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: yes we see the same error for the existing datasource as well

```
--- FAIL: TestAccDataSourceAzureAccessCredentials_ExistingSP (459.87s)
    /Users/aahelguha/hashicorp/terraform-provider-vault/vault/data_source_azure_access_credentials_test.go:91: Step 1/1 error: Error running post-apply non-refresh plan: exit status 1
        
        Error: validation failed after max_cred_validation_seconds of 300, giving up; now=2025-11-19 08:25:59.321817 +0530 IST m=+459.639157501, endTime=2025-11-19 08:23:49.822334 +0530 IST m=+453.203480251
        
          with data.vault_azure_access_credentials.test,
          on terraform_plugin_test.tf line 28, in data "vault_azure_access_credentials" "test":
          28: data "vault_azure_access_credentials" "test" {

```

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: done

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: have added both status code(500) and error message check to be more precise

internal/vault/secrets/ephemeral/azure_access_credentials.go
Zlaticanin: Should we maybe increase the time? Have we tested different times, how long does it take on average?

internal/vault/secrets/ephemeral/azure_access_credentials.go
Zlaticanin: Can we use that struct [here](https://github.com/hashicorp/terraform-provider-vault/blob/1775c9e3076b2efb8baf8229c93c06a98c990cf2/internal/vault/secrets/ephemeral/azure_access_credentials.go#L246-L262) for the credentials?

Similarly how we do it for [AWS](https://github.com/hashicorp/terraform-provider-vault/pull/2657/files#diff-75ae8d73778ae3bde114eae25c0eddb5183ec3cd98fde168e97956058b250ed0R103-R110) and [Azure static secrets](https://github.com/hashicorp/terraform-provider-vault/blob/1775c9e3076b2efb8baf8229c93c06a98c990cf2/internal/vault/secrets/ephemeral/azure_secrets_static_credentials.go#L150).

aahel: done

internal/vault/secrets/ephemeral/azure_access_credentials.go
aahel: even 600 seconds is failing. i think we would need to wait a long time 8 consecutive successes. i have observed it passes 3,4 conecutively then fails 1 and gets reset. Its not reliable
i think we should reduce  `defaultNumSequentialSuccesses` to 3 ?

Zlaticanin: Hmm... Thanks for looking into this. The 3–4 successes then one failure pattern you see makes me think the credentials are actually valid, and Azure is just throwing intermittent throttling or temporary errors. And since our logic resets the whole success count on any error, 8 might be hard to achieve. 😬 

I tried to find a reason why we chose those defaults in the first place, and it seems like ([based on the PR description here](https://github.com/hashicorp/terraform-provider-vault/pull/713)), it's only because this is what worked before. But this implementation for the validation was all the way in 2020. So it worked before, but it seems like Azure might be noisier now and returns more errors. 

I am hesitant to change the defaults because that would be a behavior change, in case some users were setting `validate_creds=true` and relied on current defaults. Can we do some investigation on what exact errors are we getting during those failures? I feel like that might help us understand whether we should improve the validation logic itself first, before deciding if the defaults need to be lowered.

Zlaticanin: I was thinking something like maybe expanding the logging [here](https://github.com/hashicorp/terraform-provider-vault/blob/ae8728c282bc92d77c0dfefdb3240a0611fb5052/internal/vault/secrets/ephemeral/azure_access_credentials.go#L410-L422) to include the error type and HTTP status code. I think that would give us more context on what is happening.

aahel: yeah i had seen the errors using a debugger while running tests. its just auth errors after a few attempts since credentials has not reached all the servers due to eventual consistency

aahel: my proposal was to reduce the default for this ephemeral resource not the datasource

Zlaticanin: Okay. If there isn't a better way to catch these error, and it's just well known eventual consistency issue with Azure, let's lower the default then. It will be clear in the docs what the default is for the ephemeral. 
Thank you for working on this. I will do a final pass tomorrow when I log on, but yeah looking good!

aahel: changed `defaultNumSequentialSuccesses` to 4

website/docs/ephemeral-resources/azure_access_credentials.html.md
```diff
@@ -0,0 +1,90 @@
...
+  configured [namespace](/docs/providers/vault/index.html#namespace).
+  *Available only for Vault Enterprise*.
+
+* `backend` - (Required) Path to the mounted Azure Secrets Engine where the role resides.
+
+* `role` - (Required) The name of the Azure role to generate credentials for.
+
+* `validate_creds` - (Optional) Whether generated credentials should be validated before being returned.
+
+* `num_sequential_successes` - (Optional) If 'validate_creds' is true, the number of sequential successes required to validate generated credentials. Defaults to 8.
```
Zlaticanin: Can we update the default here since we did change it?

Zlaticanin (approved):

CHANGELOG.md
fairclothjm: We should continue to add changelog entries to each PR in TFVP. We have not automated TFVP changelogs at this time.

During the release, we will do our best to include any missing entries: https://github.com/hashicorp/terraform-provider-vault/pull/2663. However, we recommend adding them to reduce the work during the release.

PR #2655, opened 2025-11-12
Implement ephemeral resources for gcp

vijayavelsekar:
### Description

Ephemeral resources implementation:

vault_gcp_oauth2_access_token - [Generate secret (IAM service account creds): OAuth2 access token](https://developer.hashicorp.com/vault/api-docs/secret/gcp#generate-secret-iam-service-account-creds-oauth2-access-token)

vault_gcp_service_account_key - [Generate secret (IAM service account creds): service account key](https://developer.hashicorp.com/vault/api-docs/secret/gcp#generate-secret-iam-service-account-creds-service-account-key)

### Checklist

image

### Community Note

## PCI review checklist

### DIFF
--- a/internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
+++ b/internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
@@ -23,12 +23,17 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/testutil"
 )
 
+const (
+	// Common GCP service account email pattern (without anchors)
+	gcpServiceAccountEmailPattern = `.+@.+\.iam\.gserviceaccount\.com`
+)
+
 var (
 	// Common regexp patterns used across GCP ephemeral resource tests
-	regexpGCPServiceAccountEmail = regexp.MustCompile(`^.+@.+\.iam\.gserviceaccount\.com$`)
+	regexpGCPServiceAccountEmail = regexp.MustCompile(fmt.Sprintf(`^%s$`, gcpServiceAccountEmailPattern))
 	regexpGCPServiceAccountType  = regexp.MustCompile(`(?s).*"type":\s*"service_account".*`)
 	regexpGCPPrivateKey          = regexp.MustCompile(`(?s).*"private_key":\s*"-----BEGIN PRIVATE KEY-----.*`)
-	regexpGCPClientEmail         = regexp.MustCompile(`(?s).*"client_email":\s*".+@.+\.iam\.gserviceaccount\.com".*`)
+	regexpGCPClientEmail         = regexp.MustCompile(fmt.Sprintf(`(?s).*"client_email":\s*"%s".*`, gcpServiceAccountEmailPattern))
 )
 
 // TestAccGCPServiceAccountKey_basic tests generating service account keys


internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral"
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
```
benashz, line 22: nit: please set your IDE to group the module package separately.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+// the resource's configuration, plan, and state
+func (r *GCPOAuth2AccessTokenEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"backend": schema.StringAttribute{
+				MarkdownDescription: "GCP Secret Backend to read credentials from.",
+				Required:            true,
+			},
+			"roleset": schema.StringAttribute{
+				MarkdownDescription: "GCP Secret Roleset to generate OAuth2 access token for. Mutually exclusive with `static_account` and `impersonated_account`.",
```
benashz: nit: please fold these long lines somewhere around 120 columns or less.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+// Schema defines this resource's schema which is the data that is available in
+// the resource's configuration, plan, and state
+func (r *GCPOAuth2AccessTokenEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"backend": schema.StringAttribute{
+				MarkdownDescription: "GCP Secret Backend to read credentials from.",
+				Required:            true,
+			},
+			"roleset": schema.StringAttribute{
```
benashz: Any reason this field has not been added to the consts?

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+				MarkdownDescription: "GCP Secret Roleset to generate OAuth2 access token for. Mutually exclusive with `static_account` and `impersonated_account`.",
+				Optional:            true,
+			},
+			"static_account": schema.StringAttribute{
```
benashz: Same for the field constant question above.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+				MarkdownDescription: "GCP Secret Static Account to generate OAuth2 access token for. Mutually exclusive with `roleset` and `impersonated_account`.",
+				Optional:            true,
+			},
+			"impersonated_account": schema.StringAttribute{
+				MarkdownDescription: "GCP Secret Impersonated Account to generate OAuth2 access token for. Mutually exclusive with `roleset` and `static_account`.",
+				Optional:            true,
+			},
+			"token": schema.StringAttribute{
```
benashz: I believe we already have this field as a constant; would you mind updating accordingly. Thanks!

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+	base.MustAddBaseEphemeralSchema(&resp.Schema)
+}
+
+// Metadata sets the full name for this resource
+func (r *GCPOAuth2AccessTokenEphemeralResource) Metadata(ctx context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_gcp_oauth2_access_token"
+}
+
+// isRetryableError checks if the error is retryable (e.g., resource not ready yet)
+func isRetryableError(err error) bool {
```
benashz: This looks like a generic function, can you check to see if we don't already have something like this in other code?

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+	if err == nil {
+		return false
+	}
+	errMsg := strings.ToLower(err.Error())
+	// Common error patterns that indicate the resource might not be ready yet
+	retryablePatterns := []string{
+		"unable to generate token - make sure your roleset service account and key are still valid",
```
benashz: Not sure if I would consider this to be a pattern, this feels a bit non-deterministic. I wonder if error wrapping could help us here, or perhaps inspecting the Vault response itself, you could then go beyond error string to checking the HTTP status code.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
benashz: nit: are we sure that shouldn't be spelled `role_set` - it seems like two words.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,321 @@
...
+func (r *GCPOAuth2AccessTokenEphemeralResource) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
+	var data GCPOAuth2AccessTokenModel
+	// Read Terraform prior state data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Validate that exactly one of roleset, static_account, or impersonated_account is provided
```
benashz, line 157: I am not entirely sure why need to use the booleans to increment provided count?

internal/vault/secrets/ephemeral/gcp_oauth2_access_token_test.go
```diff
@@ -0,0 +1,480 @@
...
+	nonEmpty := regexp.MustCompile(`^.+$`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
```
benashz, line 49: Would it be good to add a second step to test out multiple calls?

internal/vault/secrets/ephemeral/gcp_service_account_key.go
```diff
@@ -0,0 +1,246 @@
...
+	ServiceAccountEmail types.String `tfsdk:"service_account_email"`
+	LeaseID             types.String `tfsdk:"lease_id"`
+	LeaseDuration       types.Int64  `tfsdk:"lease_duration"`
+	LeaseStartTime      types.String `tfsdk:"lease_start_time"`
+	LeaseRenewable      types.Bool   `tfsdk:"lease_renewable"`
+}
+
+// Schema defines this resource's schema which is the data that is available in
+// the resource's configuration, plan, and state
+func (r *GCPServiceAccountKeyEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
```
benashz, line 63: Similar comments about field constants etc.

internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
```diff
@@ -0,0 +1,399 @@
...
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGCPServiceAccountKeyRolesetConfig(backend, roleset, creds, project),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify private_key_data contains expected JSON structure (matches service_account type and has private_key)
+					statecheck.ExpectKnownValue("echo.gcp_key", tfjsonpath.New("data").AtMapKey("private_key_data"), knownvalue.StringRegexp(regexp.MustCompile(`(?s).*"type":\s*"service_account".*`))),
```
benashz: Any reason we couldn't pull these compiled regexes at to package level vars?

benashz (commented):
Looking good! Just had a few comments for you.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
vijayavelsekar: @benashz as per the vault api docs its roleset, hence retaining the same naming convention. You can find the reference here https://developer.hashicorp.com/vault/api-docs/secret/gcp#roleset

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
vijayavelsekar: @benashz The aim of this provided count is to throw error in 2 scenarios as follows
1) Throw error if all the 3 attributes (RoleSet, StaticAccount and ImpersonatedAccount) are missing.
we can check this using `!hasRoleset && !hasStaticAccount && !hasImpersonatedAccount`
2) Throw error if more than 1 attribute is used out of those 3 attributes (Should be mutually exclusive). we can check this using `(hasRoleset && hasStaticAccount) || (hasRoleset && hasImpersonatedAccount) || (hasStaticAccount && hasImpersonatedAccount)`

As you can see the conditions are difficult to read, I've just simply counted the boolean fields if they exist and replaced the complex condition checking operation with just `providedCount == 0 and providedCount > 1` for the above two scenarios respectively.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,304 @@
...
+			"Error reading from Vault",
+			fmt.Sprintf("Error generating GCP OAuth2 access token from path %q after %d attempts: %s\n\n"+
+				"This may indicate that the %s %q or its associated GCP service account is not yet fully created or configured. "+
+				"Please ensure:\n"+
+				"1. The %s exists in Vault at the specified backend\n"+
+				"2. The GCP service account has been created and granted necessary permissions\n"+
+				"3. There is sufficient time between creating the %s and requesting tokens\n"+
+				"4. The Vault GCP secrets engine is properly configured",
+				tokenPath, attemptCount, retryErr, resourceType, resourceName, resourceType, resourceType),
+		)
```
benashz, line 275: I really like the detailed diagnostics here, but this code makes a lot of assumptions about the Vault API. What happens if that API changes? Would these diagnostics no longer be valid?

It feels like these messages should come from the Vault itself, rather than the TFVP client.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,304 @@
...
+	} else {
+		impersonatedAccount := data.ImpersonatedAccount.ValueString()
+		tokenPath = mount + "/impersonated-account/" + impersonatedAccount + "/token"
+		resourceType = "impersonated account"
+		resourceName = impersonatedAccount
+	}
+
+	// Configure exponential backoff with max retries
+	// Initial interval: 2s, Max interval: 30s, Max retries: 5
+	exponentialBackoff := backoff.NewExponentialBackOff()
```
benashz: nit: it's okay in Go to be little more terse with var names: suggest renaming the variable to `bo`.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,316 @@
...
+	// common fields to all ephemeral resources
+	base.BaseModelEphemeral
+
+	// fields specific to this resource
+	Mount               types.String `tfsdk:"mount"`
+	Roleset             types.String `tfsdk:"roleset"`
+	StaticAccount       types.String `tfsdk:"static_account"`
+	ImpersonatedAccount types.String `tfsdk:"impersonated_account"`
+
+	// computed fields
```
benashz, line 69: Will these always be computed fields?

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,316 @@
...
+		impersonatedAccount := data.ImpersonatedAccount.ValueString()
+		tokenPath = mount + "/impersonated-account/" + impersonatedAccount + "/token"
+		resourceType = "impersonated account"
+		resourceName = impersonatedAccount
+	}
+
+	// Configure exponential backoff with max retries
+	// Initial interval: 2s, Max interval: 30s, Max retries: 5
+	expBackoff := backoff.NewExponentialBackOff()
+	expBackoff.InitialInterval = 2 * time.Second
```
benashz: are there any concerns about terraform's internal context timeout?

Also, the initial interval seems a bit pessimistic at 2 seconds.

I also wonder if you have considered making these values configurable by the user?

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -0,0 +1,316 @@
...
+
+		// Read the OAuth2 access token from Vault (GET request)
+		secret, err := c.Logical().ReadWithContext(ctx, tokenPath)
+
+		if err != nil {
+			// Check if the error is retryable based on HTTP status code and error message
+			// Retry on 400 Bad Request with specific error message - this typically
+			// indicates the resource isn't ready yet
+			if isGCPTokenGenerationError(err) {
+				log.Printf("[DEBUG] Attempt %d failed for %s %q. Error: %s", attemptCount, resourceType, resourceName, err)
```
benashz: These feel more like trace level rather than debug.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token_test.go
```diff
@@ -0,0 +1,490 @@
...
+)
+
+var (
+	// Common regexp patterns used across GCP ephemeral resource tests
+	regexpNonEmpty = regexp.MustCompile(`^.+$`)
+)
+
+// TestAccGCPOAuth2AccessToken_roleset tests generating an OAuth2 access token
+// from a GCP roleset with all ephemeral resource attributes
+func TestAccGCPOAuth2AccessToken_roleset(t *testing.T) {
```
benashz: Is there reason we need so many tests? I wonder if mutating resource within a single test might be more realistic in terms of the UX.

internal/vault/secrets/ephemeral/gcp_service_account_key.go
```diff
@@ -0,0 +1,256 @@
...
+	var keyData map[string]interface{}
+	if err := json.Unmarshal(decodedKey, &keyData); err == nil {
+		if email, ok := keyData["client_email"].(string); ok {
+			data.ServiceAccountEmail = types.StringValue(email)
+		}
+	}
+
+	data.LeaseID = types.StringValue(vaultSecret.LeaseID)
+	data.LeaseDuration = types.Int64Value(int64(vaultSecret.LeaseDuration))
+	data.LeaseStartTime = types.StringValue(time.Now().Format(time.RFC3339))
```
benashz, line 252: Is this accurate? Shouldn't the lease start time come from Vault?

internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
```diff
@@ -0,0 +1,422 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/providertest"
+	"github.com/hashicorp/terraform-provider-vault/testutil"
+)
+
+var (
+	// Common regexp patterns used across GCP ephemeral resource tests
+	regexpGCPServiceAccountEmail = regexp.MustCompile(`^.+@.+\.iam\.gserviceaccount\.com$`)
+	regexpGCPServiceAccountType  = regexp.MustCompile(`(?s).*"type":\s*"service_account".*`)
+	regexpGCPPrivateKey          = regexp.MustCompile(`(?s).*"private_key":\s*"-----BEGIN PRIVATE KEY-----.*`)
```
benashz, line 35: Not sure I full get this? Are we implementing our own PEM decoder? If we we are doing any output validation I would suggest using the `encoding/pem` builtin package.

internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
```diff
@@ -0,0 +1,422 @@
...
+	regexpGCPClientEmail         = regexp.MustCompile(`(?s).*"client_email":\s*".+@.+\.iam\.gserviceaccount\.com".*`)
```
benashz: The last bit of the regex pattern looks very similar to pattern form regexpGCPServiceAccountEmail - you may want to use a string formatter here to share the patterns - minus the anchors of course.

internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
```diff
@@ -0,0 +1,422 @@
...
+)
+
+// TestAccGCPServiceAccountKey_roleset tests generating a service account key
+// from a GCP roleset with all ephemeral resource attributes
+func TestAccGCPServiceAccountKey_roleset(t *testing.T) {
```
benashz: Similar comment about test step encapsulation to mimic the potential UX here.

benashz (commented):
Looking good. I add some comments. Further review will be necessary.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
vijayavelsekar: Yes, these will always be computed fields because, they are output-only fields and users cannot and should not be able to set these values - they are strictly outputs from Vault's GCP secrets engine

internal/vault/secrets/ephemeral/gcp_service_account_key.go
vijayavelsekar: The LeaseStartTime is intentionally client-generated and not from Vault's API response. It's a convenience field to help users track when the lease was obtained.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
vijayavelsekar: Valid point. I'm changing this logic to constant backoff, just like the existing code for NameSpace.
1 second interval × 10 retries = ~10 seconds total
By this way, the retry operation will stay within Terraform's internal context timeout.
And Yes, it would be nice idea to have the retries configurable by the user. I'll make those changes too.

internal/vault/secrets/ephemeral/gcp_service_account_key_test.go
vijayavelsekar: No, we are not implementing our own PEM Decoder. This is just a test assertion to verify the API response contains the expected field with a PEM-formatted private key. The actual PEM parsing/validation (if needed) would be done by the consuming application, not by this test.

internal/vault/secrets/ephemeral/gcp_oauth2_access_token.go
```diff
@@ -290,24 +298,18 @@ func (r *GCPOAuth2AccessTokenEphemeralResource) Open(ctx context.Context, req ep
...
-		data.TokenTTL = types.Int64Value(int64(tokenTTL))
-	} else if tokenTTL, ok := vaultSecret.Data["token_ttl"].(int64); ok {
-		data.TokenTTL = types.Int64Value(tokenTTL)
-	} else {
-		// If token_ttl is not in the response data, set it to null
-		data.TokenTTL = types.Int64Null()
+	// Extract token_ttl if present (returned as json.Number by Vault client)
+	data.TokenTTL = types.Int64Null()
+	if ttlValue, exists := vaultSecret.Data["token_ttl"]; exists {
+		if jsonNum, ok := ttlValue.(json.Number); ok {
```
benashz, line 304: nit: typically the pattern is `if v, ok := ` where v is the result of the type assert. No need to be too verbose here.

PR #2701, opened 2025-12-10
VAULT-41068 TypeList to TypeSet conversion of network fields

vijayavelsekar:
### Description

TypeList to TypeSet conversion to handle duplicate values and spurious diff issue during terraform apply.

### Checklist

Screenshot 2025-12-10 at 10 15 36 PM
Screenshot 2025-12-10 at 10 24 52 PM

### Community Note

## PCI review checklist

### DIFF
--- a/internal/sync/secrets_sync.go
+++ b/internal/sync/secrets_sync.go
@@ -43,12 +43,10 @@ func SyncDestinationCreateUpdateWithOptions(ctx context.Context, d *schema.Resou
 			if typeSetFields != nil && typeSetFields[k] {
 				if set, ok := v.(*schema.Set); ok {
 					data[k] = set.List()
-				} else {
-					data[k] = v
+					continue
 				}
-			} else {
-				data[k] = v
 			}
+			data[k] = v
 		}
 	}
 


vault/resource_secrets_sync_vercel_destination.go
```diff
@@ -90,24 +97,24 @@ func vercelSecretsSyncDestinationResource() *schema.Resource {
 					"'preview' & 'production'.",
 			},
 			consts.FieldAllowedIPv4Addresses: {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
```
fairclothjm, line 100: Is this a breaking change? Is there any effect on existing state files generated when the fields were TypeList? We should verify that upgrading doesn't cause any issues.

fairclothjm: If we must make this change, then we can try to implement a State Upgrader to migrate the existing state data from the List format to the Set format.

fairclothjm (approved):
I believe we have not yet released the new fields for the Vercel resource so this change should be ok. LGTM!

vault/resource_secrets_sync_vercel_destination.go
vijayavelsekar: As discussed in the slack chat, this is not a breaking change and it has not been released yet.

PR #2734, opened 2026-01-08
Add metadata field to Azure dynamic roles

Zlaticanin:
This PR adds `metadata` field to dynamic roles, similarly to how it's done for Azure static roles. User can set `metadata` on role config and on the request. If a key exists in both, the value from the role's metadata takes precedence.


internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -168,6 +171,16 @@ func (r *AzureAccessCredentialsEphemeralResource) Schema(_ context.Context, _ ep
 				MarkdownDescription: "True if the duration of this lease can be extended through renewal.",
 				Computed:            true,
 			},
+			fieldRequestMetadata: schema.MapAttribute{
```
siyer-corp, line 174: Any reason why this field isn't referring to the consts?

vault/resource_azure_secret_backend_role_test.go
```diff
@@ -276,6 +289,10 @@ resource "vault_azure_secret_backend_role" "test_azure_roles" {
  description	  = "Test for Vault Provider"
  sign_in_audience = "AzureADMyOrg"
  tags             = ["team:engineering"]
+ metadata = {
```
siyer-corp, line 383: Do we need tests for metadata removal, empty maps, and adding metadata to existing roles?

internal/vault/secrets/ephemeral/azure_access_credentials_test.go
```diff
@@ -78,6 +79,15 @@ func TestAccAzureAccessCredentialsEphemeralResource_basic(t *testing.T) {
 							statecheck.ExpectKnownValue("echo.test_azure",
 								tfjsonpath.New("data").AtMapKey("lease_renewable"),
 								knownvalue.NotNull()),
+							statecheck.ExpectKnownValue("echo.test_azure",
```
siyer-corp, line 82: If the role does not have any metadata, will the response just have the metadata from the access_credentials? Should we have a test for that scenario?

internal/vault/secrets/ephemeral/azure_access_credentials.go
```diff
@@ -199,6 +212,22 @@ func (r *AzureAccessCredentialsEphemeralResource) Open(ctx context.Context, req
 	role := data.Role.ValueString()
 	credsPath := backend + "/creds/" + role
 
+	var readData map[string][]string
```
siyer-corp, line 215: Would it be good to add documentation for this logic?

internal/vault/secrets/ephemeral/azure_access_credentials.go
Zlaticanin: It is a const, just in the file, since this field was added because of the limitation of having two same fields both in role config and credentials

internal/vault/secrets/ephemeral/azure_access_credentials_test.go
Zlaticanin: Yes, if they are set. We do have a test for it?

PR #2736, opened 2026-01-09
 Deprecated fields  'organization' and 'token' with new fields and added write only fields for api_token on Vault_okta_auth_backend

sanvikam79:
### Description
The vault_okta_auth_backend is using deprecated fields:

organization - replaced by org_name

token - replaced by api_token

Implement write only fields: token and api_token on vault_okta_auth_backend

### Checklist

Screenshot 2026-01-09 at 3 55 04 PM
Screenshot 2026-01-09 at 3 49 47 PM
Screenshot 2026-01-09 at 3 49 28 PM
Screenshot 2026-01-09 at 3 48 59 PM
Screenshot 2026-01-09 at 3 48 38 PM

### Community Note

## PCI review checklist


vault/resource_okta_auth_backend.go
```diff
@@ -57,19 +58,72 @@ func oktaAuthBackendResource() *schema.Resource {
...
-			Sensitive:   true,
+			Type:          schema.TypeString,
+			Optional:      true,
+			Description:   "The Okta API token. This is required to query Okta for user group membership. If this is not supplied only locally configured groups will be enabled. Use api_token instead.",
+			Sensitive:     true,
+			Deprecated:    "Use api_token instead",
+			ConflictsWith: []string{consts.FieldAPIToken, consts.FieldAPITokenWO, consts.FieldTokenWO},
+		},
+
+		consts.FieldTokenWO: {
```
siyer-corp: We can skip adding WO fields for deprecated fields

vault/resource_okta_auth_backend.go
```diff
@@ -57,19 +58,72 @@ func oktaAuthBackendResource() *schema.Resource {
...
+			Optional:      true,
+			Description:   "Write-only Okta API token. This is required to query Okta for user group membership. If this is not supplied only locally configured groups will be enabled. Use api_token_wo instead.",
+			Sensitive:     true,
+			WriteOnly:     true,
+			Deprecated:    "Use api_token_wo instead",
+			ConflictsWith: []string{consts.FieldToken, consts.FieldAPIToken, consts.FieldAPITokenWO},
+			RequiredWith:  []string{consts.FieldTokenWOVersion},
+		},
+
+		consts.FieldTokenWOVersion: {
```
siyer-corp: Same as line 110

vault/resource_okta_auth_backend.go
```diff
@@ -403,8 +545,24 @@ func oktaAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta int
...
+	if v, ok := d.GetOk(consts.FieldAPIToken); ok {
+		configuration[consts.FieldToken] = v.(string)
+	} else if v, ok := d.GetOk(consts.FieldToken); ok {
+		configuration[consts.FieldToken] = v.(string)
+	} else if d.HasChange(consts.FieldAPITokenWOVersion) {
+		// User is using write-only api_token_wo and version changed
+		if apiTokenWo, _ := d.GetRawConfigAt(cty.GetAttrPath(consts.FieldAPITokenWO)); !apiTokenWo.IsNull() {
+			configuration[consts.FieldToken] = apiTokenWo.AsString()
+		}
+	} else if d.HasChange(consts.FieldTokenWOVersion) {
```
siyer-corp: Should get removed once the field is removed from the schema

website/docs/r/okta_auth_backend.html.md
```diff
@@ -148,6 +163,18 @@ These arguments are common across several Authentication Token resources since V
...
 
+## Ephemeral Attributes Reference
+
+The following write-only attributes are supported:
+
+* `api_token_wo` - (Optional) Write-only Okta API token. This is required to query Okta for user group membership.
+  Use this for enhanced security when you don't want the token to appear in state files. Requires `api_token_wo_version`. Conflicts with `token`, `api_token`, and `token_wo`.
+  **Note**: This property is write-only and will not be read from the API.
+
+* `token_wo` - (Optional) **Deprecated: Use `api_token_wo` instead.** Write-only Okta API token. This is required to query Okta for user group membership.
```
siyer-corp: Should get removed once the field is removed from the schema

vault/resource_okta_auth_backend.go
```diff
@@ -244,10 +333,42 @@ func parseDurationSeconds(i interface{}) (string, error) {
 	return strconv.Itoa(int(d.Seconds())), nil
 }
 
+// getOrgName returns the organization name, preferring the new field name
```
siyer-corp: Do we need this? Won't fallback be handled by using ExactlyOneOf? Only one of these would be allowed in the plan and to pass to vault.

siyer-corp (changes requested):

vault/resource_okta_auth_backend.go
sanvikam79: Done

vault/resource_okta_auth_backend.go
sanvikam79: Done

website/docs/r/okta_auth_backend.html.md
sanvikam79: Done

vault/resource_okta_auth_backend.go
sanvikam79: Done

vault/resource_okta_auth_backend.go
sanvikam79: Yes As ExactlyOneOf will Validates at plan time where as getOrgName() determines at runtime which field the user actually provided

vault/resource_okta_auth_backend.go
```diff
@@ -57,19 +58,53 @@ func oktaAuthBackendResource() *schema.Resource {
...
+
 		consts.FieldOrganization: {
-			Type:        schema.TypeString,
-			Required:    true,
-			Optional:    false,
-			Description: "The Okta organization. This will be the first part of the url https://XXX.okta.com.",
+			Type:         schema.TypeString,
+			Optional:     true,
+			Description:  "The Okta organization. This will be the first part of the url https://XXX.okta.com. Use org_name instead.",
+			Deprecated:   "Use org_name instead",
```
Zlaticanin: small nit: It would be good to write that the field will be deprecated and then note what to use instead for clarity.

vault/resource_okta_auth_backend.go
```diff
@@ -209,13 +244,48 @@ func oktaAuthBackendResource() *schema.Resource {
...
+// When the deprecated 'organization' field is removed in a future major version,
+// this function can be simplified to just call schema.ImportStatePassthroughContext
+// or the logic to set both fields can be removed.
+func oktaAuthBackendImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	// Set the resource ID from the import path
+	d.SetId(d.Id())
+
+	// Call the Read function to populate all resource attributes from Vault
+	// This includes path, description, groups, users, tokens, accessor, tune, etc.
+	diags := oktaAuthBackendRead(ctx, d, meta)
```
Zlaticanin: Question - is there a reason why we call `oktaAuthBackendRead` inside the importer as opposed to using `schema.ImportStatePassthroughContext`?

vault/resource_okta_auth_backend.go
```diff
@@ -209,13 +244,48 @@ func oktaAuthBackendResource() *schema.Resource {
...
+	if diags.HasError() {
+		return nil, fmt.Errorf("error reading okta auth backend during import: %v", diags)
+	}
+
+	// For backward compatibility: set both org_name and organization fields
+	// This ensures import works whether the user's config uses the new field (org_name)
+	// or the deprecated field (organization)
+	if orgName, ok := d.GetOk(consts.FieldOrgName); ok {
```
Zlaticanin: Do we need the same thing in the Read path, so users don't get diffs after refresh if Vault only returns `organization`?

vault/resource_okta_auth_backend.go
```diff
@@ -244,10 +314,22 @@ func parseDurationSeconds(i interface{}) (string, error) {
...
+// getOrgName returns the organization name, preferring the new field name
+func getOrgName(d *schema.ResourceData) string {
+	if v, ok := d.GetOk(consts.FieldOrgName); ok {
+		return v.(string)
+	}
+	// Fallback to deprecated field
+	if v, ok := d.GetOk(consts.FieldOrganization); ok {
+		return v.(string)
+	}
+	return ""
```
Zlaticanin, line 299: So here, if neither field is set we reuturn an empty string. Is there any chance ths could cause us to send an empty org value back to Vault and overwrite an existing one?

vault/resource_okta_auth_backend.go
```diff
@@ -401,10 +504,21 @@ func oktaAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta int
...
-		consts.FieldBaseURL:      d.Get(consts.FieldBaseURL),
-		fieldBypassOktaMFA:       d.Get(fieldBypassOktaMFA),
-		consts.FieldOrganization: d.Get(consts.FieldOrganization),
-		consts.FieldToken:        d.Get(consts.FieldToken),
+		consts.FieldBaseURL: d.Get(consts.FieldBaseURL),
+		fieldBypassOktaMFA:  d.Get(fieldBypassOktaMFA),
+		consts.FieldOrgName: getOrgName(d),
+	}
+
+	// Handle token field - check regular fields first, then write-only versions
```
Zlaticanin, line 488: Can you confirm that this won't cause TF to re-apply on every run when using `api_token_wo` unless the version is bumped?

Zlaticanin (commented):
A few questions on the initial pass. 
Was this tested manually or just with written tests?

vault/resource_okta_auth_backend.go
sanvikam79: @Zlaticanin I have attached the screen shot for the same in JIRA 
https://hashicorp.atlassian.net/browse/VAULT-41655

vault/resource_okta_auth_backend.go
sanvikam79: @Zlaticanin I have attached the screen shot for the same in JIRA
https://hashicorp.atlassian.net/browse/VAULT-41655

vault/resource_okta_auth_backend.go
sanvikam79: @Zlaticanin I have attached the screen shot for the same in JIRA
https://hashicorp.atlassian.net/browse/VAULT-41655

vault/resource_okta_auth_backend.go
sanvikam79: @Zlaticanin Yes could notice diff after refresh in the following scenario
      terraform import --> terraform refresh --> terraform apply 
but did not find diff post refresh in the following scenario
      terraform apply --> terraform refresh
 made code changes for the same

vault/resource_okta_auth_backend.go
Zlaticanin: Thank you for the explanation, that was helpful! 
So, the mismatch is happening because `Read` is always setting both `organization` and `org_name` in state which then causes the `Import` to fail when the config only specifies one of them, correct?
This fix makes sense, but I am a little bit concerned that always setting both fields in state + `DiffSuppress` is going to add some complexity.
Is it possible to use `org_name` as the primary field we store in state and handle `organization` as an alias from config? @sanvikam79

sanvikam79: @Zlaticanin 
I've implemented your suggestion as mentioned below

- org_name is the primary field - always set in state

- organization is handled as a config alias - read via GetRawConfig() in Update

- Both fields set in Read - required for import to work correctly 

    Scenario 1- terraform apply---> terraform referesh --> terraform plan --> seen no drift as the state will always have org_name and if config has deprecated field the terraform will add it automatically to state.

    Scenario 2- terraform import---> terraform referesh --> terraform plan --> seen drift as the state will always have org_name and if config has deprecated field the terraform will not add it automatically to state hence in read function setting both values

- No DiffSuppressFunc - instead, both fields have Computed: true

vault/resource_okta_auth_backend.go
```diff
@@ -401,10 +480,23 @@ func oktaAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta int
 	log.Printf("[DEBUG] Updating auth %s in Vault", path)
 
 	configuration := map[string]interface{}{
-		consts.FieldBaseURL:      d.Get(consts.FieldBaseURL),
-		fieldBypassOktaMFA:       d.Get(fieldBypassOktaMFA),
-		consts.FieldOrganization: d.Get(consts.FieldOrganization),
-		consts.FieldToken:        d.Get(consts.FieldToken),
+		consts.FieldBaseURL: d.Get(consts.FieldBaseURL),
+		fieldBypassOktaMFA:  d.Get(fieldBypassOktaMFA),
+		consts.FieldOrgName: getOrgName(d),
```
Zlaticanin, line 485: Do we want to avoid sending `org_name` at all when `getOrgName` returns empty? Since both `org_name` and `organization` are `Optional + Computed`, is it possible to hit the Update operation with neither set in the config, and would that cause to write `org_name: ""` to Vault?

vault/resource_okta_auth_backend.go
```diff
@@ -57,19 +58,55 @@ func oktaAuthBackendResource() *schema.Resource {
...
-			Type:        schema.TypeString,
-			Required:    true,
-			Optional:    false,
-			Description: "The Okta organization. This will be the first part of the url https://XXX.okta.com.",
+			Type:         schema.TypeString,
+			Optional:     true,
+			Computed:     true,
+			Description:  "The Okta organization. This will be the first part of the url https://XXX.okta.com. Use org_name instead.",
+			Deprecated:   "Use org_name instead",
+			ExactlyOneOf: []string{consts.FieldOrgName, consts.FieldOrganization},
```
Zlaticanin, line 75: Now that we have both `org_name` and `organization` as `Optional + Computed, with ExactlyOneOf`, is it possible for config to set neither? If we still require one, should we add `AtLeastOneOf` here?

vault/resource_okta_auth_backend.go
sanvikam79: ExactlyOneOf validation blocks plan/apply without the field. Hence their is no possibility for org_name to be "" even if Optional + Computed

vault/resource_okta_auth_backend.go
sanvikam79: AtLeastOneOf  does not prevents setting both fields.

Zlaticanin (approved):
Thank you for working on this! LGTM!

PR #2739, opened 2026-01-12
Add support for validate_assertion_signature and validate_response_signature in vault_saml_auth_backend resource

siyer-corp:
### Description

Add the following fields to `vault_saml_auth_backend`:

1. `validate_assertion_signature`
2. `validate_response_signature`

### Checklist

```
$ make testacc TESTARGS='-run=TestAccSAMLAuthBackend'

...
```
image

### Community Note

## PCI review checklist

### DIFF
--- a/vault/resource_saml_auth_backend.go
+++ b/vault/resource_saml_auth_backend.go
@@ -97,7 +97,7 @@ func samlAuthBackendResource() *schema.Resource {
 			consts.FieldVerboseLogging: {
 				Type:     schema.TypeBool,
 				Optional: true,
-				ForceNew: true,
+				Computed: true,
 				Description: "Log additional, potentially sensitive information " +
 					"during the SAML exchange according to the current logging level. Not " +
 					"recommended for production.",
@@ -105,13 +105,13 @@ func samlAuthBackendResource() *schema.Resource {
 			consts.FieldValidateAssertionSignature: {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				ForceNew:    true,
+				Computed:    true,
 				Description: "Whether to validate the assertion signature.",
 			},
 			consts.FieldValidateResponseSignature: {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				ForceNew:    true,
+				Computed:    true,
 				Description: "Whether to validate the response signature.",
 			},
 			consts.FieldTune: authMountTuneSchema(),
--- a/vault/resource_saml_auth_backend_test.go
+++ b/vault/resource_saml_auth_backend_test.go
@@ -234,18 +234,18 @@ func TestAccSAMLAuthBackend_booleanFlags(t *testing.T) {
 				),
 			},
 			{
-				Config: testAccSAMLAuthBackendConfig_booleansExplicit(path, true, true),
+				Config: testAccSAMLAuthBackendConfig_booleansExplicit(path, true, true, false),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(resourceName, consts.FieldVerboseLogging, "true"),
 					resource.TestCheckResourceAttr(resourceName, consts.FieldValidateAssertionSignature, "true"),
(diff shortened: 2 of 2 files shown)


vault/resource_saml_auth_backend.go
```diff
@@ -174,7 +174,16 @@ func samlAuthBackendUpdate(ctx context.Context, d *schema.ResourceData, meta int
 
 	// add boolean fields
 	for _, k := range samlBooleanAPIFields {
-		data[k] = d.Get(k)
+		// validate_assertion_signature and validate_response_signature require Vault 1.19+
+		if k == consts.FieldValidateAssertionSignature || k == consts.FieldValidateResponseSignature {
+			if provider.IsAPISupported(meta, provider.VaultVersion119) {
```
Zlaticanin: Would it make more sense here to check the Vault version first rather than inside the field loop?

vault/resource_saml_auth_backend.go
```diff
@@ -229,6 +238,12 @@ func samlAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inter
 	fields := append(samlAPIFields, samlBooleanAPIFields...)
 	for _, k := range fields {
 		if v, ok := resp.Data[k]; ok {
+			// validate_assertion_signature and validate_response_signature require Vault 1.19+
+			if k == consts.FieldValidateAssertionSignature || k == consts.FieldValidateResponseSignature {
+				if !provider.IsAPISupported(meta, provider.VaultVersion119) {
```
Zlaticanin: same here as the above

vault/resource_saml_auth_backend.go
```diff
@@ -229,6 +238,12 @@ func samlAuthBackendRead(ctx context.Context, d *schema.ResourceData, meta inter
 	fields := append(samlAPIFields, samlBooleanAPIFields...)
 	for _, k := range fields {
 		if v, ok := resp.Data[k]; ok {
+			// validate_assertion_signature and validate_response_signature require Vault 1.19+
+			if k == consts.FieldValidateAssertionSignature || k == consts.FieldValidateResponseSignature {
```
Zlaticanin: Are these fields ent only?

siyer-corp: The entire resource is ent only.

vault/resource_saml_auth_backend.go
siyer-corp: Thanks, changes are incorporated.

siyer-corp:
@Zlaticanin , I noticed that for the boolean fields in this auth method, vault does not update the values from true back to false. To address this, I've set ForceNew=true on these fields, which forces a recreation of the resource to apply the change. What are your thoughts on this workaround?

siyer-corp:
> > > > > > @Zlaticanin , I noticed that for the boolean fields in this auth method, vault does not update the values from true back to false. To address this, I've set ForceNew=true on these fields, which forces a recreation of the resource to apply the change. What are your thoughts on this workaround?
> > > > > 
> > > > > 
> > > > > Hmm. How does this behave on the plugin side, are these flags create only possibly?
> > > > 
> > > > 
> > > > It appears from the plugins code that we cannot set both flags to false once created, because of this piece of [code](https://github.com/hashicorp/vault-plugin-auth-saml/blob/main/path_config.go#L140), unsure if this is a bug.
> > > 
> > > 
> > > Not sure. I don't know much about that plugin to be honest. Were you able to dig through the code/commits to find the initial implementation PR on the plugin side?
> > 
> > 
> > Yes, found the PR [here](https://github.com/hashicorp/vault-plugin-auth-saml/commit/8d62e4458fb388fa13ef2416e26d60d0ba1a8a33#diff-c6f0305d05f9581c583eb2b15d17d32f412dcbc73809741b065ab801cc5d360eR24), I don't see an explicit reasoning why both cannot be updated to false. It says it's recommended to have both enabled if the IDP allows it.
> > Should we just let vault handle this logic and remove the forced creation of the resource here?
> 
> Yeah, I would just let Vault handle the logic.

Made the changes accordingly.

PR #2747, opened 2026-01-19
Added fields default_extensions_template , exclude_cidr_list , port to vault_ssh_secret_backend_role

sanvikam79:
Description

Added the following fields to vault_ssh_secret_backend_role:

- default_extensions_template

- exclude_cidr_list

- port

DiffSuppressFunc is used to prevent false drift detection for key_type specific fields.
SSH roles support different key types (ca, otp) with different applicable fields:
    - CA-only fields: default_extensions, default_extensions_template
    - OTP-only fields: exclude_cidr_list, port
When a field is specified in config but doesn't apply to the role's key_type,Vault ignores it. Without DiffSuppressFunc, Terraform would detect drift between config (field present) and state (field null/empty). DiffSuppressFunc suppresses this diff when the field doesn't apply to the current key_type, preventing
unnecessary plan changes and allowing users to maintain configs that work across
different key types.

### Checklist

Screenshot 2026-01-19 at 5 21 20 PM
The below screen shots are of "terraform show" post "terraform apply"
Screenshot 2026-02-13 at 11 06 35 PM
Screenshot 2026-02-13 at 1 45 46 PM
Screenshot 2026-02-13 at 1 45 59 PM

### Community Note

## PCI review checklist


internal/consts/consts.go
```diff
@@ -32,7 +32,7 @@ const (
 	FieldRequestTimeout                 = "request_timeout"
 	FieldSchema                         = "schema"
 	FieldPasswordPolicy                 = "password_policy"
-	FieldLength                         = "length"
+	FieldLength                         = "lengths"
```
siyer-corp: Can we have a separate const for the plural string? This might cause issues if "length" is being used elsewhere?

vault/resource_ssh_secret_backend_role.go
```diff
@@ -30,103 +31,127 @@ var (
...
+			Optional:    true,
+			Default:     false,
+			Description: "Specifies if the default_extensions field supports templating.",
+		},
+		consts.FieldDefaultCriticalOptions: {
 			Type:     schema.TypeMap,
 			Optional: true,
 		},
-		"allowed_users_template": {
+		consts.FieldExcludeCIDRList: {
```
siyer-corp, line 117: Can we use TypeSet here to handle duplicates here?

vault/resource_ssh_secret_backend_role.go
```diff
@@ -151,7 +176,7 @@ func sshSecretBackendRoleResource() *schema.Resource {
 							}
 						},
 					},
-					"lengths": {
+					consts.FieldLength: {
```
siyer-corp: Would be better to map _length -> FieldLengths_

vault/resource_ssh_secret_backend_role.go
sanvikam79: Done

internal/consts/consts.go
sanvikam79: Done

vault/resource_ssh_secret_backend_role.go
sanvikam79: Done

vault/resource_ssh_secret_backend_role.go
```diff
@@ -209,95 +237,112 @@ func sshSecretBackendRoleWrite(d *schema.ResourceData, meta interface{}) error {
...
 
-	if v, ok := d.GetOk("allowed_users_template"); ok {
-		data["allowed_users_template"] = v.(bool)
+	if v, ok := d.GetOk(consts.FieldDefaultCriticalOptions); ok {
+		data[consts.FieldDefaultCriticalOptions] = v
 	}
 
-	if v, ok := d.GetOk("allowed_users"); ok {
-		data["allowed_users"] = v.(string)
+	if v, ok := d.GetOk(consts.FieldExcludeCIDRList); ok {
```
siyer-corp, line 299: Can we use this [helper](https://github.com/hashicorp/terraform-provider-vault/blob/bcb1be4983c816359dbdf9956ccf384fefb62cd0/util/util.go#L100) function here instead for the Set to List conversion?

sanvikam79: Done

vault/resource_ssh_secret_backend_role.go
```diff
@@ -346,29 +387,30 @@ func sshSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
...
-		"allowed_users", "default_user", "key_id_format",
-		"max_ttl", "ttl", "algorithm_signer", "not_before_duration",
+		consts.FieldKeyType, consts.FieldAllowBareDomains, consts.FieldAllowHostCertificates,
+		consts.FieldAllowSubdomains, consts.FieldAllowUserCertificates, consts.FieldAllowUserKeyIDs,
+		consts.FieldAllowedCriticalOptions, consts.FieldAllowedDomains,
+		consts.FieldCIDRList, consts.FieldAllowedExtensions, consts.FieldDefaultExtensions,
+		consts.FieldDefaultExtensionsTemplate, consts.FieldDefaultCriticalOptions,
+		consts.FieldAllowedUsersTemplate, consts.FieldAllowedUsers, consts.FieldDefaultUser,
+		consts.FieldKeyIDFormat, consts.FieldMaxTTL, consts.FieldTTL, consts.FieldAlgorithmSigner,
+		consts.FieldNotBeforeDuration, consts.FieldPort,
```
siyer-corp: If FieldPort is missing from the Vault API for "ca", is it better to check if a value is returned before setting it to the state?

vault/resource_ssh_secret_backend_role.go
```diff
@@ -30,103 +32,130 @@ var (
...
-		"default_extensions": {
+		consts.FieldDefaultExtensions: {
 			Type:     schema.TypeMap,
 			Optional: true,
 		},
-		"default_critical_options": {
+		consts.FieldDefaultExtensionsTemplate: {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
```
siyer-corp, line 105: Can we use Computed=true here? What does vault return when we do not send this field in the config?

vault/resource_ssh_secret_backend_role.go
```diff
@@ -30,103 +32,130 @@ var (
...
+			Optional:    true,
+			Description: "List of CIDR blocks for which credentials cannot be created.",
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		consts.FieldPort: {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     22,
```
siyer-corp: Can we use Computed=true here instead?

vault/resource_ssh_secret_backend_role.go
sanvikam79: No need for computed = true as Vault returns false which matches our schema's Default: false  also we are setting vault response value to tf state

vault/resource_ssh_secret_backend_role.go
sanvikam79: Done . Changed to Computed = true to handle the case of OTP key type with no port being passed but vault returns the default.

vault/resource_ssh_secret_backend_role.go
sanvikam79: Done

vault/resource_ssh_secret_backend_role_test.go
```diff
@@ -331,3 +333,260 @@ resource "vault_ssh_secret_backend_role" "test_role" {
...
+				// Test: Invalid CIDR format
+				Config:      testAccSSHSecretBackendRoleConfig_invalidCIDR(otpRoleName, backend),
+				ExpectError: regexp.MustCompile("invalid CIDR|failed to parse"),
+			},
+			{
+				// Test: Invalid port number (too high) - Remove ExpectError to see what Vault accepts
+				Config: testAccSSHSecretBackendRoleConfig_invalidPortHigh(otpRoleName, backend),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("vault_ssh_secret_backend_role.test_role_otp", consts.FieldPort, "99999"),
+				),
```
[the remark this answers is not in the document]
sanvikam79: Done , added comment for the same

vault/resource_ssh_secret_backend_role_test.go
```diff
@@ -331,3 +333,260 @@ resource "vault_ssh_secret_backend_role" "test_role" {
...
+			},
+			{
+				// Test: Invalid port number (negative) - Remove ExpectError to see what Vault accepts
+				Config: testAccSSHSecretBackendRoleConfig_invalidPortNegative(otpRoleName, backend),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("vault_ssh_secret_backend_role.test_role_otp", consts.FieldPort, "-1"),
+				),
```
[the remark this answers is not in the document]
sanvikam79: Done , added comment for the same

vault/resource_ssh_secret_backend_role_test.go
```diff
@@ -331,3 +333,260 @@ resource "vault_ssh_secret_backend_role" "test_role" {
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDefaultExtensionsTemplate, "false"),
+				),
+			},
+			{
+				Config: testAccSSHSecretBackendRoleConfig_defaultExtensionsTemplateEmpty(roleName, backend),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDefaultExtensionsTemplate, "true"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDefaultExtensions+".%", "0"),
+				),
+			},
```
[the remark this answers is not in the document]
sanvikam79: DOne

vault/resource_ssh_secret_backend_role.go
```diff
@@ -380,6 +445,48 @@ func sshSecretBackendRoleRead(d *schema.ResourceData, meta interface{}) error {
...
+		}
+	} else {
+		// If Vault doesn't return port, set to 0
+		// This ensures proper state management when port is removed from config
+		if err := d.Set(consts.FieldPort, 0); err != nil {
+			return err
+		}
+	}
+
+	// Handle default_extensions - only set if Vault returns it
```
siyer-corp: It'll be good to rephrase this as it's common for default_extensions_template and exlude_cidr_list

sanvikam79: Done

sanvikam79: Done

PR #2753, opened 2026-01-23
Implementation in vault_password_policy to add entropy_source field.

harshit-nema:
### Description

Code added to add entropy_source field in the existing vault_password_policy system backend resource. 

### Checklist

```
$ TF_ACC=1 go test -v ./internal/vault/sys -run TestAccPasswordPolicyEntropy
...
```
Ent1 21

nonEnt

### Community Note

## PCI review checklist

### DIFF
--- a/internal/vault/sys/password_policy.go
+++ b/internal/vault/sys/password_policy.go
@@ -189,9 +189,12 @@ func (r *PasswordPolicyResource) Read(ctx context.Context, req resource.ReadRequ
 
 	data.Policy = types.StringValue(readResp.Policy)
 
-	// Handle entropy_source if supported by API version
+	// Handle entropy_source if supported by API version and present in response
 	if provider.IsAPISupported(r.Meta(), provider.VaultVersion121) {
-		data.EntropySource = types.StringValue(readResp.EntropySource)
+		// Only set entropy_source if it's not empty in the response or if it was previously configured
+		if readResp.EntropySource != "" || !data.EntropySource.IsNull() {
+			data.EntropySource = types.StringValue(readResp.EntropySource)
+		}
 	}
 
 	// write the name to state to support the import command


internal/vault/sys/password_policy.go
```diff
@@ -100,8 +113,31 @@ func (r *PasswordPolicyResource) Create(ctx context.Context, req resource.Create
 	}
 
 	vaultRequest := map[string]interface{}{
-		"policy": data.Policy.ValueString(),
+		consts.FieldPolicy: data.Policy.ValueString(),
+	}
+
+	if !data.EntropySource.IsNull() && !data.EntropySource.IsUnknown() {
+		if provider.IsAPISupported(r.Meta(), provider.VaultVersion121) {
```
siyer-corp, line 118: Can we simplify this nested if-else?

internal/vault/sys/password_policy.go
```diff
@@ -100,8 +113,31 @@ func (r *PasswordPolicyResource) Create(ctx context.Context, req resource.Create
...
+			entropySource := data.EntropySource.ValueString()
+			// Check for enterprise requirement for 'seal' option
+			if entropySource == "seal" && !provider.IsEnterpriseSupported(r.Meta()) {
+				resp.Diagnostics.AddError(
```
siyer-corp: Can we just ignore the field if vault version and ent check fails?

internal/vault/sys/password_policy.go
```diff
@@ -196,8 +232,31 @@ func (r *PasswordPolicyResource) Update(ctx context.Context, req resource.Update
 	}
 
 	vaultRequest := map[string]interface{}{
-		"policy": data.Policy.ValueString(),
+		consts.FieldPolicy: data.Policy.ValueString(),
+	}
+
+	if !data.EntropySource.IsNull() && !data.EntropySource.IsUnknown() {
```
siyer-corp, line 233: Same comments as made on create

internal/vault/sys/password_policy.go
harshit-nema: Taken care. Thanks.

internal/vault/sys/password_policy.go
harshit-nema: Changes accomodated.

internal/vault/sys/password_policy.go
harshit-nema: Taken care.

internal/vault/sys/password_policy.go
```diff
@@ -167,6 +189,14 @@ func (r *PasswordPolicyResource) Read(ctx context.Context, req resource.ReadRequ
 
 	data.Policy = types.StringValue(readResp.Policy)
 
+	// Handle entropy_source if supported by API version and present in response
+	if provider.IsAPISupported(r.Meta(), provider.VaultVersion121) {
+		// Only set entropy_source if it's not empty in the response or if it was previously configured
+		if readResp.EntropySource != "" || !data.EntropySource.IsNull() {
```
siyer-corp, line 195: Do we need an AND condition here instead of OR?

harshit-nema: We are checking on the nonEmpty or nonNull note, so if either of them is true, value should be set, so I feel OR is appropriate here.

harshit-nema: @siyer-corp Here is what I feel : 
OR : Set if Vault has value OR state had value → catches drift both ways and update the terraform state.
AND : Set only if BOTH have value → misses updating terraform state with external changes

so OR is better choice.

PR #2763, opened 2026-02-03
secrets/gcpkms: Add GCP KMS secrets engine and key management resources.

### LINKED ISSUE #2349: [New Resource]: Cloud KMS
### Description

Vault supports managing Cloud KMS. It would be beneficial to be able to manage this through Terraform.

```terraform
resource "vault_gcpkms_secret_backend" "gcpkms" {
  path        = "gcpkms"
  credentials = file("credentials.json")

  default_rotation_period = 72h
}
```

### References

[Cloud KMS](https://developer.hashicorp.com/vault/docs/secrets/gcpkms)

None

Balaji2198:
### Description

Add [GCP KMS secrets engine ](https://developer.hashicorp.com/vault/api-docs/secret/gcpkms)and key management resources.
- Implemented `vault_gcpkms_secret_backend` resource to manage GCP KMS secrets engine.
- Added `vault_gcpkms_secret_backend_key` resource for creating and managing GCP KMS keys.

Closes #2349 

### Checklist

```
$ make testacc TESTARGS='--run=TestGCPKMSSecretBackend'         
TF_ACC=1 go test --run=TestGCPKMSSecretBackend -timeout 30m ./...
?       github.com/hashicorp/terraform-provider-vault/acctestutil       [no test files]
ok      github.com/hashicorp/terraform-provider-vault/codegen   0.562s [no tests to run]
?       github.com/hashicorp/terraform-provider-vault/internal/framework/token  [no test files]
ok      github.com/hashicorp/terraform-provider-vault/internal/framework/validators     2.153s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/identity/entity  1.644s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/provider 1.993s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/ephemeral     2.431s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/spiffe        3.075s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/azure      4.688s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/ephemeral  3.702s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/gcpkms     42.914s
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/sys        5.115s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/testutil  6.218s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util      6.624s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util/mountutil    5.474s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/vault     4.243s [no tests to run]
...
```

### Community Note

## PCI review checklist


internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,363 @@
...
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldPath: schema.StringAttribute{
+				MarkdownDescription: "Path where the GCP KMS secrets engine will be mounted.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldCredentials: schema.StringAttribute{
```
siyer-corp: Do we need this field when there's already a write only field for this?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,363 @@
...
+				MarkdownDescription: "JSON-encoded GCP service account credentials. Cannot be used with `credentials_wo`.",
+				Optional:            true,
+				Sensitive:           true,
+			},
+			consts.FieldCredentialsWO: schema.StringAttribute{
```
siyer-corp, line 78: Do we need a WriteOnly=true attribute here?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,363 @@
...
+				MarkdownDescription: "Write-only JSON-encoded GCP service account credentials. Cannot be used with `credentials`.",
+				Optional:            true,
+				Sensitive:           true,
+			},
+			consts.FieldCredentialsWOVersion: schema.Int64Attribute{
```
siyer-corp, line 86: It'll be good to add a RequiredWith the WO field here?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
siyer-corp: We can also add Required=true once the credentials field is removed, that would remove the explicit checks in Create and Update

internal/vault/secrets/gcpkms/helpers_test.go
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) HashiCorp, Inc.
```
siyer-corp: We can re-use the testutil.go file to add these helpers.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,485 @@
...
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldLabels: schema.MapAttribute{
+				ElementType:         types.StringType,
+				MarkdownDescription: "Labels to apply to the key.",
+				Optional:            true,
+			}, consts.FieldRotationPeriod: schema.StringAttribute{
```
siyer-corp: nit: can move this to the next line

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,485 @@
...
+				},
+			}, consts.FieldKeyRing: schema.StringAttribute{
+				MarkdownDescription: "GCP KMS key ring resource ID (e.g., 'projects/my-project/locations/us-central1/keyRings/my-ring'). Required.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldCryptoKey: schema.StringAttribute{
+				MarkdownDescription: "Name of the crypto key to use in GCP KMS. If the crypto key does not exist, Vault will try to create it. This defaults to the Vault key name if unspecified.",
```
siyer-corp: nit: description can be split into two lines

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,485 @@
...
+
+	cli, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	backend := data.Backend.ValueString()
+	name := data.Name.ValueString()
+	keyPath := fmt.Sprintf("%s/keys/%s", backend, name)
```
siyer-corp: keyPath is re-assigned thrice here. Would be good to move it out to a separate function?

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,485 @@
...
+		// Set default values for version fields
+		data.PrimaryVersion = types.Int64Value(1)
+		data.LatestVersion = types.Int64Value(1)
+		// rotation_period should remain null/unset for asymmetric keys
+		data.RotationPeriod = types.StringNull()
+	} else {
+		// Handle rotation period for symmetric keys
+		if v, ok := secret.Data["rotation_schedule_seconds"].(float64); ok {
+			// Convert seconds back to duration string format
+			data.RotationPeriod = types.StringValue(fmt.Sprintf("%ds", int64(v)))
```
siyer-corp: Is there a reason for converting this into a String? What does Vault return here?

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	backend := data.Backend.ValueString()
+	name := data.Name.ValueString()
+	keyPath := fmt.Sprintf("%s/keys/%s", backend, name)
+
+	log.Printf("[DEBUG] Deleting GCP KMS key at %q", keyPath)
+	if _, err := cli.Logical().DeleteWithContext(ctx, keyPath); err != nil {
+		// Check if this is the known issue with asymmetric keys and rotation
+		// This is a Vault backend bug where it tries to disable rotation on asymmetric keys during deletion
```
siyer-corp, line 410: It says this is a Vault backend bug here. What would be the behavior if this error occurs here?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,351 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package gcpkms
+
+import (
+	"context"
+	"fmt"
+	"log"
```
vijayavelsekar: Instead of using log, can we use Terraform Log `"github.com/hashicorp/terraform-plugin-log/tflog"`

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,532 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package gcpkms
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
```
vijayavelsekar: Instead of using log, can we use Terraform Log `"github.com/hashicorp/terraform-plugin-log/tflog"`

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,532 @@
...
+	LatestVersion           types.Int64  `tfsdk:"latest_version"`
+	PrimaryVersion          types.Int64  `tfsdk:"primary_version"`
+	NextRotationTimeSeconds types.Int64  `tfsdk:"next_rotation_time_seconds"`
+}
+
+func (r *GCPKMSSecretBackendKeyResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_gcpkms_secret_backend_key"
+}
+
+func (r *GCPKMSSecretBackendKeyResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
```
vijayavelsekar, line 70: Missing `min_version/max_version` fields - Defined in consts but not implemented in schema. Also i see its existing in the RFC. Any reason for not implementing it ?

Balaji2198: This resource implements `gcpkms/keys/:key` API. `min_version` and `max_version` belong to a separate Vault API endpoint — `/gcpkms/keys/config/:key`. Updated the RFC to reflect that.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+
+func (r *GCPKMSSecretBackendResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var data GCPKMSSecretBackendModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// This resource owns only the /config endpoint.
+	// The mount lifecycle (create/destroy) is managed by vault_mount.
+	tflog.Debug(ctx, "vault_gcpkms_secret_backend config deleted (mount managed by vault_mount)", map[string]any{"mount": data.Mount.ValueString()})
```
sanvikam79: Do we need to call cli.Logical().DeleteWithContext(ctx, configPath) as Per the Vault GCP KMS API docs has delete api as well

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+				MarkdownDescription: "JSON-encoded GCP service account credentials. Write-only — never " +
+					"stored in Terraform state. Requires Terraform 1.11+.",
+				Required:  true,
+				Sensitive: true,
+				WriteOnly: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			consts.FieldCredentialsWOVersion: schema.Int64Attribute{
```
sanvikam79, line 86: Do we need credentials_wo_version here given that credentials_wo is already required on the resource?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+
+	tflog.Debug(ctx, "Create: credentials_wo from Config", map[string]any{
+		"is_null":    data.CredentialsWO.IsNull(),
+		"is_unknown": data.CredentialsWO.IsUnknown(),
+		"length":     len(data.CredentialsWO.ValueString()),
+	})
+
+	// Validate that credentials are actually present
+	if data.CredentialsWO.IsNull() || data.CredentialsWO.IsUnknown() || data.CredentialsWO.ValueString() == "" {
+		resp.Diagnostics.AddError(
```
sanvikam79: Should "" be allowed for CredentialsWO here? The Vault GCP KMS API docs say: “Leave this blank to use the Default Application Credentials or instance metadata authentication.”

sanvikam79 (commented):
given review comments

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+// This helper reduces code duplication between Create and Update methods.
+func buildBackendConfigFromModel(ctx context.Context, data *GCPKMSSecretBackendModel, includeCredentials bool) (map[string]interface{}, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	configData := make(map[string]interface{})
+
+	// Only include credentials when explicitly requested
+	if includeCredentials {
+		creds := data.CredentialsWO.ValueString()
+		if creds == "" {
+			diags.AddError(
```
sanvikam79: Do we need to allow "" for creds

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+		tflog.Warn(ctx, "GCP KMS backend config not found, removing from state", map[string]any{"path": configPath})
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
+	// Only update scopes if they're in the response
+	if scopes, ok := secret.Data["scopes"].([]interface{}); ok && len(scopes) > 0 {
+		scopeList := make([]string, len(scopes))
+		for i, s := range scopes {
+			scopeList[i] = s.(string)
```
sanvikam79: Should we add a diagnostic error here before asserting each entry to string?
as they do not validate the type of each element. Should we do something like this?
scope, ok := item.(string)
        if !ok {
            resp.Diagnostics.AddError(
                "Unexpected scope entry in Vault response",
                fmt.Sprintf("Expected scope entry to be a string, got %T", item),
            )
            return
        }

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	var plan, state GCPKMSSecretBackendModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Only include credentials if the version changed
+	includeCredentials := !plan.CredentialsWOVersion.Equal(state.CredentialsWOVersion)
+
```
sanvikam79, line 284: Do we need CredentialsWOVersion in addition to required CredentialsWO? If this is needed because CredentialsWO is write-only and not stored in state, could we document that explicitly?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	cli, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	mountPath := data.Mount.ValueString()
+
+	// Configure the backend with credentials
+	configPath := fmt.Sprintf("%s/config", mountPath)
```
sanvikam79, line 154: Can we put this in helper function as its used in many functions

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	configData, diags := buildBackendConfigFromModel(ctx, &data, true)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Debug(ctx, "Configuring GCP KMS backend", map[string]any{"path": configPath})
+	if _, err := cli.Logical().WriteWithContext(ctx, configPath, configData); err != nil {
+		resp.Diagnostics.AddError(
+			"Error configuring GCP KMS backend",
```
sanvikam79: can we use resp.Diagnostics.AddError(errutil.VaultCreateErr(err))

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	mountPath := data.Mount.ValueString()
+	configPath := fmt.Sprintf("%s/config", mountPath)
+
+	tflog.Debug(ctx, "Reading GCP KMS backend config", map[string]any{"path": configPath})
+	secret, err := cli.Logical().ReadWithContext(ctx, configPath)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error reading GCP KMS backend config",
```
sanvikam79: can we use resp.Diagnostics.AddError(errutil.VaultReadErr(err))

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+		return
+	}
+
+	tflog.Debug(ctx, "Updating GCP KMS backend config", map[string]any{
+		"path":                 configPath,
+		"credentials_included": includeCredentials,
+	})
+	if _, err := cli.Logical().WriteWithContext(ctx, configPath, configData); err != nil {
+		resp.Diagnostics.AddError(
+			"Error updating GCP KMS backend config",
```
sanvikam79: resp.Diagnostics.AddError(errutil.VaultUpdateErr(err))

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+	// Build the key configuration from the model
+	keyData, diags := buildKeyConfigFromModel(ctx, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Debug(ctx, "Creating GCP KMS key", map[string]any{"path": keyPath})
+	if _, err := cli.Logical().WriteWithContext(ctx, keyPath, keyData); err != nil {
+		resp.Diagnostics.AddError(
```
sanvikam79: can we use resp.Diagnostics.AddError(errutil.VaultCreateErr(err))

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+		// which GCP KMS doesn't allow. We can safely ignore this error because:
+		// 1. The key is being deleted anyway (not just having rotation disabled)
+		// 2. Asymmetric keys don't support rotation in the first place
+		// 3. The resource will be removed from Terraform state regardless
+		errStr := err.Error()
+		if strings.Contains(errStr, "failed to disable rotation") &&
+			strings.Contains(errStr, "rotation_period must not be set if purpose is ASYMMETRIC_") {
+			tflog.Warn(ctx, "Ignoring rotation disable error for asymmetric key during deletion", map[string]any{"error": err.Error()})
+			// Continue with removal from state despite the Vault error
+			// The key reference is removed from Vault's tracking even if GCP deletion partially failed
```
sanvikam79, line 420: DELETE gcpkms/keys/:key is documented to delete the key from both Vault and Google Cloud KMS. In this  we avoid the delete error and return success, which means Terraform will remove the resource from state even though the GCP-side cleanup may not have completed. Should this return a diagnostic instead of being treated as a successful delete?`

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+
+	backend := data.Mount.ValueString()
+	name := data.Name.ValueString()
+	keyPath := buildKeyPath(backend, name)
+
+	tflog.Debug(ctx, "Reading GCP KMS key", map[string]any{"path": keyPath})
+	secret, err := cli.Logical().ReadWithContext(ctx, keyPath)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error reading GCP KMS key",
```
sanvikam79: can we use resp.Diagnostics.AddError(errutil.VaultReadErr(err))

sanvikam79 (commented):
review comments provided

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+
+		resp.Diagnostics.AddError(
+			"Error deleting GCP KMS key",
+			fmt.Sprintf("Error deleting GCP KMS key at path %q: %s", keyPath, err),
+		)
+		return
+	}
+}
+
+func (r *GCPKMSSecretBackendKeyResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
sanvikam79, line 429: is import meant to be partial for this resource? 
Since ImportState only sets mount/name and Read does not populate key_ring/crypto_key, could imported resources produce incomplete state for later planning?
could this lead to drift as this leave imported state incomplete for later config comparison

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	// Copy the write-only credential value from Config into our data model.
+	// This is the only way to access write-only field values.
+	data.CredentialsWO = configModel.CredentialsWO
+
+	tflog.Debug(ctx, "Create: credentials_wo from Config", map[string]any{
+		"is_null":    data.CredentialsWO.IsNull(),
+		"is_unknown": data.CredentialsWO.IsUnknown(),
+		"length":     len(data.CredentialsWO.ValueString()),
```
vijayavelsekar: Safer to avoid credential related information in logs. Even-though its just a length, better not to give any info related to the structure of the credential also.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	if secret == nil {
+		tflog.Warn(ctx, "GCP KMS backend config not found, removing from state", map[string]any{"path": configPath})
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
+	// Only update scopes if they're in the response
+	if scopes, ok := secret.Data["scopes"].([]interface{}); ok && len(scopes) > 0 {
```
vijayavelsekar: The string 'scopes' is used directly in API response parsing. We can use FieldScopes constant, which already exist in the const file.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+		if creds == "" {
+			diags.AddError(
+				"Missing credentials",
+				"Credentials are required but the value was empty. "+
+					"This may indicate that the write-only field was not properly read from the configuration. "+
+					"Ensure you are providing valid GCP service account credentials via the credentials_wo field.",
+			)
+			return nil, diags
+		}
+		configData["credentials"] = creds
```
vijayavelsekar: The string 'credentials' is hardcoded when building backend config. Should use a constant from const file for consistency. There is an existing constant available.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+			consts.FieldCryptoKey: schema.StringAttribute{
+				MarkdownDescription: "Name of the crypto key to use in GCP KMS. If the crypto key does not exist," +
+					"Vault will try to create it. This defaults to the Vault key name if unspecified.",
+				Optional: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldPurpose: schema.StringAttribute{
+				MarkdownDescription: "Purpose of the key. Valid values: ENCRYPT_DECRYPT, ASYMMETRIC_SIGN, ASYMMETRIC_DECRYPT.",
```
vijayavelsekar: All the below values needs to be in lower case to match with the documentation and examples.
`ENCRYPT_DECRYPT, ASYMMETRIC_SIGN, ASYMMETRIC_DECRYPT`

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+	if v, ok := secret.Data[consts.FieldPurpose].(string); ok {
+		data.Purpose = types.StringValue(v)
+	}
+	if v, ok := secret.Data[consts.FieldAlgorithm].(string); ok {
+		data.Algorithm = types.StringValue(v)
+	}
+	if v, ok := secret.Data[consts.FieldProtectionLevel].(string); ok {
+		data.ProtectionLevel = types.StringValue(v)
+	}
+
```
sanvikam79, line 242: since GET gcpkms/keys/:key` returns `labels`, should Read map them back into state here? Right now labels appear to be write-only from the provider’s perspective

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+				},
+			},
+			consts.FieldCredentialsWO: schema.StringAttribute{
+				MarkdownDescription: "JSON-encoded GCP service account credentials. Write-only — never " +
+					"stored in Terraform state. Requires Terraform 1.11+.",
+				Required:  true,
+				Sensitive: true,
+				WriteOnly: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
vijayavelsekar: We don't need this plan modifier as this is not a computed.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	if _, err := cli.Logical().WriteWithContext(ctx, configPath, configData); err != nil {
+		resp.Diagnostics.AddError(
+			"Error configuring GCP KMS backend",
+			fmt.Sprintf("Error writing %q: %s", configPath, err))
+		return
+	}
+
+	// Read back the state from Vault to ensure all computed values are set
+	// Set the data in state first so Read can use it
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```
vijayavelsekar, line 171: We don't have to set the state here as its already been taken care by the read function in line 170.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
vijayavelsekar: Based on my investigation in the vault-plugin-secrets-gcp repo, this credentials_wo  field needs to be optional. Not required.

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
```diff
@@ -0,0 +1,339 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldCredentialsWOVersion, "2"),
+					// credentials_wo must remain absent from state after credential rotation
+					resource.TestCheckNoResourceAttr(resourceName, consts.FieldCredentialsWO),
+				),
+			},
+		},
+	})
+}
+
+func TestGCPKMSSecretBackend_validation(t *testing.T) {
```
vijayavelsekar: We can remove this test. Does not contribute to the code coverage.

website/docs/r/gcpkms_secret_backend.html.md
```diff
@@ -0,0 +1,136 @@
...
+page_title: "Vault: vault_gcpkms_secret_backend resource"
+sidebar_current: "docs-vault-resource-gcpkms-secret-backend"
+description: |-
+  Manages the GCP KMS secrets engine in Vault
+---
+
+# vault\_gcpkms\_secret\_backend
+
+Configures the GCP KMS secrets engine credentials and scopes in Vault. This resource **only manages
+the `<mount>/config` endpoint** — credentials and OAuth scopes. The mount itself must be created
```
sanvikam79: The docs currently say this resource manages the  endpoint, but Delete() looks like a no-op rather than a call to DELETE <mount>/config. Should we clarify here that destroy only removes the Terraform resource and leaves the Vault config in place, or update the implementation to match the documented lifecycle?

website/docs/r/gcpkms_secret_backend.html.md
```diff
@@ -0,0 +1,136 @@
...
+  Common scopes include:
+  - `https://www.googleapis.com/auth/cloudkms` - Cloud KMS access
+  - `https://www.googleapis.com/auth/cloud-platform` - Full cloud platform access
+
+## Ephemeral Attributes Reference
+
+The following write-only attributes are supported:
+
+* `credentials_wo` - (Required) JSON-encoded GCP service account credentials. This value is write-only
+  and will not be stored in Terraform state. The credentials should have the
```
sanvikam79: Vault’s API docs say leaving blank uses ADC or instance metadata auth, but this provider currently requires credentials_wo. Should the docs mention that this resource does not currently support the blank-credentials ?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	data.CredentialsWO = configModel.CredentialsWO
+
+	tflog.Debug(ctx, "Create: credentials_wo from Config", map[string]any{
+		"is_null":    data.CredentialsWO.IsNull(),
+		"is_unknown": data.CredentialsWO.IsUnknown(),
+		"length":     len(data.CredentialsWO.ValueString()),
+	})
+
+	// Validate that credentials are actually present
+	if data.CredentialsWO.IsNull() || data.CredentialsWO.IsUnknown() || data.CredentialsWO.ValueString() == "" {
```
vijayavelsekar: In addition to below comment, this field is required and not computed, so we will not have null or unknown values. So we can remove this entire condition itself.

website/docs/r/gcpkms_secret_backend.html.md
```diff
@@ -0,0 +1,136 @@
...
+  The `namespace` is always relative to the provider's
+  configured [namespace](/docs/providers/vault/index.html#namespace).
+  *Available only for Vault Enterprise*.
+
+* `mount` - (Required, Forces new resource) Path of the GCP KMS secrets engine mount. Must match the
+  `path` of a [`vault_mount`](mount.html) resource with `type = "gcpkms"`. Use
+  `vault_mount.<name>.path` here.
+
+* `credentials_wo_version` - (Required) Version number for the write-only credentials. Increment this
+  value to trigger a credential rotation. Changing this value will cause the credentials to be re-sent
```
sanvikam79, line 85: check on having this credentials_wo_version if credentials_wo is made optional can have this field else can avoid it.

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
```diff
@@ -0,0 +1,339 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config:      testGCPKMSSecretBackend_noCredentialsConfig(path),
+				ExpectError: regexp.MustCompile(`(Missing required argument|Missing required attribute)`),
+			},
+		},
+	})
+}
+
+func TestGCPKMSSecretBackend_emptyCredentials(t *testing.T) {
```
vijayavelsekar: We can remove this if we are going to allow empty string "" as per Shilpa's comment.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldMount: schema.StringAttribute{
+				MarkdownDescription: "Path where the GCP KMS secrets engine is mounted.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldName: schema.StringAttribute{
```
vijayavelsekar: Can we rename it to closely match the vault api doc ?
Instead of `FieldName` can we convert it to `FieldKey` or `FieldKeyName` ?

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,533 @@
...
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			consts.FieldCryptoKey: schema.StringAttribute{
+				MarkdownDescription: "Name of the crypto key to use in GCP KMS. If the crypto key does not exist," +
+					"Vault will try to create it. This defaults to the Vault key name if unspecified.",
+				Optional: true,
+				PlanModifiers: []planmodifier.String{
```
vijayavelsekar, line 98: Should CryptoKey be computed ? because if we do not provide CryptoKey, vault will assign the key as Crypto key.
```
cryptoKey := d.Get("crypto_key").(string)
	if cryptoKey == "" {
		cryptoKey = key
	}
```

internal/vault/secrets/gcpkms/gcpkms_key_resource_test.go
```diff
@@ -0,0 +1,517 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldLabels+".%", "2"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldLabels+".env", "test"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldLabels+".managed-by", "terraform"),
+				),
+			},
+		},
+	})
+}
+
+func TestGCPKMSSecretBackendKey_validation(t *testing.T) {
```
vijayavelsekar: This kind of validations are framework specific. So we can remove it as it does not contribute to the code coverage as well.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
Siddharth-Kannan: it is still required regardless of credentials_wo being a required field.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
Siddharth-Kannan: Tested this, Without setting state first, `resp.State` is null/empty and read needs `data.Mount` and `data.Namespace` to know where to read from Vault,  which is set by `resp.State.Set`, without which the read would fail

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
Siddharth-Kannan: I think this will be handled by vault, adding another check here might be unnecessary.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
Siddharth-Kannan: it is still required regardless of credentials_wo being a required field.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
Siddharth-Kannan: Import is effectively partial for this resource, yes. ImportState only sets mount and key_name, and I verified via direct Vault CLI that the read endpoint for this resource does not return key_ring or crypto_key in its response model. Because those fields are not exposed by the API, Read cannot populate them after import.

So imported state can remain incomplete for those attributes, and later config comparison may show diffs that are a consequence of partial import rather than true remote drift. That said, this is a limitation of the current Vault API/import shape rather than something we can fully address in this resource without changing the import format or schema strategy.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
Siddharth-Kannan: This seems to be a drawback on the GCP KMS plugin side, where asymmetric keys are not deleted on terraform destroy. Throwing an error in this case will also cause error to be thrown by all the existing asymmetric keys, This should be resolved on the plugin side.

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
Siddharth-Kannan: This still contributes to code coverage, I think it is fine to have it.

sanvikam79 (dismissed):

siyer-corp:
Before merging this, are we considering migrating the mount resource to the new framework and enable mounting via the gcpkms_secret_backend resource for this?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,431 @@
...
+	tflog.Debug(ctx, "Reading GCP KMS backend config", map[string]any{"path": configPath})
+	secret, err := cli.Logical().ReadWithContext(ctx, configPath)
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.VaultReadErr(err))
+		return
+	}
+
+	if secret == nil {
+		tflog.Warn(ctx, "GCP KMS backend config not found, removing from state", map[string]any{"path": configPath})
+		resp.State.RemoveResource(ctx)
```
sanvikam79: Executing `resp.State.RemoveResource` when secret == nil removes from tf.state but the mount might still be alive in Vault. Do we need to keep the state alive when only config is nil and mount is alive?

sanvikam79 (changes requested):

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,431 @@
...
+				Optional:  true,
+				Sensitive: true,
+				WriteOnly: true,
+			},
+			consts.FieldCredentialsWOVersion: schema.Int64Attribute{
+				MarkdownDescription: "Version number for the write-only credentials. Increment this value to trigger a credential rotation. " +
+					"Changing this value will cause the credentials to be re-sent to Vault during the next apply.",
+				Required: true,
+				Validators: []validator.Int64{
+					int64validator.AlsoRequires(path.MatchRoot(consts.FieldCredentialsWO)),
```
sanvikam79: Omitting credentials_wo (as it is Optional) might lead to an AlsoRequires validator exception since credentials_wo_version is Required. Should credentials_wo be made Required to match the enforced behaviour?

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
```diff
@@ -0,0 +1,721 @@
...
+				ImportStateVerifyIgnore: []string{
+					consts.FieldCredentialsWO,
+					consts.FieldCredentialsWOVersion,
+				},
+			},
+		},
+	})
+}
+
+func TestAccGCPKMSSecretBackend_mountConfigImmutable(t *testing.T) {
```
sanvikam79, line 176: Do we need test for destroy/recreate behaviour ?

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
```diff
@@ -0,0 +1,721 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, path),
+					resource.TestCheckResourceAttrSet(resourceName, consts.FieldAccessor),
+				),
+			},
+			{
+				Config: testGCPKMSSecretBackend_initialConfig(updatedPath, credentials),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, updatedPath),
+					resource.TestCheckResourceAttrSet(resourceName, consts.FieldAccessor),
```
sanvikam79, line 294: Should we confirm if the accessor from the new path after remount  is correctly fetched rather than just non-empty?

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
Siddharth-Kannan: Credentials_wo was made to be optional as it can be left empty to use the application default credentials. On looking into it, it does not make sense to add an AlsoRequires validator.

internal/vault/secrets/gcpkms/gcpkms_backend_resource_test.go
Siddharth-Kannan: Good catch
having computed alone without UseStateForUnknown made remount set a new accessor for this test case, which was not verified, Adding UseStateForUnknown fixed this. Made changes in mount.go to include these attributes for the required fields. and added a step to verify that it is correctly fetched.

internal/vault/secrets/gcpkms/gcpkms_backend_resource.go
```diff
@@ -0,0 +1,434 @@
...
+					"or instance metadata authentication. Requires Terraform 1.11+.",
+				Optional:  true,
+				Sensitive: true,
+				WriteOnly: true,
+			},
+			consts.FieldCredentialsWOVersion: schema.Int64Attribute{
+				MarkdownDescription: "Version number for the write-only credentials. Increment this value to trigger a credential rotation. " +
+					"Changing this value will cause the credentials to be re-sent to Vault during the next apply. " +
+					"Set `credentials_wo = \"\"` to rotate to Default Application Credentials.",
+				Required: true,
```
sanvikam79, line 90: Removing AlsoRequires is correct for the credentials_wo = "" ADC case, but when credentials_wo is omitted entirely, ValueString() on a null returns ""  so Vault still receives credentials = "" and silently uses ADC.
Also on Update, if the user bumps credentials_wo_version but omits credentials_wo, the code sends credentials = "" to Vault, overwriting previously set service account credentials with ADC silently.

Siddharth-Kannan: Fixed by adding a !credentials_wo.IsNull() guard in buildBackendConfigFromModel — credentials are now only included in the Vault write payload when credentials_wo is explicitly set in config. This matches the SDKv2 vault_gcp_secret_backend behaviour, which also only sends credentials when non-empty (if credentials != "").

With the fix, credentials are only sent to Vault when credentials_wo is explicitly set (non-null). A null value means the user omitted the field entirely and Vault's stored credentials should remain unchanged. Explicitly setting credentials_wo = "" still correctly signals ADC intent.

internal/vault/secrets/gcpkms/gcpkms_key_resource.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+
+	backend := data.Mount.ValueString()
+	name := data.KeyName.ValueString()
+	keyPath := buildKeyPath(backend, name)
+
+	tflog.Debug(ctx, "Deleting GCP KMS key", map[string]any{"path": keyPath})
+	if _, err := cli.Logical().DeleteWithContext(ctx, keyPath); err != nil {
+		// Check if this is the known issue with asymmetric keys and rotation
+		// This is a Vault backend bug where it tries to disable rotation on asymmetric keys during deletion
```
siyer-corp, line 410: Did we open a bug for this issue on the plugin side? Can we link that JIRA here for future reference?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ FEATURES:
```
siyer-corp, line 53: We're missing a changelog entry for the new GCPKMS resources added. Can we add that?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ FEATURES:
 
 IMPROVEMENTS: 
 
+* Added Plugin Framework `mount_helpers` for Vault mount operations to support incremental migration from SDKv2. These internal helpers provide CRUD operations for self managing mounts for resources using Plugin Framework types. ([#2917](https://github.com/hashicorp/terraform-provider-vault/pull/2917))
```
siyer-corp: Let's reference this PR as part of the GCPKMS changelog entry, as it's not necessarily a customer facing change.

siyer-corp (commented):
Looking great, almost there. Left some nit comments.

sanvikam79 (dismissed):
LGTM

siyer-corp (approved):

PR #2769, opened 2026-02-09
Add support for GCP Cloud KMS managed keys

siyer-corp:
### Description

Add support to `vault_managed_keys` resource for

- GCP Cloud KMS managed keys with parameters: `credentials, project, region, key_ring, crypto_key, crypto_key_version` and `algorithm`.

### Checklist

image

### Community Note

## PCI review checklist

### DIFF
--- a/vault/resource_managed_keys.go
+++ b/vault/resource_managed_keys.go
@@ -428,10 +428,10 @@ func managedKeysGCPConfigSchema() schemaMap {
 			Type:     schema.TypeString,
 			Required: true,
 			Description: "The signature algorithm to be used with the key. " +
-				"Supported values: EC_SIGN_P256_SHA256, EC_SIGN_P384_SHA384, " +
-				"RSA_SIGN_PSS_2048_SHA256, RSA_SIGN_PSS_3072_SHA256, RSA_SIGN_PSS_4096_SHA256, " +
-				"RSA_SIGN_PSS_4096_SHA512, RSA_SIGN_PKCS1_2048_SHA256, RSA_SIGN_PKCS1_3072_SHA256, " +
-				"RSA_SIGN_PKCS1_4096_SHA256, RSA_SIGN_PKCS1_4096_SHA512",
+				"Supported values: ec_sign_p256_sha256, ec_sign_p384_sha384, " +
+				"rsa_sign_pss_2048_sha256, rsa_sign_pss_3072_sha256, rsa_sign_pss_4096_sha256, " +
+				"rsa_sign_pss_4096_sha512, rsa_sign_pkcs1_2048_sha256, rsa_sign_pkcs1_3072_sha256, " +
+				"rsa_sign_pkcs1_4096_sha256, rsa_sign_pkcs1_4096_sha512",
 		},
 	}
 
--- a/website/docs/r/managed_keys.html.md
+++ b/website/docs/r/managed_keys.html.md
@@ -214,16 +214,16 @@ The following arguments are supported:
 * `crypto_key_version` - (Optional) The version of the key to use. (Default: 1)
 
 * `algorithm` - (Required) The signature algorithm to be used with the key. Supported values are:
-  - `EC_SIGN_P256_SHA256`
-  - `EC_SIGN_P384_SHA384`
-  - `RSA_SIGN_PSS_2048_SHA256`
-  - `RSA_SIGN_PSS_3072_SHA256`
-  - `RSA_SIGN_PSS_4096_SHA256`
-  - `RSA_SIGN_PSS_4096_SHA512`
-  - `RSA_SIGN_PKCS1_2048_SHA256`
-  - `RSA_SIGN_PKCS1_3072_SHA256`
-  - `RSA_SIGN_PKCS1_4096_SHA256`
-  - `RSA_SIGN_PKCS1_4096_SHA512`
+  - `ec_sign_p256_sha256`
+  - `ec_sign_p384_sha384`
+  - `rsa_sign_pss_2048_sha256`
+  - `rsa_sign_pss_3072_sha256`
+  - `rsa_sign_pss_4096_sha256`
+  - `rsa_sign_pss_4096_sha512`
+  - `rsa_sign_pkcs1_2048_sha256`
+  - `rsa_sign_pkcs1_3072_sha256`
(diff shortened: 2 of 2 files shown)


testutil/testutil.go
```diff
@@ -269,6 +269,16 @@ func GetTestPKCSCreds(t *testing.T) (string, string, string) {
 	return v[0], v[1], v[2]
 }
 
+func GetTestGCPKeyRing(t *testing.T) string {
```
vijayavelsekar, line 272: Can we combine these two functions into one single function like below ?
By this way we are reduce function all overhead by making single call instead of two calls in the acceptance test case.

```
func GetTestGCPKeyRingAndRegion(t *testing.T) (string, string) {
	v := SkipTestEnvUnset(t, "GOOGLE_KEY_RING", "GOOGLE_REGION")
	return v[0], v[1]
}
```

vault/resource_managed_keys_test.go
```diff
@@ -278,3 +280,327 @@ resource "vault_managed_keys" "test" {
...
+//
+// The final variable specifies that this test can only be run locally
+func TestManagedKeysGCP(t *testing.T) {
+	testutil.SkipTestEnvUnset(t, "TF_ACC_LOCAL")
+
+	name := acctest.RandomWithPrefix("gcp-keys")
+	resourceName := "vault_managed_keys.test"
+
+	credentials, project := testutil.GetTestGCPCreds(t)
+	keyRing := testutil.GetTestGCPKeyRing(t)
```
vijayavelsekar, line 299: If the functions are combined, we can get the values like this. So just in one pass, we can retrieve two values. Same goes for other locations in the same file.

`keyRing, region := testutil.GetTestGCPKeyRingAndRegion(t)`

website/docs/r/managed_keys.html.md
```diff
@@ -164,6 +164,33 @@ The following arguments are supported:
   the HSM.
 
 
+### GCP Cloud KMS Parameters
+
+**Note** this provider is available only with Vault Enterprise Plus (HSMs).
+
+* `name` - (Required) A unique lowercase name that serves as identifying the key.
+
+* `credentials` - (Required) The GCP service account credentials in JSON format.
```
vijayavelsekar: For all the fields here, you can add the additional details about the environment variables hear as well, just like the description in the schema. OR on second thought, we can have the detailed environment variable details here in the docs and keep the schema description short. Whichever works for you.

vault/resource_managed_keys.go
```diff
@@ -365,6 +382,62 @@ func managedKeysAzureConfigSchema() schemaMap {
...
+		consts.FieldName: {
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "A unique lowercase name that serves as " +
+				"identifying the key",
+		},
+		consts.FieldCredentials: {
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "The path to the credentials JSON file to use for " +
```
vijayavelsekar: Try testing the alternative way of setting the environment variables and see if we don't have to pass the Credentials field.

vault/resource_managed_keys.go
```diff
@@ -365,6 +382,62 @@ func managedKeysAzureConfigSchema() schemaMap {
...
+				"authenticating to GCP. Alternatively set via the GOOGLE_CREDENTIALS " +
```
vijayavelsekar: I've also tried manual testing by ignoring one field at a time and got below results for them. Just see if we need to re-write the logic to pull data from these environment variables if not sent via terraform fields. If we are going to accept the values from Environment variables as well, those fields needs to be options.
```
credential
│       * either credentials parameter or either GOOGLE_CREDENTIALS or GOOGLE_APPLICATION_CREDENTIALS env var is required


project
│       * either project parameter or GOOGLE_PROJECT is required


key_ring
│       * key_ring is required


region
│       * either region parameter or GOOGLE_REGION env var is required


crypto_key
│       * crypto_key is required


algorithm
│       * algorithm is required
```

testutil/testutil.go
siyer-corp: Kept this separate in case both are required to be set separately.

website/docs/r/managed_keys.html.md
siyer-corp: Removed the env variables from the descriptions as they are only valid for Vault APIs and not for terraform.

vault/resource_managed_keys.go
```diff
@@ -365,6 +382,62 @@ func managedKeysAzureConfigSchema() schemaMap {
...
 
+func managedKeysGCPConfigSchema() schemaMap {
+	s := schemaMap{
+		consts.FieldName: {
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "A unique lowercase name that serves as " +
+				"identifying the key",
+		},
+		consts.FieldCredentials: {
```
vijayavelsekar, line 393: Would it be more appropriate to mark this field sensitive or convert it to a write-only field ?

siyer-corp: This is not possible due to a limitation from the Terraform Plugin SDK v2. Sensitive: true on any attribute inside a TypeSet/TypeList block redacts the entire block's output, not just that one attribute. This can be resolved if we migrate the resource to the new framework.

vault/resource_managed_keys.go
```diff
@@ -365,6 +382,62 @@ func managedKeysAzureConfigSchema() schemaMap {
...
+		consts.FieldCryptoKey: {
+			Type:     schema.TypeString,
+			Required: true,
+			Description: "The name of the GCP Cloud KMS key. If no existing key " +
+				"exists and allow_generate_key is true, Vault will generate a key with this name",
+		},
+		consts.FieldCryptoKeyVersion: {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
```
vijayavelsekar, line 419: Suppose if I have version "2" in the config initially and applied it. Then later I wanted to remove this attribute expecting it to default to "1" as per the description, this will not show the diff right ? Should we have Default "1" attribute added here ?

siyer-corp: I don't believe that is applicable here. The crypto_key_version will be managed by GCP, not vault. In the initial config, it's optional so that the user can map the version present in GCP for the particular key. If it's a new key, vault will map it to 1.

PR #2778, opened 2026-02-13
secrets/gcpkms: Add GCP KMS secrets engine ephemeral resources.

Balaji2198:
### Description

Add [GCP KMS secrets engine ](https://developer.hashicorp.com/vault/api-docs/secret/gcpkms) ephemeral resources.
1. **`vault_gcpkms_encrypt`** - Encrypts plaintext data using a GCP KMS symmetric encryption key via Vault's GCP KMS secrets engine.
2. **`vault_gcpkms_decrypt`** - Decrypts ciphertext that was encrypted using a GCP KMS symmetric encryption key via Vault's GCP KMS secrets engine.
3. **`vault_gcpkms_reencrypt`** - Re-encrypts existing ciphertext with the latest version of a GCP KMS symmetric encryption key via Vault's GCP KMS secrets engine.
4. **`vault_gcpkms_sign`** - Signs a digest using a GCP KMS asymmetric signing key via Vault's GCP KMS secrets engine.

### Checklist

```
$ make testacc TESTARGS='-run=TestAccGCPKMS'
TF_ACC=1 go test -run=TestAccGCPKMS -timeout 30m ./...
?       github.com/hashicorp/terraform-provider-vault/acctestutil       [no test files]
ok      github.com/hashicorp/terraform-provider-vault/codegen   1.205s [no tests to run]
?       github.com/hashicorp/terraform-provider-vault/internal/framework/token  [no test files]
ok      github.com/hashicorp/terraform-provider-vault/internal/framework/validators     1.731s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/identity/entity  2.343s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/provider 2.290s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/ephemeral     2.756s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/spiffe        3.397s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/azure      3.057s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/ephemeral  147.287s
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/gcpkms     5.694s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/sys        6.371s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/testutil  5.042s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util      3.672s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/util/mountutil    6.627s [no tests to run]
ok      github.com/hashicorp/terraform-provider-vault/vault     4.171s [no tests to run]

...
```

### Community Note

## PCI review checklist


internal/vault/secrets/ephemeral/gcpkms_sign.go
```diff
@@ -0,0 +1,120 @@
...
+	c, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	path := fmt.Sprintf("%s/sign/%s", data.Mount.ValueString(), data.Name.ValueString())
+
+	requestData := map[string]interface{}{
+		"digest": data.Digest.ValueString(),
```
vijayavelsekar: I think we can replace all hardcoded string literals to constants whereever necessary.

website/docs/ephemeral-resources/gcpkms_secret_decrypt.html.md
```diff
@@ -0,0 +1,130 @@
...
+}
+
+ephemeral "vault_gcpkms_decrypt" "data" {
+  mount      = vault_mount.gcpkms.path
+  name       = vault_gcpkms_secret_backend_key.encryption_key.name
+  ciphertext = var.encrypted_data
+}
+
+output "decrypted_secret" {
+  value     = ephemeral.vault_gcpkms_decrypt.data.plaintext
```
vijayavelsekar: I don't think we can use ephemeral resource in output. It will throw error. Probably we can provide a different example. Something like this https://github.com/hashicorp/terraform-provider-vault/pull/2778/changes#diff-016e81a534eae7974a26f2180f144377089a0911c82d87ef3967316c726258a0R50

website/docs/ephemeral-resources/gcpkms_secret_encrypt.html.md
```diff
@@ -0,0 +1,132 @@
...
+  plaintext = base64encode("my secret message")
+}
+
+ephemeral "vault_gcpkms_decrypt" "recovered" {
+  mount      = vault_gcpkms.path
+  name       = vault_gcpkms_secret_backend_key.encryption_key.name
+  ciphertext = ephemeral.vault_gcpkms_encrypt.secret.ciphertext
+}
+
+output "decrypted_value" {
```
vijayavelsekar: Again same as my above comment. Output will not work for ephemeral resource. Try a different example

website/docs/ephemeral-resources/gcpkms_secret_sign.html.md
```diff
@@ -0,0 +1,187 @@
...
+ephemeral "vault_gcpkms_sign" "signature" {
+  mount_id    = vault_mount.gcpkms.id
+  mount       = vault_mount.gcpkms.path
+  name        = vault_gcpkms_secret_backend_key.signing_key.name
+  digest      = local.digest
+  key_version = 1
+}
+
+# Ephemeral values cannot be used in output blocks — use a check block instead.
+check "signature_produced" {
```
vijayavelsekar: I'm not sure how frequently this check block is used by customers in production. But its definitely useful for us the developers to test ephemeral resources. Do you think a different example like the one below would be more appropriate for the customers use-case ?

https://github.com/hashicorp/terraform-provider-vault/pull/2778/changes#diff-016e81a534eae7974a26f2180f144377089a0911c82d87ef3967316c726258a0R50

website/docs/ephemeral-resources/gcpkms_secret_sign.html.md
```diff
@@ -0,0 +1,187 @@
...
+
+ephemeral "vault_gcpkms_sign" "jwt_signature" {
+  mount_id    = vault_mount.gcpkms.id
+  mount       = vault_mount.gcpkms.path
+  name        = vault_gcpkms_secret_backend_key.jwt_key.name
+  digest      = local.jwt_digest
+  key_version = 1
+}
+
+check "jwt_signature_produced" {
```
vijayavelsekar: Same comment as my above one https://github.com/hashicorp/terraform-provider-vault/pull/2778/changes#r2915896551

internal/vault/secrets/ephemeral/helpers_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) HashiCorp, Inc.
```
vijayavelsekar: Can we have all these helpers in the existing  `testutil.go` file ?

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource_test.go
```diff
@@ -0,0 +1,356 @@
...
+  algorithm        = "symmetric_encryption"
+  protection_level = "software"
+%s
+}
+
+ephemeral "vault_gcpkms_encrypt" "test" {
+  mount_id  = vault_mount.test.id
+  mount     = vault_mount.test.path
+  name      = vault_gcpkms_secret_backend_key.test.name
+  plaintext = base64encode("test plaintext data")
```
vijayavelsekar, line 177: If this resource does not work without the resource config, its better to add depends_on to create an explicit edge. I will explain this in todays daily call.

Balaji2198: The `name` attribute in the `vault_gcpkms_encrypt` resource already creates an indirect dependency on the `vault_gcpkms_secret_backend_key` resource. So the Terraform ensures that the backend key is created before the encrypt resource is applied.

internal/vault/secrets/gcpkms/gcpkms_sign_ephemeral_resource.go
```diff
@@ -0,0 +1,120 @@
...
+		return
+	}
+
+	path := fmt.Sprintf("%s/sign/%s", data.Mount.ValueString(), data.Name.ValueString())
+
+	requestData := map[string]interface{}{
+		consts.FieldDigest: data.Digest.ValueString(),
+	}
+
+	if !data.KeyVersion.IsNull() {
```
vijayavelsekar, line 100: The code checks if key_version is null even though it's marked as Required. I think we should be good to remove this check.

internal/vault/secrets/ephemeral/gcpkms_decrypt.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (r *GCPKMSDecryptEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldMount: schema.StringAttribute{
+				MarkdownDescription: "Path where the GCP KMS secrets engine is mounted.",
+				Required:            true,
+			},
+			consts.FieldName: schema.StringAttribute{
```
vijayavelsekar: Should we have this field name as FieldKey or FieldKeyName as it would make more sense to the user and more close to the API docs ? let me know your thoughts.
If you are planning to make this change, we need to do it in all the ephemeral resources.

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource.go
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package ephemeralsecrets
+
+import (
+	"context"
+	"fmt"
+
```
vijayavelsekar, line 9: We can add some `tflog` wherever necessary in all the resources ?

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource_test.go
```diff
@@ -0,0 +1,356 @@
...
+//
+// To run these tests, set the following environment variables:
+// - GOOGLE_CREDENTIALS: GCP service account JSON credentials with KMS permissions
+// - GOOGLE_KMS_KEY_RING: Full GCP KMS key ring path (e.g., "projects/my-project/locations/us-central1/keyRings/my-keyring")
+//
+// Without these environment variables, the tests will be skipped.
+//
+// Note: These tests create real GCP KMS keys and perform actual encryption/decryption operations.
+
+func TestAccGCPKMSDecrypt_basic(t *testing.T) {
```
vijayavelsekar, line 37: nit: Instead of having separate test case for without AAD and with AAD and with key_version. Can we combine all these three in to one single test that has both AAD and key_version ? That will cover all the code that we need right ?

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
```diff
@@ -0,0 +1,356 @@
...
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGCPKMSDecryptConfig(backend, keyName, credentials, keyRing),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.plaintext", tfjsonpath.New("data").AtMapKey("plaintext"), knownvalue.StringRegexp(testutil.RegexpBase64)),
```
sanvikam79: Could we assert that the decrypted value matches the original plaintext passed in the encrypt block instead? As written, these tests would still pass if decrypt returned the wrong value as long as it still looked like base64.

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
```diff
@@ -0,0 +1,356 @@
...
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGCPKMSDecryptWithAADConfig(backend, keyName, aad, credentials, keyRing),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.plaintext", tfjsonpath.New("data").AtMapKey("plaintext"), knownvalue.StringRegexp(testutil.RegexpBase64)),
```
sanvikam79: Could we assert that the decrypted value matches the original plaintext passed in the encrypt block instead? As written, these tests would still pass if decrypt returned the wrong value as long as it still looked like base64.

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
```diff
@@ -0,0 +1,356 @@
...
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGCPKMSDecryptWithKeyVersionConfig(backend, keyName, keyVersion, credentials, keyRing),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.plaintext", tfjsonpath.New("data").AtMapKey("plaintext"), knownvalue.StringRegexp(testutil.RegexpBase64)),
```
sanvikam79: Could we assert that the decrypted value matches the original plaintext passed in the encrypt block instead? As written, these tests would still pass if decrypt returned the wrong value as long as it still looked like base64.

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
```diff
@@ -0,0 +1,356 @@
...
+
+func TestAccGCPKMSDecrypt_namespace(t *testing.T) {
+	credentials, keyRing := testutil.GetTestGCPKMSCreds(t)
+
+	getSteps := func(backend, keyName, ns, credentials, keyRing string) []resource.TestStep {
+		return []resource.TestStep{
+			{
+				Config: testAccGCPKMSDecryptNsConfig(backend, keyName, ns, credentials, keyRing),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.plaintext", tfjsonpath.New("data").AtMapKey("plaintext"), knownvalue.StringRegexp(testutil.RegexpBase64)),
```
sanvikam79: Could we assert that the decrypted value matches the original plaintext passed in the encrypt block instead? As written, these tests would still pass if decrypt returned the wrong value as long as it still looked like base64.

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
```diff
@@ -0,0 +1,356 @@
...
+				},
+			},
+		},
+	})
+}
+
+func TestAccGCPKMSDecrypt_withKeyVersion(t *testing.T) {
+	credentials, keyRing := testutil.GetTestGCPKMSCreds(t)
```
sanvikam79: Should we either pass key_version to vault_gcpkms_decrypt here or rename this test? Right now the KeyVersion case sets key_version only on the encrypt step, so it does not actually exercise the decrypt resource’s key_version handling.

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource_test.go
```diff
@@ -0,0 +1,356 @@
...
+  credentials_wo_version = 1
+%s
+}
+
+resource "vault_gcpkms_secret_backend_key" "test" {
+  mount            = vault_mount.test.path
+  name             = "%s"
+  key_ring         = "%s"
+  purpose          = "encrypt_decrypt"
+  algorithm        = "symmetric_encryption"
```
sanvikam79, line 168: These acceptance tests only cover symmetric encrypt_decrypt keys. should we add an asymmetric-decrypt case as well?

internal/vault/secrets/gcpkms/gcpkms_encrypt_ephemeral_resource_test.go
```diff
@@ -0,0 +1,332 @@
...
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGCPKMSEncryptWithKeyVersionConfig(backend, keyName, plaintext, keyVersion, credentials, keyRing),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.ciphertext", tfjsonpath.New("data").AtMapKey("ciphertext"), knownvalue.StringRegexp(testutil.RegexpBase64)),
```
sanvikam79, line 57: Could we assert key_version here as well, instead of only checking that ciphertext is returned? as per the vault api doc it returns key_version as well

sanvikam79 (commented):
Provided review comments

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource_test.go
itsjiyapatel: Combined the test cases

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource.go
itsjiyapatel: Added necessary logs

internal/vault/secrets/ephemeral/gcpkms_decrypt_test.go
itsjiyapatel: key_version should be passed only for asymmetric keys as for symmetric keys, Cloud KMS chooses the correct version automatically.

internal/vault/secrets/gcpkms/gcpkms_decrypt_ephemeral_resource_test.go
itsjiyapatel: `asymmetric_decrypt` requires a public key to encrypt the data externally which is then decrypted using Vault's `/decrypt` endpoint. Since each test run creates a new GCP KMS key with a unique public-private key pair, this cannot be tested in acceptance tests. Manual testing verifies that the resource supports asymmetric_decrypt functionality correctly.

internal/vault/secrets/gcpkms/gcpkms_encrypt_ephemeral_resource_test.go
itsjiyapatel: added the assertion.

vijayavelsekar (approved):
LGTM

sanvikam79 (approved):

PR #2540, opened 2025-07-31
Fix VAULT_NAMESPACE env var not being set to provider

helenfufu:
### Description

This PR sets the provider namespace at the end of `setClient()` to the result of `GetResourceDataStr(d, consts.FieldNamespace, "VAULT_NAMESPACE", "")` introduced with #2382 so that child namespace paths are honored. It also adds a few type assertion checks to avoid panics in the recently added `GetResourceData*` functions.

This fixes a particular regression case described in https://hashicorp.atlassian.net/browse/VAULT-38208 where:
* In 4.8.0, the `VAULT_NAMESPACE` (`admin/gcp` in our example) gets prepended [here](https://github.com/hashicorp/terraform-provider-vault/blob/e072990b9201524031c802f3a2b1378ea0789e15/internal/provider/meta.go#L106) in `GetNSClient()`, so our namespace gets set to `admin/gcp/test` as desired.
* In 5.0.0 after the multiplexing, `VAULT_NAMESPACE` is not being prepended on that line, so our namespace gets set to `test` and we end up trying to create the policy in a (non-existent) namespace under the root parent namespace. Hence the 403 because our token only has access to the admin namespace.

Note on testing: 
- I have added regression unit test cases to `TestGetClient` in `meta_test.go`. 
- I created a follow-up ticket to write a regression acceptance test for the specific customer case above: https://hashicorp.atlassian.net/browse/VAULT-38637. I had a WIP (detailed in the ticket description) but it's not quite working. Instead, have tested locally that the above scenario is fixed.

#### Local testing

<details><summary>Reproduced success on TFVP 4.8.0</summary>

```zsh
[vault-38208-vault-policy-namespace-bugfix][~/workspace/terraform-provider-vault/scratch/vault-38208-vault-policy-namespace-local]$ terraform init
Initializing the backend...
Initializing provider plugins...
- Finding hashicorp/vault versions matching "4.8.0"...
- Installing hashicorp/vault v4.8.0...
- Installed hashicorp/vault v4.8.0 (signed by HashiCorp)
Terraform has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that Terraform can guarantee to make the same selections by default when
you run "terraform init" in the future.

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
[...]

PR #2799, opened 2026-02-27
fix: respect set_namespace_from_token=false in setClient

bpaquet:
PR #2540 added an unconditional `d.Set(FieldNamespace)` call in the final block of `setClient()`. This bypasses the `set_namespace_from_token` guard, causing the token's namespace to leak into ResourceData even when `set_namespace_from_token=false`. The leaked namespace then gets prepended to the resource namespace, producing doubled paths like `org/staging/org/staging` instead of `org/staging`.

Fix this by tracking whether the namespace was derived from the token and skipping the redundant d.Set in the final block for that case. The first block already handles `d.Set` with the proper `set_namespace_from_token` guard. client.SetNamespace is still called unconditionally so API calls target the correct namespace regardless.

Add tests covering setClient namespace handling for both `set_namespace_from_token=true` and `set_namespace_from_token=false`.

### DIFF
--- a/internal/provider/meta.go
+++ b/internal/provider/meta.go
@@ -349,6 +349,9 @@ func (p *ProviderMeta) setClient() error {
 		client.SetToken(token)
 	}
 
+	// namespaceFromToken tracks whether the namespace was derived from the
+	// auth token rather than being explicitly configured by the user.
+	var namespaceFromToken bool
 	if namespace == "" && tokenNamespace != "" {
 		// set the provider namespace to the token's namespace
 		// this is here to ensure that do not break any configurations that are relying on the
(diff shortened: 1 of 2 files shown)


iniinikoski:
Thanks for the fix @bpaquet ! Bumped into this as well - nasty... They should not break things like this, honestly...

jgrumboe:
We're also blocked upgrading from 5.1 to 5.4+ provider version.
Is there anay ETA when this can be merged?

jgrumboe:
FYI @bpaquet 
@iniinikoski opened an IBM support ticket to help fixing this problem. His ticket details plus summary is here https://github.com/iniinikoski/vault-namespace-bug-report/blob/master/PROVIDER_BUG_REPORT.md
He also found out that this PR is unfortunately incomplete, https://github.com/iniinikoski/vault-namespace-bug-report/blob/master/PROVIDER_BUG_REPORT.md#why-pr-2799-is-insufficient

Also pinging @drewmullen (as you're one of the top contributors the last week and seem to working for IBM Hashicorp).
There should also be an IBM internal Jira Ticket VAULT-45613 about this.
Maybe you can help us get this fixed 🙇

drewmullen:
hey all - i dont have any ownership or influence over this codebase. however, i did try to bubble this up. ive seen internal movement on it so I hope it should get attention soon. appreciate your patience

vijayavelsekar:
Below PR has fixed this issue and merged to main.
https://github.com/hashicorp/terraform-provider-vault/pull/2926
This fix will be part TFVP v5.11.0 that is planned for end of July 2026.

vijayavelsekar:
Closing this PR in favour of https://github.com/hashicorp/terraform-provider-vault/pull/2926

PR #2622, opened 2025-10-08
set deny_null_bind default to true

aahel:
### Description
Currently if the user does not set `deny_null_bind` in tf config we send false to the vault api. This creates an insecure config.
In this pr we are setting  `deny_null_bind` to `true` if its not provided in tf config
This prevents users from bypassing authentication when providing an empty password. It also makes it compliant with  vault's default behaviour.

jira:- https://hashicorp.atlassian.net/browse/SECVULN-29861

**Manual testing**

Testing that  `deny_null_bind` gets set to `true` when not provided in tf config

```
terraform {
  required_providers {
    vault = {
      source = "hashicorp/vault"
      version = "~> 4.0"
    }
  }
}

provider "vault" {
  address = var.address
  token   = var.token
}

# Minimal LDAP auth backend to test deny_null_bind default behavior
resource "vault_ldap_auth_backend" "test" {
  path = "ldap-test"
  url  = "ldap://test.example.com:389"
  
  # Minimal required config
  binddn   = "cn=admin,dc=test,dc=com"
  bindpass = "test-password"
  
  # deny_null_bind is NOT specified - should default to true
}

# Output to verify the default behavior
output "deny_null_bind_value" {
  value = vault_ldap_auth_backend.test.deny_null_bind
  description = "Should show 'true' due to secure default"
}

output "ldap_backend_info" {
  value = {
    path           = vault_ldap_auth_backend.test.path
    accessor       = vault_ldap_auth_backend.test.accessor
    deny_null_bind = vault_ldap_auth_backend.test.deny_null_bind
  }
}
```

**Before the change**

Screenshot 2025-10-08 at 4 22 52 PM

**After the Change**

Screenshot 2025-10-08 at 4 16 25 PM

Screenshot 2025-10-08 at 5 20 03 PM

### Checklist
- [ ] Added [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md) entry (only for user-facing changes)
- [ ] Acceptance tests where run against all supported Vault Versions

### Output from acceptance testing:

Screenshot 2025-10-08 at 5 07 10 PM

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

## PCI review checklist

[...]

PR #2819, opened 2026-03-11
Kerberos auth method initial implementation of resources and ephemeral resources

### LINKED ISSUE #1723: Feature request: Allow configuration of kerberos authentication backend
There currently doesn't seem to be a way to configure the Kerberos authentication backend with this provider.

We'd need resources like `vault_kerberos_auth_backend`.

vijayavelsekar:
### Description

Initial implementation for below resources and ephemeral resource.

1. vault_kerberos_auth_backend_config (Resource)
2. vault_kerberos_auth_backend_ldap_config (Resource)
3. vault_kerberos_auth_backend_group (Resource)
4. vault_kerberos_auth_backend_login (Ephemeral Resource)

> **Note:-**
> This PR updates the `TestDataSourceAuthBackends` acceptance test because it runs concurrently with Kerberos test cases. During concurrent execution, the auth backends created by the Kerberos tests conflict with those used by the data source auth backend tests. To avoid this conflict, the `TestDataSourceAuthBackends` test has been updated to validate only the backends that it explicitly creates.

Closes #1723

### Checklist

```
go test -v ./internal/vault/auth/kerberos -run="TestAccKerberosAuthBackendConfig|TestAccKerberosAuthBackendLDAPConfig|TestAccKerberosAuthBackendGroup" -count=1
=== RUN   TestAccKerberosAuthBackendConfig_basic
--- PASS: TestAccKerberosAuthBackendConfig_basic (4.82s)
=== RUN   TestAccKerberosAuthBackendConfig_update
--- PASS: TestAccKerberosAuthBackendConfig_update (8.17s)
=== RUN   TestAccKerberosAuthBackendConfig_updateAndReplacement
--- PASS: TestAccKerberosAuthBackendConfig_updateAndReplacement (10.54s)
=== RUN   TestAccKerberosAuthBackendConfig_validationErrors
--- PASS: TestAccKerberosAuthBackendConfig_validationErrors (3.69s)
=== RUN   TestAccKerberosAuthBackendConfig_importErrors
--- PASS: TestAccKerberosAuthBackendConfig_importErrors (5.14s)
=== RUN   TestAccKerberosAuthBackendConfig_namespace
--- PASS: TestAccKerberosAuthBackendConfig_namespace (7.36s)
=== RUN   TestAccKerberosAuthBackendGroup_basic
--- PASS: TestAccKerberosAuthBackendGroup_basic (8.02s)
=== RUN   TestAccKerberosAuthBackendGroup_defaultCheck
--- PASS: TestAccKerberosAuthBackendGroup_defaultCheck (3.79s)
=== RUN   TestAccKerberosAuthBackendGroup_pathChange
--- PASS: TestAccKerberosAuthBackendGroup_pathChange (8.97s)
=== RUN   TestAccKerberosAuthBackendGroup_nameChange
--- PASS: TestAccKerberosAuthBackendGroup_nameChange (7.49s)
=== RUN   TestAccKerberosAuthBackendGroup_noPolicies
--- PASS: TestAccKerberosAuthBackendGroup_noPolicies (4.02s)
=== RUN   TestAccKerberosAuthBackendGroup_namespace
--- PASS: TestAccKerberosAuthBackendGroup_namespace (7.62s)
=== RUN   TestAccKerberosAuthBackendGroup_invalidNamespace
--- PASS: TestAccKerberosAuthBackendGroup_invalidNamespace (2.34s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_basic
--- PASS: TestAccKerberosAuthBackendLDAPConfig_basic (5.26s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_update
--- PASS: TestAccKerberosAuthBackendLDAPConfig_update (8.21s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_bindPassUpdate
--- PASS: TestAccKerberosAuthBackendLDAPConfig_bindPassUpdate (6.31s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_defaultCheck
--- PASS: TestAccKerberosAuthBackendLDAPConfig_defaultCheck (3.98s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_pathChange
--- PASS: TestAccKerberosAuthBackendLDAPConfig_pathChange (7.84s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_validationErrors
--- PASS: TestAccKerberosAuthBackendLDAPConfig_validationErrors (2.89s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_batchTokenWithNumUses
--- PASS: TestAccKerberosAuthBackendLDAPConfig_batchTokenWithNumUses (2.55s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_configNotFound
--- PASS: TestAccKerberosAuthBackendLDAPConfig_configNotFound (5.61s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_importErrors
--- PASS: TestAccKerberosAuthBackendLDAPConfig_importErrors (4.98s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_allFields
--- PASS: TestAccKerberosAuthBackendLDAPConfig_allFields (3.84s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_aliasMetadata
    kerberos_ldap_config_resource_test.go:440: Vault server version "1.21.0+ent"
--- PASS: TestAccKerberosAuthBackendLDAPConfig_aliasMetadata (7.00s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_enableSAMAccountNameLogin
    kerberos_ldap_config_resource_test.go:478: Vault server version "1.21.0+ent"
--- PASS: TestAccKerberosAuthBackendLDAPConfig_enableSAMAccountNameLogin (6.54s)
=== RUN   TestAccKerberosAuthBackendLDAPConfig_namespace
--- PASS: TestAccKerberosAuthBackendLDAPConfig_namespace (9.03s)
=== RUN   TestAccKerberosAuthBackendLoginEphemeralResource_basic
--- PASS: TestAccKerberosAuthBackendLoginEphemeralResource_basic (55.75s)
=== RUN   TestAccKerberosAuthBackendLoginEphemeralResource_withOptions
--- PASS: TestAccKerberosAuthBackendLoginEphemeralResource_withOptions (55.70s)
=== RUN   TestAccKerberosAuthBackendLoginEphemeralResource_namespace
--- PASS: TestAccKerberosAuthBackendLoginEphemeralResource_namespace (56.85s)
=== RUN   TestAccKerberosAuthBackendLoginEphemeralResource_invalidKeytab
--- PASS: TestAccKerberosAuthBackendLoginEphemeralResource_invalidKeytab (3.01s)
=== RUN   TestAccKerberosAuthBackendLoginEphemeralResource_invalidNamespace
--- PASS: TestAccKerberosAuthBackendLoginEphemeralResource_invalidNamespace (12.94s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/kerberos      340.904s
```

### Community Note

## PCI review checklist


internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/validators"
+)
+
+const (
+	fieldMount              = consts.FieldMount
+	fieldKeytab             = consts.FieldKeytab
+	fieldServiceAccount     = consts.FieldServiceAccount
+	fieldRemoveInstanceName = consts.FieldRemoveInstanceName
+	fieldAddGroupAliases    = consts.FieldAddGroupAliases
+)
```
harshit-nema, line 27: Should the variables be declared in the consts file ?

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,382 @@
...
+func (r *kerberosAuthBackendConfigResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages the Kerberos authentication method configuration in Vault.\n\n" +
+			"**Note:** Vault does not support deleting auth backend configurations via the API. " +
+			"When this resource is destroyed or replaced (e.g., when changing the `mount`), " +
+			"it is only removed from Terraform state. The configuration remains in Vault until " +
+			"the auth mount itself is deleted.",
+		Attributes: map[string]schema.Attribute{
+			fieldMount: schema.StringAttribute{
+				Optional:    true,
```
harshit-nema: Should we use Required=true in place of optional=true and Computed =true. Default and validator shouldn't be used. Vault will provide its own validation.

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,382 @@
...
+				Default:     stringdefault.StaticString("kerberos"),
+				Description: "Path where the Kerberos auth method is mounted.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					validators.PathValidator(),
+				},
+			},
+			fieldKeytab: schema.StringAttribute{
```
harshit-nema: Good to _wo prefix for write-only fields for better readability and uniformity with other resource.

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,382 @@
...
+
+	var apiModel kerberosAuthBackendConfigAPIModel
+	if err := model.ToAPIModel(respData, &apiModel); err != nil {
+		diags.AddError("Unable to translate Vault response data", err.Error())
+		return diags
+	}
+
+	data.ServiceAccount = types.StringValue(apiModel.ServiceAccount)
+
+	if apiModel.RemoveInstanceName {
```
harshit-nema, line 360: What if we avoid **if condition ** here ?? As per current implementation : 
“false from Vault” -> "not returned" -> "will keep old state"

Should we directly put  ?
data.RemoveInstanceName = types.BoolValue(apiModel.RemoveInstanceName)

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,382 @@
...
+		data.RemoveInstanceName = types.BoolValue(apiModel.RemoveInstanceName)
+	}
+	if apiModel.AddGroupAliases {
```
harshit-nema, line 363: What if we avoid **if condition ** here ?? As per current implementation : 
“false from Vault” -> "not returned" -> "will keep old state"

should we directly put : 
data.AddGroupAliases = types.BoolValue(apiModel.AddGroupAliases)

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,523 @@
...
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldMount, path),
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldServiceAccount, serviceAccount),
+					resource.TestCheckNoResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldRemoveInstanceName),
+					resource.TestCheckNoResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldAddGroupAliases),
+				),
+			},
+		},
+	})
+}
+
```
harshit-nema, line 54: Should Update,Basic test, TestAccKerberosAuthBackendConfig_import be as (Different test step in same test case )?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,523 @@
...
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldRemoveInstanceName, "true"),
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldAddGroupAliases, "true"),
+				),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendConfig_keytabUpdate tests updating the keytab
+func TestAccKerberosAuthBackendConfig_keytabUpdate(t *testing.T) {
```
harshit-nema: Shall Update cases like : 
- TestAccKerberosAuthBackendConfig_pathChange
- TestAccKerberosAuthBackendConfig_keytabUpdate
be put as different test steps under single test case ?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,523 @@
...
+				ExpectError:       regexp.MustCompile(`Invalid import ID format`),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendConfig_namespace tests configuration with namespace (Enterprise only)
+// Note: This test is currently skipped as vault_auth_backend resource doesn't properly support
+// namespaces in the Plugin Framework. This test serves as documentation for the namespace field.
+func TestAccKerberosAuthBackendConfig_namespace(t *testing.T) {
```
harshit-nema, line 206: TestAccKerberosAuthBackendConfig_namespace, TestAccKerberosAuthBackendConfig_importWithNamespace 
Shall these namespace test case be put in single test case (Different test steps) ??

harshit-nema (commented):
Review comments for config_resource.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+		Attributes: map[string]schema.Attribute{
+			fieldMount: schema.StringAttribute{
+				Optional:    true,
+				Computed:    true,
+				Default:     stringdefault.StaticString("kerberos"),
+				Description: "Path where the Kerberos auth method is mounted.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
```
harshit-nema: It would be better if we let vault validation being applied rather than declaring validators in schema ?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+func (r *kerberosAuthBackendLDAPConfigResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages LDAP configuration for the Kerberos authentication method in Vault.\n\n" +
+			"**Note:** Vault does not support deleting auth backend LDAP configurations via the API. " +
+			"When this resource is destroyed or replaced (e.g., when changing the `mount`), " +
+			"it is only removed from Terraform state. The configuration remains in Vault until " +
+			"the auth mount itself is deleted.",
+		Attributes: map[string]schema.Attribute{
+			fieldMount: schema.StringAttribute{
+				Optional:    true,
```
harshit-nema: Aren't mount always required ?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					validators.PathValidator(),
+				},
+			},
+			fieldURL: schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				Default:  stringdefault.StaticString("ldap://127.0.0.1"),
```
harshit-nema, line 147: We shouldnt be declaring default. Let vault handle the default, rather than declaring in TFVP package.
Similar suggestion for : fieldUserAttr
fieldUserFilter
fieldGroupFilter
fieldGroupAttr
fieldTLSMinVersion
fieldTLSMaxVersion
fieldDenyNullBind
fieldRequestTimeout
fieldConnectionTimeout
fieldDereferenceAliases

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+				Sensitive:   true,
+				Description: "LDAP password for searching for the user DN (write-only). Must be used together with bindpass_wo_version.",
+				Validators: []validator.String{
+					stringvalidator.AlsoRequires(path.MatchRoot(fieldBindPassWOVersion)),
+				},
+			},
+			fieldBindPassWOVersion: schema.StringAttribute{
+				Optional:    true,
+				Description: "Version identifier for bindpass updates. Change to trigger password update. Must be used together with bindpass_wo.",
+				Validators: []validator.String{
```
harshit-nema: Rather than validator, should we use **RequiredWith: []string{fieldBindPassWO}** ?
Similar suggestion for : 
- fieldClientTLSCertWO
- fieldClientTLSCertWOVersion
- fieldClientTLSKeyWO
- fieldClientTLSKeyWOVersion

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+
+	resp.Diagnostics.Append(r.writeConfig(ctx, &plan, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (r *kerberosAuthBackendLDAPConfigResource) writeConfig(ctx context.Context, plan *kerberosAuthBackendLDAPConfigModel, config *kerberosAuthBackendLDAPConfigModel) diag.Diagnostics {
```
harshit-nema: To optimise code, should writeConfig wrapper be removed, and writeConfigWithState be directly called ?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+		return nil, diags
+	}
+
+	// Convert API model to map for Vault request
+	var data map[string]interface{}
+	if err := mapstructure.Decode(apiModel, &data); err != nil {
+		diags.AddError("Failed to encode LDAP config API model", err.Error())
+		return nil, diags
+	}
+
```
harshit-nema, line 484: Just a thought, will it be possible to avoid decode-then-delete pattern ? 
Alternate approach : 
- Build [data := map[string]interface{}{}]
- Add each supported field conditionally.
- Gate versioned fields before insertion, not after.
- Populate token fields into a temporary map, then merge only supported entries.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,804 @@
...
+	// Decode response into API model
+	var apiModel kerberosAuthBackendLDAPConfigAPIModel
+	if err := model.ToAPIModel(respData, &apiModel); err != nil {
+		diags.AddError("Unable to translate Vault response data", err.Error())
+		return diags
+	}
+
+	tfModel.URL = types.StringValue(apiModel.URL)
+
+	if apiModel.BindDN != "" {
```
harshit-nema, line 582: On the note of edge case when as per following steps if  : 
- Terraform state already had BindDN = "cn=vault,ou=Users,dc=example,dc=com"
- and Vault now returns binddn = ""
- then this block does nothing
- so the old state value survives incorrectly

Alternate approach suggestion as happening in **token.go**  (114 line) file : 
`model.TokenTTL = types.Int64Null()
if apiModel.TokenTTL != 0 {
    model.TokenTTL = types.Int64Value(apiModel.TokenTTL)
}`

What if we try ? 
`tfModel.BindDN = types.StringNull()
if apiModel.BindDN != "" {
    tfModel.BindDN = types.StringValue(apiModel.BindDN)
}`

Similar suggestion for : UserDN, GroupDN, Certificate, UPNDomain, etc.

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
```diff
@@ -0,0 +1,1037 @@
...
+func TestAccKerberosAuthBackendLDAPConfig_aliasMetadata(t *testing.T) {
+	path := acctest.RandomWithPrefix("kerberos")
+	url := testLDAPURL
+	bindDN := testBindDN
+	userDN := testUserDN
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion121)
```
harshit-nema, line 440: Is aliasMetadata working for CE or Ent vault version ? If only ent, then version check is required here.

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
```diff
@@ -0,0 +1,1037 @@
...
+	bindDN := testBindDN
+	userDN := testUserDN
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctestutil.TestAccPreCheck(t) },
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccKerberosAuthBackendLDAPConfigConfig_batchTokenWithNumUses(path, url, bindDN, userDN),
+				ExpectError: regexp.MustCompile(`'token_type' cannot be 'batch' or 'default_batch'`),
```
harshit-nema, line 270: For the scenario "batch tokens cannot have limited use count", the expected error looks like not matching with the test case. Is this the full error message or can we add some extra detail ?

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
```diff
@@ -0,0 +1,1037 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMount, path),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldEnableSamaccountnameLogin, "false"),
+				),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendLDAPConfig_namespace tests configuration with namespace (Enterprise only)
+func TestAccKerberosAuthBackendLDAPConfig_namespace(t *testing.T) {
```
harshit-nema, line 503: Will it be optimised to add namespace and import namespace under same test case? 
Also invalid namespace test case can also be added for better test coverage.

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
```diff
@@ -0,0 +1,1037 @@
...
+					resource.TestCheckResourceAttr(resourceName, consts.FieldBindDN, bindDN),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldBindPassWOVersion, "v2"),
+				),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendLDAPConfig_import tests importing the resource
+func TestAccKerberosAuthBackendLDAPConfig_import(t *testing.T) {
```
harshit-nema: Will it be optimal to add TestAccKerberosAuthBackendLDAPConfig_import and testAccKerberosAuthBackendLDAPConfigConfig_basic under same test case ?

harshit-nema (commented):
Sharing few review comments for kerberos_ldap_config_resource.

internal/vault/auth/kerberos/kerberos_config_resource_test.go
vijayavelsekar: You are right. I think we can remove this separate import test and add it to the basic and update test function as a last step.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: RequiredWith attribute is not available in the new plugin framework. This is the option they provide.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: Both are being used. For Create we use `writeConfig` and for update operation we use `writeConfigWithState`. By this way we can reuse the same code for different purpose.
writeConfigWithState passes state in addition to use it for checking the previous state value of write only version field during update operation.

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
vijayavelsekar: If the token type is batch or default batch, we are not allowed to give token_num_uses. We are testing this scenario here.

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
vijayavelsekar: Usually we don't test CE right ? we are testing against Enterprise only.

internal/vault/auth/kerberos/kerberos_config_resource.go
vijayavelsekar: This condition check is required.
Senario 1:
During create, If config does not have remove_instance_name. There will be drift between the actual plan and the value from vault. Plan has null and vault response have false.

Senario 2:
If remove_instance_name is true during create, we have true in state and true in vault. No problem now. After we change to false. We don't have to update this value in in populateDataModelFromApi, because its already having false in the config. Supose if we remove this field instead of making it a false, plan will have null and this config will also have null and there wont be any issues, if we don't have this condition, we will have false writtent to the config object and there will be drift again.

internal/vault/auth/kerberos/kerberos_config_resource.go
vijayavelsekar: Same comment as above.

https://github.com/hashicorp/terraform-provider-vault/pull/2819/changes#r2930459305

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: All the fields that you have listed are Non Zero/Non Null default values being set in the vault side. So we have to provide explicit default values in the schema. Otherwise, If we remove the field from terraform, state will have null, which will not be practically same as the vault.
If the default value is zero or null or false, what you have suggested would be fine.
_Note: in the new framework, if we have default, we need to have computed (mandatory). Otherwise, while running, we will get error message asking use to added computed._

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: This code executes in different way. `tfModel.BindDN` will already be having empty string value when `apiModel.BindDN` is equal to empty string.
Same explanation holds from the comment https://github.com/hashicorp/terraform-provider-vault/pull/2819/changes#r2930459305

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: This approach will increase the number of lines of code as I have a huge list of fields. 
Using decode and delete will reduce the lines of code and provides more clarity while reading.
I took reference from the cloud foundry on the same. We can discuss with the team also about this and we can decide.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
harshit-nema: How about providing 'omitempty' params in the struct declaration?  It will not be saved in state during the pass, but vault will return non-empty default for it.

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	base.BaseModel
+
+	Mount    types.String `tfsdk:"mount"`
+	Name     types.String `tfsdk:"name"`
+	Policies types.Set    `tfsdk:"policies"`
+}
+
+// kerberosAuthBackendGroupAPIModel describes the Vault API response structure.
+type kerberosAuthBackendGroupAPIModel struct {
+	Policies []string `json:"policies" mapstructure:"policies"`
```
harshit-nema, line 65: As per doc, Policies are Comma-separated list of policies associated to the group, but we are using slice []string here, which is feasible too, but can we give a thought to use String here ??

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	// Add the common base schema
+	base.MustAddBaseSchema(&resp.Schema)
+}
+
+// Create is called during the terraform apply command.
+func (r *kerberosAuthBackendGroupResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var config kerberosAuthBackendGroupModel
+
+	// Read Terraform config data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
```
harshit-nema: Reading data from Plan will be better approach and future proof, if there will be defaults, computed or write-only attributes. 
`resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)`

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	vaultClient, err := client.GetClient(ctx, r.Meta(), config.Namespace.ValueString())
+	if err != nil {
+		diags.AddError(errutil.ClientConfigureErr(err))
+		return false, diags
+	}
+
+	mount := strings.Trim(config.Mount.ValueString(), "/")
+	name := strings.Trim(config.Name.ValueString(), "/")
+	groupPath := r.groupPath(mount, name)
+
```
harshit-nema: L219-L228 looks redudant. The same is present in writeGroup method.

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	vaultClient, err := client.GetClient(ctx, r.Meta(), config.Namespace.ValueString())
+	if err != nil {
+		diags.AddError(errutil.ClientConfigureErr(err))
+		return diags
+	}
+
+	mount := strings.Trim(config.Mount.ValueString(), "/")
+	name := strings.Trim(config.Name.ValueString(), "/")
+	groupPath := r.groupPath(mount, name)
+
```
harshit-nema: This piece of code is also present in readConfig. Should we optimise it ?

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	// Save data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &config)...)
+}
+
+// Update is called during the terraform apply command.
+func (r *kerberosAuthBackendGroupResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var config kerberosAuthBackendGroupModel
+
+	// Read Terraform config data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
```
harshit-nema: Reading data from Plan will be better approach and future proof, if there will be defaults, computed or write-only attributes.
`resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)`

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+
+	// Read group from Vault
+	tflog.Debug(ctx, fmt.Sprintf("Reading Kerberos group from '%s'", groupPath))
+	groupResp, err := vaultClient.Logical().ReadWithContext(ctx, groupPath)
+	if err != nil {
+		diags.AddError(errutil.VaultReadErr(err))
+		return false, diags
+	}
+	if groupResp == nil {
+		tflog.Warn(ctx, fmt.Sprintf("Kerberos group at '%s' not found, removing from state", groupPath))
```
harshit-nema, line 233: Should we add :  `resp.Diagnostics.AddError(errutil.VaultReadResponseNil()) ` along with warning ?

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	return matches[1], nil
+}
+
+// getApiModel builds the Vault API request map from the Terraform data model.
+func (r *kerberosAuthBackendGroupResource) getApiModel(ctx context.Context, data *kerberosAuthBackendGroupModel) (map[string]any, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	var policies []string
+	diags.Append(data.Policies.ElementsAs(ctx, &policies, false)...)
```
harshit-nema: If policies is unknown or cannot be converted, this will silently produce an empty value. Should we append/return the diagnostics and/or only build the field when policies is known ?

Example : 

> if !data.Policies.IsNull() && !data.Policies.IsUnknown() {
> 		var elements []string
> 		elementsDiags := data.Policies.ElementsAs(ctx, &elements, false)
> 		diags.Append(elementsDiags...)
> 		if diags.HasError() {
> 			return nil, diags
> 		}
> 		if len(elements) > 0 {
> 			policiesStr := strings.Join(elements, ",")
> 			vaultRequest[consts.FieldPolicies] = policiesStr
> 		}

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,395 @@
...
+	// Convert policies from API model to Terraform model
+	if len(apiModel.Policies) > 0 {
+		policies, setDiags := types.SetValueFrom(ctx, types.StringType, apiModel.Policies)
+		diags.Append(setDiags...)
+		if diags.HasError() {
+			return diags
+		}
+		config.Policies = policies
+	}
+
```
harshit-nema, line 406: When Vault returns an empty policies list, this function leaves config.Policies unchanged. During Read/Refresh that can preserve stale policies in state even after policies were removed in Vault.
Better to explicitly set data.Policies to types.SetNull(types.StringType) (or an empty set, depending on desired semantics) when apiModel.Policies is empty.

> else {
> 		// When Vault returns no policies, clear the Terraform state to avoid stale values.
>         config.Policies = types.SetNull(types.StringType)
> }

internal/vault/auth/kerberos/kerberos_group_resource_test.go
```diff
@@ -0,0 +1,437 @@
...
+				Config:      testAccKerberosAuthBackendGroupConfig_invalidMount(groupName),
+				ExpectError: regexp.MustCompile("no handler for route|unsupported path|route entry not found"),
+			},
+			// Test missing group name
+			{
+				Config:      testAccKerberosAuthBackendGroupConfig_missingName(backend),
+				ExpectError: regexp.MustCompile(`Missing required argument|The argument "name" is required`),
+			},
+		},
+	})
```
harshit-nema: Should it be good test coverage if missingMount test scenario is also included here ?

harshit-nema (commented):
Review comments for kerberos_group_resource

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package kerberos
+
```
harshit-nema, line 5: How about renaming the resource name to : kerberos_login_ephemeral_resource for better readability?

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+// kerberosPrivateData holds data that needs to be passed from Open to Close
+type kerberosPrivateData struct {
+	Accessor  string `json:"accessor"`
+	Namespace string `json:"namespace"`
+}
+
+type kerberosAuthBackendLoginModel struct {
+	base.BaseModelEphemeral
+
+	// Input parameters
```
harshit-nema, line 48: Kerberos API doc doesnt mention these input params. Are these taken from Kerberos plugin reference ?

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+	resp.TypeName = req.ProviderTypeName + "_kerberos_auth_backend_login"
+}
+
+func (e *kerberosAuthBackendLoginEphemeral) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Performs Kerberos authentication and returns a Vault token. This is an ephemeral resource - credentials and tokens are not persisted to state.",
+		Attributes: map[string]schema.Attribute{
+			consts.FieldMount: schema.StringAttribute{
+				Required:    true,
+				Description: "Path where the Kerberos auth method is mounted. Defaults to 'kerberos'.",
```
harshit-nema, line 81: How about using optional=true, computed=true for FieldMount ??

internal/vault/auth/kerberos/kerberos_login.go
```diff
@@ -0,0 +1,365 @@
...
+func (e *kerberosAuthBackendLoginEphemeral) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemeral.OpenResponse) {
+	var config kerberosAuthBackendLoginModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	c, err := client.GetClient(ctx, e.Meta(), config.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Error getting client", err.Error())
```
harshit-nema: Should we use `resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))` rather than hardcoded error ?

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+		config.Orphan = types.BoolValue(false)
+	}
+
+	// Convert policies
+	if len(secret.Auth.Policies) > 0 {
+		policies, diags := types.SetValueFrom(ctx, types.StringType, secret.Auth.Policies)
+		resp.Diagnostics.Append(diags...)
+		if !resp.Diagnostics.HasError() {
+			config.Policies = policies
+		}
```
harshit-nema, line 252: Should we add code for return early if resp.Diagnostics.HasError() ?
```
if resp.Diagnostics.HasError() {
			return
		}
```

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+		config.Policies = types.SetNull(types.StringType)
+	}
+
+	// Convert token policies
+	if len(secret.Auth.TokenPolicies) > 0 {
+		tokenPolicies, diags := types.SetValueFrom(ctx, types.StringType, secret.Auth.TokenPolicies)
+		resp.Diagnostics.Append(diags...)
+		if !resp.Diagnostics.HasError() {
+			config.TokenPolicies = tokenPolicies
+		}
```
harshit-nema, line 264: Should we add code for return early if resp.Diagnostics.HasError() ?

```
if resp.Diagnostics.HasError() {
			return
		}
```

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+		config.TokenPolicies = types.SetNull(types.StringType)
+	}
+
+	// Convert identity policies
+	if len(secret.Auth.IdentityPolicies) > 0 {
+		identityPolicies, diags := types.SetValueFrom(ctx, types.StringType, secret.Auth.IdentityPolicies)
+		resp.Diagnostics.Append(diags...)
+		if !resp.Diagnostics.HasError() {
+			config.IdentityPolicies = identityPolicies
+		}
```
harshit-nema, line 276: Should we add code for return early if resp.Diagnostics.HasError() ?

```
if resp.Diagnostics.HasError() {
			return
		}
```

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,365 @@
...
+		config.IdentityPolicies = types.SetNull(types.StringType)
+	}
+
+	// Convert metadata
+	if len(secret.Auth.Metadata) > 0 {
+		metadata, diags := types.MapValueFrom(ctx, types.StringType, secret.Auth.Metadata)
+		resp.Diagnostics.Append(diags...)
+		if !resp.Diagnostics.HasError() {
+			config.Metadata = metadata
+		}
```
harshit-nema, line 288: Should we add code for return early if resp.Diagnostics.HasError() ?

```
if resp.Diagnostics.HasError() {
			return
		}
```

internal/vault/auth/kerberos/kerberos_login_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+	config := testutil.SkipTestEnvUnsetKerberos(t)
+
+	mount := acctest.RandomWithPrefix("kerberos-mount")
+	nonEmpty := regexp.MustCompile(`^.+$`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion116)
```
harshit-nema: Are we keeping check : applicable for version >=  vault v1.16  ? And not the lower ones ?

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+					statecheck.ExpectKnownValue("echo.test_kerberos",
+						tfjsonpath.New("data").AtMapKey("entity_id"),
+						knownvalue.StringRegexp(nonEmpty)),
+				},
+			},
+		},
+	})
+}
+
+func testAccKerberosAuthBackendLoginEphemeralResourceConfig_namespace(namespace, mount string, config *testutil.KerberosTestConfig) string {
```
harshit-nema: Will it be wholesome to add test case for invalid namespace as well ?

testutil/testutil.go
```diff
@@ -95,6 +95,68 @@ func FatalTestEnvUnset(t *testing.T, envVars ...string) []string {
 	return handleTestEnvUnsetF(t, t.Fatalf, envVars...)
 }
 
+// SkipTestEnvUnsetKerberos skips the test if any of the required Kerberos environment variables
+// are empty/unset. Returns a KerberosTestConfig struct with all the values.
+func SkipTestEnvUnsetKerberos(t *testing.T) *KerberosTestConfig {
```
harshit-nema: Instead of writing env variables in testUtil, can we directly use testutil.SkpTestEnvUnset() in test file ?

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+)
+
+// TestAccKerberosAuthBackendLoginEphemeralResource_basic tests the Kerberos login ephemeral resource
+// Note: This test requires a properly configured Kerberos environment with:
+// - A valid keytab file
+// - A krb5.conf configuration file
+// - A configured LDAP backend
+// - A running KDC (Key Distribution Center)
+//
+// Required environment variables:
```
harshit-nema, line 92: Instead of writing env variables in testUtil, can we directly use testutil.SkpTestEnvUnset() in test file ? Like  : 
testutil.SkipTestEnvUnset(t,
  "VAULT_TEST_KERBEROS_KEYTAB_PATH",
  "VAULT_TEST_KERBEROS_KRB5CONF_PATH"
.....and other env variables...so on)

harshit-nema (commented):
Ephemeral resource : Kerberos_login review. 
Also will there be any modifications required for vault.erb file for this new implementation on Vault Kerberos auth method

internal/vault/auth/kerberos/kerberos_group_resource.go
vijayavelsekar: If I do so I get below error. Because the vault returns them in the form of array of string.
Note that this struct is to hold the value coming from the [Vault API response](https://developer.hashicorp.com/vault/api-docs/auth/kerberos#read-kerberos-ldap-group)(GET method) not for creation of the group, which is POST method.
```
Error: json: cannot unmarshal array into Go struct field
kerberosAuthBackendGroupAPIModel.policies of type string
```

internal/vault/auth/kerberos/kerberos_group_resource.go
vijayavelsekar: Not required, because in the Null scenario where we do not provide the policies in the config file, ElementsAs() successfully converts Null → empty slice []string{}
And for the Unknown scenario, this function will be called during `Apply phase` only, so there is no way it will be unknown during apply phase. By any changes if its unknown/causes any error, it has also been handled by properly adding the diagnostic errors.

internal/vault/auth/kerberos/kerberos_group_resource.go
vijayavelsekar: This is a nice finding 👍 . It helps in `terraform refresh`.

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
vijayavelsekar: Yes. They are needed to generate the SPNEGO token first from kerberos. And then a Vault API request will be set using the generated token to get the vault token.

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	return matches[1], nil
+}
+
+// getApiModel builds the Vault API request map from the Terraform data model.
+func (r *kerberosAuthBackendConfigResource) getApiModel(config *kerberosAuthBackendConfigModel) (map[string]any, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	vaultRequest := map[string]any{
+		consts.FieldKeytab:             config.Keytab.ValueString(),
```
siyer-corp: Does this need to be config.KeytabWO?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,775 @@
...
+				Sensitive:   true,
+				Description: "LDAP password for searching for the user DN (write-only). Must be used together with bindpass_wo_version.",
+				Validators: []validator.String{
+					stringvalidator.AlsoRequires(path.MatchRoot(consts.FieldBindPassWOVersion)),
+				},
+			},
+			consts.FieldBindPassWOVersion: schema.StringAttribute{
+				Optional:    true,
+				Description: "Version identifier for bindpass updates. Change to trigger password update. Must be used together with bindpass_wo.",
+				Validators: []validator.String{
```
siyer-corp: Should this be int? Applies to all wo_version fields

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,775 @@
...
+			apiModel.BindPass = config.BindPassWO.ValueString()
+			tflog.Debug(ctx, "Bindpass version changed or new resource, updating bindpass")
+		}
+	}
+
+	if !plan.UserDN.IsNull() {
+		apiModel.UserDN = plan.UserDN.ValueString()
+	}
+
+	apiModel.UserAttr = plan.UserAttr.ValueString()
```
siyer-corp: We have potential to clean up the code here, we should group these like it's done [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/vault/auth/cloudfoundry/cf_auth_backend_config_resource.go#L310)

website/docs/ephemeral-resources/kerberos_auth_backend_login.html.md
```diff
@@ -0,0 +1,297 @@
...
+```hcl
+resource "vault_auth_backend" "kerberos" {
+  type = "kerberos"
+  
+  tune {
+    passthrough_request_headers = ["Authorization"]
+  }
+}
+
+ephemeral "vault_kerberos_auth_backend_login" "login" {
```
siyer-corp: Will this work if mount is a Required field in the schema?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,775 @@
...
+				Description: "Version identifier for client TLS key updates. Must be used together with client_tls_key_wo.",
+				Validators: []validator.String{
+					stringvalidator.AlsoRequires(path.MatchRoot(consts.FieldClientTLSKeyWO)),
+				},
+			},
+			consts.FieldDiscoverDN: schema.BoolAttribute{
+				Optional:    true,
+				Description: "Use anonymous bind to discover bind DN of a user. Default: false.",
+			},
+			consts.FieldDenyNullBind: schema.BoolAttribute{
```
siyer-corp, line 285: Can we check if we tested the scenario that caused the [bug](https://github.com/hashicorp/terraform-provider-vault/pull/2622) in ldap_auth?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,775 @@
...
+	resp.Diagnostics.Append(r.writeConfig(ctx, &plan, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (r *kerberosAuthBackendLDAPConfigResource) writeConfig(ctx context.Context, plan *kerberosAuthBackendLDAPConfigModel, config *kerberosAuthBackendLDAPConfigModel) diag.Diagnostics {
+	return r.writeConfigWithState(ctx, plan, config, nil)
```
siyer-corp: Can we move the code from writeConfigWithState into here as this function is just calling it?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,775 @@
...
+	// Populate model from API response
+	populateDiags := r.populateDataModelFromApi(ctx, data, resp.Data)
+	diags.Append(populateDiags...)
+	return true, diags
+}
+
+// populateDataModelFromApi maps the Vault API response to the Terraform data model.
+func (r *kerberosAuthBackendLDAPConfigResource) populateDataModelFromApi(ctx context.Context, tfModel *kerberosAuthBackendLDAPConfigModel, respData map[string]interface{}) diag.Diagnostics {
+	var diags diag.Diagnostics
+
```
siyer-corp, line 555: We have potential to clean up the code here, we should group these like it's done [here](https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/vault/auth/cloudfoundry/cf_auth_backend_config_resource.go#L310)

testutil/testutil.go
siyer-corp: Makes sense, since they are not called from any other resource and only used for kerberos_ephemeral tests, should be reasonable to move this in there?

siyer-corp (commented):
Provided some comments on the first pass. For the ldap_config_resource, can we check if the functionality is similar to ldap_auth resource?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: yes. We don't have that bug as we are setting the default as true.

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
vijayavelsekar: Done (ref: line number 558 to 568)

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
vijayavelsekar: @harshit-nema & @siyer-corp Need your help
As we are keeping other resources mount as required, I am following same pattern here. Any reason, why only ephemeral mount has to be optional and computed ?
If this has to be optional + computed, then the remaining 3 resources should follow same pattern ?

Refer Resource usage in kerberos_auth_backend_login.html.md

siyer-corp: It makes sense to make it optional + computed if we want to support default mount fallback, but having it required should not be a problem.

website/docs/ephemeral-resources/kerberos_auth_backend_login.html.md
```diff
@@ -0,0 +1,271 @@
...
+  keytab          = filebase64("/path/to/vault.keytab")
+  service_account = "vault/localhost@EXAMPLE.COM"
+}
+
+resource "vault_kerberos_auth_backend_ldap_config" "ldap" {
+  mount               = vault_auth_backend.kerberos.path
+  url                 = "ldap://localhost:389"
+  binddn              = "cn=admin,dc=example,dc=com"
+  bindpass_wo         = "admin-password"
+  bindpass_wo_version = "1"
```
siyer-corp: This would be an int. Can remove the quotes.

website/docs/ephemeral-resources/kerberos_auth_backend_login.html.md
```diff
@@ -0,0 +1,271 @@
...
+  service_account = "vault/localhost@EXAMPLE.COM"
+}
+
+resource "vault_kerberos_auth_backend_ldap_config" "ldap" {
+  namespace           = vault_namespace.app.path
+  mount               = vault_auth_backend.kerberos.path
+  url                 = "ldap://localhost:389"
+  binddn              = "cn=admin,dc=example,dc=com"
+  bindpass_wo         = "admin-password"
+  bindpass_wo_version = "1"
```
siyer-corp: This would be an int. Can remove the quotes.

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
siyer-corp: That given, yes, all resources must follow either Required or Optional+Computed(which falls back to default)

siyer-corp (commented):
As discussed, let's check if changes in `vault/data_source_auth_backends_test.go` are required. Provided some additional comments.

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
vijayavelsekar: Okay. Then I'll keep them required. Thanks

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,732 @@
...
+				Default:     stringdefault.StaticString("never"),
+				Description: "When aliases should be dereferenced on search operations. Accepted values are 'never', 'finding', 'searching', 'always'. Default: 'never'",
+			},
+			consts.FieldMaxPageSize: schema.Int64Attribute{
+				Optional:    true,
+				Description: "If set to a value greater than 0, the LDAP backend will use the LDAP server's paged search control. Default: 0.",
+			},
+			consts.FieldEnableSamaccountnameLogin: schema.BoolAttribute{
+				Optional:    true,
+				Description: "If true, matching sAMAccountName attribute values will be allowed to login when upndomain is defined. Default: false. **Note:** This field is only supported in Vault 1.19.0 and above. Do not configure this attribute if your Vault version is below 1.19.0.",
```
siyer-corp: Can we make this consise? We can update the note to say _"Requires Vault 1.19.0+"_

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,732 @@
...
+				Description: "Version identifier for client TLS certificate updates. Change to trigger certificate update. Must be used together with client_tls_cert_wo.",
+				Validators: []validator.Int64{
+					int64validator.AlsoRequires(path.MatchRoot(consts.FieldClientTLSCertWO)),
+				},
+			},
+			consts.FieldClientTLSKeyWO: schema.StringAttribute{
+				Optional:    true,
+				WriteOnly:   true,
+				Sensitive:   true,
+				Description: "Client certificate key to provide to the LDAP server, must be x509 PEM encoded (write-only). Must be used together with client_tls_key_wo_version.",
```
siyer-corp: Can we split these into multi-line for better readability?

internal/vault/auth/ephemeral/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,367 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package ephemeralauth
```
siyer-corp: Can we move these two files inside the kerberos directory?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,477 @@
...
+			{
+				Config:      testAccKerberosAuthBackendConfigConfig_emptyKeytab(path, serviceAccount),
+				ExpectError: regexp.MustCompile(`data does not contain keytab|string must not be empty`),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendConfig_runtimeErrors tests runtime errors
+func TestAccKerberosAuthBackendConfig_runtimeErrors(t *testing.T) {
```
siyer-corp: Do we need this as it's testing vault functionality, not TFVP?

internal/vault/auth/kerberos/kerberos_ldap_config_resource_test.go
```diff
@@ -0,0 +1,979 @@
...
+			{
+				Config:      testAccKerberosAuthBackendLDAPConfigConfig_clientKeyWithoutVersion(path, url, bindDN, userDN),
+				ExpectError: regexp.MustCompile(`Attribute "client_tls_key_wo_version" must be specified when`),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendLDAPConfig_runtimeErrors tests runtime errors
+func TestAccKerberosAuthBackendLDAPConfig_runtimeErrors(t *testing.T) {
```
siyer-corp: Do we need this as it's testing vault functionality, not TFVP?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,477 @@
...
+			{
+				Config:      testAccKerberosAuthBackendConfigConfig_nonExistentBackend(nonExistentPath, serviceAccount),
+				ExpectError: regexp.MustCompile(`error writing|no handler for route|unsupported path`),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendConfig_importErrors tests import validation errors
+func TestAccKerberosAuthBackendConfig_importErrors(t *testing.T) {
```
siyer-corp, line 160: Do we control the import ID? When will these scenarios occur?

internal/vault/auth/kerberos/kerberos_group_resource_test.go
```diff
@@ -0,0 +1,451 @@
...
+				PreConfig: func() {
+					os.Unsetenv(consts.EnvVarVaultNamespaceImport)
+				},
+				PlanOnly: true,
+			},
+		},
+	})
+}
+
+func TestAccKerberosAuthBackendGroup_invalid(t *testing.T) {
```
siyer-corp: Do we need this as it's testing vault functionality, not TFVP?

internal/vault/auth/kerberos/kerberos_group_resource_test.go
```diff
@@ -0,0 +1,451 @@
...
+			// Test missing group name
+			{
+				Config:      testAccKerberosAuthBackendGroupConfig_missingName(backend),
+				ExpectError: regexp.MustCompile(`Missing required argument|The argument "name" is required`),
+			},
+		},
+	})
+}
+
+func TestAccKerberosAuthBackendGroup_invalidNamespace(t *testing.T) {
```
siyer-corp, line 280: Do we need this as it's testing vault functionality, not TFVP?

website/docs/r/kerberos_auth_backend_ldap_config.html.md
```diff
@@ -0,0 +1,327 @@
...
+  (within its lifetime); 0 means unlimited. If you require the token to have the ability to 
+  create child tokens, you will need to set this value to 0.
+
+* `token_type` - (Optional) The type of token that should be generated. Can be `service`,
+  `batch`, or `default` to use the mount's tuned default (which unless changed will be
+  `service` tokens). For token store roles, there are two additional possibilities:
+  `default-service` and `default-batch` which specify the type to return unless the client
+  requests a different type at generation time.
+
+* `alias_metadata` - (Optional) A map of string to string that will be set as metadata on
```
harshit-nema, line 305: its for vault ent >=1.21

siyer-corp (commented):
As discussed, can we refactor the tests accordingly?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
vijayavelsekar: yes, while doing `terraform import vault_kerberos_auth_backend_config.config auth/kerberos/config`
We can give some invalid path. And this covers approximately 3% code coverage in ImportState function, specifically this error part.
```
resp.Diagnostics.AddError(
			"Invalid import ID format",
			fmt.Sprintf("Expected format: 'auth/<mount>/config', got: '%s'", req.ID),
		)
```

internal/vault/auth/kerberos/kerberos_group_resource_test.go
vijayavelsekar: You are right, but it help me cover the test coverage after WriteWithContext line. So it covers around 2% code.

internal/vault/auth/kerberos/kerberos_config_resource.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	var apiModel kerberosAuthBackendConfigAPIModel
+	if err := model.ToAPIModel(respData, &apiModel); err != nil {
+		diags.AddError("Unable to translate Vault response data", err.Error())
+		return diags
+	}
+
+	config.ServiceAccount = types.StringValue(apiModel.ServiceAccount)
+
```
Siddharth-Kannan, line 359: These fields are only written to state when the API returns true. When Vault returns false, the field stays as `types.BoolNull()` (the Go zero value), which means after terraform import both fields will be missing from state — causing a plan diff if the user has them explicitly set in config.

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,360 @@
...
+				Config: testAccKerberosAuthBackendConfigConfig_full(path, serviceAccount1, false, false),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldMount, path),
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldServiceAccount, serviceAccount1),
+					resource.TestCheckNoResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldKeytabWO),
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldRemoveInstanceName, "false"),
+					resource.TestCheckResourceAttr("vault_kerberos_auth_backend_config.config", consts.FieldAddGroupAliases, "false"),
+				),
+			},
+			{
```
Siddharth-Kannan, line 75: Consider changing this step to `true, false` so the import verify step covers both the true and false cases for these bool fields in a single step.
 The import step is called after the second step so if there is a false field here, the test case does not pass.

internal/vault/auth/kerberos/kerberos_group_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/model"
+)
+
+var (
+	// Regex patterns for parsing import path
+	kerberosAuthBackendGroupMountFromPathRegex = regexp.MustCompile("^auth/(.+)/groups/.+$")
+	kerberosAuthBackendGroupNameFromPathRegex  = regexp.MustCompile("^auth/.+/groups/(.+)$")
+)
+
+// Ensure the implementation satisfies the resource.ResourceWithConfigure interface
```
Siddharth-Kannan: The group resource is missing two compile-time interface guards that the other two resources have.  
```
var (
	_ resource.Resource                = (*kerberosAuthBackendLDAPConfigResource)(nil)
	_ resource.ResourceWithConfigure   = (*kerberosAuthBackendLDAPConfigResource)(nil)
	_ resource.ResourceWithImportState = (*kerberosAuthBackendLDAPConfigResource)(nil)
)
```
Should we also add the other 2 guards in this file as well?

internal/vault/auth/kerberos/kerberos_ldap_config_resource.go
```diff
@@ -0,0 +1,746 @@
...
+		tfModel.MaxPageSize = types.Int64Value(apiModel.MaxPageSize)
+	} else {
+		tfModel.MaxPageSize = types.Int64Null()
+	}
+
+	if apiModel.EnableSAMAccountNameLogin {
+		tfModel.EnableSAMAccountNameLogin = types.BoolValue(apiModel.EnableSAMAccountNameLogin)
+	}
+
+	// Populate token fields using the token package helper
```
Siddharth-Kannan, line 657: On Vault < 1.21, alias_metadata is correctly omitted from the write payload, but will `token.PopulateTokenModelFromAPI` then overwrite it with null on read-back — causing a persistent plan diff for users who have this field set? The CF auth backend role resource handles this with a **save/restore** pattern around **PopulateTokenModelFromAPI** — is the same needed here?

internal/vault/auth/kerberos/kerberos_config_resource_test.go
```diff
@@ -0,0 +1,360 @@
...
+				ImportStateVerify: false,
+				ExpectError:       regexp.MustCompile(`Invalid import ID format`),
+			},
+		},
+	})
+}
+
+// TestAccKerberosAuthBackendConfig_namespace tests configuration with namespace (Enterprise only)
+// Note: This test is currently skipped as vault_auth_backend resource doesn't properly support
+// namespaces in the Plugin Framework. This test serves as documentation for the namespace field.
```
Siddharth-Kannan: Is this comment still valid?
This comment says the test is currently skipped, but there is no t.Skip() call.

internal/vault/auth/kerberos/kerberos_login_ephemeral_resource.go
```diff
@@ -0,0 +1,367 @@
...
+		return
+	}
+
+	// Populate output fields from auth response
+	config.ClientToken = types.StringValue(secret.Auth.ClientToken)
+	config.Accessor = types.StringValue(secret.Auth.Accessor)
+	config.LeaseDuration = types.Int64Value(int64(secret.Auth.LeaseDuration))
+	config.Renewable = types.BoolValue(secret.Auth.Renewable)
+	config.EntityID = types.StringValue(secret.Auth.EntityID)
+
```
Siddharth-Kannan: Is there a reason for using if else for `Config.Orphan`? Can we not populate it the same way we are doing it for the other fields above?

internal/vault/auth/kerberos/kerberos_config_resource.go
vijayavelsekar: This is the design decision that we have made for now. If the user wants to unset the value they have to  remove it from the config.

internal/vault/auth/kerberos/kerberos_config_resource_test.go
vijayavelsekar: There is no false block to cover in the code.

Siddharth-Kannan (approved):

siyer-corp (approved):

PR #2830, opened 2026-03-17
Add vault_generic_endpoint ephemeral resource with auth and wrap_info support

### LINKED ISSUE #2542: [Enhancement]: return full Vault response with `vault_generic_endpoint`
### Description

Currently `vault_generic_endpoint` only returns the `.data` part of the response which is not always useful - such as when trying to create a token through `auth/token/create/<role>` where the `.data` field of the response is in fact `null` and the interesting data is returned in the `.auth` field.

Since the resource is called "generic", it should support all use cases, not just those which return `.data` in the response.

My use case specifically would be a workaround for #2541 as with neither this feature nor #2541 it is impossible to create an entity_alias-bound token using this provider.

* vault_generic_endpoint

```terraform
resource "vault_generic_endpoint" "my_app_token" {
  path = "auth/token/create/app"
  write_fields = ["auth"]
  disable_read = true
  disable_delete = true
  ignore_absent_fields = true

  data_json = <<EOT
{
  "entity_alias": "my-app"
}
EOT
}
```

### References

https://developer.hashicorp.com/vault/api-docs/auth/token#sample-response-1

No

itsjiyapatel:
### Description

The field extraction logic only checked response.Data, missing the response.Auth and response.WrapInfo structures that Vault returns for authentication and response wrapping operations.

Enhanced field extraction in [internal/vault/generic/generic_endpoint.go](vscode-webview://0dn3e0ob670p47lr38vkei6r50vhhq05hm8q3s7rbo2rtfpdt7sj/internal/vault/generic/generic_endpoint.go:1):

Added extraction from response.Auth (for authentication tokens)
Added extraction from response.WrapInfo (for wrapped responses)
Implemented proper field mapping (e.g., "token" → "client_token" from Auth)
Added path_wrap_ttl parameter to enable Vault response wrapping via X-Vault-Wrap-TTL header

Closes #2542

### Checklist

```
$ go test -v -count=1 -run TestAccGenericEndpointEphemeral ./internal/vault/generic
...
```
image

### Community Note

## PCI review checklist


internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,313 @@
...
+			consts.FieldPath: schema.StringAttribute{
+				MarkdownDescription: "Full path to the Vault endpoint that will be written",
+				Required:            true,
+			},
+			consts.FieldDataJSON: schema.StringAttribute{
+				MarkdownDescription: "JSON-encoded data to write.",
+				Required:            true,
+				Sensitive:           true,
+			},
+			"write_fields": schema.ListAttribute{
```
siyer-corp: Can we use consts here instead of string literals?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error Writing to Vault",
+			fmt.Sprintf("error writing to %s: %s", path, err),
+		)
+		return
+	}
+
+	// Debug: log what the response actually contains
+	if response != nil {
```
siyer-corp: Should we remove these as they may contain sensitive information?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,313 @@
...
+			if resp.Diagnostics.HasError() {
+				return
+			}
+		}
+
+		for _, writeField := range writeFields {
+			// 1) initial check for response.Data
+			if response.Data != nil {
+				if v, ok := response.Data[writeField]; ok {
+					log.Printf("[DEBUG] %s found in response.Data", writeField)
```
siyer-corp: We should probably remove all statements that logs sensitive data.

siyer-corp (commented):
Added a few comments on first pass. Also, are we making changes to the resource_generic_endpoint as well?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// Process write_fields if provided
+	writeDataMap := make(map[string]string)
+	writeData := make(map[string]interface{})
+
+	// Prepare wrap info map if present
+	var wrapMap map[string]interface{}
+	if response != nil && response.WrapInfo != nil {
+		if wb, err := json.Marshal(response.WrapInfo); err == nil {
+			_ = json.Unmarshal(wb, &wrapMap)
```
siyer-corp: If json.Unmarshal fails, wrapMap will be nil and fields won't be extracted, but users won't know why their write_fields weren't populated. Can we add a warning here when it fails?

```
if err := json.Unmarshal(wb, &wrapMap); err != nil {
        resp.Diagnostics.AddWarning(
            "Failed to parse WrapInfo",
            fmt.Sprintf("Could not extract wrap_info fields: %s", err),
        )
    }
```

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+	}
+
+	// Prepare auth map if present
+	var authMap map[string]interface{}
+	if response != nil && response.Auth != nil {
+		if ab, err := json.Marshal(response.Auth); err == nil {
+			_ = json.Unmarshal(ab, &authMap)
```
siyer-corp: Can we add error handling here as well?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,291 @@
...
+			if resp.Diagnostics.HasError() {
+				return
+			}
+		}
+
+		for _, writeField := range writeFields {
+			// 1) initial check for response.Data
+			if response.Data != nil {
+				if v, ok := response.Data[writeField]; ok {
+					writeData[writeField] = v
```
siyer-corp: It looks like this piece of code is duplicated across, can we extract it to a function?

```
if vs, ok := v.(string); ok {
    writeDataMap[writeField] = vs
} else {
    vBytes, _ := json.Marshal(v)
    writeDataMap[writeField] = string(vBytes)
}
```

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,291 @@
...
+		var writeFields []string
+		if !data.WriteFields.IsNull() && !data.WriteFields.IsUnknown() {
+			resp.Diagnostics.Append(data.WriteFields.ElementsAs(ctx, &writeFields, false)...)
+			if resp.Diagnostics.HasError() {
+				return
+			}
+		}
+
+		for _, writeField := range writeFields {
+			// 1) initial check for response.Data
```
siyer-corp: Can we make these comments more descriptive, it'd be good to add why and when this will be executed.

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,442 @@
...
+	path := instanceState.ID
+
+	if path != instanceState.Attributes["path"] {
+		return fmt.Errorf("id doesn't match path")
+	}
+	if path != "identity/lookup/entity" {
+		return fmt.Errorf("unexpected secret path")
+	}
+
+	write_data_count := instanceState.Attributes["write_data.%"]
```
siyer-corp: nit: It'd be good to use camelCase here instead of snake casing.

website/docs/ephemeral-resources/generic_endpoint.html.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+### DR Secondary Token Generation
+
+```hcl
+# Generate DR secondary token with response wrapping
+ephemeral "vault_generic_endpoint" "dr_secondary_token" {
+  path          = "sys/replication/dr/primary/secondary-token"
+  disable_read  = true
```
siyer-corp: Is this field missing from the schema?

siyer-corp (commented):
Provided some comments to go through 😄

siyer-corp (commented):
Looking good! :clap::clap:

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
+// Copyright (c) HashiCorp, Inc.
```
harshit-nema, line 1: File name should have ephemeral word for better readability.

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+// Metadata sets the full name for this resource
+func (r *GenericEndpointEphemeralResource) Metadata(ctx context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_generic_endpoint"
+}
+
+// addToWriteData is a helper function that adds a value to both writeData and writeDataMap.
+// It handles string values directly and marshals non-string values to JSON.
+func addToWriteData(writeData map[string]interface{}, writeDataMap map[string]string, field string, value interface{}) {
+	writeData[field] = value
+	if vs, ok := value.(string); ok {
```
harshit-nema, line 101: Shall the word be more meaningful rather than "vs" ?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+	var data GenericEndpointEphemeralModel
+
+	// Read Terraform configuration data into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	c, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
```
harshit-nema: Should we rename "c" with descriptive word?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// Process write_fields if provided
+	writeDataMap := make(map[string]string)
+	writeData := make(map[string]interface{})
+
+	// Prepare wrap info map if present
+	var wrapMap map[string]interface{}
+	if response != nil && response.WrapInfo != nil {
+		if wb, err := json.Marshal(response.WrapInfo); err == nil {
```
harshit-nema, line 179: "wb", naming convention. Would be more readable with descriptive word.

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+					fmt.Sprintf("Could not extract wrap_info fields: %s", err),
+				)
+			}
+		}
+	}
+
+	// Prepare auth map if present
+	var authMap map[string]interface{}
+	if response != nil && response.Auth != nil {
+		if ab, err := json.Marshal(response.Auth); err == nil {
```
harshit-nema, line 192: "ab", naming convention, similarly at other places. Should it be readable if descriptive word is provided ?

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+				resp.Diagnostics.AddWarning(
+					"Failed to parse Auth",
+					fmt.Sprintf("Could not extract auth fields: %s", err),
+				)
+			}
+		}
+	}
+
+	if response != nil && (response.Data != nil || wrapMap != nil || authMap != nil || response.LeaseDuration != 0 || response.LeaseID != "") {
+		// Get write_fields list
```
harshit-nema, line 203: Should we add diagnostic warning when write_fields is specified but no response data is available to extract ? This would handle the edge case.

internal/vault/generic/generic_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+		}
+
+		for _, writeField := range writeFields {
+			// 1) Check response.Data first - this is the primary data payload returned by most Vault endpoints.
+			// It is always checked first for any write_field requested by the user.
+			// As most Vault API responses return their data in the Data field (e.g., secret values, configuration).
+			if response.Data != nil {
+				if v, ok := response.Data[writeField]; ok {
+					addToWriteData(writeData, writeDataMap, writeField, v)
```
harshit-nema, line 218: In addToWriteData, its observed that writeData and writeDataMap are storing same data. Shall we remove one parameter to avoid redundancy ? 
writeData and writeDataMap are Terraform features and not from Vault right ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+  data_json = jsonencode({ password = "%s" })
+  write_fields = ["token", "accessor"]
+}
+
+# Echo provider: authenticated with the ephemeral token
+provider "vault" {
+  alias            = "echo_test"
+  address          = "%s"
+  token            = ephemeral.vault_generic_endpoint.u1_token.write_data["token"]
+  skip_child_token = true
```
harshit-nema: Rather than using skip_child_token, shall we create policy granting specific permission to read the ephemeral data ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+  depends_on           = [vault_auth_backend.userpass]
+  path                 = "auth/%s/users/%s"
+  ignore_absent_fields = true
+  data_json = jsonencode({
+    password = "%s"
+  })
+}
+
+ephemeral "vault_generic_endpoint" "u1_token" {
+  path      = "auth/%s/login/%s"
```
harshit-nema: Is ephemeral working without using `id` in single execution ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+// (token, ttl, creation_time, etc.) are successfully extracted from the response.WrapInfo structure.
+func TestAccGenericEndpointEphemeral_wrapInfo(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGenericEndpointEphemeral_wrapInfoConfig(),
```
harshit-nema, line 99: What are the asserts or verification targeted for this test step ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+// This verifies that write_fields = ["wrap_info"] extracts the full WrapInfo structure as JSON.
+func TestAccGenericEndpointEphemeral_wrapInfoSpecialField(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGenericEndpointEphemeral_wrapInfoSpecialFieldConfig(),
```
harshit-nema, line 165: What are we asserting here ??

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
+			// Step 2: Add ephemeral resource with "auth" special field
+			{
+				Config: testAccGenericEndpointEphemeral_authSpecialFieldConfig(mount, username, password),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
```
harshit-nema, line 221: Shall we add the assertion statements for this test case ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+				Config: testAccGenericEndpointEphemeral_infraConfig(mount, username, password),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
+			// Step 2: Add ephemeral resource with "token" alias
+			{
+				Config: testAccGenericEndpointEphemeral_tokenAliasConfig(mount, username, password),
```
harshit-nema, line 258: What are the assertions targeted here ? It would be wholesome if assertions are added here.

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -0,0 +1,518 @@
...
+// This verifies that fields like lease_duration and lease_id can be extracted from the response root.
+func TestAccGenericEndpointEphemeral_topLevelFields(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGenericEndpointEphemeral_topLevelFieldsConfig(),
```
harshit-nema, line 286: Would calling ephemeral block directly be completing the test? What is the behavior in manual testing ? + assertions to be added.
Similar observation for : TestAccGenericEndpointEphemeral_extractFromData, TestAccGenericEndpointEphemeral_complexTypes

harshit-nema (commented):
Review comments are provided. On the generic note, it would be better to use declarative variable name and assertion in all the added test cases.

internal/vault/generic/generic_endpoint.go
siyer-corp: I think terse var names should be OK if their scope is small. If the var is used far away from the declaration we should consider having longer names.

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -173,9 +199,22 @@ func TestAccGenericEndpointEphemeral_authSpecialField(t *testing.T) {
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
+			// Step 3: Remove the ephemeral login resource before test cleanup.
+			{
+				Config: testAccGenericEndpointEphemeral_infraConfig(mount, username, password),
 				ConfigPlanChecks: resource.ConfigPlanChecks{
```
harshit-nema: Since the comment says "Remove the ephemeral login resource before test cleanup.", so why should we be calling config here ?? Calling testAccGenericEndpointEphemeral_infraConfig looks redudant.

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -208,9 +250,22 @@ func TestAccGenericEndpointEphemeral_tokenAlias(t *testing.T) {
...
+					// Verify the token field is extracted using the alias
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("token"), knownvalue.NotNull()),
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
+			// Step 3: Remove the ephemeral login resource before test cleanup.
```
harshit-nema: Since the comment says "Remove the ephemeral login resource before test cleanup.", so should we be calling config here ?? Calling testAccGenericEndpointEphemeral_infraConfig looks redudant.

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -138,9 +154,16 @@ func TestAccGenericEndpointEphemeral_wrapInfoSpecialField(t *testing.T) {
...
 		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGenericEndpointEphemeral_wrapInfoSpecialFieldConfig(),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify the entire wrap_info object is extracted as a JSON string
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("wrap_info"), knownvalue.NotNull()),
```
harshit-nema: Is the value wrap_info a default returned by vault ? If no, should we assert on expected value rather than on not null ?

internal/vault/generic/generic_endpoint_ephemeral_test.go
```diff
@@ -79,15 +84,26 @@ func TestAccGenericEndpointEphemeral(t *testing.T) {
...
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGenericEndpointEphemeral_wrapInfoConfig(),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify wrap_info fields are extracted from response.WrapInfo
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("token"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("ttl"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("creation_time"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("wrapped_accessor"), knownvalue.NotNull()),
+				},
```
harshit-nema, line 109: Should we provide assertion on the exact values rather than on not null ? Are these the defaults returned by vault or set during the request ?

internal/vault/generic/generic_endpoint_ephemeral.go
```diff
@@ -182,7 +199,7 @@ func (r *GenericEndpointEphemeralResource) Open(ctx context.Context, req ephemer
 		}
 	}
 
-	if response != nil && (response.Data != nil || wrapMap != nil || authMap != nil || response.LeaseDuration != 0 || response.LeaseID != "") {
+	if response != nil {
```
harshit-nema, line 202: Since the !nil conditions are removed, is the code handling the !nil and =nil condition for the removed parameters as expected ?

internal/vault/generic/generic_endpoint_ephemeral_test.go
```diff
@@ -229,9 +284,17 @@ func TestAccGenericEndpointEphemeral_topLevelFields(t *testing.T) {
...
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGenericEndpointEphemeral_topLevelFieldsConfig(),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify top-level fields are extracted
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("lease_duration"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("lease_id"), knownvalue.NotNull()),
```
harshit-nema, line 292: Are lease_duration and lease_id carrying defaults from vault ? If not, should we provide value check rather than null check here ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -250,9 +313,17 @@ func TestAccGenericEndpointEphemeral_extractFromData(t *testing.T) {
...
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGenericEndpointEphemeral_extractFromDataConfig(),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify fields from response.Data are extracted
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("accessor"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("policies"), knownvalue.NotNull()),
```
harshit-nema: Are 'accessor' and 'policies' carrying defaults from vault ? If not, should we provide value check rather than null check here ?

internal/vault/generic/generic_endpoint_test.go
```diff
@@ -271,9 +342,16 @@ func TestAccGenericEndpointEphemeral_complexTypes(t *testing.T) {
...
 		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGenericEndpointEphemeral_complexTypesConfig(),
+				ConfigStateChecks: []statecheck.StateCheck{
+					// Verify complex types (arrays) are properly extracted and JSON-encoded
+					statecheck.ExpectKnownValue("echo.test", tfjsonpath.New("data").AtMapKey("policies"), knownvalue.NotNull()),
```
harshit-nema: Should we provide expected value check rather than null check here ?

internal/vault/generic/generic_endpoint_ephemeral.go
itsjiyapatel: Yes since we are checking the nested fields individually as well before access

internal/vault/generic/generic_endpoint_test.go
itsjiyapatel: We can add assertions for the fields but the values are computed by vault and hence cannot add expected value checks here.

itsjiyapatel: Depending on the fields with exact values as per config we can add assertions for those such as creation_path and ttl

internal/vault/generic/generic_endpoint_ephemeral.go
```diff
@@ -0,0 +1,343 @@
...
+		// Get write_fields list
+		var writeFields []string
+		if !data.WriteFields.IsNull() && !data.WriteFields.IsUnknown() {
+			resp.Diagnostics.Append(data.WriteFields.ElementsAs(ctx, &writeFields, false)...)
+			if resp.Diagnostics.HasError() {
+				return
+			}
+		}
+
+		foundFields := make(map[string]bool)
```
siyer-corp: Is this being used? It looks like we're just setting the values in the map to true if the fields are found, but never check against the values for throwing the warning. It would eventually go to the warning if all of the if conditions fail. So we can probably get rid of this map.

internal/vault/generic/generic_endpoint_ephemeral.go
```diff
@@ -0,0 +1,343 @@
...
+						continue
+					}
+				}
+			}
+
+			// 6) Check top-level response fields for lease and renewal information.
+			// When User requests lease_duration, lease_id, or renewable and the response includes lease data.
+			// Some Vault responses (especially dynamic secrets) include lease information at the top level
+			// rather than in Data, Auth, or WrapInfo. These fields control secret lifecycle and renewal.
+			topLevel := map[string]interface{}{
```
siyer-corp, line 276: Are these top level fields always returned by Vault?

website/docs/ephemeral-resources/generic_endpoint.html.md
```diff
@@ -0,0 +1,205 @@
...
+  Write to a generic Vault endpoint and extract response data as ephemeral values.
+---
+
+# vault\_generic\_endpoint (Ephemeral)
+
+Writes to a generic Vault endpoint and extracts response data as ephemeral values that are not stored in Terraform state.
+This is useful for authentication flows, token generation, or any Vault operation
+where you need to extract sensitive data without persisting it to state.
+
+The ephemeral resource supports extracting fields from three response structures:
```
siyer-corp: This seems redundant as this info is already provided in detail in line 137

website/docs/ephemeral-resources/generic_endpoint.html.md
```diff
@@ -0,0 +1,205 @@
...
+   - Full structure: Use `"wrap_info"` in `write_fields`
+3. **response.Auth** - Authentication response fields
+   - Individual fields: `accessor`, `policies`, `lease_duration`
+   - Token field: `"token"` maps to `client_token`
+   - Full structure: Use `"auth"` in `write_fields`
+4. **Top-level response** - `lease_duration`, `lease_id`, `renewable`
+
+## Important Notes
+
+~> **Important** Ephemeral resources are designed for sensitive data that should not be stored in Terraform state. However, the data will still appear in console output when Terraform runs and may be included in plan files if secrets are interpolated into resource attributes. Protect these artifacts accordingly.
```
siyer-corp: Can we change “will still appear in console output” to “can still appear in console output”

siyer-corp (approved):
Thank you, LGTM!

weitzj:
Yeah! Thank you very much. ;)

PR #2860, opened 2026-04-01
add length based id generation

vivekpandeyHashi:
### Description
Bug: https://hashicorp.atlassian.net/browse/VAULT-31541

Fix ID parsing for secrets_sync_association resource

### Problem
The current ID format for vault_secrets_sync_association uses regex-based parsing with keyword delimiters:

`{destType}/dest/{name}/mount/{mount}/secret/{secretName}
`
This fails when field values contain the keywords "dest", "mount", or "secret". For example:

name = "my/mount" → ID becomes aws-kms/dest/my/mount/mount/kv/secret/token
Regex incorrectly parses "my/" as part of destType instead of name

### Solution
Implemented length-based ID format that uses explicit field lengths for parsing:

```
{len1},{len2},{len3},{len4}:{destType}:{name}:{mount}:{secretName}
Example: 7,8,2,5:aws-kms:my-mount:kv:token
```

The parser uses the length values to extract fields, making it immune to keyword collisions and special characters (including colons in field values).

### Changes

- Updated ID generation in secretsSyncAssociationWrite()
- Added isLengthBasedIDFormat() for format detection (checks if ID starts with digit)
- Added parseLengthBasedID() for parsing new format
- Updated syncAssociationFieldsFromID() to support both formats

### Backward Compatibility
Fully backward compatible - old format IDs continue to work

Automatic format detection: digit = new format, letter = old format
No state migration required
Existing resources unaffected

### DIFF
--- a/vault/resource_secrets_sync_association.go
+++ b/vault/resource_secrets_sync_association.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"strconv"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -117,9 +119,11 @@ func secretsSyncAssociationWrite(ctx context.Context, d *schema.ResourceData, me
 	}
 	log.Printf("[DEBUG] Wrote association to %q", path)
 
-	// ex: gh/dest/gh-dest-1/mount/kv/secret/token
-	// unique. each destination should only have one association for a particular accessor and secret
-	id := fmt.Sprintf("%s/dest/%s/mount/%s/secret/%s", destType, name, mount, secretName)
+	// Length-based ID format: {len1},{len2},{len3},{len4}:{destType}:{name}:{mount}:{secretName}
+	// Example: 7,8,2,5:aws-kms:my-mount:kv:token
+	id := fmt.Sprintf("%d,%d,%d,%d:%s:%s:%s:%s",
+		len(destType), len(name), len(mount), len(secretName),
+		destType, name, mount, secretName)
 	d.SetId(id)
 
 	return secretsSyncAssociationRead(ctx, d, meta)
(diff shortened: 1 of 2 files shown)


vault/resource_secrets_sync_association.go
```diff
@@ -117,9 +119,11 @@ func secretsSyncAssociationWrite(ctx context.Context, d *schema.ResourceData, me
 	}
 	log.Printf("[DEBUG] Wrote association to %q", path)
 
-	// ex: gh/dest/gh-dest-1/mount/kv/secret/token
-	// unique. each destination should only have one association for a particular accessor and secret
-	id := fmt.Sprintf("%s/dest/%s/mount/%s/secret/%s", destType, name, mount, secretName)
+	// Length-based ID format: {len1},{len2},{len3},{len4}:{destType}:{name}:{mount}:{secretName}
+	// Example: 7,8,2,5:aws-kms:my-mount:kv:token
+	id := fmt.Sprintf("%d,%d,%d,%d:%s:%s:%s:%s",
```
vijayavelsekar, line 124: Now that we are updating the id format, will there be any change in the `terraform import` ? Should we document that in the md doc file ?

vault/resource_secrets_sync_association_test.go
```diff
@@ -93,3 +93,260 @@ resource "vault_secrets_sync_association" "test" {
 
 	return ret
 }
+
+// TestSyncAssociationFieldsFromID_NewFormat tests the new length-based ID format
+func TestSyncAssociationFieldsFromID_NewFormat(t *testing.T) {
```
vijayavelsekar, line 98: Are the existing acceptance tests passes without skipping in all the necessary versions of vault ?

vault/resource_secrets_sync_association.go
vivekpandeyHashi: I don't think so
we will have backward compatibility for the old format as well
while reading we can identify the format and handle this accordingly
@vijayavelsekar

vault/resource_secrets_sync_association_test.go
vivekpandeyHashi: yes

vault/resource_secrets_sync_association.go
vijayavelsekar: @vivekpandeyHashi I've tried terraform import, the id field is not user friendly as the user have to count the length and form the id during terraform import. 

`terraform import vault_secrets_sync_association.gcp_token 14,11,4,16:vercel-project:vercel-dest:kvv2:gai/secret/aihub`

Can we try to implement without the length fields ? Or we will have to find some other way in such a way that terraform import is easier for the users.
`Ex: vercel-project:vercel-dest:kvv2:gai/secret/aihub`

vivekpandeyHashi: Hi @vijayavelsekar 
Thanks for trying out.
why user have to count the length? in which scenario?

vijayavelsekar: @vivekpandeyHashi When the user newly going to onboard their vault infrastructure, in our case the secret sync configuration into terraform to maintain their infra via terraform, the user will use terraform import command, where they have to provide the id value as I've mentioned in my previous comment. This will populate the state-file with the values based on their vault configuration.
Let me know if you need a call, so I can explain with an example.

vijayavelsekar (changes requested):
Alternate approach identified.
For backward compatibility we can have the existing functionality as it is and to fix this issue we can use [Resource Identity](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/identity).
But this approach needs to be discussed as this can be done only from terraform version 1.12.0 or later.

Another approach identified is to have comman seperated id as `type,destination,mount,secretName`

PR #2863, opened 2026-04-01
TFVP-Initial-Implementation-Sys-Backend-Group-Policy-Application

itsjiyapatel:
### Description
Implements a new Terraform resource vault_config_group_policy_application for managing Vault's group policy application mode configuration. This provides a type-safe, declarative way to control how identity group policies are applied across namespace hierarchies, replacing the need to use vault_generic_endpoint for this critical security configuration.

Resource: vault_config_group_policy_application at internal/vault/sys/config/group_policy_application.go

Key Features:
- Full CRUD operations with proper error handling
- Root/admin namespace restriction with validation
- Two supported modes: within_namespace_hierarchy (default) and any
- Singleton resource pattern with ID "sys/config/group-policy-application"
- Import support with ID validation
- Enterprise namespace support
- Delete operation resets to default mode

Configuration Modes:
- within_namespace_hierarchy: Groups can only apply policies within their namespace hierarchy (default, more restrictive)
- any: Groups can apply policies from any namespace (less restrictive, increased flexibility)

Relates OR Closes #0000

### Checklist

```
$ go test -v -count=1 -run TestAccConfigGroupPolicyApplication 
=== RUN   TestAccConfigGroupPolicyApplication
--- PASS: TestAccConfigGroupPolicyApplication (0.88s)
=== RUN   TestAccConfigGroupPolicyApplication_update
--- PASS: TestAccConfigGroupPolicyApplication_update (1.52s)
=== RUN   TestAccConfigGroupPolicyApplication_invalidMode
--- PASS: TestAccConfigGroupPolicyApplication_invalidMode (0.23s)
=== RUN   TestAccConfigGroupPolicyApplication_invalidNamespace
--- PASS: TestAccConfigGroupPolicyApplication_invalidNamespace (0.24s)
=== RUN   TestAccConfigGroupPolicyApplication_providerNamespaceUnset
--- PASS: TestAccConfigGroupPolicyApplication_providerNamespaceUnset (0.24s)
=== RUN   TestAccConfigGroupPolicyApplication_importInvalidID
--- PASS: TestAccConfigGroupPolicyApplication_importInvalidID (0.57s)
=== RUN   TestAccConfigGroupPolicyApplication_explicitRootNamespace
--- PASS: TestAccConfigGroupPolicyApplication_explicitRootNamespace (0.67s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/sys/config 5.594s

...
```

### Community Note

## PCI review checklist


internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	if !provider.IsEnterpriseSupported(r.Meta()) {
+		diagnostics.AddError(
+			"Enterprise Feature Required",
+			"Group policy application configuration is only available in Vault Enterprise.",
+		)
+		return false
+	}
+
+	// Version check
+	if !provider.IsAPISupported(r.Meta(), provider.VaultVersion1138) {
```
vijayavelsekar: TFVP is supported from vault version 1.15.0+
So we should be good to remove this check.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+				},
+			},
+			consts.FieldGroupPolicyApplicationMode: schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				MarkdownDescription: "Mode for group policy application. Must be either \"within_namespace_hierarchy\" or \"any\". " +
+					"Defaults to \"within_namespace_hierarchy\". " +
+					"\"within_namespace_hierarchy\" means policies only apply when the token authorizing a request was created in the same namespace as the group, or a descendant namespace. " +
+					"\"any\" means group policies apply to all members of a group, regardless of what namespace the request token came from.",
+				Default: stringdefault.StaticString(modeWithinNamespaceHierarchy),
```
vijayavelsekar: Have you checked if the api works with similar default if this field is not sent ?

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+				Validators: []validator.String{
+					stringvalidator.OneOf(modeWithinNamespaceHierarchy, modeAny),
+				},
+			},
+			consts.FieldNamespace: schema.StringAttribute{
```
vijayavelsekar: Instead of directly using this namspace field, try to use `base.MustAddBaseSchema(&resp.Schema)` to be in sync with other resource.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+			},
+			consts.FieldGroupPolicyApplicationMode: schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				MarkdownDescription: "Mode for group policy application. Must be either \"within_namespace_hierarchy\" or \"any\". " +
+					"Defaults to \"within_namespace_hierarchy\". " +
+					"\"within_namespace_hierarchy\" means policies only apply when the token authorizing a request was created in the same namespace as the group, or a descendant namespace. " +
+					"\"any\" means group policies apply to all members of a group, regardless of what namespace the request token came from.",
+				Default: stringdefault.StaticString(modeWithinNamespaceHierarchy),
+				Validators: []validator.String{
```
vijayavelsekar: I think its good to avoid validations to avoid any future changes if there is any additional mode gets added in vault side.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+
+	// Read Terraform plan data into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Validate enterprise and version requirements
+	if !r.validateEnterpriseAndVersion(&resp.Diagnostics) {
```
vijayavelsekar: We can remove these validations to keep the resource simple as they will be handled in the vault side anyway.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	vaultClient, ok := r.getClientForNamespace(ctx, namespace, &resp.Diagnostics)
+	if !ok {
+		return
+	}
+
+	// Get the actual namespace from the client
+	actualNamespace := vaultClient.Namespace()
+
+	// Validate actual namespace - must be root or administrative
+	if !r.validateActualNamespace(actualNamespace, &resp.Diagnostics) {
```
vijayavelsekar: Same as my above comment

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/base"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/client"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/errutil"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+	"github.com/hashicorp/vault/api"
+)
+
+const (
+	// Singleton resource ID
+	configGroupPolicyApplicationID = "config"
```
vijayavelsekar: Terraform resource usually would have path `/sys/config/group-policy-application` as ID.
Hence same will be used during Import.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	// Write configuration to Vault
+	if !r.writeGroupPolicyConfig(ctx, vaultClient, mode, &resp.Diagnostics, errutil.VaultCreateErr) {
+		return
+	}
+
+	// Set the singleton ID
+	data.ID = types.StringValue(configGroupPolicyApplicationID)
+
+	// Set namespace to the actual namespace from the client
+	data.Namespace = types.StringValue(actualNamespace)
```
vijayavelsekar: At this point `data.Namespace` will already have the value from the plan. Setting it seems redundant.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+
+	mode := data.GroupPolicyApplicationMode.ValueString()
+
+	// Write configuration to Vault
+	if !r.writeGroupPolicyConfig(ctx, vaultClient, mode, &resp.Diagnostics, errutil.VaultUpdateErr) {
+		return
+	}
+
+	// Always set namespace explicitly (it's computed)
+	data.Namespace = types.StringValue(actualNamespace)
```
vijayavelsekar: Once after adding MustAddBaseSchema as per my comment in the schema section, this will not be computed and we don't have to set it as they will be in the plan already.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	if !r.validateActualNamespace(actualNamespace, &resp.Diagnostics) {
+		return
+	}
+
+	// Reset to default value instead of deleting
+	if !r.writeGroupPolicyConfig(ctx, vaultClient, modeWithinNamespaceHierarchy, &resp.Diagnostics, errutil.VaultUpdateErr) {
+		return
+	}
+
+	// If the logic reaches here, it implicitly succeeded and will remove
```
vijayavelsekar, line 280: Add a diagnostic warning here, saying that this resource is not actually deletable but resets to default. Something like that.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	// the resource from state if there are no other errors.
+}
+
+// ImportState implements the import functionality for this resource
+func (r *ConfigGroupPolicyApplicationResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// The import ID should always be "config"
+	if req.ID != configGroupPolicyApplicationID {
```
vijayavelsekar: import ID should be path `/sys/config/group-policy-application`

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,424 @@
...
+	if !r.writeGroupPolicyConfig(ctx, vaultClient, modeWithinNamespaceHierarchy, &resp.Diagnostics, errutil.VaultUpdateErr) {
+		return
+	}
+
+	// If the logic reaches here, it implicitly succeeded and will remove
+	// the resource from state if there are no other errors.
+}
+
+// ImportState implements the import functionality for this resource
+func (r *ConfigGroupPolicyApplicationResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
vijayavelsekar, line 285: Did we manually verify if the import works for root and admin namespace ?

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion1138)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccConfigGroupPolicyApplicationConfig_basic("invalid_mode"),
+				ExpectError: regexp.MustCompile(`Attribute group_policy_application_mode value must be one of`),
```
vijayavelsekar: we can remove this validation in the resource and check for the error message from the vault.

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+)
+
+// TestAccConfigGroupPolicyApplication tests basic resource creation and import
+func TestAccConfigGroupPolicyApplication(t *testing.T) {
+	resourceName := "vault_config_group_policy_application.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion1138)
```
vijayavelsekar: These version gating are not required in all the test cases as the tfvp is supported from v1.15.0+

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccConfigGroupPolicyApplicationConfig_basic("any"),
+				ExpectError: regexp.MustCompile(`(Invalid Namespace|Enterprise Feature Required)`),
```
vijayavelsekar: I don't think the `Expected error needs Enterprise Feature Required` is needed, because if its not enterprise, this test will be skipped in the first place.

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccConfigGroupPolicyApplicationConfig_basic("any"),
+			},
+			{
+				ResourceName:  "vault_config_group_policy_application.test",
+				ImportState:   true,
+				ImportStateId: "invalid-id",
+				ExpectError:   regexp.MustCompile(`Import ID must be "config"`),
```
vijayavelsekar: The import ID must be the path

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+			},
+			{
+				// Delete the resource - this should reset to default
+				Config: testAccConfigGroupPolicyApplicationConfig_empty(),
+			},
+			{
+				// Re-import to verify it was reset to default
+				Config: testAccConfigGroupPolicyApplicationConfig_basic("within_namespace_hierarchy"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldGroupPolicyApplicationMode, "within_namespace_hierarchy"),
```
vijayavelsekar: This does not test actual delete as you are setting it back to `within_namespace_hierarchy`. Also the `testAccConfigGroupPolicyApplicationConfig_empty() `function is dummy function that does not do any delete. I think its better not to have a separate test case for this, as every time a test case is completed, the delete method is being called and that confirms our testing already.

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,321 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateId:     "config",
+			},
+		},
+	})
+}
+
+// TestAccConfigGroupPolicyApplication_importVerifyNamespace tests import with namespace verification
+func TestAccConfigGroupPolicyApplication_importVerifyNamespace(t *testing.T) {
```
vijayavelsekar: This test seems redundant as the above one.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+**Important:** This is a singleton resource - only one instance can exist per Vault cluster. The resource must be managed from the root or administrative namespace.
+
+**Note** this feature is available only with Vault Enterprise 1.13.8+.
+
+## Example Usage
+
+### Basic Usage (Root Namespace)
+
+```hcl
+terraform {
```
vijayavelsekar: In examples, no need to provide terraform block

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+  required_providers {
+    vault = {
+      source = "hashicorp/vault"
+    }
+  }
+}
+
+provider "vault" {
```
vijayavelsekar: not required in examples

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+  # VAULT_ADDR, VAULT_TOKEN, VAULT_NAMESPACE
+  # Or specified directly:
+  address   = "http://127.0.0.1:8200"
+  token     = "your-token-here"
+  namespace = "admin"
+}
+
+resource "vault_config_group_policy_application" "config" {
+  group_policy_application_mode = "within_namespace_hierarchy"
+  # namespace is inherited from provider configuration
```
vijayavelsekar: I think its better to stick to the usage of this particular resource and you can provide an example of providing namespace using the namespace field. We can remove the vault provider usage.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+  # Example: export VAULT_TOKEN="your-token-here"
+}
+
+# Configure group policy application mode in root namespace
+# Using the default mode
+resource "vault_config_group_policy_application" "test" {
+  group_policy_application_mode = "within_namespace_hierarchy"
+}
+
+# Outputs
```
vijayavelsekar: Output blocks are not required.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+Alternatively, using environment variables:
+
+```bash
+export VAULT_ADDR="http://127.0.0.1:8200"
+export VAULT_TOKEN="your-token-here"
+export VAULT_NAMESPACE="admin"
+```
+
+```hcl
+provider "vault" {
```
vijayavelsekar: vault provider is not required for the examples. Its better to stick to providing example to the current resource usage.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+  Set this to `""` to explicitly target the root namespace. The value should not contain leading or trailing forward slashes.
+  *Available only for Vault Enterprise*.
+
+## Attributes Reference
+
+In addition to the arguments above, the following attributes are exported:
+
+* `id` - The resource ID (always `"config"`).
+
+* `namespace` - The namespace where the configuration is managed.
```
vijayavelsekar: You can remove namespace from here as you have already covered it in the Argument reference.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Import
+
+The group policy application configuration can be imported using the ID `config`:
```
vijayavelsekar: ID should be path

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Namespace Requirements
+
+This resource can only be managed from:
+- The root namespace (empty string `""`)
+- The administrative namespace (`"admin"`)
+
+Attempting to manage this resource from any other namespace will result in an error.
+
+### Administrative Namespace
```
vijayavelsekar: The informations in this section `Administrative Namespace` is out of scope of this document. We should be good to remove it.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,188 @@
...
+
+storage "raft" {
+  path = "./vault-data/raft"
+  node_id = "node1"
+}
+```
+
+## Version Requirements
+
+- Requires Vault Enterprise 1.13.8 or later
```
vijayavelsekar: TFVP itself is supported from v1.15.0+

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ BREAKING CHANGES:
 
 FEATURES:
 
+* **New Resource**: `vault_config_group_policy_application` - Manages the global group policy application mode for Vault Enterprise. Controls how policies attached to identity groups are applied across namespace boundaries. Supports `within_namespace_hierarchy` (default) and `any` modes. Requires Vault Enterprise 1.13.8+. ([#2863](https://github.com/hashicorp/terraform-provider-vault/pull/2863))
```
vijayavelsekar: You need to update your branch with the latest changes from main and fix the location of your change log entry. Please be extra cautious while doing merge or rebase. Make sure we don't loose any previous entries.

internal/vault/sys/config/group_policy_application.go
itsjiyapatel: yes manually tested for both the namespaces and also for both the group_policy_application_modes

internal/vault/sys/config/group_policy_application.go
itsjiyapatel: Vault API requires `group_policy_application_mode` to be sent, will change the field to required

internal/consts/consts.go
```diff
@@ -64,6 +64,8 @@ const (
 	FieldData                               = "data"
 	FieldDisableRead                        = "disable_read"
 	FieldName                               = "name"
+	FieldValues                             = "values"
```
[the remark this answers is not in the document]
vijayavelsekar: @itsjiyapatel Should we remove this constant ?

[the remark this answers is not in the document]
itsjiyapatel: This constant was not added as a part of this PR, it is used in other resources I don't think we should be removing it.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,104 @@
...
+  group_policy_application_mode = "any"
+}
+```
+
+### Explicit Root Namespace
+
+```hcl
+resource "vault_config_group_policy_application" "config" {
+  group_policy_application_mode = "within_namespace_hierarchy"
+  namespace                     = ""
```
vijayavelsekar: For root namespace we don't have to provide the namespace field.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,104 @@
...
+This resource is a singleton, meaning only one instance can exist per Vault cluster. Multiple `vault_config_group_policy_application` resources in your Terraform configuration will conflict.
+
+### Deletion Behavior
+
+When this resource is destroyed via `terraform destroy`, it does not delete the configuration from Vault. Instead, it resets the mode to the default value `within_namespace_hierarchy`. This ensures that Vault always has a valid group policy application mode configured.
+
+### Namespace Requirements
+
+This resource can only be managed from:
+- The root namespace (empty string `""`)
```
vijayavelsekar: For root namespace we don't have to provide the namespace field.

internal/consts/consts.go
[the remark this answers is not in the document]
vijayavelsekar: Then not required.

internal/provider/meta.go
```diff
@@ -41,6 +41,7 @@ var (
 	VaultVersion111  = version.Must(version.NewSemver(consts.VaultVersion111))
 	VaultVersion112  = version.Must(version.NewSemver(consts.VaultVersion112))
 	VaultVersion113  = version.Must(version.NewSemver(consts.VaultVersion113))
+	VaultVersion1138 = version.Must(version.NewSemver(consts.VaultVersion1138))
```
vijayavelsekar: As we are not using it anymore, we can remove it.

internal/consts/consts.go
```diff
@@ -949,6 +950,7 @@ const (
 	VaultVersion111  = "1.11.0"
 	VaultVersion112  = "1.12.0"
 	VaultVersion113  = "1.13.0"
+	VaultVersion1138 = "1.13.8"
```
vijayavelsekar: As we are not using it anymore, we can remove it.

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/base"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/client"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/errutil"
+	"github.com/hashicorp/vault/api"
+)
+
+const (
+	// Singleton resource ID - uses the full path for consistency with Vault API
+	configGroupPolicyApplicationID = "/sys/config/group-policy-application"
```
vijayavelsekar: Instead of duplicating this in path(), we can have either constant or path()

itsjiyapatel: added ConfigGroupPolicyApplicationPath constant

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,307 @@
...
+	namespace := data.Namespace.ValueString()
+
+	// Get Vault client
+	vaultClient, ok := r.getClientForNamespace(ctx, namespace, &resp.Diagnostics)
+	if !ok {
+		return
+	}
+
+	// Reset to default value instead of deleting
+	if !r.writeGroupPolicyConfig(ctx, vaultClient, modeWithinNamespaceHierarchy, &resp.Diagnostics, errutil.VaultUpdateErr) {
```
sanvikam79: Wrong error function passed to writeGroupPolicyConfig in Delete()
errutil.VaultUpdateErr is passed here, but this is the Delete method

sanvikam79 (changes requested):

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,307 @@
...
+			consts.FieldID: schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The resource ID (always \"sys/config/group-policy-application\").",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			consts.FieldGroupPolicyApplicationMode: schema.StringAttribute{
+				Required: true,
+				MarkdownDescription: "Mode for group policy application. Must be either \"within_namespace_hierarchy\" or \"any\". " +
```
sanvikam79, line 81: Do we need to wire the constants to a schema validator something like this
 Validators: []validator.String{
        stringvalidator.OneOf(modeWithinNamespaceHierarchy, modeAny),
    },
So error caught at Plan time

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,240 @@
...
+func TestAccConfigGroupPolicyApplication_invalidMode(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccConfigGroupPolicyApplicationConfig_basic("invalid_mode"),
+				ExpectError: regexp.MustCompile(`group_policy_application_mode must be either`),
```
sanvikam79: If stringvalidator.OneOf(modeWithinNamespaceHierarchy, modeAny) to the schema attribute is not added and rely on vault response for error then can we  loosen the regexp to match something 
ExpectError: regexp.MustCompile(`group_policy_application_mode`)

internal/vault/sys/config/group_policy_application.go
```diff
@@ -0,0 +1,307 @@
...
+					"\"within_namespace_hierarchy\" means policies only apply when the token authorizing a request was created in the same namespace as the group, or a descendant namespace. " +
+					"\"any\" means group policies apply to all members of a group, regardless of what namespace the request token came from.",
+			},
+		},
+		MarkdownDescription: "Manages the global group policy application mode for Vault Enterprise. " +
+			"This resource controls how policies attached to identity groups are applied across namespace boundaries. " +
```
sanvikam79: Should we specify this resource is mainly for ACL policies rather than just specify policies as per the vault-api doc and should it also mention "will be replicated between primary and secondaries — primaries cannot have a different policy application mode to secondaries" as per vault api-doc

internal/vault/sys/config/group_policy_application_test.go
```diff
@@ -0,0 +1,240 @@
...
+				ImportState:   true,
+				ImportStateId: "invalid-id",
+				ExpectError:   regexp.MustCompile(fmt.Sprintf(`Import ID must be "%s"`, config.ConfigGroupPolicyApplicationPath)),
+			},
+		},
+	})
+}
+
+// TestAccConfigGroupPolicyApplication_modeAny tests the "any" mode specifically
+func TestAccConfigGroupPolicyApplication_modeAny(t *testing.T) {
```
sanvikam79: This test case is redundant with TestAccConfigGroupPolicyApplication_update can we move the consts.FieldID check into the any step inside _update ?

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Attributes Reference
+
+In addition to the arguments above, the following attributes are exported:
+
+* `id` - The resource ID (always `"/sys/config/group-policy-application"`).
+
+## Import
+
+The group policy application configuration can be imported using the path `/sys/config/group-policy-application`:
```
sanvikam79: The import command shows /sys/config/group-policy-application with a leading /, but ConfigGroupPolicyApplicationPath = "sys/config/group-policy-application has no leading slash.
with /sys/config/group-policy-application import might fail.

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,103 @@
...
+* `namespace` - (Optional) Target namespace. Must be root (`""`) or administrative (`"admin"`) namespace. If omitted, the provider's configured namespace is used.
+  Set this to `""` to explicitly target the root namespace. The value should not contain leading or trailing forward slashes.
+  *Available only for Vault Enterprise*.
+
+## Attributes Reference
+
+In addition to the arguments above, the following attributes are exported:
+
+* `id` - The resource ID (always `"/sys/config/group-policy-application"`).
+
```
sanvikam79, line 74: same comment having leading "/" unlike in the code

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,103 @@
...
+  group_policy_application_mode = "within_namespace_hierarchy"
+  namespace                     = "admin"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `group_policy_application_mode` - (Optional) Mode for group policy application. Must be either `within_namespace_hierarchy` or `any`. Defaults to `within_namespace_hierarchy`.
```
sanvikam79: this field is "Required" in code but here it is mentioned as "Optional"

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,103 @@
+---
+layout: "vault"
+page_title: "Vault: vault_config_group_policy_application resource"
+sidebar_current: "docs-vault-resource-config-group-policy-application"
+description: |-
+  Manages the global group policy application mode in Vault Enterprise.
+---
+
+# vault\_config\_group\_policy\_application
+
```
sanvikam79, line 10: Should we mention these in doc as per vault-api doc

1. The group policy application mode only applies to ACL policies and no longer affects Sentinel RGPs for Vault ≥ 1.13.8, 1.14.4, 1.15.0.

2. This configuration will be replicated between primary and secondaries — primaries cannot have a different mode to secondaries.

website/docs/r/config_group_policy_application.html.md
itsjiyapatel: Changed the import command to sys/config/group-policy-application

internal/vault/sys/config/group_policy_application.go
itsjiyapatel: Changed the error function to `errutil.VaultDeleteErr`

internal/vault/sys/config/group_policy_application_test.go
itsjiyapatel: Removed the test case and added `resource.TestCheckResourceAttr(resourceName, consts.FieldID, config.ConfigGroupPolicyApplicationPath),` to `TestAccConfigGroupPolicyApplication_update` test

internal/vault/sys/config/group_policy_application_test.go
itsjiyapatel: Changed the ExpectError to `regexp.MustCompile(group_policy_application_mode)`

internal/vault/sys/config/group_policy_application.go
itsjiyapatel: Edited the description to include the same.

internal/vault/sys/config/group_policy_application.go
itsjiyapatel: The validation is handled by the Vault API itself

website/docs/r/config_group_policy_application.html.md
```diff
@@ -0,0 +1,111 @@
...
+In addition to the arguments above, the following attributes are exported:
+
+* `id` - The resource ID (always `"sys/config/group-policy-application"`).
+
+## Import
+
+The group policy application configuration can be imported using the path `sys/config/group-policy-application`:
+
+```
+$ terraform import vault_config_group_policy_application.config /sys/config/group-policy-application
```
sanvikam79: do we need to remove the leading "/" in the command as well?

itsjiyapatel: Yes the import id is sys/config/group-policy-application

sanvikam79 (approved):

PR #2707, opened 2025-12-16
VAULT-37696: vault_plugin support for enterprise plugins

chpag:
### Description

This PR allow to use vault_plugin ressource for Vault Enterprise Plugin
As the sha256 should not be defined for enterprise plugin,  it has been changed from mandatory to Optional
command attributes has also been marked as Optional as Enterprise plugin ignore the value
An additional check has been added in CustomizeDiff for
   - if enterprise plugin (plugin version ending with "+ent")
      - sha256 should be empty
      - command should be empty
   - else (not enterprise plugin)
      - sha256 should not be empty
      - command should not be empty

pluginIDRegex Regular Expression has been updated to allow "+" character in the version to allow "+ent"

when plugin is Enterprise, exclude sha256 from the attribute retrieved by pluginRead(), so there is no mismatch between what is defined in the hcl code and what is put on the state file

Update also build.yml to add Enteprise plugin download and ENV variables set

Closes #2518

### Checklist
- [X] Added [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md) entry (only for user-facing changes)
- [X] Acceptance tests where run against all supported Vault Versions

### Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

- [X] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [X] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

PR #2865, opened 2026-04-02
Add support of OS Secrets Engine

siyer-corp:
### Description

This PR adds Terraform provider support for Vault’s OS Secrets Engine, including backend, host, and account resources. It covers the core configuration flow for mounting the engine, registering SSH targets, and managing rotated account credentials.

- Adds `vault_os_secret_backend` to manage OS Secrets Engine backend configuration, including settings such as ssh_host_key_trust_on_first_use and backend-level config updates.

- Adds `vault_os_secret_backend_host` to manage target hosts under the OS Secrets Engine, including address, port, and host-specific rotation defaults supported by Vault.

- Adds `vault_os_secret_backend_account` to manage accounts on registered hosts, including onboarding credentials, optional rotation settings, and readback of rotation metadata such as last_vault_rotation, next_vault_rotation.

### Checklist

```

...
```
image

### Community Note

## PCI review checklist


internal/vault/secrets/os/os_secret_backend_resource.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+func (r *OSSecretBackendResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_os_secret_backend"
+}
+
+func (r *OSSecretBackendResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldPath: schema.StringAttribute{
```
vijayavelsekar: This has to be `mount` field.

website/docs/r/os_secret_backend.html.md
```diff
@@ -0,0 +1,80 @@
...
+### Advanced Configuration
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os-prod"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path                             = vault_mount.os.path
```
vijayavelsekar: Same here. This has to be renamed as `mount` through out this document.

internal/vault/secrets/os/os_secret_backend_resource_test.go
```diff
@@ -0,0 +1,280 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccOSSecretBackendConfig_basic(path),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldPath, path),
+					resource.TestCheckResourceAttr(resourceName, "max_versions", "5"),
+					resource.TestCheckResourceAttr(resourceName, "ssh_host_key_trust_on_first_use", "true"),
+				),
+			},
+			// TODO: Fix import test - currently fails because resource is destroyed between steps
```
vijayavelsekar: Is this issue fixed ? are we good to enable import step now ?

internal/vault/secrets/os/os_secret_backend_account_resource.go
```diff
@@ -0,0 +1,509 @@
...
+	_, err = cli.Logical().WriteWithContext(ctx, path, requestData)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Error writing OS secret backend account to %q", path),
+			err.Error(),
+		)
+		return
+	}
+
+	// Set the ID
```
vijayavelsekar: Do we need this comment ?

internal/vault/secrets/os/os_secret_backend_account_resource.go
```diff
@@ -0,0 +1,509 @@
...
+		return
+	}
+
+	mount := data.Mount.ValueString()
+	host := data.Host.ValueString()
+	name := data.Name.ValueString()
+	path := fmt.Sprintf("%s/hosts/%s/accounts/%s", mount, host, name)
+
+	// Build the request data
+	requestData := make(map[string]interface{})
```
vijayavelsekar, line 248: I see that we can reuse the requestData population and request to vault in Create and Update function.

internal/vault/secrets/os/os_secret_backend_account_resource.go
```diff
@@ -0,0 +1,509 @@
...
+	_, err = cli.Logical().DeleteWithContext(ctx, path)
+	if err != nil && !util.Is404(err) {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Error deleting OS secret backend account at %q", path),
+			err.Error(),
+		)
+	}
+}
+
+func (r *OSSecretBackendAccountResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
vijayavelsekar, line 454: This ImportState implementation differs from our usual approach. Can we align it with the standard pattern so the code remains concise and consistent?

Also, since the import operation will automatically invoke the Read function, we may not need to call it explicitly here. In fact, doing so could result in the Read being executed twice—once within this function and again after it completes.

Lastly, we should ensure that this works correctly for resources within a namespace as well.

internal/vault/secrets/os/os_secret_backend_account_resource_test.go
```diff
@@ -0,0 +1,539 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldMount, mount),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldHost, hostName),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldName, accountName),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldUsername, username),
+					resource.TestCheckResourceAttr(resourceName, "rotation_period", "172800"),
+					resource.TestCheckNoResourceAttr(resourceName, "rotation_window"),
+					resource.TestCheckResourceAttr(resourceName, consts.FieldVerifyConnection, "false"),
+				),
+			},
```
vijayavelsekar, line 81: we can have some import test step as well for coverage within ImportState function.

internal/vault/secrets/os/os_secret_backend_account_resource.go
```diff
@@ -0,0 +1,509 @@
+// Copyright IBM Corp. 2016, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package os
+
+import (
+	"context"
+	"fmt"
```
vijayavelsekar, line 8: It would be better to use `tflog`

internal/vault/secrets/os/os_secret_backend_host_resource.go
```diff
@@ -0,0 +1,406 @@
+// Copyright IBM Corp. 2016, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package os
+
+import (
+	"context"
+	"fmt"
```
vijayavelsekar, line 8: Good to have `tflog`

internal/vault/secrets/os/os_secret_backend_host_resource.go
```diff
@@ -0,0 +1,406 @@
...
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	mount := data.Mount.ValueString()
+	name := data.Name.ValueString()
+	path := fmt.Sprintf("%s/hosts/%s", mount, name)
+
+	// Build the request data
+	requestData := make(map[string]interface{})
```
vijayavelsekar: Can we reuse requestData poplulation and writing to vault be reused between this function and Create ?

internal/vault/secrets/os/os_secret_backend_host_resource.go
```diff
@@ -0,0 +1,406 @@
...
+	data.Mount = types.StringValue(mount)
+	data.Name = types.StringValue(name)
+
+	cli, err := client.GetClient(ctx, r.Meta(), data.Namespace.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.ClientConfigureErr(err))
+		return
+	}
+
+	r.readHostFromVault(ctx, cli, &data, &resp.Diagnostics)
```
vijayavelsekar: Should we manually call the readHostFromVault ? because after this function, it will be called anyway. Can we try our regular approach ?

internal/vault/secrets/os/os_secret_backend_resource.go
```diff
@@ -0,0 +1,338 @@
...
+				Computed:            true,
+				Default:             booldefault.StaticBool(false),
+			},
+		},
+		MarkdownDescription: "Manages the configuration of an existing OS Secrets Engine mount in Vault.",
+	}
+	base.MustAddBaseSchema(&resp.Schema)
+}
+
+func (r *OSSecretBackendResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
vijayavelsekar, line 159: Opportunities for reusing code for read and write operations.

internal/vault/secrets/os/os_secret_backend_resource.go
```diff
@@ -0,0 +1,338 @@
...
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r *OSSecretBackendResource) Delete(_ context.Context, _ resource.DeleteRequest, _ *resource.DeleteResponse) {
+}
+
+func (r *OSSecretBackendResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Use the built-in helper to set the path attribute from the import ID
+	resource.ImportStatePassthroughID(ctx, path.Root(consts.FieldPath), req, resp)
```
vijayavelsekar: Does this implementation work for resources created inside a namespace ?

website/docs/r/os_secret_backend.html.md
```diff
@@ -0,0 +1,80 @@
...
+* `max_versions` - (Optional, Computed) The maximum number of versions to keep. When omitted, Vault applies its server-side default of `10`. If you later remove the field from configuration, Vault retains the current value. Set to `0` to explicitly store zero in Vault.
+
+* `ssh_host_key_trust_on_first_use` - (Optional) If `true`, SSH host keys will be trusted on first use (TOFU). If `false`, host keys must be explicitly configured. Defaults to `false`.
+
+## Import
+
+OS Secret backend can be imported using the `path`, e.g.
+
+```
+$ terraform import vault_os_secret_backend.os os
```
vijayavelsekar, line 94: we need to check if this works with TERRAFORM_VAULT_NAMESPACE_IMPORT

website/docs/r/os_secret_backend_account.html.md
```diff
@@ -0,0 +1,193 @@
...
+### Account with Rotation Configuration
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: This one also has to be `mount` and has to be fixed through out all the examples

website/docs/r/os_secret_backend_host.html.md
```diff
@@ -0,0 +1,146 @@
...
+### Basic Host Configuration
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: Here as well `mount` and in all the examples

internal/vault/secrets/os/os_secret_backend_account_resource_test.go
siyer-corp: ImportState is being covered in all resources

website/docs/r/os_secret_backend.html.md
siyer-corp: This is tested manually

website/docs/r/os_secret_backend_account.html.md
```diff
@@ -0,0 +1,193 @@
...
+### Basic Account Configuration
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: This field has to be `mount`.

website/docs/r/os_secret_backend_account.html.md
vijayavelsekar: This field has to be `mount`.

website/docs/r/os_secret_backend_account.html.md
```diff
@@ -0,0 +1,193 @@
...
+### Account with Scheduled Rotation
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: This field has to be `mount`.

website/docs/r/os_secret_backend_host.html.md
vijayavelsekar: This field has to be `mount`.

website/docs/r/os_secret_backend_host.html.md
```diff
@@ -0,0 +1,146 @@
...
+### Advanced Host Configuration
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: This field has to be `mount`.

website/docs/r/os_secret_backend_host.html.md
```diff
@@ -0,0 +1,146 @@
...
+### Host with SSH Host Key
+
+```hcl
+resource "vault_mount" "os" {
+  path = "os"
+  type = "vault-plugin-secrets-os"
+}
+
+resource "vault_os_secret_backend" "os" {
+  path = vault_mount.os.path
```
vijayavelsekar: This field has to be `mount`.

internal/vault/secrets/os/os_secret_backend_account_resource.go
```diff
@@ -0,0 +1,514 @@
...
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldHost), host)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldName), name)...)
+
+	// Password must be set after import since it's write-only
+	// Set it to a placeholder that will force user to update
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldPasswordWO), "IMPORT_PLACEHOLDER_UPDATE_REQUIRED")...)
+
+	ns := os.Getenv(consts.EnvVarVaultNamespaceImport)
+	if ns != "" {
+		tflog.Info(ctx,
```
vijayavelsekar, line 471: nit: good to have tflog - Info/Debug in other lifecycle methods as well.

internal/vault/secrets/os/os_secret_backend_host_resource.go
```diff
@@ -0,0 +1,385 @@
...
+		)
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldMount), mount)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldName), name)...)
+
+	ns := os.Getenv(consts.EnvVarVaultNamespaceImport)
+	if ns != "" {
+		tflog.Info(ctx,
```
vijayavelsekar, line 409: nit: good to have tflog - Info/Debug in other lifecycle methods as well.

internal/vault/secrets/os/os_secret_backend_resource.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func (r *OSSecretBackendResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	mount := req.ID
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(consts.FieldMount), mount)...)
+
+	ns := os.Getenv(consts.EnvVarVaultNamespaceImport)
+	if ns != "" {
+		tflog.Info(ctx,
```
vijayavelsekar, line 307: nit: good to have tflog - Info/Debug in other lifecycle methods as well.

internal/vault/secrets/os/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"regexp"
+)
+
+var (
+	// hostIDRe matches: {mount}/hosts/{name}
+	hostIDRe = regexp.MustCompile(`^([^/]+)/hosts/([^/]+)$`)
+	// accountIDRe matches: {mount}/hosts/{host}/accounts/{name}
+	accountIDRe = regexp.MustCompile(`^([^/]+)/hosts/([^/]+)/accounts/([^/]+)$`)
```
kpcraig: We're guilty of writing a regex for this on the plugin side too, but in addition to being different to the one in vault (e.g., https://github.com/hashicorp/vault/blob/f1fc9d3fc0175ea93d821564329accd6c8f84b27/sdk/framework/path.go#L20 and https://github.com/hashicorp/vault-plugin-secrets-os/blob/3e2e9afbc7697f79be6d7c56b335b278b8a95b14/path_accounts_rotate.go#L22) do we have any tests to convince ourselves we're capturing what we expect in all cases?

Also, is this the usual paradigm for extracting this information? (It might be, I'm not a TFVP expert)

kpcraig: @vinay-gopalan might have some insight here?

siyer-corp: @kpcraig Thanks for the callout. I see the parsing on Vault side is stricter than `[^/]+.` I notice other resources use more lenient patterns ([example](https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_database_secret_backend_static_role.go#L20)).

We could update the host and account name parsing to match Vault's GenericNameRegex pattern `(\w(?:(?:[\w-.]+)?\w)?)` for consistency. This would catch invalid names during import before calling Vault's API, providing clearer error messages. The mount path validation would remain unchanged?

I can update the pattern and add test coverage if you think this alignment makes sense.

vinay-gopalan: I took a closer look at the logic. I wanted to see if it was possible to avoid regexes to get the relevant info during imports. It looks like we will still need the regexes for the account and host resources. We won't need one for the backend resource, as we can just use the mount (currently being done in your code so that's good)

Since we can't avoid the regexes, I would lean towards Kay's suggestion, to map these regexes as close to Vault as possible. We have used lenient patterns in the past like you mentioned above, but that has lead to edge case bugs where users input weird values that can break our flows. I would say mirroring as close to Vault and adding some test coverage on the regexes should give us confidence.

On a side note, what is the idea behind setting the write-only field to state on an [import here](https://github.com/hashicorp/terraform-provider-vault/pull/2865/changes#diff-a89fac664904e42a77c60c5104dfa735b788e208e31f3cfa2a691cbc0c08f268R471)? Write-only fields shouldn't ever be written to state, even when placeholders as that is not what they are meant for. Is this line required?

siyer-corp: Thanks @vinay-gopalan for the review! I've updated the regex to closely match with Vault. Great catch on the Write-only field in the import, that was an inadvertent AI addition. That's removed now. Let me know if you have any other concerns.

.github/workflows/build.yml
```diff
@@ -301,6 +322,23 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-golang-
 
+      - name: Register OS plugin in Vault
```
fairclothjm, line 325: We are going to need this PR to be merged before we can ship the OS Secrets engine resources to production: https://github.com/hashicorp/terraform-provider-vault/pull/2707

That said, should we prioritize that PR so it can be used in this one instead of this step?

vijayavelsekar (approved):
LGTM

PR #2877, opened 2026-04-13
Add support for ephemeral vault_token resource with type and entity_alias fields

Siddharth-Kannan:
### Description

Created new ephemeral resource for resource_token and added support for type and entity_alias fields.
Added acceptance tests for the ephemeral resource.
Added a deprecation message for the sdkv2 implementation of the token resource. 

Performed testing across vault versions 1.15.0-1.21.0

### Checklist

```
$ terraform-provider-vault % go test -v -count=1 -run TestAccToken ./internal/vault/auth/ephemeral/
=== RUN   TestAccToken_basic
--- PASS: TestAccToken_basic (2.83s)
=== RUN   TestAccToken_batch
--- PASS: TestAccToken_batch (1.31s)
=== RUN   TestAccToken_withPolicies
--- PASS: TestAccToken_withPolicies (3.09s)
=== RUN   TestAccToken_orphan
--- PASS: TestAccToken_orphan (2.97s)
=== RUN   TestAccToken_withRole
--- PASS: TestAccToken_withRole (3.64s)
=== RUN   TestAccToken_wrapped
--- PASS: TestAccToken_wrapped (3.97s)
=== RUN   TestAccToken_wrappedBatch
--- PASS: TestAccToken_wrappedBatch (2.58s)
=== RUN   TestAccToken_withEntityAlias
--- PASS: TestAccToken_withEntityAlias (4.13s)
=== RUN   TestAccToken_batchTokenAutoDetectionViaRole
--- PASS: TestAccToken_batchTokenAutoDetectionViaRole (2.11s)
=== RUN   TestAccToken_full
--- PASS: TestAccToken_full (2.88s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/internal/vault/auth/ephemeral     30.430s
...
```

### Community Note

## PCI review checklist


CHANGELOG.md
```diff
@@ -9,11 +9,13 @@ FEATURES:
...
 * **New Resource**: `vault_plugin_runtime` for managing plugin runtimes in Vault's plugin runtimes catalog. Requires Vault 1.15 or later.([#2835](https://github.com/hashicorp/terraform-provider-vault/pull/2835/))
+* **New Ephemeral Resource**: `vault_token` for creating Vault tokens with automatic revocation. Supports service and batch tokens, entity alias association which were not supported in the SDKv2 resource. ([#2877](https://github.com/hashicorp/terraform-provider-vault/pull/2877))
 
 IMPROVEMENTS:
 
 * `vault_pki_secret_backend_config_acme`: Added new fields that control the PKI ACME challenge worker IP ranges that they can connect. ([#2839]https://github.com/hashicorp/terraform-provider-vault/pull/2839)
 * Add support for metadata fields in `azure_access_credentials` and `resource_azure_secret_backend_role` resources. ([#2734](https://github.com/hashicorp/terraform-provider-vault/pull/2734)
+* `resource/vault_token`: Added deprecation warning to guide users toward the new ephemeral `vault_token` resource for better security and batch token support.
```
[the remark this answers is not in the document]
Siddharth-Kannan: Looking at the changeling across other instances, I see that it is referenced as resource/.... . therefore I think it's still fine to use the existing changeling comment itself.

internal/vault/auth/ephemeral/token.go
```diff
@@ -0,0 +1,514 @@
...
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// Schema defines this resource's schema which is the data that is available in
+// the resource's configuration, plan, and state
+func (r *TokenEphemeralResource) Schema(_ context.Context, _ ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldID: schema.StringAttribute{
+				MarkdownDescription: "The ID of the client token. Can only be specified by a root token. The ID provided may not contain a '.' character and should not start with the 's.' prefix.",
```
siyer-corp: Would be good to add that this is not the resource identifier, and that it's an input field.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,514 @@
...
+			return
+		}
+		data.ClientToken = types.StringValue(tokenResp.Auth.ClientToken)
+		accessor = tokenResp.Auth.Accessor
+		leaseID = tokenResp.LeaseID
+		data.LeaseDuration = types.Int64Value(int64(tokenResp.Auth.LeaseDuration))
+
+		// Handle batch tokens (no accessor)
+		if accessor == "" && tokenResp.Auth.ClientToken != "" && strings.HasPrefix(tokenResp.Auth.ClientToken, "hvb.") {
+			accessor = tokenResp.RequestID
```
siyer-corp: Can we document why we're using RequestID here?

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,514 @@
...
+		data.Accessor = types.StringValue(accessor)
+	}
+	if leaseID != "" {
+		data.LeaseID = types.StringValue(leaseID)
+	}
+
+	// Set additional computed fields from auth response
+	if !wrapped && tokenResp.Auth != nil {
+		// Token policies
+		if len(tokenResp.Auth.TokenPolicies) > 0 {
```
siyer-corp, line 410: Should we add an else block here?

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,514 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package ephemeralauth
```
siyer-corp, line 4: We're missing the docs for this resource.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: No, we don't need to add an else block. After verifying the Vault response model, I confirmed that wrapped token responses do not contain an auth object, and token_policies, entity_id, or orphan fields. These fields are marked as Computed: true in the schema, which means they default to null values when not explicitly set. Adding an else block to explicitly set them to null would be redundant.

Additionally, I've added test coverage to verify this behaviour - the tests confirm that these fields are properly null for wrapped tokens without requiring explicit initialisation.

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,685 @@
...
+
+provider "echo" {
+  data = ephemeral.vault_token.test
+}
+
+resource "echo" "test" {}
+`, policyName)
+}
+
+func testTokenConfig_withMetadata() string {
```
siyer-corp: This config seems unused. Is there supposed to be a test around this?

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,685 @@
...
+
+provider "echo" {
+  data = ephemeral.vault_token.test
+}
+
+resource "echo" "test" {}
+`
+}
+
+func testTokenConfig_withNumUses() string {
```
siyer-corp: This config seems unused. Is there supposed to be a test around this?

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,685 @@
...
+
+provider "echo" {
+  data = ephemeral.vault_token.test
+}
+
+resource "echo" "test" {}
+`, roleName)
+}
+
+func testTokenConfig_periodic() string {
```
siyer-corp: This config seems unused. Is there supposed to be a test around this?

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,685 @@
...
+
+provider "echo" {
+  data = ephemeral.vault_token.test
+}
+
+resource "echo" "test" {}
+`, policyName)
+}
+
+func testTokenConfig_withCustomID(customID string) string {
```
siyer-corp: This config seems unused. Is there supposed to be a test around this?

website/docs/ephemeral-resources/token_ephemeral_resource.html.md
```diff
@@ -0,0 +1,238 @@
...
+
+```hcl
+resource "vault_identity_entity" "app" {
+  name     = "app-entity"
+  policies = ["app-policy"]
+}
+
+resource "vault_identity_entity_alias" "app_alias" {
+  name           = "app-alias"
+  mount_accessor = vault_auth_backend.token.accessor
```
siyer-corp: Should we have a `vault_auth_backend` that's being referenced here?

website/docs/ephemeral-resources/token_ephemeral_resource.html.md
```diff
@@ -0,0 +1,238 @@
...
+}
+
+ephemeral "vault_token" "with_entity" {
+  entity_alias = vault_identity_entity_alias.app_alias.name
+  policies     = ["app-policy"]
+  ttl          = "1h"
+}
+```
+
+## Argument Reference
```
siyer-corp, line 146: Should we document `mount_id` as well which is part of the Ephemeral Schema

internal/vault/auth/ephemeral/token_ephemeral_resource.go
siyer-corp: Does vault return null or an empty slice for non-wrapped token policies?

website/docs/ephemeral-resources/token_ephemeral_resource.html.md
```diff
@@ -0,0 +1,266 @@
...
+# Token role must allow the entity alias
+resource "vault_token_auth_backend_role" "app_role" {
+  role_name              = "app-role"
+  allowed_policies       = ["app-policy"]
+  allowed_entity_aliases = [vault_identity_entity_alias.app_alias.name]
+}
+
+# Step 2: After all resources exist, use them with an ephemeral token
+ephemeral "vault_token" "with_entity" {
+  role_name    = vault_token_auth_backend_role.app_role.role_name
```
vijayavelsekar, line 137: Instead of executing them in two different steps can we do it in single shot by using mount_id to create a dependency ?

website/docs/ephemeral-resources/token_ephemeral_resource.html.md
```diff
@@ -0,0 +1,266 @@
...
+  role_name              = "app-role"
+  allowed_policies       = ["app-policy"]
+  orphan                 = true
+  token_period           = "86400"
+  renewable              = true
+  token_explicit_max_ttl = "115200"
+}
+
+# Step 2: After the role exists, use it with an ephemeral token
+ephemeral "vault_token" "app_token" {
```
vijayavelsekar, line 50: Instead of executing them in two different steps can we do it in single shot by using mount_id to create a dependency ?

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,619 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	"github.com/hashicorp/terraform-provider-vault/internal/providertest"
+)
+
+// TestAccToken_basic confirms that a basic service token can be created
+//
+// Uses the Echo Provider to test values set in ephemeral resources
+// see documentation here for more details:
+// https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/ephemeral-resources#using-echo-provider-in-acceptance-tests
+func TestAccToken_basic(t *testing.T) {
```
vijayavelsekar, line 28: can we have a test case to create token inside a namespace ?

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,517 @@
...
+
+	// Read Terraform configuration into the model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Set default values only if null or known empty string (not if unknown)
+	if data.DisplayName.IsNull() || (!data.DisplayName.IsUnknown() && data.DisplayName.ValueString() == "") {
+		data.DisplayName = types.StringValue("token")
```
vijayavelsekar: Instead of this can we just simply make DisplayName with default "token" in the schema ?

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+	if data.Type.IsNull() || (!data.Type.IsUnknown() && data.Type.ValueString() == "") {
+		data.Type = types.StringValue("service")
```
vijayavelsekar: Not sure if this is right.
Because, document says `Defaults to the type specified by the role configuration named by role_name`.
We need to check below scenarios.

**Scenario 1:** (If role_name is provided)
If type is not set, does it takes type specified by the role configuration as per the doc ?

**Scenario 2:** (If role_name is not set)
If type is not set, what type its taking in vault ? we need to check this via vault cli

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	resp.Private.SetKey(ctx, "private_data", privateBytes)
+
+	// Set the result
+	resp.Diagnostics.Append(resp.Result.Set(ctx, &data)...)
+}
+
+// Close revokes the token when it's no longer needed
+func (r *TokenEphemeralResource) Close(ctx context.Context, req ephemeral.CloseRequest, resp *ephemeral.CloseResponse) {
```
vijayavelsekar, line 456: Question: If we have a wrapped token, should we revoke the parent token along with the wrapped token ? are we handling that ?

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: Ephemeral resources use ephemeral/schema which does NOT have a Default field or default modifiers, I can change the explicitly put strings into constants instead.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: I have tested these cases via vault-CLI, in scenario 1: the type specified by the role is taken, and in scenario 2: the type is automatically set to service. 
Note: Vault does not return the `type` field in the response structure, We are using the token prefix (`hvs/hvb`) and computing this field.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: No, we don't revoke the wrapping token - only the wrapped (inner) token. The wrapping token is single-use and consumed when unwrapped, so it doesn't need explicit revocation. The code correctly stores and revokes the wrapped_accessor , which is the accessor of the actual token that was created.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: I wasn't able to create a token with no policies, if we dont give any policy it always inherits the creating token's policies, even when i specified no-parent. and root token has the root policy automatically so thats passed down as well. So i couldn't verify if vault would return null or an empty slice, but i don't think a token can exist without any policy.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
```diff
@@ -0,0 +1,517 @@
...
+		accessor = tokenResp.Auth.Accessor
+		leaseID = tokenResp.LeaseID
+		data.LeaseDuration = types.Int64Value(int64(tokenResp.Auth.LeaseDuration))
+
+		// Handle batch tokens (no accessor)
+		// Batch tokens don't have an accessor field in the Auth response, but we use
+		// the RequestID as a unique identifier for tracking purposes. This is not used
+		// for revocation since batch tokens cannot be revoked via API.
+		if accessor == "" && tokenResp.Auth.ClientToken != "" && strings.HasPrefix(tokenResp.Auth.ClientToken, "hvb.") {
+			accessor = tokenResp.RequestID
```
vijayavelsekar: Do we really need this ? Because accessor is of no use in Batch token. Also we are not going to track this anywhere as they are ephemeral.

Also do we really need this much granular checks for clientToken ? is there any specific reason for this check ?
`strings.HasPrefix(tokenResp.Auth.ClientToken, "hvb.")`

internal/vault/auth/ephemeral/token_ephemeral_resource_test.go
```diff
@@ -0,0 +1,655 @@
...
+
+provider "echo" {
+  data = ephemeral.vault_token.test
+}
+
+resource "echo" "test" {}
+`
+}
+
+func testTokenConfig_roleOnly(roleName string) string {
```
vijayavelsekar: We are good to remove this as we are not using it.

internal/vault/auth/ephemeral/token_ephemeral_resource.go
Siddharth-Kannan: As pointed out, it doesn't need to have accessor, this was added for the original SDKv2 resource, and I had assumed that it would still be needed to track batch tokens, but upon testing without it, I verified that its not needed. Also the hvb. test could also be simplified to use a simpler check to see if the accessor is empty.

vijayavelsekar (approved):
LGTM

PR #2885, opened 2026-04-19
TFVP: implement support for Agent Registry

Chris-Paris-HashiCorp:
### Description

Jira: https://hashicorp.atlassian.net/browse/VAULT-44159

### Checklist

```

...
```

### Community Note

## PCI review checklist


Chris-Paris-HashiCorp:
# Agent Registration Resource Implementation

## Overview

This document describes the implementation of the `vault_agent_registration` resource for the Terraform Vault Provider. This resource enables management of agent registrations in Vault Enterprise, allowing users to register Vault agents with specific identity entities and configure ceiling policies that limit the maximum permissions an agent can obtain.

## Feature Summary

**Resource Name**: `vault_agent_registration`
**Vault Version Required**: 2.0.0+
**Vault Edition**: Enterprise only
**API Endpoints**:
- `agent-registry/register` (create)
- `agent-registry/registration/display-name/{name}` (read by name, import)
- `agent-registry/registration/id/{id}` (read/update/delete by ID)

## Implementation Details

### Files Added

1. **`internal/vault/sys/agent_registration_resource.go`** (486 lines)
   - Main resource implementation using Terraform Plugin Framework
   - Implements full CRUD lifecycle (Create, Read, Update, Delete)
   - Includes import functionality
   - Handles namespace support for Vault Enterprise

2. **`internal/vault/sys/agent_registration_resource_test.go`** (467 lines)
   - Comprehensive acceptance test suite
   - Tests cover: basic CRUD, import, policies, namespaces, updates
   - 8 test scenarios with multiple test steps each

3. **`website/docs/r/agent_registration.html.md`** (186 lines)
   - Complete user-facing documentation
   - Multiple usage examples (basic, with policies, namespaces, etc.)
   - Full argument and attribute reference
   - Import instructions and usage notes

### Files Modified

1. **`internal/consts/consts.go`**
   - Added 4 new field constants for agent registration:
     - `FieldCeilingPolicyIdentifiers`
     - `FieldNoDefaultCeilingPolicy`
     - `FieldCreationTime`
     - `FieldLastUpdatedTime`

2. **`internal/provider/fwprovider/provider.go`**
   - Registered `NewAgentRegistrationResource` in the provider's Resources() method

3. **`CHANGELOG.md`**
   - Added feature entry under Unreleased > FEATURES section

## Resource Schema

### Required Fields
- `display_name` (string) - Unique identifier for the agent, requires replacement on change
- `entity_id` (string) - ID of the identity entity to associate with the agent

### Optional Fields
- `namespace` (string) - Vault Enterprise namespace
- `ceiling_policy_identifiers` (list of strings) - Policy names defining maximum permissions
- `no_default_ceiling_policy` (bool) - Opt out of default ceiling policy (default: false)
- `description` (string) - Human-readable description

### Computed Fields
- `id` (string) - Resource identifier (same as display_name)
- `creation_time` (string) - RFC3339 timestamp of creation
- `last_updated_time` (string) - RFC3339 timestamp of last update

## Key Implementation Features

### 1. Ceiling Policy Management
- Supports multiple ceiling policies per agent
- Automatically filters out default policies when reading state
- Policies define maximum permissions (ceiling) for the agent
- Agent cannot exceed ceiling even if entity/token policies grant more

### 2. Namespace Support
- Full support for Vault Enterprise namespaces
- Import format: `namespace/display_name` for namespaced resources
- Import format: `display_name` for root namespace resources

### 3. Import Functionality
- Simple import using display name
- Namespace-aware import with custom ID parsing
- Properly reconstructs state from Vault API

### 4. State Management
- Read-after-write pattern for Create and Update operations
- Proper 404 handling (removes from state)
- Computed fields updated from API responses
- Write-only field handling (none in this resource)

### 5. Plan Modifiers
- `display_name` requires replacement (immutable identifier)
- Computed fields use `UseStateForUnknown` for stability

## Test Coverage

### Test Scenarios
1. **Basic CRUD** - Create, read, update, delete with import
2. **With Policies** - Agent with ceiling policies
3. **No Default Policy** - Opt out of default ceiling policy
4. **With Description** - Description field and updates
5. **Update Policies** - Changing ceiling policies
6. **Requires Replace** - Verify display_name changes force replacement
7. **Namespace** - Agent registration in custom namespace
8. **Multiple Policies** - Agent with multiple ceiling policies

### Test Patterns
- Uses `resource.Test` with multiple steps
- Import verification with custom ID functions
- Proper PreCheck for Enterprise features
- Comprehensive attribute checking
- Dependency management (entities, policies, namespaces)

## API Integration

### Endpoints Used
- **Create**: `POST agent-registry/register`
- **Read by Name**: `GET agent-registry/registration/display-name/{display_name}` (used during import)
- **Read by ID**: `GET agent-registry/registration/id/{id}` (used after create/update)
- **Update**: `PUT agent-registry/registration/id/{id}`
- **Delete**: `DELETE agent-registry/registration/id/{id}`

### Request/Response Handling
- Uses `api.Client.Logical()` for all operations
- Proper error handling with context-aware messages
- JSON marshaling/unmarshaling for API models
- Namespace injection via client configuration

## Design Decisions

### 1. Display Name as Identifier
- Chose `display_name` over auto-generated IDs for user-friendliness
- Makes import intuitive: `terraform import vault_agent_registration.example my-agent`
- Requires replacement on change (immutable identifier)

### 2. Policy Filtering
- Default policies automatically filtered from state
- Users only see policies they explicitly configured
- Prevents drift from Vault's automatic default policy application

### 3. Separate API Model
- `AgentRegistrationModel` for Terraform state
- `AgentRegistrationAPIModel` for Vault API
- Clean separation of concerns
- Easier to handle field transformations

### 4. Framework Choice
- Uses Terraform Plugin Framework (modern approach)
- Type-safe schema definitions
- Better plan modifier support
- Improved diagnostics

## Usage Examples

### Basic Usage
```hcl
resource "vault_identity_entity" "agent" {
  name     = "my-agent-entity"
  policies = ["default"]
}

resource "vault_agent_registration" "example" {
  display_name = "my-agent"
  entity_id    = vault_identity_entity.agent.id
}
```

### With Ceiling Policies
```hcl
resource "vault_policy" "agent_ceiling" {
  name = "agent-ceiling-policy"
  policy = <<EOT
path "secret/data/*" {
  capabilities = ["read"]
}
EOT
}

resource "vault_agent_registration" "example" {
  display_name               = "my-agent"
  entity_id                  = vault_identity_entity.agent.id
  ceiling_policy_identifiers = [vault_policy.agent_ceiling.name]
}
```

## Best Practices

### For Users
1. Always create the identity entity before the agent registration
2. Use descriptive display names for easy identification
3. Configure ceiling policies to limit agent permissions
4. Use descriptions to document agent purpose
5. Consider namespace organization for multi-tenant setups

### For Developers
1. Follow TFVP resource development patterns
2. Use read-after-write for consistency
3. Implement comprehensive test coverage
4. Document all fields thoroughly
5. Handle Enterprise features gracefully

## Security Considerations

1. **Ceiling Policies**: Act as a security boundary - agents cannot exceed these permissions
2. **Entity Association**: Links agent to identity system for audit trails
3. **Namespace Isolation**: Enterprise namespaces provide tenant isolation
4. **No Sensitive Data**: Resource doesn't handle passwords or tokens directly

## Design Decision: Resource Only (No Data Source)

This implementation includes only the `vault_agent_registration` resource, not a corresponding data source. This was a deliberate design decision based on the following considerations:

### Why No Data Source?

1. **Primary Use Case**: Agent registrations are primarily created and managed by Terraform, not read from existing registrations. The typical workflow is:
   - Create identity entity
   - Register agent with that entity
   - Configure ceiling policies
   
2. **Import Covers Read Scenarios**: The resource's import functionality handles the main "read existing" use case:
   ```bash
   terraform import vault_agent_registration.example my-agent
   ```

3. **Limited Query Patterns**: The Vault API for agent registrations doesn't support list operations or complex queries. You need to know the exact `display_name` to read an agent registration, which makes a data source less useful than typical data sources that support discovery.

4. **State Management**: Agent registrations are stateful resources that Terraform should manage. Reading them as data sources could lead to confusion about ownership and lifecycle management.

5. **Scope Limitation**: This initial implementation focuses on the core management workflow. A data source can be added later if user feedback indicates a strong need.

### When a Data Source Might Be Useful

A data source could be valuable in future versions for scenarios like:
- Reading agent registrations created outside Terraform
- Discovering agents in a namespace for reporting
- Referencing agent details in other resources
- Audit and compliance workflows

If these use cases emerge, a data source can be added in a future release without breaking changes.

## Future Enhancements

Potential improvements for future versions:
1. **Data source** for reading agent registrations (if user demand emerges)
2. **List data source** for discovering all agents in a namespace (if API supports it)
3. Bulk import support for multiple agents
4. Agent status monitoring
5. Integration with agent auto-auth workflows
6. Policy template support

## References

- Vault API Documentation: `agent-registry/*` endpoints
- Terraform Plugin Framework: https://developer.hashicorp.com/terraform/plugin/framework
- TFVP Development Guide: Internal documentation
- Vault Enterprise Features: https://www.vaultproject.io/docs/enterprise

## Changelog Entry

```markdown
* **New Resource**: `vault_agent_registration` for managing agent registrations in Vault Enterprise.
  Allows registering Vault agents with specific identity entities and configuring ceiling policies
  that limit maximum agent permissions. Requires Vault 2.0.0+.
```

## Testing Instructions

### Prerequisites
- Vault Enterprise 2.0.0 or later
- Valid Vault Enterprise license
- Terraform 1.0+

### Running Tests
```bash
# Run all agent registration tests
TF_ACC=1 go test ./internal/vault/sys -v -run TestAccAgentRegistration

# Run specific test
TF_ACC=1 go test ./internal/vault/sys -v -run TestAccAgentRegistration_basic

# Run with namespace tests (requires Enterprise)
TF_ACC=1 VAULT_NAMESPACE=root go test ./internal/vault/sys -v -run TestAccAgentRegistration_namespace
```

### Manual Testing
```bash
# Initialize Terraform
terraform init

# Plan changes
terraform plan

# Apply configuration
terraform apply

# Import existing agent
terraform import vault_agent_registration.example my-agent

# Verify state
terraform show
```

## Conclusion

This implementation provides a complete, production-ready resource for managing Vault agent registrations. It follows all TFVP best practices, includes comprehensive testing, and provides clear documentation for users. The resource integrates seamlessly with Vault's identity system and policy framework to enable secure, manageable agent deployments.

website/docs/r/agent_registration.html.md
```diff
@@ -0,0 +1,187 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault/index.html#namespace).
+  *Available only for Vault Enterprise*.
+
+* `display_name` - (Required) The display name for the agent registration. This is used as the unique identifier for the agent. Changing this will force a new resource to be created.
```
davidadeleon: I think outlining the requirement that this is unique per namespace is important here. Also the wording around it being `the unique identifier for the agent` is a bit confusing. I think you mean that this is the unique identifier for the resource in Terraform and not necessarily in Vault. I think spelling that out, clears things up here.

Chris-Paris-HashiCorp: updated

jasonpilz (dismissed):
LGTM, nice job

internal/vault/sys/agent_registration_resource_test.go
```diff
@@ -0,0 +1,450 @@
...
+// TestAccAgentRegistration_basic tests the basic CRUD lifecycle of an agent registration
+func TestAccAgentRegistration_basic(t *testing.T) {
+	displayName := acctest.RandomWithPrefix("test-agent")
+	resourceName := "vault_agent_registration.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLTE(t, provider.VaultVersion200)
```
Balaji2198: The documentation states that this feature requires Vault `2.0.1` or later, but the current implementation checks for `2.0.0`. Can we align the version check with the requirement?

internal/vault/sys/agent_registration_resource_test.go
```diff
@@ -0,0 +1,450 @@
...
+func TestAccAgentRegistration_withPolicies(t *testing.T) {
+	displayName := acctest.RandomWithPrefix("test-agent")
+	policyName := acctest.RandomWithPrefix("test-policy")
+	resourceName := "vault_agent_registration.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestAccPreCheck(t)
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLTE(t, provider.VaultVersion200)
```
Balaji2198: same comment as above(`2.0.1 `version check). please update this for all the testcases.

internal/vault/sys/agent_registration_resource.go
```diff
@@ -0,0 +1,481 @@
...
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldID: schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Unique identifier for the agent registration.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			consts.FieldDisplayName: schema.StringAttribute{
```
Balaji2198, line 94: The documentation specifies that the `display_name` must not contain forward slashes (`/`), since `/` is used as a delimiter during import. However, the current schema definition does not enforce this restriction. Do we need to add explicit schema validation here, or is this constraint already handled on the Vault side?

internal/vault/sys/agent_registration_resource_test.go
```diff
@@ -0,0 +1,450 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, consts.FieldDisplayName, displayName2),
+				),
+			},
+		},
+	})
+}
+
+// TestAccAgentRegistration_namespace tests agent registration in a namespace
+func TestAccAgentRegistration_namespace(t *testing.T) {
```
Balaji2198, line 326: can we add a testcase that covers the duplicate `display_name` scenario?

internal/vault/sys/agent_registration_resource.go
```diff
@@ -0,0 +1,481 @@
...
+// https://developer.hashicorp.com/terraform/plugin/framework/resources#metadata-method
+func (r *AgentRegistrationResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_agent_registration"
+}
+
+// Schema defines this resource's schema which is the data that is available in
+// the resource's configuration, plan, and state
+//
+// https://developer.hashicorp.com/terraform/plugin/framework/resources#schema-method
+func (r *AgentRegistrationResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
```
Balaji2198, line 84: Can we add the `owner` field in the schema? this is defined in the [Vault API](https://github.com/hashicorp/web-unified-docs/pull/2485/changes) specification but is missing from the current implementation.

internal/vault/sys/agent_registration_resource.go
```diff
@@ -0,0 +1,481 @@
...
+	EntityID               types.String `tfsdk:"entity_id"`
+	CeilingPolicies        types.List   `tfsdk:"ceiling_policies"`
+	NoDefaultCeilingPolicy types.Bool   `tfsdk:"no_default_ceiling_policy"`
+	Description            types.String `tfsdk:"description"`
+	CreationTime           types.String `tfsdk:"creation_time"`
+	LastUpdatedTime        types.String `tfsdk:"last_updated_time"`
+}
+
+// AgentRegistrationAPIModel describes the Vault API data model.
+type AgentRegistrationAPIModel struct {
```
Balaji2198, line 62: Can we add the `owner` field in the API model as well?

internal/vault/sys/agent_registration_resource.go
```diff
@@ -0,0 +1,481 @@
...
+		}
+	}
+}
+
+func (r *AgentRegistrationResource) registerPath() string {
+	return "agent-registry/register"
+}
+
+func (r *AgentRegistrationResource) registrationByNamePath(name string) string {
+	return fmt.Sprintf("agent-registry/registration/display-name/%s", name)
```
Balaji2198: Can we make sure that `display_name` is URL‑encoded in the path construction so special characters are handled correctly?
```suggestion
	return fmt.Sprintf("agent-registry/registration/display-name/%s", url.PathEscape(name))
```

internal/vault/sys/agent_registration_resource.go
```diff
@@ -0,0 +1,481 @@
...
+
+	readResp, err := client.Logical().ReadWithContext(ctx, path)
+	if err != nil {
+		diags.AddError(
+			errutil.VaultReadErr(err),
+		)
+		return
+	}
+
+	if readResp == nil {
```
Balaji2198: Should we handle the nil response in Read the same way as [plugin_runtime_resource](https://github.com/hashicorp/terraform-provider-vault/blob/main/internal/vault/sys/plugin_runtime_resource.go#L207), by removing the state when data is empty?
This would ensure Terraform gracefully clears the resource instead of returning an error.

Balaji2198 (commented):
Thanks for working on this. I've added few comments.

internal/vault/sys/agent_registration_resource.go
fcallegari7: Great catch! We have a ticket to add the owner field as a follow-up next week. We decided to do that as a follow-up to not increase the scope of this PR.
https://hashicorp.atlassian.net/browse/VAULT-45188

internal/vault/sys/agent_registration_resource.go
fcallegari7: Great catch! We have a ticket to add the owner field as a follow-up next week. We decided to do that as a follow-up to not increase the scope of this PR.
https://hashicorp.atlassian.net/browse/VAULT-45188

internal/vault/sys/agent_registration_resource_test.go
fcallegari7: @Balaji2198 I checked with the team and the feature was released in vault `2.0.0`. Could you please share a link to the documentation?
@Chris-Paris-HashiCorp would you know why the documentation states the `2.0.1` requirement?

internal/vault/sys/agent_registration_resource.go
fcallegari7: I'm leaning towards adding the regex validation here to be extra safe

internal/vault/sys/agent_registration_resource_test.go
fcallegari7: @Chris-Paris-HashiCorp could you please confirm this requirement? What is the expected behavior for duplicate `display_name`?

internal/vault/sys/agent_registration_resource_test.go
Chris-Paris-HashiCorp: I recall that these Vault features missed the 2.0.0 release. We can check the new TFVP features against Vault 2.0.0 and make sure the docs match reality.

internal/vault/sys/agent_registration_resource.go
Chris-Paris-HashiCorp: I think the / restriction exists only in Terraform's interface to it. Vault itself is fine with / in display_name. Terraform (at least the TFVP part of it) uses a convention where a / in the name is a delimiter for a namespace specification. The Vault API itself has namespace specified in a different way, and as a result, / is fine in the name in Vault.

Clearly this is not ideal alignment. I'm open to ideas but I think it probably doesn't make sense to add a restriction in Vault itself just because a particular consumer of Vault (Terraform) can't handle the / character where Vault can.

internal/vault/sys/agent_registration_resource_test.go
Chris-Paris-HashiCorp: We can add that test case. The expected result of trying to create a new registration record with a display_name already used by another registration record in the namespace is an error from Vault. So we can write a test that expects the error.

internal/vault/sys/agent_registration_resource_test.go
fcallegari7: Thank you for confirming it Chris. I've added a VaultVersion201 and updated the tests. The team has confirmed the feature was fully released in 2.0.1.

internal/vault/sys/agent_registration_resource.go
```diff
@@ -412,15 +446,20 @@ func (r *AgentRegistrationResource) readFromVault(ctx context.Context, client *a
 		return
 	}
 
-	if readResp == nil {
+	if readResp == nil || readResp.Data == nil {
```
Chris-Paris-HashiCorp, line 443: Nice catch here.

internal/vault/sys/agent_registration_resource.go
```diff
@@ -92,6 +100,9 @@ func (r *AgentRegistrationResource) Schema(ctx context.Context, req resource.Sch
 			consts.FieldDisplayName: schema.StringAttribute{
 				MarkdownDescription: "Human-readable name for the agent registration.",
 				Required:            true,
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(displayNameRegexp, "must not contain forward slashes (/)"),
+				},
```
Chris-Paris-HashiCorp: If we restrict the `/` character but also do the bifurcated import (by ID or by name), then we still cannot manage display-names that have slashes in them, from within TFVP. I thought adding the ability to import by ID obviated the need for rejecting the `/` in display-name.

It was my opinion that we should do one or other improvement but not both.

However, it would be easy to revert the restriction if a customer complains that their display name of "this agent turns on/off the lights" (which seems reasonable, not a fanciful edge case that we'll never see).

In other words, I'm agreeable to including the restriction, and we can reevaluate later if a customer complains.

internal/vault/sys/agent_registration_resource.go
```diff
@@ -249,7 +260,30 @@ func (r *AgentRegistrationResource) Read(ctx context.Context, req resource.ReadR
...
 	}
 
-	r.readFromVault(ctx, client, &data, &resp.Diagnostics, true)
+	// Prefer reading by ID, fall back to display_name for import
+	var path string
+	if !data.ID.IsNull() && !data.ID.IsUnknown() && data.ID.ValueString() != "" {
+		path = r.registrationByIDPath(data.ID.ValueString())
+	} else {
+		path = r.registrationByNamePath(data.DisplayName.ValueString())
+	}
```
Chris-Paris-HashiCorp: Well this order turns things around. It says the default is import by ID, not dependent on what the string looks like. I thought we would do something like (pseudocode):

`if _, ok := parse.IsUUID(thestring); ok { import by ID } else {import  by displayname}`

Chris-Paris-HashiCorp:
For to add that once we settle on the final import behavior, we also need to update the documentation parts of this PR).

internal/vault/sys/agent_registration_resource_test.go
fcallegari7: We restrict duplicate `display_name` in the same namespace. However, if you try to register an agent with an existing `display_name` and don't pass an `id`, it will lookup the registration by `display_name` and update the registration that already exists.

That functionality is already covered in `TestAccAgentRegistration_withDescription` as that already updates a registration by `display_name`.

We added a test `TestAccAgentRegistration_duplicateDisplayNameAcrossNamespaces` that checks for allowing duplicate `display_name` across namespaces.

Chris-Paris-HashiCorp: Confirmed. My earlier comment was wrong. Thank you Fabio.

Chris-Paris-HashiCorp:
I resolved the parsing challenge stemming from using a /-delimited string to denote both a nested namespace and a display_name that might have one or more / in it. Turned out that this fundamentally, untenable, and also this is not the pattern elsewhere in the TFVP. With the latest patches, the namespace must be specified by environment variable. The name is then just the name. If it parses as a UUID, then we are importing from vault by ID. Otherwise, we import by display_name. This completely solves the problem where / characters within the display_name (totally legal in vault) would cause misparses in TFVP. This is also more consistent with the rest of the provider.

BTW there are detailed commit comments on the latest round of commits.

Chris-Paris-HashiCorp:
I think we got everything this time. 


From the main fixing commit message:

Import Agent Registry records by UUID or display_name
Detect whether the import ID is a record UUID or a display_name and read
from the corresponding Vault endpoint, instead of splitting the import ID
on "/" to derive a namespace.

A display_name may contain "/", which made the previous rightmost-"/"
split ambiguous: records whose display_name contained "/" could not be
imported, even though Vault, its CLI, and its API allow them. Splitting
also forced an inline namespace convention unique to this resource.

ImportState now treats the import ID as the verbatim Vault identifier and
parses it with go-uuid (the package Vault uses to generate record ids):
a successful parse reads by id, otherwise it reads by display_name. The
namespace is taken from TERRAFORM_VAULT_NAMESPACE_IMPORT, matching
base.WithImportByID and the rest of the provider. This makes every valid
record importable, with import-by-UUID as the reliable path for
display_names containing "/".

Promote go-uuid from an indirect to a direct dependency. Remove the now
unused base.ParseImportID helper and its test.

Add acceptance tests for import-by-UUID and import-by-UUID of a record
whose display_name contains "/", fix the namespace import test to use the
env var, and add a Vault-free unit test for the UUID classifier.

@Balaji2198 @fcallegari7

internal/vault/sys/agent_registration_resource.go
Chris-Paris-HashiCorp: The namespace/display_name convention introduced by this PR was the origin of the ambiguous parsing problem, and isn't present in the rest of the provider. Later commits removed the ability of specifying namespace with a ns/ prefix -- we are not consistent with the rest of the provider -- namespace must be specified by environment variable, so there's no problem with / in display_name now.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ BREAKING CHANGES:
 FEATURES:
 
 * **New Resources**: Add support for OS Secrets Engine with `vault_os_secret_backend`, `vault_os_secret_backend_host`, and `vault_os_secret_backend_account` resources for managing operating system credentials via SSH. Requires Vault 2.0.0+. ([#2865](https://github.com/hashicorp/terraform-provider-vault/pull/2865))
+* **New Resource**: `vault_agent_registration` for managing Agent Registry records in Vault Enterprise. Allows registering Vault agents with specific identity entities and configuring ceiling policies that limit maximum agent permissions. Requires Vault 2.0.1+.
```
Balaji2198: Can we move this changelog to the Unreleased Features section?

Chris-Paris-HashiCorp: Done. I made the complimentary change in the [oauth PR](https://github.com/hashicorp/terraform-provider-vault/pull/2890).

Balaji2198 (approved):

PR #2923, opened 2026-06-11
fix(auth_login_aws): assume aws_role_arn only once for non-web-identity credentials

### LINKED ISSUE #2922: [Bug]: auth_login_aws assumes aws_role_arn twice, breaking any role that can't assume itself
### Terraform Core Version

1.14.8

5.9.0

### Vault Server Version

1.20.4

### Affected Resource(s)

 * provider "vault" auth_login_aws block

### Expected Behavior

When `auth_login_aws` is given both an `aws_profile` (whose profile already assumes a role) and an `aws_role_arn`, the provider should assume the target role once to obtain credentials for the Vault login.

### Actual Behavior

The provider assumes `aws_role_arn` twice:

 1. Once implicitly, by handing the role ARN to the AWS SDK credential chain while resolving base credentials.
 2. Once explicitly, in its own `sts:AssumeRole` call right before logging in to Vault.

 Because step 1 already lands the session as the target role, step 2 then tries to assume the same role from that same role. This only succeeds if the role's trust policy permits self-assumption. For any normal role that cannot assume itself (i.e. almost all cross-account roles), the login fails.

```shell
Error: failed to get AWS credentials required for Vault login,
err=failed to assume role arn:aws:iam::ACCOUNT_ID:role/ci/some-target-role:
operation error STS: AssumeRole, https response error StatusCode: 400,
RequestID: ..., api error AccessDenied: User:
arn:aws:sts::ACCOUNT_ID:assumed-role/some-target-role/... is not authorized
to perform: sts:AssumeRole on resource: arn:aws:iam::ACCOUNT_ID:role/some-target-role


(Depending on whether an external ID is involved in the chain, the first failure can instead surface as a ValidationError on externalId
 — same root cause: the duplicate assume.)
```

Works — single assume (no aws_role_arn, the profile does the assume):

 ```hcl
   # ~/.aws/config
   [profile terraform]
   sso_start_url  = https://EXAMPLE.awsapps.com/start
   sso_region     = us-east-1
   sso_account_id = ACCOUNT_ID
   sso_role_name  = Administrator
   region         = us-east-1
 ```

 ```hcl
   provider "vault" {
     address = "https://vault.example.com"
     auth_login_aws {
       aws_profile = "terraform"
       role        = "some-vault-role"
     }
   }
 ```

 Fails — double assume (profile assumes a role, and aws_role_arn set):

 ```hcl
   # ~/.aws/config
   [profile main]
   sso_start_url  = https://EXAMPLE.awsapps.com/start
   sso_region     = us-east-1
   sso_account_id = ACCOUNT_ID
   sso_role_name  = Administrator
   region         = us-east-1

   [profile terraform]
   source_profile = main
   role_arn       = arn:aws:iam::ACCOUNT_ID:role/ci/some-runner-role
 ```

 ```hcl
   provider "vault" {
     address = "https://vault.example.com"
     auth_login_aws {
       aws_profile  = "terraform"
       aws_role_arn = "arn:aws:iam::ACCOUNT_ID:role/ci/some-target-role"
       role         = "some-vault-role"
     }
   }
 ```

### Steps to Reproduce

1. Configure an AWS profile that assumes a role (e.g. source_profile + role_arn, or an SSO profile that resolves to a role).
2. In `auth_login_aws`, set both `aws_profile` (to that profile) and `aws_role_arn` (to a target role that cannot assume itself — e.g. a cross-account role whose trust policy lists your runner role but not itself).
3. Run terraform plan.
4. Login fails with AccessDenied … is not authorized to perform: sts:AssumeRole on the target role (it's trying to assume itself).
5. Point `aws_role_arn` at a role that can assume itself, and it succeeds — confirming the duplicate assume.

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

This is the exact shape of a CI credential chain: an instance profile (or SSO login) assumes a "runner" role, and that runner role then assumes into other accounts via aws_role_arn. The intended chain is:

 ```
   base creds (SSO / instance profile)
     → assume runner role        (via the profile)
       → assume target role      (via aws_role_arn)   ← used for Vault login
 ```

 The bug inserts a phantom extra hop: the target role is assumed once by the SDK chain and then again by the provider's explicit call, so the real chain executed is:

 ```
   … → assume target role (SDK chain) → assume target role AGAIN (explicit)   ← fails
 ```

The single source of the duplicate is `getCredentialsConfig` in `internal/provider/auth_aws.go` passing `aws_role_arn` to `awsutil.WithRoleArn(...)` (the SDK-chain assume) in addition to the explicit `stsClient.AssumeRole(...)` already performed in Login. The provider already has a helper, `manualAssumeRoleARN()`, that decides whether the explicit assume will run; the fix is to skip the SDK-chain assume whenever that helper says the explicit path will handle it. A one-line guard resolves it; PR to follow.

### References

* GH-2766  (umbrella AWS Auth Method regression since v5.7.0; the IRSA report in that thread is the same double-assume for
 web-identity credentials)
* GH-2679  (the awsutil/v2 migration that introduced the second assume)
* GH-2519  (proposed full rewrite of auth_login_aws; would also resolve this)

Yes

nomeelnoj:
Fixes `auth_login_aws` assuming `aws_role_arn` twice when credentials are sourced from a profile or another non-web-identity chain. `getCredentialsConfig` handed `aws_role_arn` to the AWS SDK while `getLoginData` independently ran a manual STS `AssumeRole`, so the target role had to assume itself — which fails for cross-account roles that disallow self-assumption. Regression from the v5.7.0 awsutil/v2 migration.

## Fix

Gate `WithRoleArn`/`WithRoleSessionName` in `getCredentialsConfig` behind `!manualAssumeRoleARN()`. Non-web-identity sources defer to the existing manual `AssumeRole` (assumed once). Web-identity (IRSA) is unchanged — the SDK credential chain assumes the role, so `aws_role_arn` stays set.

## Testing

Updated the two `TestAuthLoginAWS_getCredentialsConfig` cases that asserted the old double-assume behavior, and adjusted `TestAuthLoginAWS_RoleAssumption` to cover both paths: web-identity (role ARN passed to the SDK) and non-web-identity (role ARN deferred to the manual assume). `gofmt`/`go vet` clean; all AWS auth unit tests pass. Verified against a live cross-account role.

Closes #2922

### Checklist
- [X] Added [CHANGELOG](https://github.com/hashicorp/terraform-provider-vault/blob/master/CHANGELOG.md) entry (only for user-facing changes)

```

...
```

### Community Note

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,14 @@ FEATURES:
 * Add support for `pkcs12_bundle` and `jks_bundle` formats in `vault_pki_secret_backend_cert`, `vault_pki_secret_backend_root_cert`, `vault_pki_secret_backend_root_sign_intermediate`, and `vault_pki_secret_backend_sign` ([#2908](https://github.com/hashicorp/terraform-provider-vault/pull/2908)). Requires Vault 2.1+.
 * `vault_policy`: Added `allow_overwrite` to optionally prevent overwriting Vault policies.([#2895](https://github.com/hashicorp/terraform-provider-vault/pull/2895))
 
-IMPROVEMENTS: 
+IMPROVEMENTS:
 
 * `resource/vault_token`: Added deprecation warning to guide users toward the new ephemeral `vault_token` resource for better security and batch token support. ([#2877](https://github.com/hashicorp/terraform-provider-vault/pull/2877))
 * Replaced backend with mount in `vault_aws_access_credentials` resource's documentation and improved descriptions for a few other parameters.([#2911](https://github.com/hashicorp/terraform-provider-vault/pull/2911))
+
+BUG FIXES:
+
+* `provider`: Fix `auth_login_aws` assuming `aws_role_arn` twice when credentials are sourced from a profile or other non-web-identity chain, which caused the target role to assume itself and fail for roles that do not permit self-assumption. ([#2923](https://github.com/hashicorp/terraform-provider-vault/pull/2923))
 
 ## 5.9.0 (April 22, 2026)
 
--- a/internal/provider/auth_aws.go
+++ b/internal/provider/auth_aws.go
@@ -478,10 +478,11 @@ func (l *AuthLoginAWS) getCredentialsConfig(logger hclog.Logger) (*awsutil.Crede
 	if v, ok := l.params[consts.FieldAWSRegion].(string); ok && v != "" {
 		opts = append(opts, awsutil.WithRegion(v))
 	}
-	if v, ok := l.params[consts.FieldAWSRoleARN].(string); ok && v != "" {
+	_, manualAssume := l.manualAssumeRoleARN()
+	if v, ok := l.params[consts.FieldAWSRoleARN].(string); ok && v != "" && !manualAssume {
 		opts = append(opts, awsutil.WithRoleArn(v))
 	}
-	if v, ok := l.params[consts.FieldAWSRoleSessionName].(string); ok && v != "" {
+	if v, ok := l.params[consts.FieldAWSRoleSessionName].(string); ok && v != "" && !manualAssume {
 		opts = append(opts, awsutil.WithRoleSessionName(v))
 	}
 	if v, ok := l.params[consts.FieldAWSWebIdentityTokenFile].(string); ok && v != "" {
--- a/internal/provider/auth_aws_test.go
+++ b/internal/provider/auth_aws_test.go
@@ -404,12 +404,12 @@ func TestAuthLoginAWS_getCredentialsConfig(t *testing.T) {
 				},
 			},
 			logger: hclog.NewNullLogger(),
+			// Non-web-identity creds: aws_role_arn is assumed manually in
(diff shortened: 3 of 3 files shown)


nomeelnoj:
This also sidesteps an unrelated `awsutil` bug: the old path sent `ExternalId=""` to STS (rejected with a `ValidationError`) when no external ID was set. By assuming the role once via the manual path, we no longer hit that path. Underlying bug fixed upstream in hashicorp/go-secure-stdlib#186 to ensure proper functionality in other tools leveraging the lib.

nomeelnoj:
Any possibility of getting this reviewed? Our CI pipeline is using a legacy / hacked together solution for Vault auth and converting to AWS-based auth would be a HUGE relief of techdebt and add a ton of reliability--we'd prefer not to run a custom fork of the provider.

Please let me know if anything needs to be addressed or modified.  The failing Jira sync build I do not know how to debug.

PR #2960, opened 2026-07-08
add rotate-on-read option on ldap secrets

vivekpandeyHashi:
### Description
Support for Vault's openldap rotate-on-read feature
Adds a rotate_on_read configuration field to the LDAP secrets engine and static roles, enabling on-demand password rotation at the time of credential read. This complements (not replaces) the existing scheduled rotation mechanism.

### DIFF
--- a/vault/resource_ldap_secret_backend.go
+++ b/vault/resource_ldap_secret_backend.go
@@ -272,11 +272,11 @@ func createUpdateLDAPConfigResource(ctx context.Context, d *schema.ResourceData,
 
 	// get rotate-on-read fields
 	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
-		if v, ok := d.GetOk(consts.FieldRotateOnRead); ok {
-			data[consts.FieldRotateOnRead] = v
+		if d.HasChange(consts.FieldRotateOnRead) {
+			data[consts.FieldRotateOnRead] = d.Get(consts.FieldRotateOnRead)
 		}
-		if v, ok := d.GetOk(consts.FieldRotateOnReadCooldown); ok {
-			data[consts.FieldRotateOnReadCooldown] = v
+		if d.HasChange(consts.FieldRotateOnReadCooldown) {
+			data[consts.FieldRotateOnReadCooldown] = d.Get(consts.FieldRotateOnReadCooldown)
 		}
 	}
 
--- a/vault/resource_ldap_secret_backend.go
+++ b/vault/resource_ldap_secret_backend.go
@@ -150,6 +150,18 @@ func ldapSecretBackendResource() *schema.Resource {
 			ConflictsWith: []string{consts.FieldBindPass, consts.FieldBindPassWO},
 			ExactlyOneOf:  []string{consts.FieldBindPass, consts.FieldBindPassWO, consts.FieldSelfManaged},
 		},
+		consts.FieldRotateOnRead: {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
+			Description: "If true, static role credentials are rotated on each read. Acts as a default for all static roles. Requires Vault Enterprise ≥ 2.1.0.",
+		},
+		consts.FieldRotateOnReadCooldown: {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+			Description: "Minimum seconds between rotate-on-read rotations. Acts as a default cooldown for all static roles. Requires Vault Enterprise ≥ 2.1.0.",
+		},
 	}
 	resource := provider.MustAddMountMigrationSchema(&schema.Resource{
 		CreateContext: provider.MountCreateContextWrapper(createUpdateLDAPConfigResource, provider.VaultVersion112),
@@ -258,6 +270,16 @@ func createUpdateLDAPConfigResource(ctx context.Context, d *schema.ResourceData,
 		automatedrotationutil.ParseAutomatedRotationFields(d, data)
 	}
 
+	// get rotate-on-read fields
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
+		if v, ok := d.GetOk(consts.FieldRotateOnRead); ok {
+			data[consts.FieldRotateOnRead] = v
+		}
+		if v, ok := d.GetOk(consts.FieldRotateOnReadCooldown); ok {
+			data[consts.FieldRotateOnReadCooldown] = v
+		}
+	}
+
 	// get self-managed boolean
 	if provider.IsAPISupported(meta, provider.VaultVersion200) && provider.IsEnterpriseSupported(meta) {
 		if d.HasChange(consts.FieldSelfManaged) {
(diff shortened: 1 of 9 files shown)


vault/resource_ldap_secret_backend.go
```diff
@@ -258,6 +270,16 @@ func createUpdateLDAPConfigResource(ctx context.Context, d *schema.ResourceData,
 		automatedrotationutil.ParseAutomatedRotationFields(d, data)
 	}
 
+	// get rotate-on-read fields
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
+		if v, ok := d.GetOk(consts.FieldRotateOnRead); ok {
```
ankitsutharhashicorp: Can we add condition for update scenario? Like,

`d.HasChange(consts.FieldRotateOnRead)`

`d.HasChange(consts.FieldRotateOnReadCooldown)`

vault/resource_ldap_secret_backend.go
```diff
@@ -348,6 +370,19 @@ func readLDAPConfigResource(ctx context.Context, d *schema.ResourceData, meta in
 		}
 	}
 
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
```
ankitsutharhashicorp, line 373: If these two fields are not set explicitly, should they be returned in output of read response?

vault/resource_ldap_secret_backend.go
```diff
@@ -258,6 +270,16 @@ func createUpdateLDAPConfigResource(ctx context.Context, d *schema.ResourceData,
...
 
+	// get rotate-on-read fields
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
+		if v, ok := d.GetOk(consts.FieldRotateOnRead); ok {
+			data[consts.FieldRotateOnRead] = v
+		}
+		if v, ok := d.GetOk(consts.FieldRotateOnReadCooldown); ok {
+			data[consts.FieldRotateOnReadCooldown] = v
+		}
+	}
```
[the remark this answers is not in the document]
vivekpandeyHashi: GetOkExists is deprecated but we can use d.HasChange here

vault/resource_ldap_secret_backend.go
vivekpandeyHashi: yes, vault also returns the default value

website/docs/r/ldap_secret_backend_static_role.html.md
```diff
@@ -77,6 +77,17 @@ The following arguments are supported:
 * `password_policy` - (Optional) Name of the password policy to use to generate passwords for this role.
   Requires Vault 2.1.0+.
 
+* `rotate_on_read` - (Optional) If true, credentials are rotated on each read. When set, this overrides
+  the engine-level `rotate_on_read` default. When unset, the role inherits the engine-level value.
+  Removing this attribute after it has been set is not supported without recreating the role.
```
vijayavelsekar, line 82: Does vault does not support unsetting these fields ?
SDK v2 resources support removing attributes by sending API requests with the field omitted or set to a default value.

vivekpandeyHashi: No, unsetting rotate_on_read is not supported.
once it's been set to true or false, you can't unset it back to nil. You can only toggle between true and false.

currently for this field or existing field omitting the on an update simply preserves the existing value. If the field was previously true, it stays true. If it was nil (never set), it stays nil.

vault/resource_ldap_secret_backend.go
```diff
@@ -258,6 +270,16 @@ func createUpdateLDAPConfigResource(ctx context.Context, d *schema.ResourceData,
 		automatedrotationutil.ParseAutomatedRotationFields(d, data)
 	}
 
+	// get rotate-on-read fields
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
+		if d.HasChange(consts.FieldRotateOnRead) {
```
vivekpandeyHashi, line 275: Yes, these fields can be updated any time.

vault/resource_ldap_secret_backend_static_role.go
```diff
@@ -141,6 +152,16 @@ func createUpdateLDAPStaticRoleResource(ctx context.Context, d *schema.ResourceD
...
 	}
 
+	// get rotate-on-read role-level overrides
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
+		if d.HasChange(consts.FieldRotateOnRead) {
+			data[consts.FieldRotateOnRead] = d.Get(consts.FieldRotateOnRead)
+		}
+		if d.HasChange(consts.FieldRotateOnReadCooldown) {
+			data[consts.FieldRotateOnReadCooldown] = d.Get(consts.FieldRotateOnReadCooldown)
+		}
```
ankitsutharhashicorp, line 162: My suggestion is whenever user tries to set these fields with vault version below 2.1.0, it should return relevant error for these both the fields. In addition to this, you can check if update causing to remove this field (i.e. nil) remove this, again throw error.

vault/resource_ldap_secret_backend_static_role.go
```diff
@@ -187,6 +208,19 @@ func readLDAPStaticRoleResource(ctx context.Context, d *schema.ResourceData, met
 		}
 	}
 
+	if provider.IsAPISupported(meta, provider.VaultVersion210) && provider.IsEnterpriseSupported(meta) {
```
ankitsutharhashicorp, line 211: This code block is redundant:

1. **During create/update**: Version check ensures these fields are only sent to Vault 2.1.0+
2. **During read**: The [for loop above](https://github.com/hashicorp/terraform-provider-vault/pull/2960/changes#diff-106def9fd3cdaf4291ab380ac4a4ddf92bd44d4fe3741a934e88f1e20baa124dL176) already reads these fields from `ldapSecretBackendStaticRoleFields`
3. **Backward compatibility**: The `if val, ok := resp.Data[field]; ok` check in the loop handles cases where Vault doesn't return these fields (older versions or when not set)

To fix this add `consts.FieldRotateOnRead` and `consts.FieldRotateOnReadCooldown` to the `ldapSecretBackendStaticRoleFields` array.

vault/resource_ldap_secret_backend_static_role.go
vivekpandeyHashi: vault does not return any error on unknown fields for these API's so we should continue this in TFVP as well and ignore the unknown fields

vault/resource_ldap_secret_backend_static_role.go
vivekpandeyHashi: adding field in ldapSecretBackendStaticRoleFields will break
createUpdateLDAPStaticRoleResource

while creating/update
Two problems:
1. No version gate: the loop would send rotate_on_read/rotate_on_read_cooldown to Vault < 2.1.0 or non-Enterprise
2. GetOk vs HasChange: GetOk returns false for zero values (false bool, 0 int), so clearing these fields (e.g. removing them from config) would never be sent. The separate block uses HasChange + Get to handle that correctly.

stuti-sr (commented):
Can you please check for test case for Unsetting rotate_on_read after it's been enabled and Setting rotate_on_read_cooldown without rotate_on_read?

stuti-sr (dismissed):
LGTM!!

vijayavelsekar (dismissed):
LGTM

stuti-sr (approved):

vijayavelsekar (approved):

PR #2974, opened 2026-07-20
Adding access token support for Terraform

coleheinbaugh:
### Description
This PR adds Terraform support for the new access token path being added to secrets-engine-azure-enterprise. It creates an ephemeral access token resource that can be used in Terraform.

### Checklist
- [x] Acceptance tests where run against all supported Vault Versions (This is only available for Vault 2.2+ and tests will be skipped if Vault is below this version)

```
➜  terraform-provider-vault git:(VAULT-46319/cbh-azure-access-token-support) ✗ go test ./internal/vault/secrets/ephemeral/... -run TestAccAzureAccessToken -v -timeout 30m
=== RUN   TestAccAzureAccessToken_basic
    azure_access_token_test.go:38: Vault server version "2.1.0-beta1+ent"
--- PASS: TestAccAzureAccessToken_basic (18.39s)
=== RUN   TestAccAzureAccessToken_invalidRole
    azure_access_token_test.go:69: Vault server version "2.1.0-beta1+ent"
--- PASS: TestAccAzureAccessToken_invalidRole (0.54s)
=== RUN   TestAccAzureAccessToken_invalidMount
    azure_access_token_test.go:93: Vault server version "2.1.0-beta1+ent"
--- PASS: TestAccAzureAccessToken_invalidMount (0.16s)
=== RUN   TestAccAzureAccessToken_customRetry
    azure_access_token_test.go:120: Vault server version "2.1.0-beta1+ent"
--- PASS: TestAccAzureAccessToken_customRetry (29.00s)
=== RUN   TestAccAzureAccessToken_namespace
    azure_access_token_test.go:154: Vault server version "2.1.0-beta1+ent"
--- PASS: TestAccAzureAccessToken_namespace (27.32s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/internal/vault/secrets/ephemeral	76.109s
```

### Community Note

## PCI review checklist


internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+// TODO: replace terraform_data workaround with vault_azure_secret_backend once
+// the token/ endpoint ships in an official Vault release.
+func testAccAzureAccessTokenInvalidRoleConfig(backend string, conf *testutil.AzureTestConf) string {
+	return fmt.Sprintf(`
+resource "terraform_data" "azure_mount" {
+  input = "%[1]s"
+
+  provisioner "local-exec" {
+    command = "vault secrets enable -path=%[1]s vault-plugin-secrets-azure && vault write %[1]s/config subscription_id='%[2]s' tenant_id='%[3]s' client_id='%[4]s' client_secret='%[5]s'"
```
coleheinbaugh: Because the access token isn’t out yet on the main branch, but I still need to test if this E2E test works, I have to spin up vault and register my version of the Azure secrets engine plugin: `vault-plugin-secrets-azure`. When the PR is ready to merge I will change it to use the proper plugin version.

cbiguet:
@coleheinbaugh we should make sure to include a version check here. FYI the 2.1.0 release got renamed to 2.2.0 this week, so the resource should check if the Vault version is at least 2.2.0 (and return a clear error if not).

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -0,0 +1,276 @@
...
+	role := acctest.RandomWithPrefix("tf-role")
+	nonEmpty := regexp.MustCompile(`^.+$`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion121)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
```
fairclothjm, line 41: Can we remove this?

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -0,0 +1,276 @@
...
+	conf.Scope = testutil.SkipTestEnvUnset(t, "AZURE_ROLE_SCOPE")[0]
+
+	backend := acctest.RandomWithPrefix("tf-test-azure")
+	role := acctest.RandomWithPrefix("tf-role")
+	nonEmpty := regexp.MustCompile(`^.+$`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion121)
```
fairclothjm:
```suggestion
			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion221)
```

fairclothjm: You may need to create a new var VaultVersion221 in the consts file

internal/vault/secrets/ephemeral/azure_access_token_test.go
coleheinbaugh: I think this is actually important. My understanding is that without it we can't check if the access token was created. Ephemeral resources (which the access token is) are not stored in state, so we can't really check if they are there or not. echo on protoV6 allows us to copy the ephemeral resource and save it so we can check if it is non-empty later

internal/vault/secrets/ephemeral/azure_access_token_test.go
fairclothjm: Sorry, should be VaultVersion220

coleheinbaugh: Done, changed to skip unless lower than 2.2.0

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	if !r.Meta().IsAPISupported(provider.VaultVersion220) {
+		resp.Diagnostics.AddError(
+			"Feature Not Supported",
+			"vault_azure_access_token requires Vault version 2.2.0 or later. "+
+				"Current Vault version: "+r.Meta().GetVaultVersion().String(),
+		)
+		return
+	}
```
coleheinbaugh, line 179: Added version checking to use the resource here @cbiguet

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/client"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/errutil"
+	"github.com/hashicorp/terraform-provider-vault/internal/framework/model"
+	"github.com/hashicorp/terraform-provider-vault/internal/provider"
+	"github.com/hashicorp/vault/api"
+)
+
+const (
+	fieldAccessToken  = "access_token"
+	fieldExtExpiresIn = "ext_expires_in"
```
robmonte, line 31: I am shocked that internal/consts doesn't already contain an "access_token" variable. It's a nit but there are many hits for "access_token" in the repo though, maybe fast-follow this change with a replacement for them all to make a new shared constant?

website/docs/ephemeral-resources/azure_access_token.html.md
```diff
@@ -0,0 +1,85 @@
...
+sidebar_current: "docs-vault-ephemeral-azure-access-token"
+description: |-
+  Generate ephemeral Azure OAuth2 access tokens from Vault static role credentials
+
+---
+
+# vault\_azure\_access\_token (Ephemeral)
+
+Generates ephemeral Azure OAuth2 access tokens from the Vault Azure Secrets engine using static role credentials.
+These tokens are not stored in Terraform state.
```
robmonte: Very nitpick but should probably be either "Secrets Engine" or "secrets engine"?

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: Yeah that makes sense, I can add a TODO comment of a reminder to add it in a following PR.

website/docs/ephemeral-resources/azure_access_token.html.md
coleheinbaugh: great call. Looks like I did this in a few places. Based on what I saw when talking about a specific secrets engine we use capitols (i.e. Azure Secrets Engine), but when talking about them in general we use "secrets engine". This example should now be fixed to show "Azure Secrets Engine"

internal/consts/consts.go
```diff
@@ -1002,6 +1002,7 @@ const (
 	VaultVersion201  = "2.0.1"
 	VaultVersion203  = "2.0.3"
 	VaultVersion210  = "2.1.0"
+	VaultVersion220  = "2.2.0"
```
fairclothjm, line 1029: This already exists on main, so you will need to rebase your branch

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+	fieldExpiresIn    = "expires_in"
+)
+
+// Ensure the implementation satisfies the ephemeral.EphemeralResource interface.
+var _ ephemeral.EphemeralResource = &AzureAccessTokenEphemeralResource{}
+
+// NewAzureAccessTokenEphemeralResource returns the implementation for this resource to be
+// imported by the Terraform Plugin Framework provider.
+var NewAzureAccessTokenEphemeralResource = func() ephemeral.EphemeralResource {
```
fairclothjm: Can we make this a normal function? I see no reason for it to be a var. In general, we should try to avoid using global variables.

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+	resp.Diagnostics.Append(resp.Result.Set(ctx, &data)...)
+}
+
+// credPropagationRetries is the number of times to retry
+const credPropagationRetries = 4
+
+// credPropagationDelay is the wait between retries.
+const credPropagationDelay = 4 * time.Second
+
+func requestAzureAccessToken(ctx context.Context, cli *api.Client, mount, role, scope string) (*AzureAccessTokenAPIModel, error) {
```
fairclothjm: can we add a doc comment to this function?

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+	data.ExtExpiresIn = types.Int64Value(tokenResp.ExtExpiresIn)
+
+	resp.Diagnostics.Append(resp.Result.Set(ctx, &data)...)
+}
+
+// credPropagationRetries is the number of times to retry
+const credPropagationRetries = 4
+
+// credPropagationDelay is the wait between retries.
+const credPropagationDelay = 4 * time.Second
```
fairclothjm, line 219: should we go ahead and make these configurable? I imagine we will get requests for that.

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+			// token/ returns a 400 when the static credential has not yet been
+			// provisioned. Reading static-creds/ initializes it on first access.
+			// This only happens once — on every subsequent Open() the credential
+			// already exists and token/ succeeds on the first attempt.
+			if !initialized && strings.Contains(errStr, "rotate the role once before token generation") {
+				tflog.Info(ctx, fmt.Sprintf(
+					"Static credential for role %q not yet provisioned; initializing via static-creds/",
+					role,
+				))
+				staticCredsPath := fmt.Sprintf("%s/static-creds/%s", mount, role)
```
fairclothjm, line 259: I am a little surprised by this. I don't necessarily disagree but are there any downsides to implicitly making the call to /static-creds?

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+						"retrying in %s (attempt %d/%d)",
+					credPropagationDelay, attempt+1, credPropagationRetries,
+				))
+				continue
+			}
+			// token/ returns a 400 when the static credential has not yet been
+			// provisioned. Reading static-creds/ initializes it on first access.
+			// This only happens once — on every subsequent Open() the credential
+			// already exists and token/ succeeds on the first attempt.
+			if !initialized && strings.Contains(errStr, "rotate the role once before token generation") {
```
fairclothjm: Checking a string like this isn't ideal. This is probably ok for now, but for the future the most robust design would be:

Vault returns something like:
- HTTP 400 or 409
  - body includes a stable code such as static_credentials_not_initialized
  - vault/api exposes structured access to that code
- provider checks the code and handles initialization

internal/vault/secrets/ephemeral/azure_access_token_test.go
fairclothjm: got it, thanks!

internal/vault/secrets/ephemeral/azure_access_token_test.go
fairclothjm: Sounds good. We should be sure to change this to just use the native azure mount and config resources before merging this PR.

internal/vault/secrets/ephemeral/azure_access_token.go
fairclothjm: For now, maybe we could add helpers similar to isGCPTokenGenerationError(), use errors.As() and maybe http.StatusBadRequest and then narrow message matching to the specific Vault response entry?

fairclothjm (commented):
Looking great! I left a round of comments

internal/vault/secrets/ephemeral/azure_access_token.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		secret, err := cli.Logical().WriteWithContext(ctx, path, map[string]interface{}{
+			"scope": scope,
+		})
+		if err != nil {
+			errStr := err.Error()
+			// AADSTS7000215 means the client secret exists in Vault but has not
+			// yet propagated across Azure AD. Retry after a short wait.
+			if strings.Contains(errStr, "AADSTS7000215") {
```
vinay-gopalan: can we pull the error code out to a variable?

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion220)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"echo": echoprovider.NewProviderServer(),
+		},
```
vinay-gopalan: nit: I believe we can remove the `echo` provider from these tests since it's unused

```suggestion
```

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -0,0 +1,275 @@
...
```
vinay-gopalan, line 122: nit: same as above
```suggestion
```

vinay-gopalan (commented):
Looking great! Had a couple more nits/comments

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: Good call. I’ve made it configurable

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: Unfortunately I think it needs to be done if the delayed credential field is used there might not be a credential, so we’d have to call /static-creds in that scenario.

internal/vault/secrets/ephemeral/azure_access_token_test.go
coleheinbaugh: Echo needs to stay. It is the only way that we can retain the information in an ephemeral resource. The tests fail without it. I think it is used in the tests in places like this:
```
ConfigStateChecks: []statecheck.StateCheck{
					statecheck.ExpectKnownValue("echo.azure_token", tfjsonpath.New("data").AtMapKey("access_token"), knownvalue.StringRegexp(nonEmpty)),
					statecheck.ExpectKnownValue("echo.azure_token", tfjsonpath.New("data").AtMapKey("token_type"), knownvalue.StringExact("Bearer")),
					statecheck.ExpectKnownValue("echo.azure_token", tfjsonpath.New("data").AtMapKey("expires_in"), knownvalue.NotNull()),
					statecheck.ExpectKnownValue("echo.azure_token", tfjsonpath.New("data").AtMapKey("ext_expires_in"), knownvalue.NotNull()),
				},
```

internal/vault/secrets/ephemeral/azure_access_token_test.go
coleheinbaugh: see comment above

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: The convention in this repo seems to be that all the ephemeral resources are vars. I think this is for easy testing so they can be replaced.

Ex: 
```
var NewAzureStaticCredsEphemeralSecretResource = func() ephemeral.EphemeralResource {
	return &AzureStaticCredsEphemeralSecretResource{}
}

var NewGCPOAuth2AccessTokenEphemeralResource = func() ephemeral.EphemeralResource {
	return &GCPOAuth2AccessTokenEphemeralResource{}
}

var NewAWSStaticAccessCredentialsEphemeralSecretResource = func() ephemeral.EphemeralResource {
	return &AWSStaticAccessCredentialsEphemeralSecretResource{}
}
```

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: good call - done.

internal/vault/secrets/ephemeral/azure_access_token.go
fairclothjm: We should not use go anti-patterns like global variables as a workaround for stubbing out test implementations. That said, I don't see any of those being used in that manner so I see no reason for any of them to be a var.

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: Made some functions to extract that

internal/vault/secrets/ephemeral/azure_access_token.go
coleheinbaugh: Ok, that makes sense. I’ve changed it to a normal function

internal/vault/secrets/ephemeral/azure_access_token_test.go
vinay-gopalan: @coleheinbaugh sorry, I meant we can remove the echo provider blocks specifically from `TestAccAzureAccessToken_invalidRole` and `TestAccAzureAccessToken_invalidMount`, since those tests don't use the echo provider. They will fail on config apply time

internal/vault/secrets/ephemeral/azure_access_token_test.go
coleheinbaugh: removed

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -293,41 +263,39 @@ ephemeral "vault_azure_access_token" "token" {
...
+  namespace = vault_namespace.test.path
+  path      = "%[2]s"
+  type      = "vault-plugin-secrets-azure"
+}
 
-  provisioner "local-exec" {
-    when    = destroy
-    command = "VAULT_NAMESPACE=%[1]s vault secrets disable ${self.output}"
-  }
+resource "vault_generic_endpoint" "azure_config" {
```
fairclothjm: Let's try to use the native resource `vault_azure_secret_backend` instead.

internal/vault/secrets/ephemeral/azure_access_token_test.go
```diff
@@ -174,116 +168,92 @@ func TestAccAzureAccessToken_namespace(t *testing.T) {
...
+resource "vault_mount" "%[1]s" {
+  path = "%[2]s"
+  type = "vault-plugin-secrets-azure"
+}
 
-  provisioner "local-exec" {
-    when    = destroy
-    command = "vault secrets disable ${self.output}"
-  }
+resource "vault_generic_endpoint" "%[1]s_config" {
```
fairclothjm: Use `vault_azure_secret_backend` instead

fairclothjm (approved):
LGTM

PR #2979, opened 2026-07-27
Add tfvp support for tpms, tpmgroups, and tpm auth config & roles

gauravmitra1:
### Description
Adds Terraform support for Vault TPM authentication introduced in Vault Enterprise 2.2.0.

This PR adds four Plugin Framework resources covering TPM identity registration, grouping, auth backend configuration, and role management:

- `vault_identity_tpm` — CRUD and import support for individual TPM identity records
- `vault_identity_tpm_group` — CRUD and import support for TPM groups
- `vault_tpm_auth_backend_config` — Configuring a mounted TPM auth backend
- `vault_tpm_auth_backend_role` — CRUD and import support for TPM auth roles

## Special notes
- Each resource uses Vault 2.2.0 API version checks in the Create function
- All duration fields of type `TypeDurationSecond` in Vault (which technically support either duration strings or integers) **only support integer seconds** per TFVP guidelines to avoid state drift.
- The public key (`tpm_ek_public_key`) for `vault_identity_tpm` is immutable and can only be set on create, so the field uses `RequiresReplace` to ensure any changes destroy and recreate the resource. However, this technically includes formatting-only differences such as a trailing `\n`.
  > For example, if a user writes `"-----END PUBLIC KEY-----"` in the config (without a trailing `\n`), the plan won't equal state because Vault always returns keys with a trailing newline via Go's `pem.EncodeToMemory`.

  Since triggering a resource replacement for a whitespace-only formatting difference is poor UX, we use `RequiresReplaceIf` with `normalizePEM` to handle the comparison at the replace decision layer. Whitespace-only differences normalize to the same value, so no replacement is triggered. The tradeoff is that a whitespace-only change produces a one-time in-place update plan. After applying, state reflects the config value exactly and subsequent plans are empty. This works because Terraform sends the same public key value and Vault accepts this without an error, so this update is harmless.

## Abandoned ideas

**Why not a custom type with `StringSemanticEquals`:**
The TFVP code patterns recommend a custom type with `StringSemanticEquals` for API normalization scenarios ([Gotcha #6 in TFVP skills](https://github.com/hashicorp/terraform-provider-vault-skills/blob/main/skills/tfvp-code-patterns/SKILL.md?plain=1#L379-L383)). This was implemented but did not suppress the diff. The Plugin Framework only invokes `StringSemanticEquals` after Read/Create/Update when comparing prior state vs new state and it is was never called during planning. For a `Required` non-`Computed` field, Terraform's core planning engine compares config vs state with raw string equality, bypassing any custom type logic entirely. This field cannot be `Computed` since the user must always supply it.

**Why not a normalizing plan modifier:**
A `PlanModifyString` that rewrites the plan value to its normalized form would solve it, but the Framework rejects plan values that differ from the config value for `Required` non-`Computed` fields with "Provider produced invalid plan".

`RequiresReplaceIf` seems to be the only viable approach for a required, immutable field.

### Checklist
- [x] Acceptance tests where run against all supported Vault Versions (well, they were run against a few supported versions)

**Vault 2.2.0**
```
Running tool: /usr/local/go/bin/go test -test.fullpath=true -timeout 30s -run ^(TestAccIdentityTPM_versionGate|TestAccIdentityTPM|TestAccIdentityTPM_pemNormalization|TestAccIdentityTPM_metadata)$ github.com/hashicorp/terraform-provider-vault/internal/vault/identity -v

=== RUN   TestAccIdentityTPM_versionGate
=== PAUSE TestAccIdentityTPM_versionGate
=== RUN   TestAccIdentityTPM
=== PAUSE TestAccIdentityTPM
=== RUN   TestAccIdentityTPM_pemNormalization
=== PAUSE TestAccIdentityTPM_pemNormalization
=== RUN   TestAccIdentityTPM_metadata
=== PAUSE TestAccIdentityTPM_metadata
=== CONT  TestAccIdentityTPM_versionGate
=== CONT  TestAccIdentityTPM_pemNormalization
=== CONT  TestAccIdentityTPM_metadata
=== CONT  TestAccIdentityTPM
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:64: Vault server version "2.2.0+ent"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:35: Vault server version "2.2.0+ent"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:35: Vault version >= "2.2.0"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:181: Vault server version "2.2.0+ent"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:258: Vault server version "2.2.0+ent"
--- SKIP: TestAccIdentityTPM_versionGate (0.08s)
--- PASS: TestAccIdentityTPM_pemNormalization (1.88s)
--- PASS: TestAccIdentityTPM_metadata (2.53s)
--- PASS: TestAccIdentityTPM (2.82s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/internal/vault/identity	3.727s
```

**Tests on 2.0.4 skip or pass accordingly**
```
Running tool: /usr/local/go/bin/go test -test.fullpath=true -timeout 30s -run ^(TestAccIdentityTPM_versionGate|TestAccIdentityTPM|TestAccIdentityTPM_pemNormalization|TestAccIdentityTPM_metadata)$ github.com/hashicorp/terraform-provider-vault/internal/vault/identity -v

=== RUN   TestAccIdentityTPM_versionGate
=== PAUSE TestAccIdentityTPM_versionGate
=== RUN   TestAccIdentityTPM
=== PAUSE TestAccIdentityTPM
=== RUN   TestAccIdentityTPM_pemNormalization
=== PAUSE TestAccIdentityTPM_pemNormalization
=== RUN   TestAccIdentityTPM_metadata
=== PAUSE TestAccIdentityTPM_metadata
=== CONT  TestAccIdentityTPM_versionGate
=== CONT  TestAccIdentityTPM_pemNormalization
=== CONT  TestAccIdentityTPM_metadata
=== CONT  TestAccIdentityTPM
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:35: Vault server version "2.0.4"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:64: Vault server version "2.0.4"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:64: Vault version < "2.2.0"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:181: Vault server version "2.0.4"
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:181: Vault version < "2.2.0"
--- SKIP: TestAccIdentityTPM (0.10s)
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:258: Vault server version "2.0.4"
--- SKIP: TestAccIdentityTPM_pemNormalization (0.06s)
    /terraform-provider-vault/internal/vault/identity/tpm_resource_test.go:258: Vault version < "2.2.0"
--- SKIP: TestAccIdentityTPM_metadata (0.02s)
--- PASS: TestAccIdentityTPM_versionGate (0.46s)
PASS
ok  	github.com/hashicorp/terraform-provider-vault/internal/vault/identity	1.416s
```

**Manual import testing**
See script to test import and configs in [vault-tools](https://github.com/hashicorp/vault-tools/tree/main/users/cbontempo/tfvp/tpm/tpm-auth) 
```bash
  $ terraform import vault_tpm_auth_backend_config.config "auth/tpm-test/config"
vault_tpm_auth_backend_config.config: Importing from ID "auth/tpm-test/config"...
vault_tpm_auth_backend_config.config: Import prepared!
  Prepared vault_tpm_auth_backend_config for import
vault_tpm_auth_backend_config.config: Refreshing state...

Import successful!

```
```bash
  $ terraform import vault_identity_tpm.server1 "sha256-f485a95a7dfe5994dbb37fdc66bfb18687efa1d941510262f44ae9535dfa1d3b"
vault_identity_tpm.server1: Importing from ID "sha256-f485a95a7dfe5994dbb37fdc66bfb18687efa1d941510262f44ae9535dfa1d3b"...
vault_identity_tpm.server1: Import prepared!
  Prepared vault_identity_tpm for import
vault_identity_tpm.server1: Refreshing state...

Import successful!

```
```bash
  $ terraform import vault_identity_tpm_group.servers "93d4056c-20b9-2574-7d3c-41e1ff864a78"
vault_identity_tpm_group.servers: Importing from ID "93d4056c-20b9-2574-7d3c-41e1ff864a78"...
vault_identity_tpm_group.servers: Import prepared!
  Prepared vault_identity_tpm_group for import
vault_identity_tpm_group.servers: Refreshing state...

Import successful!

```
```bash
  $ terraform import vault_tpm_auth_backend_role.specific_tpms "tpm-test/role/specific-tpms-role"
vault_tpm_auth_backend_role.specific_tpms: Importing from ID "tpm-test/role/specific-tpms-role"...
vault_tpm_auth_backend_role.specific_tpms: Import prepared!
  Prepared vault_tpm_auth_backend_role for import
vault_tpm_auth_backend_role.specific_tpms: Refreshing state... [name=specific-tpms-role]

Import successful!

```

### Community Note

## PCI review checklist


internal/vault/identity/tpm_resource.go
```diff
@@ -0,0 +1,305 @@
...
+
+	data.Name = types.StringValue(readResp.Name)
+	data.TPMEKPublicKey = types.StringValue(readResp.TPMEKPublicKey)
+	data.Disabled = types.BoolValue(readResp.Disabled)
+	data.TPMID = types.StringValue(readResp.TPMID)
+
+	return nil
+}
+
+func (r *IdentityTPMResource) getAPIModel(ctx context.Context, data *IdentityTPMModel) (map[string]any, diag.Diagnostics) {
```
gauravmitra1, line 324: **Low**: The `getAPIModel()` method is defined but never called. The `Create()` and `Update()` methods build the request body inline instead (lines 107-113, 184-190). This is inconsistent with `tpm_role_resource.go` and `tpm_group_resource.go` which both call their `getAPIModel()` methods. Consider either using this method or removing it for consistency.

internal/vault/identity/tpm_resource_test.go
```diff
@@ -0,0 +1,131 @@
+// Copyright IBM Corp. 2016, 2026
```
gauravmitra1, line 1: **Low**: Copyright year ends in 2026, which is in the future. Should this be 2025 or the current year?

gauravmitra1 (commented):
### Bob review

The PR adds comprehensive support for TPM (Trusted Platform Module) resources with three new Terraform resources: `vault_identity_tpm`, `vault_identity_tpm_group`, and `vault_tpm_auth_backend_role`. The implementation follows the Terraform Plugin Framework patterns and includes proper CRUD operations, import support, and comprehensive acceptance tests. However, there are a few minor issues to address:

1. **Copyright year anomaly**: All new files have copyright ending in 2026, which is in the future
2. **Unused method**: `tpm_resource.go` defines `getAPIModel()` but never uses it, creating inconsistency with other resources
3. **Minor code quality**: Some opportunities for consistency improvements

Overall, the code is well-structured with proper error handling, nil checks, and follows existing patterns in the codebase.

internal/vault/identity/tpm_resource.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource_test.go
```diff
@@ -0,0 +1,237 @@
...
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			// Step 1: Create a role bound to a TPM ID.
+			{
+				Config: testAccTPMAuthRoleWithTPMIDConfig(mount, roleName, tpmName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "mount", mount),
+					resource.TestCheckResourceAttr(resourceName, "name", roleName),
+					resource.TestCheckResourceAttr(resourceName, "tpm_ids.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "tpmgroup_ids.#", "0"),
```
ncabatoff, line 59: We should provide a group id as well.

internal/vault/auth/tpm/tpm_role_resource_test.go
```diff
@@ -0,0 +1,237 @@
...
+						plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionDestroy),
+					},
+				},
+			},
+		},
+	})
+}
+
+// TestAccTPMAuthRoleNamespace verifies that the role resource works correctly
+// when deployed inside a Vault namespace (Enterprise only).
```
ncabatoff:
```suggestion
// when deployed inside a Vault namespace.
```

The whole feature is ent-only, so the godoc was a bit confusing.

internal/vault/identity/tpm_group_resource.go
```diff
@@ -0,0 +1,294 @@
...
+type IdentityTPMGroupModel struct {
+	base.BaseModel
+
+	Name         types.String `tfsdk:"name"`
+	MemberTPMIDs types.Set    `tfsdk:"member_tpm_ids"`
+}
+
+type identityTPMGroupAPIModel struct {
+	Name         string   `json:"name" mapstructure:"name"`
+	MemberTPMIDs []string `json:"member_tpm_ids" mapstructure:"member_tpm_ids"`
```
ncabatoff: There is also a map[string]string metadata field.

internal/vault/identity/tpm_resource.go
```diff
@@ -0,0 +1,301 @@
...
+	TPMEKPublicKey types.String `tfsdk:"tpm_ek_public_key"`
+	Disabled       types.Bool   `tfsdk:"disabled"`
+	TPMID          types.String `tfsdk:"tpm_id"`
+}
+
+type identityTPMAPIModel struct {
+	Name           string `json:"name" mapstructure:"name"`
+	TPMEKPublicKey string `json:"tpm_ek_public_key" mapstructure:"tpm_ek_public_key"`
+	Disabled       bool   `json:"disabled" mapstructure:"disabled"`
+	TPMID          string `json:"id" mapstructure:"id"`
```
ncabatoff: There's also a metadata map on this type.

internal/vault/identity/tpm_group_resource.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource_test.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource_test.go
gauravmitra1: Fixed

internal/vault/identity/tpm_resource.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource.go
```diff
@@ -0,0 +1,347 @@
...
+	token.TokenModel
+
+	Mount       types.String `tfsdk:"mount"`
+	Name        types.String `tfsdk:"name"`
+	DisplayName types.String `tfsdk:"display_name"`
+	TPMIDs      types.Set    `tfsdk:"tpm_ids"`
+	TPMGroupIDs types.Set    `tfsdk:"tpmgroup_ids"`
+}
+
+type tpmRoleAPIModel struct {
```
ncabatoff, line 57: Need to add `cert_ttl`.

internal/vault/auth/tpm/tpm_role_resource_test.go
```diff
@@ -179,6 +179,28 @@ resource "vault_tpm_auth_backend_role" "test" {
...
+MCowBQYDK2VwAyEAu5YIWbS0JtKO6mgJrmMa24RHTACn2BF3OOd9N7BxtIA=
+-----END PUBLIC KEY-----
+EOT
+}
+
+resource "vault_tpm_auth_backend_role" "test" {
+  mount        = vault_auth_backend.tpm.path
+  name         = %q
+  tpm_ids      = [vault_identity_tpm.test.tpm_id]
+  tpmgroup_ids = ["example-group-id"]
```
ncabatoff: Why not use the same approach here as you do for tpm_ids?  It's also ok to skip the vault_identity_tpm.test resource in this config and just use a made-up tpmid, but I'd like it to be consistent.

internal/vault/identity/tpm_resource_test.go
```diff
@@ -56,9 +58,25 @@ func TestAccIdentityTPM(t *testing.T) {
 				},
 			},
 			{
-				Config: testAccIdentityTPMConfig(name, tpmPublicKeyOne, true),
+				Config: testAccIdentityTPMConfig(renamedName, tpmPublicKeyOne, false),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "name", renamedName),
+					resource.TestCheckResourceAttr(resourceName, "disabled", "false"),
+					resource.TestCheckResourceAttrSet(resourceName, "tpm_id"),
```
ncabatoff: Can we assert this has the same value as the old tpm_id?  Just to distinguish between cases "renamed" and "created new tpm with renamedName".

gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource_test.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_role_resource.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_auth_backend_config_resource_test.go
```diff
@@ -0,0 +1,69 @@
...
+			acctestutil.TestEntPreCheck(t)
+			acctestutil.SkipIfAPIVersionLT(t, provider.VaultVersion210)
+		},
+		ProtoV5ProviderFactories: providertest.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccTPMAuthBackendConfigConfig(mount, "10m"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "mount", mount),
+					resource.TestCheckResourceAttr(resourceName, "cert_ttl", "10m"),
```
ncabatoff: Since cert_ttl isn't a config field (it shoudl be default_cert_ttl), this makes me think you haven't tested these changes locally.  Or you did, but you didn't tweak the PreCheck and your vault version was interepreted as <2.1.0.

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
```diff
@@ -0,0 +1,257 @@
...
+
+type TPMAuthBackendConfigResource struct {
+	base.ResourceWithConfigure
+}
+
+type TPMAuthBackendConfigModel struct {
+	base.BaseModel
+
+	Mount   types.String `tfsdk:"mount"`
+	CertTTL types.String `tfsdk:"cert_ttl"`
```
ncabatoff: There are actually 3 config fields, and none of them are cert_ttl.

internal/vault/auth/tpm/tpm_auth_backend_config_resource_test.go
gauravmitra1: Fixed

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
gauravmitra1: Fixed

ncabatoff (commented):
Looks good, but let's hold off on merging it for a little while longer, in case we find more things we want to tweak in tpm.

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
```diff
@@ -0,0 +1,320 @@
...
+		return
+	}
+
+	if !r.Meta().IsAPISupported(provider.VaultVersion220) {
+		resp.Diagnostics.AddError(
+			"Feature Not Supported",
+			fmt.Sprintf("TPM auth backend requires Vault version %s or later. Current Vault version: %s", provider.VaultVersion220, r.Meta().GetVaultVersion().String()),
+		)
+		return
+	}
```
hellobontempo, line 104: Followed the pattern in `ca_generated_resource.go` and `os_secret_backend_host_server.go` and added version gate to the `Create` function:
https://github.com/hashicorp/terraform-provider-vault/blob/3aa27da33f03a96486d1a1aa0264a44c3af42dbe/internal/vault/secrets/kmip/ca_generated_resource.go#L137-L145

https://github.com/hashicorp/terraform-provider-vault/blob/522d38b63002f0a8cbc81eb5177d74d9b9d03c35/internal/vault/secrets/os/os_secret_backend_host_resource.go#L257-L265

internal/consts/consts.go
```diff
@@ -1009,7 +1038,8 @@ const (
 	VaultVersion200  = "2.0.0"
 	VaultVersion201  = "2.0.1"
 	VaultVersion203  = "2.0.3"
-	VaultVersion210  = "2.1.0-beta1"
+	VaultVersion210  = "2.1.0"
```
ncabatoff, line 1041: Might want to pull in changes from main, I believe these are already present there.

internal/vault/identity/tpm_resource.go
```diff
@@ -106,6 +105,15 @@ func (r *IdentityTPMResource) Create(ctx context.Context, req resource.CreateReq
 		return
 	}
 
+	// Check if Vault version supports TMP auth (requires 2.2.0+)
```
ncabatoff: I can't seem to make a suggestion, not sure why.  Anyway, typo (TPM not TMP).

website/docs/r/identity_tpm.html.md
```diff
@@ -0,0 +1,68 @@
...
+  name             = "production-server-01"
+  tpm_ek_public_key = <<-EOT
+    -----BEGIN PUBLIC KEY-----
+    MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA...
+    -----END PUBLIC KEY-----
+  EOT
+  
+  disabled = false
+  
+  metadata = {
```
ncabatoff, line 29: I realized in reviewing this that we don't test metadata in our tpm resource tests, could you add coverage for that please?

website/docs/r/identity_tpm.html.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault/index.html#namespace).
+   *Available only for Vault Enterprise*.
```
ncabatoff: Everything tpm related is ent-only, seems weird to mention it for this field in particular.

website/docs/r/identity_tpm.html.md
```diff
@@ -0,0 +1,68 @@
...
+
+* `name` - (Required) Name of the TPM record.
```
ncabatoff: Not required.

website/docs/r/identity_tpm_group.html.md
```diff
@@ -0,0 +1,72 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace` - (Optional) The namespace to provision the resource in.
+  The value should not contain leading or trailing forward slashes.
+  The `namespace` is always relative to the provider's configured [namespace](/docs/providers/vault/index.html#namespace).
+   *Available only for Vault Enterprise*.
+
+* `name` - (Required) Name of the TPM group.
```
ncabatoff: Again not required, vault will generate a name if not supplied.

website/docs/r/tpm_auth_backend_role.html.md
```diff
@@ -0,0 +1,121 @@
...
+* `mount` - (Required) TPM auth backend mount path.
+
+* `name` - (Required) Name of the TPM role.
+
+* `display_name` - (Optional) Display name for the role. Defaults to the role name.
+
+* `cert_ttl` - (Optional) Certificate TTL for the TPM role. Accepts duration format strings
+  (e.g., "720h", "30d"). If not specified, uses the backend's default.
+
+* `tpm_ids` - (Optional) Set of TPM record IDs authorized to authenticate with this role.
```
ncabatoff, line 65: Both this field and tpmgroup_ids are optional, but at least one of them must be provided.

website/docs/r/tpm_auth_backend_role.html.md
```diff
@@ -0,0 +1,121 @@
...
+  These are the unique IDs assigned by Vault to TPM records (SHA256 of the EK public key).
+
+* `tpmgroup_ids` - (Optional) Set of TPM group IDs authorized to authenticate with this role.
+
+### Common Token Arguments
```
ncabatoff, line 70: Missing alias_metadata

website/docs/r/identity_tpm.html.md
hellobontempo: I agree, I was using SPIFFE docs as an example written and wasn't sure if this was just standard in TFVP?Here: https://github.com/hashicorp/terraform-provider-vault/blob/8e6af7b258d717580982682c53cf7b67234681c3/website/docs/r/spiffe_auth_backend_config.html.md#L44-L47

But it is probably just copy/pasta. I'll move the enterprise requirement to the important note at the top:
https://github.com/hashicorp/terraform-provider-vault/blob/6dc2fba331c4e37cf8ccd28eb5940c2756210714/website/docs/r/identity_tpm.html.md#L14

internal/consts/consts.go
hellobontempo: I'm not sure I follow you, I have the latest changes from main. After removing Gaurav's `-beta` change, there is no longer a diff for the VaultVersion consts. For the Field consts above, Gaurav had added some of them, but I added more and grouped them together in commit https://github.com/hashicorp/terraform-provider-vault/pull/2979/changes/a3023b6883c31c8043783fa0400f0386452a91fd 

I grouped them which I noticed someone did it for the SAML auth fields: 
https://github.com/hashicorp/terraform-provider-vault/blob/a3023b6883c31c8043783fa0400f0386452a91fd/internal/consts/consts.go#L833-L834

internal/vault/identity/tpm_resource.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (r *IdentityTPMResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_identity_tpm"
+}
+
+func (r *IdentityTPMResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			consts.FieldName: schema.StringAttribute{
+				Required:    true,
```
hellobontempo: If name is not required, should I change this to `Optional: true` @ncabatoff ? Same question for the tpm group resource

internal/consts/consts.go
ncabatoff: Yeah, I was just talking about the VaultVersion stuff.

internal/vault/identity/tpm_resource.go
ncabatoff: Yes.  Also Computed I think.

website/docs/r/tpm_auth_backend_role.html.md
hellobontempo: Great catch!

internal/vault/identity/tpm_resource.go
hellobontempo: Good call - looks like it based on this https://hashicorp.atlassian.net/wiki/spaces/VAULT/pages/2661089281/TFVP+Development+Best+Practices#When-to-Use-Computed:-true

internal/vault/identity/tpm_resource.go
hellobontempo: 🤦 Fixed

internal/vault/identity/tpm_resource.go
```diff
@@ -127,22 +127,12 @@ func (r *IdentityTPMResource) Create(ctx context.Context, req resource.CreateReq
...
 	}
 
-	resourcePath := r.path(&data)
-	if _, err := vaultClient.Logical().WriteWithContext(ctx, resourcePath, requestBody); err != nil {
-		resp.Diagnostics.AddError(errutil.VaultCreateErr(err))
-		return
-	}
-
-	secret, err := vaultClient.Logical().ReadWithContext(ctx, resourcePath)
+	resourcePath := "identity/tpm"
```
hellobontempo, line 145: Confirmed newly added test coverage errors prior to these changes

```
=== RUN   TestAccIdentityTPM
tpm_resource_test.go:63: Vault server version "2.2.0+ent"
tpm_resource_test.go:60: Step 1/6 error: Error running apply: exit status 1

        Error: Unable to Create Resource

          with vault_identity_tpm.test,
          on terraform_plugin_test.tf line 12, in resource "vault_identity_tpm" "test":
          12: resource "vault_identity_tpm" "test" {

        An unexpected error occurred while attempting to read the resource. Please
        retry the operation or report this issue to the provider developers.

        HTTP Error: Error making API request.

        URL: PUT http://127.0.0.1:8200/v1/identity/tpm/name
        Code: 405. Errors:

        * 1 error occurred:
        	* unsupported operation


--- FAIL: TestAccIdentityTPM (0.60s)
FAIL
FAIL	github.com/hashicorp/terraform-provider-vault/internal/vault/identity	1.606s

**Comment by hellobontempo** [comment]
The TFVP PR review Bob skill provided the following feedback, I've iterated on the skill's feedback several times so I've included reasoning for why I dismissed most of it. Number 3 was the only new addition and I addressed it in 16b8154f818e26626f6426ac0b05d021f3cebe5a

Here's the review summary in markdown:

---

## PR Review: VAULT-46452 — TPM Auth Backend Support

*4 new resources · Plugin Framework · Reviewed against TFVP checklist*

### Findings Detail

**1 — `omitempty` on `ca_lifetime` / `ca_soft_expiry`**
Raised the concern that setting either field to `0` would silently drop it from the request body sent to Vault.
> **Dismissed.** `0` is not a meaningful value for these fields — Vault ignores it and keeps its own default. The `omitempty` is intentional. `default_cert_ttl` deliberately lacks `omitempty` because `0` is a valid value that means "inherit from backend config."

---

**2 — `identity_tpm` Create: no explicit post-write Read**
`IdentityTPMResource.Create` populates state directly from the Create response rather than doing a follow-up `ReadWithContext`.
> **Dismissed.** The `identity/tpm` POST endpoint returns the full resource representation in its response. The acceptance tests confirm this: Step 1 of `TestAccIdentityTPM` checks all fields and asserts `ExpectEmptyPlan()` on post-apply refresh, which would detect any drift if the Create response were incomplete.

---

**3 — `TPMAuthRoleResource.ImportState` performs a redundant Vault read**
`ImportState` was calling `client.GetClient`, reading the role from Vault, and calling `resp.State.Set` — resulting in two Vault reads on every import and diverging from the pattern used by every other two-identifier resource in the codebase (e.g. `CFAuthBackendRoleResource`, `kerberosAuthBackendGroupResource`).
> **Fixed.** Removed the extra Vault read. `ImportState` now only parses the import ID, sets `mount` and `name` via `resp.State.SetAttribute`, and optionally sets `namespace` from the environment variable. The framework calls `Read` automatically afterward to populate the rest.

---

**4 — `tpmgroup_ids` field name**
The name `tpmgroup_ids` (no underscore between "tpm" and "group") looks inconsistent relative to `tpm_ids` and `member_tpm_ids`.
> **Dismissed.** The field is correct. This is validated by the test in `tpm_role_resource_test.go` that creates a `vault_identity_tpm_group` resource, sets `tpmgroup_ids` on the role, and then checks `tpmgroup_ids.# = "1"` with `ExpectEmptyPlan()` on post-apply refresh. If the field name didn't match what Vault expects, either the write would silently drop it (Vault would return an empty list) or Vault would error. Either way the test would fail.
---

**5 — Version check only in Create, not Update**
> **Dismissed.** Matches the established pattern for all other version-gated resources in the provider.

---

**6 — Config Delete warning text**
`TPMAuthBackendConfigResource.Delete` emits a warning rather than deleting anything from Vault.
> **Dismissed.** Exactly mirrors the `RadiusAuthBackendConfigResource.Delete` pattern. The TPM auth config endpoint cannot be independently cleared from its mount, so a state-only removal with a warning is the correct behavior.

---

**7 & 8 — `metadata` null-vs-empty-map and set count assertions**
Raised concerns about potential drift between a null config value and an empty map returned from Vault, and whether checking `member_tpm_ids.#` = `"0"` against a null set was fragile.
> **Dismissed.** Both behaviors are directly covered by acceptance tests with `ExpectEmptyPlan()` on post-apply refresh. `TestAccIdentityTPM_metadata` and `TestAccIdentityTPMGroup_metadata` both verify that omitting `metadata` produces no drift.

---

**9 — Inline path in `ImportState` instead of `r.path()`**
> **Dismissed (moot).** Code removed as part of the fix for item 3.

---

**10 — `FieldDisabled` constant**
Queried whether `FieldDisabled = "disabled"` already existed in `consts.go`.
> **Dismissed.** The constant is genuinely new, it did not exist before this PR.

---

**11 — Changelog blank line**
> **Dismissed.** Not actually a problem.

---

**12 & 13 — `alias_metadata` documentation and token schema**
Queried whether `alias_metadata` was actually present in the schema for the TPM role resource.
> **Dismissed.** `tokenSchema()` in `internal/framework/token/token.go` defines `alias_metadata` and `MustAddBaseAndTokenSchemas` merges it into any resource that calls it. The documentation is accurate.

---

### What Looks Good

- All four resources have compile-time `resource.ResourceWithImportState` assertions
- Read-after-Write in `TPMAuthRoleResource` Create and Update
- `resp.State.RemoveResource(ctx)` on 404 in all Read functions
- `resource.ParallelTest` used throughout; `TestEntPreCheck` and `SkipIfAPIVersionLT` correctly gate all tests
- `stringplanmodifier.RequiresReplaceIf` with PEM normalization for `tpm_ek_public_key`
- `stringplanmodifier.UseStateForUnknown()` on computed IDs (`tpm_id`, `tpm_group_id`)
- Token fields wired via shared `token.MustAddBaseAndTokenSchemas` / `PopulateTokenAPIFromModel`
- Import steps in all tests with `ImportStateVerify: true` and `ExpectEmptyPlan()` drift detection
- Provider registration, docs, sidebar, and changelog all present
- No hardcoded credentials or sensitive values

File: internal/vault/auth/tpm/tpm_auth_backend_role_resource.go [hunk 23]
```diff
@@ -0,0 +1,369 @@
...
+				Optional:    true,
+				Computed:    true,
+				Description: "Display name for the role. Defaults to the role name.",
+			},
+			consts.FieldCertTTL: schema.Int64Attribute{
+				Optional:    true,
+				Computed:    true,
+				Description: "Certificate TTL for the TPM role, in seconds. Zero means inherit the global `default_cert_ttl` from the backend config.",
+			},
+			consts.FieldTPMIDs: schema.SetAttribute{
```
ncabatoff, line 99: Can we add a validator that ensures that at least one of tpmids or tpmgroupids is present?

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
```diff
@@ -0,0 +1,369 @@
+// Copyright IBM Corp. 2016, 2026
```
ncabatoff, line 1: Given the name of the resource, probably the filename should be tpm_auth_backend_role_resource.go.

ncabatoff (commented):
PR is looking almost good to go.  Love how thorough your tests are!

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
hellobontempo: Oof, yeah doesn't look like we're consistent with naming conventions 🫠 I'll rename

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
```diff
@@ -0,0 +1,292 @@
...
+	configPath := r.path(&data)
+	configResp, err := vaultClient.Logical().WriteWithContext(ctx, configPath, requestBody)
+	if err != nil {
+		resp.Diagnostics.AddError(errutil.VaultCreateErr(err))
+		return
+	}
+	if configResp == nil {
+		resp.Diagnostics.AddError(errutil.VaultReadResponseNil())
+		return
+	}
```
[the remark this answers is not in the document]
hellobontempo: done in 4983c935e0083b5280ec78aa636bac41db35248c

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
hellobontempo: Done https://github.com/hashicorp/terraform-provider-vault/pull/2979/changes/55b56bfb24795b362a955f868dfd2dda5e012678

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
hellobontempo: Added in 6f94e7bc2903dfe3b1f33354e0b95aa5d1243b78!

ncabatoff (dismissed):

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
```diff
@@ -0,0 +1,302 @@
...
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Check if Vault version supports TPM auth (requires 2.2.0+)
+	if !r.Meta().IsAPISupported(provider.VaultVersion220) {
+		resp.Diagnostics.AddError(
+			"Feature Not Supported",
+			fmt.Sprintf("TPM auth backend requires Vault version %s or later. Current Vault version: %s", provider.VaultVersion220, r.Meta().GetVaultVersion().String()),
```
robmonte: `GetVaultVersion` is capable of returning a nil so this could panic. Instead of saying what the current version is in the message, I think it's best to stick to only stating the required minimum Vault version to avoid the potential nil dereference.

Similar comment for several other places in the PR where the `VaultVersion220` is checked:
* TPMAuthBackendRoleResource
* IdentityTPMGroupResource
* IdentityTPMResource

robmonte (commented):
I only have the one finding and a question.

I notice there's no Enterprise checks with `IsEnterpriseSupported()`, should there be any for these features?

internal/vault/auth/tpm/tpm_auth_backend_config_resource.go
hellobontempo: Great catch! Completely missed the enterprise check, added those and removed the current version read. Thank you!

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
```diff
@@ -0,0 +1,387 @@
...
+			consts.FieldCertTTL: schema.Int64Attribute{
+				Optional:    true,
+				Computed:    true,
+				Description: "Certificate TTL for the TPM role, in seconds. Zero means inherit the global `default_cert_ttl` from the backend config.",
+			},
+			consts.FieldTPMIDs: schema.SetAttribute{
+				ElementType: types.StringType,
+				Optional:    true,
+				Validators: []validator.Set{
+					setvalidator.AtLeastOneOf(
```
robmonte, line 103: Sets can be an empty set so `setvalidator.SizeAtLeast(1)` should be added to the list of validators.

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
```diff
@@ -0,0 +1,387 @@
...
+						path.MatchRoot(consts.FieldTPMGroupIDs),
+					),
+				},
+				Description: "Set of TPM record IDs authorized to authenticate with this role.",
+			},
+			consts.FieldTPMGroupIDs: schema.SetAttribute{
+				ElementType: types.StringType,
+				Optional:    true,
+				Validators: []validator.Set{
+					setvalidator.AtLeastOneOf(
```
robmonte, line 115: `setvalidator.SizeAtLeast(1)` here too

internal/vault/identity/tpm_group_resource.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+	}
+
+	// Update data model with values from Vault
+	data.Name = types.StringValue(readResp.Name)
+	data.TPMGroupID = types.StringValue(readResp.TPMGroupID)
+
+	if len(readResp.MemberTPMIDs) == 0 {
+		data.MemberTPMIDs = types.SetNull(types.StringType)
+	} else {
```
robmonte, line 339: Similar to the above comments, since it can be an empty set, we shouldn't default to a null value when the length is zero. 

```golang
if len(readResp.MemberTPMIDs) == 0 {
  if data.MemberTPMIDs.IsNull() {
      data.MemberTPMIDs = types.SetNull(types.StringType)
  } else {
      data.MemberTPMIDs, _ = types.SetValueFrom(ctx, types.StringType, []string{})
  }
```

You can set it to the empty set if the length is zero but not null.

internal/vault/auth/tpm/tpm_auth_backend_role_resource.go
hellobontempo: Ah, I thought we needed to support `= []` explicitly to clear the field in Vault, but I realize omitting the field from the config already achieves that.

internal/vault/identity/tpm_group_resource.go
hellobontempo: Makes sense! Missed this when I picked up the PR from the original author. Thanks for catching.

robmonte (approved):
Looks good to me!

PR #2988, opened 2026-08-07
feat: set a standard Terraform User-Agent on all Vault requests

### LINKED ISSUE #2987: [Bug]: Provider does not respect TF_APPEND_USER_AGENT convention, instead uses default Golang User-Agent
### Terraform Core Version

1.15.8

5.10.1

### Vault Server Version

2.0.3

### Affected Resource(s)

All resources

### Expected Behavior

When using the vault Terraform provider, the `User-Agent` is set to the default Terraform user-agent which includes the Terraform version, the provider version, etc. Additionally, if the `TF_APPEND_USER_AGENT` environment variable is set, the value is included on each request.

### Actual Behavior

All requests are made using the default Golang `User-Agent` header: `Go-http-client/1.1`.

N/A

```hcl
terraform {
  required_providers {
    vault = {
      source  = "hashicorp/vault"
      version = "5.10.1"
    }
  }
}

provider "vault" {
  # address and token come from VAULT_ADDR / VAULT_TOKEN
  skip_child_token       = true
  skip_get_vault_version = true
}

resource "vault_mount" "test" {
  path = "ua-repro"
  type = "kv"
}
```

#### 1. Start a Vault dev server

```bash
vault server -dev -dev-root-token-id=root -dev-listen-address=127.0.0.1:8200
```

Leave this running. In a second shell:

```bash
export VAULT_ADDR=http://127.0.0.1:8200
export VAULT_TOKEN=root
```

#### 2. Enable a file audit device

```bash
vault audit enable file file_path=/tmp/vault-audit.log
```

#### 3. Ensure `user-agent` is an audited request header

```bash
vault write sys/config/auditing/request-headers/user-agent hmac=false
```

Vault only records request headers that are explicitly registered here, and `hmac=false` is what
makes the value readable rather than hashed. On Vault v2.0.3 this is **already the default** —
`vault read sys/config/auditing/request-headers` shows `user-agent` (plus `correlation-id` and
`x-correlation-id`) pre-registered with `hmac:false` — so the command above is a harmless
re-assertion. Including it anyway for older Vault versions, where the header is not audited by
default and the log will show `"headers": null` without it.

#### 4. Run Terraform
```bash
terraform init
terraform apply -auto-approve
```

#### 6. Read the User-Agent out of the audit log

```bash
jq -r 'select(.type=="request") | "\(.request.path)\t\(.request.headers["user-agent"] // "-")"' /tmp/vault-audit.log
sys/mounts/ua-repro	["Go-http-client/1.1"]
```

#### 7. Repeat with environment variables
Repeating the above steps with various environment variable set:
| Env | Resulting `user-agent` |
|---|---|
| *(none)* | `Go-http-client/1.1` |
| `TF_APPEND_USER_AGENT=my-wrapper/9.9.9` | `Go-http-client/1.1` — **ignored** |
| `VAULT_HEADERS={"User-Agent":"my-tooling/1.2.3"}` | `my-tooling/1.2.3` |
| both of the above | `my-tooling/1.2.3` |

### Debug Output
N/A

### Panic Output
N/A

### Important Factoids
Because the Vault SDK is used to construct the request, the `VAULT_HEADERS` environment variable _is_ respected and can be used a provider specific work-around by setting `VAULT_HEADERS={"User-Agent":"..."}`. 

### References

https://github.com/hashicorp/vault/issues/31063

Yes

lyoung-confluent:
### Description
Closes #2987 by setting a default Terraform `User-Agent` header on each Vault request, ex:
```
Terraform/1.15.8 (+https://www.terraform.io) Terraform-Plugin-SDK/2.40.1 terraform-provider-vault/5.10.1
```

### Checklist

```

...
```

### Community Note

## PCI review checklist

### DIFF
--- a/internal/provider/meta.go
+++ b/internal/provider/meta.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/hashicorp/terraform-provider-vault/helper"
 	"github.com/hashicorp/terraform-provider-vault/internal/consts"
+	providerversion "github.com/hashicorp/terraform-provider-vault/version"
 )
 
 const (
@@ -69,7 +70,23 @@ type ProviderMeta struct {
 	resourceData *schema.ResourceData
 	clientCache  map[string]*api.Client
 	vaultVersion *version.Version
+	userAgent    string
 	mu           sync.RWMutex
+}
+
+// getUserAgent returns the User-Agent to send with every Vault request.
+//
+// The value is normally set during provider configuration, where the Terraform
+// version is known. ProviderMeta can also be constructed directly (notably in
+// tests), so fall back to building the string without it rather than sending no
+// User-Agent at all. Both paths go through schema.Provider.UserAgent, which
+// appends TF_APPEND_USER_AGENT when it is set.
+func (p *ProviderMeta) getUserAgent() string {
+	if p.userAgent != "" {
+		return p.userAgent
+	}
+
+	return (&schema.Provider{}).UserAgent(providerversion.ProviderName, providerversion.ProviderVersion())
 }
 
 // GetClient returns the providers default Vault client.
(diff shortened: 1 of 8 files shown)


internal/provider/meta.go
```diff
@@ -268,6 +285,16 @@ func (p *ProviderMeta) setClient() error {
 			parsedHeaders.Add(name.(string), header["value"].(string))
```
Chris-Paris-HashiCorp, line 285: Not really part of your PR, but I notice that we're `Add`ing headers to the `parsedHeaders`, so if there is overlap in the keys between parsedHeaders and headers, we'll have duplicates. Should this be `parsedHeaders.Set`?

Chris-Paris-HashiCorp (approved):
Changes LGTM. But one preexisting thing I noticed:

PR #3006, opened 2026-08-18
VAULT-49205: vault_terraform_cloud_secret_backend: return early if mount removed 

Himnish-Nadiminti:
### Description

This PR updates `terraformCloudSecretBackendRead` so that if the secrets mount has been deleted from Vault out-of-band, the Read function returns immediately after `readMount` clears the resource ID, instead of falling through to a stray `GET <backend>/config` call against the deleted mount. It also adds a `util.Is404` guard to `Delete` so that `terraform destroy` succeeds cleanly when the mount has already been removed from Vault.

Closes [VAULT-46286](https://hashicorp.atlassian.net/browse/VAULT-46286?atlOrigin=eyJpIjoiZWMzNmIyZTViZmMzNDYzM2E1Yzk3OWQ0NTU2MzY5NjQiLCJwIjoiaiJ9)

## Read fix

When a `vault_terraform_cloud_secret_backend` mount is deleted from Vault out-of-band, `readMount` correctly sets `d.SetId("")` to signal the resource is gone — but returns `nil` with no error, causing execution to fall through to a stray `GET <backend>/config` call against the now-deleted mount.

This adds a 3-line early return guard after `readMount`: if the mount check already cleared the resource ID, we return immediately. The config read is never attempted.

* **Before:** `readMount` removes the resource from state, execution falls through, `GET /v1/terraform/config` hits the deleted mount, resource is silently removed via the config-level nil check.
* **After:** `readMount` removes the resource from state, early return fires, no further requests are made.

## Delete fix

Without the `util.Is404` guard, `terraform destroy` would error with `Error unmounting Terraform Cloud backend` if the mount was already deleted out-of-band. The guard treats a 404 on unmount as a no-op and removes the resource from state cleanly.

## Unit tests
image

## go test
image

### Checklist

```

...
```
image

### Community Note

## PCI review checklist

[VAULT-46286]: https://hashicorp.atlassian.net/browse/VAULT-46286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bcsamrudh (dismissed):
LGTM

stuti-sr (changes requested):
none of the test cases are passing in shared snapshot

vault/resource_terraform_cloud_secret_backend.go
stuti-sr, line 361: util.Is404(err) is missing.
eg : resource_approle_auth_backend_role.go:321

stuti-sr, line 215: util.Is404(err) is missing.
eg : resource_approle_auth_backend_role.go:321

Himnish-Nadiminti: The code already handles it perfectly:

readMount → mount gone → d.SetId("") → our early return fires → done
If mount exists but /config returns 404 → ParseRawResponseAndCloseBody returns (nil, nil) → err == nil → falls to if secret == nil → d.SetId("") → done
Both 404 paths are already covered. The util.Is404 guard on Read is not required.

Himnish-Nadiminti: Added the guard as done in the other file.
Along with unit test (positive and negative)

vault/resource_terraform_cloud_secret_backend_test.go
```diff
@@ -269,3 +273,201 @@ resource "vault_terraform_cloud_secret_backend" "test" {
   token_wo_version  = %d
 }`, path, version)
 }
+
+// newUnitTestMeta spins up an httptest server with the given handler, creates
+// a Vault API client pointed at it, and returns a *provider.ProviderMeta
+// wired to that client (skip_get_vault_version=true).
+func newUnitTestMeta(t *testing.T, handler http.HandlerFunc) interface{} {
```
stuti-sr: naming convention is not correct.

Himnish-Nadiminti: Updated the naming convention.

suraj-simha (commented):
LGTM, Can we add a changelog?

vivekpandeyHashi (approved):

stuti-sr (approved):
Looks good to me!!

bcsamrudh (approved):
LGTM

PR #3007, opened 2026-08-18
VAULT-18118/suppress kv-v2 type diff on import to prevent ForceNew

sgsource:
## Problem

Importing a `kv-v2` mount caused a destructive `ForceNew` destroy on the next `terraform plan`.

Vault always returns `type=kv` + `options.version=2` for kv-v2 mounts. On a cold import, `readMount`'s normalization block (which rewrites `kv` → `kv-v2`) is gated on `d.GetOk(FieldType)`, which returns `false` when `d` is empty. So `kv` lands raw in state. Since `FieldType` has `ForceNew: true`, the next plan sees `state=kv` vs `config=kv-v2` and schedules a destroy.

## Fix

Add `DiffSuppressFunc` on two fields in `MountResource()`:

- **`FieldType`**: suppresses `kv` ↔ `kv-v2` diff, but only in the `state=kv, config=kv-v2` direction and only when `options.version == "2"` is present in state. This prevents silently hiding a genuine v1/v2 misconfiguration.
- **`FieldOptions`**: suppresses the `options.version` removal and map count diff when config uses `kv-v2` without an explicit `options` block.

## Tests

9 acceptance tests in `import_mount_test.go` covering:

| Case | Scenario |
|---|---|
| 1 | kv v1 import baseline — no suppressor fires |
| 2 | non-kv mount import — no suppressor fires |
| 3 | core bug: `kv-v2` config import → no destroy |
| 4 | `kv`+`version=2` explicit config import → no destroy |
| 5 | kv v1 mount, suppressor correctly doesn't fire |
| 6 | v1 mount + `kv-v2` config → real ForceNew surfaces |
| 7 | `kv-v2` + redundant `version=2` → plan clean |
| 8 | state=`kv-v2`, config changes to `kv`+`version=2` → ForceNew correct (suppressor one-way only) |
| 9 | update unrelated field on `kv-v2` → no type drift on subsequent plans |

Fixes: VAULT-18118


vault/resource_mount.go
```diff
@@ -183,6 +183,50 @@ func getMountSchema(excludes ...string) schemaMap {
 }
 
 func MountResource() *schema.Resource {
+	rSchema := getMountSchema()
+
+	// on cold import, vault returns type=kv+opts.version=2; config says kv-v2.
+	// suppress that one-way alias only — never suppress kv-v2→kv (would hide a downgrade)
+	if tyField, ok := rSchema[consts.FieldType]; ok {
```
vinay-gopalan, line 190: Instead of adding these at the top of the resource, should we instead specify these on the fields themselves where the schema is defined? (within `getMountSchema`) That is typically the pattern I have seen. For example:
```
consts.FieldType: {
			Required:    true,
			ForceNew:    true,
			Description: "Type of the backend, such as 'aws'",
			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {}
		}
```

sgsource: Hey, thanks for noting this. That's what I thought originally. That's the convention (usually) instead of surgically injecting this chunk of code. However, `getMountSchema` is referred to in other files like `resource_aws_secret_backend`, `resource_gcp_secret_backend`, `resource_consul_secret_backend`, and I was thinking since this fix is only for Terraform kv-v2, we should try to limit the scope to here.

fairclothjm (commented):
Can we add a changelog entry?

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ IMPROVEMENTS:
 
 BUG FIXES:
 
+* `vault_mount`: Fix spurious `ForceNew` destroy when importing a `kv-v2` mount. Vault returns `type=kv`+`options.version=2` on import; a `DiffSuppressFunc` now suppresses the `kv`/`kv-v2` alias diff so the next plan is clean. ([#3007](https://github.com/hashicorp/terraform-provider-vault/pull/3007))
```
fairclothjm: This needs to go under the `Unreleased` section above

vault/resource_mount.go
```diff
@@ -183,6 +183,50 @@ func getMountSchema(excludes ...string) schemaMap {
...
+				return optMap["version"] == "2"
+			}
+			return false
+		}
+	}
+
+	if optField, ok := rSchema[consts.FieldOptions]; ok {
+		optField.DiffSuppressFunc = func(key, stateVal, cfgVal string, d *schema.ResourceData) bool {
+			cfgTy := d.Get(consts.FieldType).(string)
+			isKVMount := cfgTy == "kv" || cfgTy == "kv-v2"
```
fairclothjm: nit: Should we just short-circuit here and return early if we are not kv or kv-v2? Then I think we wouldn't need the `isKVMount` checks below?

vault/resource_mount.go
```diff
@@ -183,6 +183,50 @@ func getMountSchema(excludes ...string) schemaMap {
 }
 
 func MountResource() *schema.Resource {
+	rSchema := getMountSchema()
+
+	// on cold import, vault returns type=kv+opts.version=2; config says kv-v2.
+	// suppress that one-way alias only — never suppress kv-v2→kv (would hide a downgrade)
+	if tyField, ok := rSchema[consts.FieldType]; ok {
+		tyField.DiffSuppressFunc = func(key, stateTy, cfgTy string, d *schema.ResourceData) bool {
+			if stateTy == "kv" && cfgTy == "kv-v2" {
```
fairclothjm: nit: We could invert this if check. In Go it's generally considered more idiomatic to handle the exceptional case first and return early, reducing nesting and keeping the happy path less indented.

vault/resource_mount.go
vinay-gopalan: That makes sense, thanks for the noting the tradeoffs

vault/resource_mount.go
sgsource: Thanks for pointing this out. Updated to be more idiomatic.

vault/resource_mount.go
sgsource: I agree. Updated. It looks so much better now. Thanks!

vault/import_mount_test.go
```diff
@@ -13,6 +14,9 @@ import (
 	"github.com/hashicorp/terraform-provider-vault/testutil"
 )
 
+const resourceName = "vault_mount.test"
+
+// Case 1: kv v1 import baseline.
```
fairclothjm: Can we remove these "Case 1", "Case 2", etc? That isn't really standard practice for a permanent test suite. Let's just have the test names and doc comments describe the scenario.

sgsource: I see why that doesn't make sense for an ongoing test suite. Will update.

fairclothjm (approved):
LGTM! Great work!

PR #3011, opened 2026-08-19
vault_cert_auth_backend_role: add write-only certificate support

Niyathi24:
### Description
This PR adds `certificate_wo` and `certificate_wo_version` write-only fields to the `vault_cert_auth_backend_role` resource, allowing users to supply a CA certificate via an ephemeral resource without it being stored in Terraform state. 

Relates [VAULT-44145](https://hashicorp.atlassian.net/browse/VAULT-44145)

### Checklist
- [] Acceptance tests where run against all supported Vault Versions

```
$ TF_ACC=1 go test ./vault/... -run TestAccCertAuthBackend_WriteOnly -v -timeout 30m -tags testonly
=== RUN   TestAccCertAuthBackend_WriteOnly
--- PASS: TestAccCertAuthBackend_WriteOnly (1.19s)
PASS
ok      github.com/hashicorp/terraform-provider-vault/vault     1.952s
```

### Community Note

## PCI review checklist

[VAULT-44145]: https://hashicorp.atlassian.net/browse/VAULT-44145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 5.12.0 (Unreleased)
 
+IMPROVEMENTS:
+
+* `vault_cert_auth_backend_role`: Add `certificate_wo` and `certificate_wo_version` write-only fields to allow ephemeral resource values, to supply the CA certificate. The `certificate` field is now `Computed` and `ForceNew` has been removed enabling in-place updates when the certificate changes instead of resource replacement.
+
+
 BUG FIXES:
 
 * `vault_terraform_cloud_secret_backend`: Fix logic gap in `Read` where execution would fall through to a stray `GET <backend>/config` call after `readMount` detected the mount was deleted out-of-band and cleared the resource ID. Add `util.Is404` guard to `Delete` so that `terraform destroy` succeeds cleanly when the mount has already been removed from Vault. ([#3006](https://github.com/hashicorp/terraform-provider-vault/pull/3006))
@@ -23,7 +28,6 @@ FEATURES:
 
 IMPROVEMENTS:
 
-* `vault_cert_auth_backend_role`: Add `certificate_wo` and `certificate_wo_version` write-only fields to support ephemeral resource values, preventing CA certificate data from being stored in Terraform state.
 * Migrated AWS provider dependency from `aws-sdk-go` (v1) to `aws-sdk-go-v2` for improved performance and maintainability. ([#2882](https://github.com/hashicorp/terraform-provider-vault/pull/2882))
 * `vault_identity_entity_alias`: Add support for `external_id` and `issuer` fields. Available only for Vault Enterprise. ([#2994](https://github.com/hashicorp/terraform-provider-vault/pull/2994))
 * `vault_aws_auth_backend_config_identity`: Add support for `canonical_arn` as a valid value for the `iam_alias` parameter. Requires Vault 1.16+. ([#2982](https://github.com/hashicorp/terraform-provider-vault/pull/2982))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
+* `vault_cert_auth_backend_role`: Add `certificate_wo` and `certificate_wo_version` write-only fields to support ephemeral resource values, preventing CA certificate data from being stored in Terraform state.
 * Migrated AWS provider dependency from `aws-sdk-go` (v1) to `aws-sdk-go-v2` for improved performance and maintainability. ([#2882](https://github.com/hashicorp/terraform-provider-vault/pull/2882))
 * `vault_identity_entity_alias`: Add support for `external_id` and `issuer` fields. Available only for Vault Enterprise. ([#2994](https://github.com/hashicorp/terraform-provider-vault/pull/2994))
 * `vault_aws_auth_backend_config_identity`: Add support for `canonical_arn` as a valid value for the `iam_alias` parameter. Requires Vault 1.16+. ([#2982](https://github.com/hashicorp/terraform-provider-vault/pull/2982))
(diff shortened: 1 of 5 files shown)


website/docs/r/cert_auth_backend_role.html.md
```diff
@@ -42,7 +62,9 @@ The following arguments are supported:
 
 * `name` - (Required string) Name of the role
 
-* `certificate` - (Required string) CA certificate used to validate client certificates
+* `certificate` - (Optional string) CA certificate used to validate client certificates. Conflicts with `certificate_wo`
```
ncabatoff: Can we document that exactly one of `certificate` and `certificate_wo` are required?  Right now it appears that both are optional.

vault/resource_cert_auth_backend_role.go
```diff
@@ -65,9 +65,23 @@ func certAuthBackendRoleResource() *schema.Resource {
 			ForceNew: true,
 		},
 		consts.FieldCertificate: {
-			Type:     schema.TypeString,
-			Required: true,
-			ForceNew: true,
```
ncabatoff, line 70: ForceNew does seem inappropriate here.  Did you do any digging to see if there was a reason behind it, like looking at the commit/PR that introduced it and seeing if there was any discussion?

vault/resource_cert_auth_backend_role.go
```diff
@@ -354,62 +384,74 @@ func certAuthResourceRead(_ context.Context, d *schema.ResourceData, meta interf
...
+		d.Set(consts.FieldCertificate, resp.Data[consts.FieldCertificate])
+	}
 	d.Set("display_name", resp.Data["display_name"])
 
 	// Vault sometimes returns these as null instead of an empty list.
 	if resp.Data["allowed_names"] != nil {
 		d.Set("allowed_names",
 			schema.NewSet(
-				schema.HashString, resp.Data["allowed_names"].([]interface{})))
+				schema.HashString, resp.Data["allowed_names"].([]interface{}),
```
ncabatoff: Please revert the gratuitous reformatting.

vault/resource_cert_auth_backend_role.go
Niyathi24: I traced `ForceNew: true` back to the original commit that created this file in 2018. It was added as part of the initial implementation with no specific discussion. Since `certAuthResourceUpdate` handles certificate updates in place, and removing ForceNew is required for certificate_wo rotation to work without destroying the resource, I thought it might be best to remove it.

ncabatoff: Yes, agreed, just applying Chesterton’s Fence.

Niyathi24: Makes sense, thank you!

vault/resource_cert_auth_backend_role.go
Niyathi24: It's updated now, I think it was automatically formatting it previously.

vault/resource_cert_auth_backend_role.go
```diff
@@ -354,7 +385,9 @@ func certAuthResourceRead(_ context.Context, d *schema.ResourceData, meta interf
 		return diag.FromErr(err)
 	}
 
-	d.Set("certificate", resp.Data["certificate"])
+	if _, usingWO := d.GetOk(consts.FieldCertificateWOVersion); !usingWO {
```
Balaji2198: `d.GetOk()` should not be used inside a Read function. During `terraform import`, the state is initially empty, so this guard can incorrectly skip setting `certificate`. Since the SDK already handles nullifying `WriteOnly` fields after apply, Can we please remove the condition and unconditionally set `certificate` from the API response?

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
+* `vault_cert_auth_backend_role`: Add `certificate_wo` and `certificate_wo_version` write-only fields to support ephemeral resource values, preventing CA certificate data from being stored in Terraform state.
```
Balaji2198: Can we move this changelog to the unreleased section?

vault/resource_cert_auth_backend_role.go
```diff
@@ -65,9 +65,24 @@ func certAuthBackendRoleResource() *schema.Resource {
 			ForceNew: true,
 		},
 		consts.FieldCertificate: {
```
Balaji2198, line 67: Removing `ForceNew` changes the resource lifecycle from replacement to in-place updates, which could affect existing customers relying on the current behaviour. If this change is intentional, can we please document it in the changelog and relevant docs?

vault/resource_cert_auth_backend_role.go
```diff
@@ -65,9 +65,24 @@ func certAuthBackendRoleResource() *schema.Resource {
...
+			Type:          schema.TypeString,
+			Optional:      true,
+			ConflictsWith: []string{consts.FieldCertificateWO},
+			AtLeastOneOf:  []string{consts.FieldCertificate, consts.FieldCertificateWO},
+		},
+		consts.FieldCertificateWO: {
+			Type:          schema.TypeString,
+			Optional:      true,
+			WriteOnly:     true,
+			Sensitive:     true,
```
Balaji2198: `certificate_wo` has both `WriteOnly: true` and `Sensitive: true`, which appear to be redundant/mutually exclusive semantics. Could we remove `Sensitive` and rely on `WriteOnly` here?

Balaji2198 (commented):
Thanks for working on this! Added few comments.

vault/resource_cert_auth_backend_role.go
Niyathi24: Done, removed the guard and set certificate unconditionally. Also added `Computed`: true to `certificate` to avoid perpetual drift when `certificate_wo` is used. Updated test and docs to match.

Balaji2198 (approved):

PR #3016, opened 2026-08-24
fix: update max_open_connections default from 2 to 4

Damini008:
## Description

Updates the default value of `max_open_connections` in the database 
secrets backend connection schema from `2` to `4`, aligning it with 
the documented default and Vault's own default for the same feature.

## Problem

The schema default was set to `2` but the provider documentation and 
Vault's native database secrets engine both specify `4` as the default. 
This mismatch caused confusion for users who relied on the documented 
default when configuring database connections without explicitly setting 
this value.

Reported in ZD 134520.

## Changes

- Updated `Default` value from `2` to `4` in `connectionStringResource` 
  schema definition
- Updated test assertions in `resource_database_secret_backend_connection_test.go` 
  to reflect the corrected default
- Updated test assertions in `resource_database_secrets_mount_test.go` 
  to reflect the corrected default

## Testing

- `go test ./vault` passes with no failures

Fixes VAULT-22953
Jira: https://hashicorp.atlassian.net/browse/VAULT-22953


fairclothjm (commented):
Can we add a changelog entry?

charlesn-hc (approved):

PR #3028, opened 2026-08-31
Add linux/s390x build target for IBM Z / LinuxONE

zisom-hc:
## Summary

Add `s390x` to the `goarch` build matrix in `build.yml` so the provider is compiled and released for `linux/s390x` (IBM Z / LinuxONE).

## Change

``.github/workflows/build.yml`` — 6 lines added, 1 line changed:

```yaml
goarch: [ "386", "amd64", "arm", "arm64", "s390x" ]   # + s390x
exclude:
  ...
  + - goos: darwin
  +   goarch: s390x
  + - goos: freebsd
  +   goarch: s390x
  + - goos: windows
  +   goarch: s390x
```

## Why the exclude entries are required

Go does not support `darwin/s390x`, `freebsd/s390x`, or `windows/s390x`. Running `go build` for any of those combinations exits with:

```
unsupported GOOS/GOARCH pair <os>/s390x
```

The GitHub Actions matrix would fail on those entries without the excludes — they are not optional guards, they are required to prevent build errors.

## Motivation

IBM Z / LinuxONE hardware runs `linux/s390x`. Users running Terraform on mainframes cannot install this provider today because no `linux/s390x` binary is included in releases. This change unblocks that use case with no impact on any existing platform.

## Testing

- `go tool dist list | grep s390x` confirms `linux/s390x` is a supported Go target
- `GOOS=linux GOARCH=s390x go build ./...` completes successfully against this codebase
- `GOOS=darwin GOARCH=s390x go build` confirmed to fail — validating the exclude entries are necessary


kpcraig (dismissed):

.github/workflows/build.yml
```diff
@@ -107,7 +107,7 @@ jobs:
       fail-fast: true
       matrix:
         goos: [ freebsd, windows, linux, darwin ]
-        goarch: [ "386", "amd64", "arm", "arm64" ]
+        goarch: [ "386", "amd64", "arm", "arm64", "s390x" ]
```
vijayavelsekar, line 110: I see that we need to update `.release/terraform-provider-vault-artifacts.hcl` file as well.

vijayavelsekar (changes requested):
I see that we need to update `.release/terraform-provider-vault-artifacts.hcl` file as well.

.github/workflows/build.yml
zisom-hc: Updated it now

zisom-hc:
> I see that we need to update `.release/terraform-provider-vault-artifacts.hcl` file as well.

updated it now; thanks for noticing that! Please re-review @vijayavelsekar / @kpcraig

vijayavelsekar (approved):
LGTM

PR #3030, opened 2026-09-02
Fix FlattenVaultDuration util to return "" instead of "0s"

hellobontempo:
`FlattenVaultDuration` previously returned `"0s"` for zero, nil, and unrecognized inputs. This was incorrect for Vault's `TypeDurationSecond` fields, where an integer value of `0` means the field is unset, not that its duration is literally zero. Any resource that stored `"0s"` in state for an omitted field would produce a perpetual plan diff on subsequent runs.

The fix changes all zero/nil/unrecognized paths to return `""` instead. The conversion logic was also refactored to collect the integer seconds into a single variable before the zero check and the `ShortDur` call, rather than duplicating those calls across each type case. The output for all non-zero inputs is unchanged.

From what I can tell, this bug hasn't caused any issues with any of the following parameters because of how `FlattenVaultDuration` was called. In all three cases the bug was latent, either the output was discarded or the zero input never arrived.

`flattenAuthMethodTune` - `default_lease_ttl`, `max_lease_ttl`
`mergeAuthMethodTune` immediately overwrites whatever FlattenVaultDuration produced with the user's config value. So even if FlattenVaultDuration stored "0s" in the raw tune map, it was thrown away before ever reaching state. 

SPIFFE - `bundle_refresh_hint`, `key_lifetime`, `ttl`
PKI - `not_before_duration`
Vault always returns non-zero defaults for these fields. Since the engine initializes them to real values FlattenVaultDuration never actually received 0 from Vault for these fields in practice. 

### Description

Relates OR Closes #0000

### Checklist

```

...
```

### Community Note

## PCI review checklist


victorr (approved):
Thanks for fixing this! 👍

PR #3049, opened 2026-09-17
Add write-only data_json_wo to vault_generic_endpoint

shaman007:
## Summary
- Add `data_json_wo` and `data_json_wo_version` to `vault_generic_endpoint` so arbitrary endpoint payloads (e.g. third-party plugin account configuration) can use ephemeral values without storing JSON in Terraform state.
- Follow the existing `vault_kv_secret_v2` write-only pattern (`GetRawConfigAt` on create/version change; do not populate `data_json` from Vault read when write-only is configured).

## Test plan
- [x] `TF_ACC=1 go test ./vault -run TestAccGenericEndpoint_data_json_wo -timeout 30m` against local Vault dev
- [ ] Provider CI acceptance tests

Made with [Cursor](https://cursor.com)


p-honik:
Hi Andrey,

I came across #3049 (data_json_wo on vault_generic_endpoint) while working through a similar problem — writing a third-party Vault plugin's config without the payload landing in state.

Noticed you closed it the same day you opened it, before review. Would you mind sharing why? Trying to figure out if it was abandoned, hit a blocker, or you're planning to resubmit it in a different form — would help me decide whether to pick it up or take another approach.

Thanks

shaman007:
Yes, because of the requirement of signing a CLA. I don't know if I want / if I can legally do that yet. I had tested the code, and it works and does the thing. The code itself is trivial, so if you'll do a similar thing faster, please go ahead. With CLA in our company it'll take 2-3 weeks to figure out.

